PR #21, opened 2018-08-14
it is a start

seguler:


zezha-msft:
@seguler please push to dev branch instead, we have updated the tests and Travis configurations so master branch no longer builds.

zezha-msft:
@seguler I'm rebasing your change to merge in, with #23

PR #61, opened 2018-09-26
Prjain azcopy

prjain-msft:


prjain-msft:
Need to exclude symlink support added for sync / copy command; added exclude blobType support to exclude blobs of given blobType

PR #149, opened 2019-01-04
oauth: accept external client id

yugangw-msft:
Right now, azcopy uses its own application id for token refreshing which will not work if the refresh token was acquired by external apps, such as Azure CLI and POSH. This PR will fix this problem, hence enable Azure CLI to host azcopy for faster storage data operations
I checked out other routines in the same file, I believe only the `RefreshTokenWithUserCredential` requires this adjustments.

//CC @williexu, @jiacfan, @zezha-msft

### DIFF
--- a/common/oauthTokenManager.go
+++ b/common/oauthTokenManager.go
@@ -309,6 +309,7 @@ type OAuthTokenInfo struct {
 	ActiveDirectoryEndpoint string `json:"_ad_endpoint"`
 	Identity                bool   `json:"_identity"`
 	IdentityInfo            IdentityInfo
+	ClientID                string `json:"_client_id"`
 }
 
 // IdentityInfo contains info for MSI.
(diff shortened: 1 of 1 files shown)


jiacfan:
> package common

The version in dev branch is the latest oauthTokenManager.go, would you mind merge this optimization with dev branch, and we can merge this PR to dev branch. Thanks!

---
Refers to: common/oauthTokenManager.go:21 in a77abc1. [](commit_id = a77abc116c96b0a655ab273ccea8cb3b55e589fc, deletion_comment = False)

common/oauthTokenManager.go
```diff
@@ -413,9 +414,14 @@ func (credInfo *OAuthTokenInfo) RefreshTokenWithUserCredential(ctx context.Conte
 		return nil, err
 	}
 
+	var applicationId = credInfo.ClientID
```
jiacfan: Would you please move this applicationId default value assigning logic to GetTokenInfo, which is factory method to get OAuthTokenInfo? And in this method, e.g NewServicePrincipalTokenFromManualToken can use ClientId from credInfo, as it now contains ClientId . Thanks!

jiacfan:
> func (uotm *UserOAuthTokenManager) GetTokenInfo(ctx context.Context) (*OAuthTokenInfo, error) {

Here is where token is initially get from. Please move the ClientID default value assigning to here. Thanks!

---
Refers to: common/oauthTokenManager.go:95 in a77abc1. [](commit_id = a77abc116c96b0a655ab273ccea8cb3b55e589fc, deletion_comment = False)

PR #159, opened 2019-01-10
Fix sparse data detection logic for uploading local file to Azure File.

jiacfan:
Fix sparse data detection logic for uploading local file to Azure File, which could cause data integrity issue.


zezha-msft:
Fixes #160

PR #174, opened 2019-01-19
Refactor uploads for DRY and performance

JohnRusk:
Refactors the STE upload code to avoid repetition and improve performance. (The performance changes primarily amount to moving from Memory Mapped Files to explict reads using standard File Read functions, and the concurrency changes that we already have in Dev from the download changes, to separate chunkfunc creation from chunkfunc execution.)

Key changes are:
1. Refactors the code for DRY, using new localToRemote and uploader types.
2. Stops using memory mapped files for uploads. (They remain in use for job part files etc).
3. Re-uses the same memory usage capping that we already have for downloads.
4. Adds makes the chunk state logger work for uploads.
5. Reads files sequentially (since sometimes that helps performance, and it never hurts).  _The only exception to sequential reading, in this new code, is if there is a retry, and which point it will go back to the file to refetch the data_
6. Tidies up a few things (e.g. failure messages for active transfers)


JohnRusk:
Reviewed offline by @zezha-msft and @jiacfan

PR #190, opened 2019-02-01
Add a nodelete flag for non-interactive deletion avoidance

N6UDP:
I needed the ability to run sync without deleting files on the destination if they were removed on the source non-interactively so I added a flag.  Of course I'm open to other options / renaming the flag.
Could also probably use a test or two.


zezha-msft:
@N6UDP Thank you so much for the PR! I've included your change in #225.

PR #195, opened 2019-02-04
Job completion messages

rickle-msft:


zezha-msft:
@rickle-msft there seems to be an extra commit in your PR: 2a8a70d. Could you remove it?

PR #219, opened 2019-02-20
Transfer will now fail if the source is modified after the transfer i…

rickle-msft:
…s scheduled


zezha-msft:
Will open another PR.

PR #226, opened 2019-02-22
 Store src and dest roots separately from transfer paths to save spac…

rickle-msft:
…e in plan file


zezha-msft:
Will open another PR.

Issue #209, opened 2019-02-13
Deploying a large number of files fails.


iclanton:
### Which version of the AzCopy was used? 

AzCopy 10.0.7-Preview

### Which platform are you using? (ex: Windows, Mac, Linux)

Windows

### What command did you run?

`azcopy copy D:\a\r1\a\build-artifacts\DeployArtifacts\* https://<ACCOUNT>.blob.core.windows.net/<PATH>/<SASTOKEN> --recursive=true`

### What problem was encountered?

The tool stalls for several minutes and then prints:

```
runtime: VirtualAlloc of 1048576 bytes failed with errno=1455
    + CategoryInfo          : NotSpecified: (runtime: Virtua...with errno=1455:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
    + PSComputerName        : localhost

atal error: out of memory
runtime stack:
runtime.throw(0xb47012, 0xd)
    /usr/local/go/src/runtime/panic.go:616 +0x88
runtime.sysMap(0xc0a8050000, 0x100000, 0x1, 0xfcb298)
    /usr/local/go/src/runtime/mem_windows.go:122 +0x13b
runtime.(*mheap).sysAlloc(0xfb1b20, 0x100000, 0x3f212e533d79c81f)
    /usr/local/go/src/runtime/malloc.go:470 +0xdb
runtime.(*mheap).grow(0xfb1b20, 0x40, 0x0)
    /usr/local/go/src/runtime/mheap.go:907 +0x67
runtime.(*mheap).allocSpanLocked(0xfb1b20, 0x40, 0xfcb2a8, 0x5d2fe20)
    /usr/local/go/src/runtime/mheap.go:820 +0x308
runtime.(*mheap).alloc_m(0xfb1b20, 0x40, 0x101, 0x0)
    /usr/local/go/src/runtime/mheap.go:686 +0x126

(with a lot more)
```

### How can we reproduce the problem in the simplest way?

We're deploying CDN artifacts for a large live service. We're deploying ~10,500 files, totaling ~5GB to a number of different Azure storage accounts during a CI release process. It's not the easiest repro. I'll try to contact you directly.

### Have you found a mitigation/solution?

No. This actually seems to be a regression as of a few weeks ago. The update warning forced us to update the tool to the latest version because otherwise the VSTS pipeline would fail, so we're unable to roll back to a previous version.

JohnRusk:
[...]

PR #228, opened 2019-02-22
Reduce memory usage, especially with many small files

JohnRusk:
For the small-files scenario:

* Reduce max slice count in pools
* Set GC Percent based on allocation patterns in this app. (I.e. lots of permanently-allocated big buffers up front, then nothing changes much... but we still want those relatively small later changes to be GC, even if they are small relative to the RAM occupied by the big buffers).

And, related, for big files:
* Gradually prune out contents of idle pools


JohnRusk:
This change, coupled with #216, should address #209.

PR #230, opened 2019-02-22
Uncomment sync test that was previously flaky

zezha-msft:


zezha-msft:
Smoke tests are being flaky.

Issue #253, opened 2019-03-04
Outdated README and issues scripting install


lawrencegripper:
Firstly, thanks this is awesome and working great for speeding up my copies. One issue I have is around install. 

Current `readme` instructs to install as follows:

```
wget -O azcopyv10.tar https://aka.ms/downloadazcopy-v10-linux
tar -xf azcopyv10.tar
cd azcopy_linux_amd64_10.0.4
./azcopy
```

However this fails as the tar contains `azcopy_linux_amd64_10.0.8` not `azcopy_linux_amd64_10.0.8`. 

The larger frustration is that I'm attempting to script the running of AzCopy as part of a `cloud-init` machine setup and including the version number inside the `tar` without providing a way to pin my install to a version means scripts break. 

As a workaround I've taken the url of the redirect and used this `https://azcopyvnext.azureedge.net/release20190301/azcopy_linux_amd64_10.0.8.tar.gz` then extracting using `tar -xf azcopyv10.tar --strip-components=1`. Can this URI be relied on to remain available?

In terms of a fix, would you be happy for me to submit a PR to use [GoReleaser](https://goreleaser.com/) to build the necessary binaries and publish them to the `releases` page on the github repository? I can commit the `.goreleaser.yaml` file but would need to CI configured with a `githubtoken` etc for the publish to work. 

I've done something similar with [azbrowse](https://github.com/lawrencegripper/azbrowse/releases) it automatically generates release notes, docker, brew, scoop and snap installers too if you want to set them up.

zezha-msft:
Hi @lawrencegripper, thanks for reaching out and providing feedbacks!

Yes, the download links will remain available. In general, I'd suggest to stick to a particular version for scripting purposes. 

I really appreciate your offer to add GoReleaser into this project, but I don't think it will be able to meet all of our requirements. To start, the Windows executable needs to be signed before releasing. Also, we want to switch to Azure DevOps for CI (since Travis is a bit slow and unreliable sometimes), so maybe we could do this work a bit later. Thanks a lot for the suggestion though, I'll keep it mind!

I've updated the [README](https://github.com/Azure/azure-storage-azcopy/commit/098e0fce81f5e71fe13d65bd0d785bd6dae320ec).

lawrencegripper:
Sure would you like me to update the README with details on how to pin to a release and using the extract method I mentioned when scripting things?

[...]

PR #269, opened 2019-03-11
Update README with installation for repeatable scripts

lawrencegripper:
Hi, 

I've had a go at adding some install instructions for pinning to a version of `azcopy` for use in scripts. How does it look? Hopefully useful to others, let me know if you'd like any edits/changes. 

Related to #253

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,11 +66,18 @@ To pin to a version of `azcopy`
 
 2. Now in your scripts download the version from this URL, unpack it and use as before. 
 
+Linux:
 ```
 
 wget -O azcopyv10.tar https://azcopyvnext.azureedge.net/release20190301/azcopy_linux_amd64_10.0.8.tar.gz
 tar -xf azcopyv10.tar --strip-components=1
 ./azcopy
+```
+Windows:
+```
+Invoke-WebRequest https://azcopyvnext.azureedge.net/release20190517/azcopy_windows_amd64_10.1.2.zip -OutFile azcopyv10.zip
+<<Unzip here>>
+
 ```
 
 > Note: `--strip-components=1` on the `tar` command removes the top level folder containing the version name and instead extracts the binary directly into the current folder. This allows the script to be updated with a new version of `azcopy` by only updating the `wget` URL.
--- a/README.md
+++ b/README.md
@@ -49,6 +49,32 @@ cd .\azcopy_windows_amd64_10.0.8
 .\azcopy.exe
 ```
 
+### Installation for repeatable scripts
+
+The version of `azcopy` available at `https://aka.ms/downloadazcopy-v10-linux` will be updated over time. If you plan on using `azcopy` as part of a script it's best to pin to a known version so future updates don't affect your scripts. 
+
+To pin to a version of `azcopy`
+
+1. Run `curl -v https://aka.ms/downloadazcopy-v10-linux` and review the output to locate the current version URL. Here is an example output from `curl` showing the `301 Moved Permanently` response.
+
+```
+...
+< HTTP/1.1 301 Moved Permanently
+< Location: https://azcopyvnext.azureedge.net/release20190301/azcopy_linux_amd64_10.0.8.tar.gz
+...
+```
+
+2. Now in your scripts download the version from this URL, unpack it and use as before. 
+
+```
+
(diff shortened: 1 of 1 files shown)


zezha-msft:
Thanks @lawrencegripper!

@seguler could you please review the doc change?

JohnRusk:
Hi Lawrence,

I see this PR has slipped though the cracks.  I'll discuss it with the rest of the team later in the week. (I know we didn't want to encourage pinning before GA.  Now that we are GA, I want to check in with the rest of the team to confirm the extent to which we are comfortable with pinning.  I agree with your point that folks want to be able to script against a predictable version).

README.md
```diff
@@ -49,6 +49,32 @@ cd .\azcopy_windows_amd64_10.0.8
 .\azcopy.exe
 ```
 
+### Installation for repeatable scripts
+
+The version of `azcopy` available at `https://aka.ms/downloadazcopy-v10-linux` will be updated over time. If you plan on using `azcopy` as part of a script it's best to pin to a known version so future updates don't affect your scripts. 
+
+To pin to a version of `azcopy`
+
+1. Run `curl -v https://aka.ms/downloadazcopy-v10-linux` and review the output to locate the current version URL. Here is an example output from `curl` showing the `301 Moved Permanently` response.
```
JohnRusk: Windows equivalent is this PowerShell command:
` (curl https://aka.ms/downloadazcopy-v10-windows -MaximumRedirection 0 -ErrorAction silentlycontinue).RawContent`

JohnRusk:
@lawrencegripper I had a chat with the team, and yes, we think your PR is a good idea and we should go ahead with it.  

One question: before we merge the PR, might you be able to update it so that, in addition to the Linux commands, it also shows Windows commands?  E.g. to find the versioned URL this works in PowerShell 

` (curl https://aka.ms/downloadazcopy-v10-windows -MaximumRedirection 0 -ErrorAction silentlycontinue).RawContent`

README.md
lawrencegripper: I've added this one in now - good call

lawrencegripper:
@JohnRusk no problem, I've attempted to update the script but I can't find any universal way to do the equivalent of `--strip-compontents=1` from `tar` in windows. 

Do you know of a way to do this with `Expand-Archive` - [docs here](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.archive/Expand-Archive?view=powershell-6).

My suggestion would be that the packages releases don't put the exe in the zip under a versioned path and this is why this step is required. If the zip path didn't contain the version is then things would be simpler.

JohnRusk:
Hi @lawrencegripper. Sorry about the delayed reply.  I don't know of any way to replicate --strip-components directly. But, you could make a child folder with a unique name (e.g. from a guid) so that it won't clash with anything else. Expand-Archive into that child folder.  Cd into there then cd az* (or whatever will take you into the version AzCopy folder that's inside the zip), then copy * ..\.., then cd ..\.., then delete the child directory.

zezha-msft:
Since our README has changed dramatically, @normesta could you please incorporate this PR into the docs website instead?

normesta:
Yep. I'll have a look at this one. Thx!

lawrencegripper:
Cool, I'll close this PR now then.

normesta:
Hi all - I'm a bit confused about the scenario here. It seems logical to want to pin a script to a specific version of AzCopy, but these steps don't seem to be doing that. I'm sure I'm missing something here, but these steps appear to use a PowerShell command to obtain the latest URL of the script. That would by definition be the latest updated version of the script right? That's exactly what we wouldn't want to do. Can you please explain how this gives the developer / administrator the URL of a specific version and not just the latest version which is what it seems like we're getting here? Thx!

JohnRusk:
Norm, there are two parts to the scripts. The first part, as you have seen, discovereds the underlying URL of the latest version. So you have a fixed URL for that specific version. The second part is a download script … that uses that specific URL.  So the second part is pinned to the version.

(whereas if you write a download script that uses the aka.ms link, it's not pinned to a version).

normesta:
Thank you. Yes this sequence makes sense. Finding the current version happens once. Then, for each use of the script downloads that specific version of AzCopy. Why not just store a version of AzCopy for use with a script. Why always download AzCopy as part of running the script?

zezha-msft:
@normesta in a typical CI environment (like Travis), environments are completely wiped after a run and new jobs start from brand new environments, thus they have to download AzCopy every time.

PR #325, opened 2019-04-08
Clarify --overwrite parameter in README.md

MattKotsenas:
Clarify that to avoid overwriting files, the `--overwrite` parameter should be `false`.


zezha-msft:
Thanks @MattKotsenas!

PR #326, opened 2019-04-10
Enable blob type and tier related testing, and optimize param validation.

jiacfan:


jiacfan:
Check https scheme is used for s2s copy source and destination. Enable s2s testing related to blob type and blob tier.

PR #334, opened 2019-04-16
Fix/third party notices file

JohnRusk:
Update this file to the final version


JohnRusk:
Just force-pushed to fix a error in what the branch was based on

PR #335, opened 2019-04-17
Fix/flag names

JohnRusk:
Reviewers: the changes, at least to test code, were more extensive that we would have liked, due to the changing of the block size to be specified in MiB.  

PR contains two commits: one with the block size changes, and one with all the other (simple) renames.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,34 @@
 # Change Log
 
+## Version 10.0.10 (GA)
+
+### Breaking changes
+
+1. The `--block-size` parameter has been replaced by `--block-size-mb`. The old parameter took a number of _bytes_; the
+   new one takes a number of Megabytes (MiB).
+1. The following command line parameters have been renamed, for clarity
+    * `--output` is now `--output-type`
+    * `--md5-validation` is now called `--check-md5`
+    * `--s2s-source-change-validation` is now called `--s2s-detect-source-changed`
+    * `--s2s-invalid-metadata-handle` is is now called `--s2s-handle-invalid-metadata`
+    * `--quota` (in the `make` command) is now called `--quota-gb`. Note that the values were always in GB, the new name
+      simply clarifies that fact
+
+### New features
+
+1. AzCopy is now able to be configured to use older API versions. This enables (limited) support for Azure Stack.
+
+### Bug fixes
+
+1. AzCopy v10 now works correctly with ADLS Gen 2 folders that contain spaces in their names
+1. When cancelling with CRTL-C, status of in-progress transfers is now correctly recorded.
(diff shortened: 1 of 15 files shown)


cmd/make.go
```diff
@@ -214,6 +214,6 @@ func init() {
 		},
 	}
 
-	makeCmd.PersistentFlags().Uint32Var(&rawArgs.quota, "quota", 0, "specifies the maximum size of the share in gigabytes, 0 means you accept the file service's default quota.")
+	makeCmd.PersistentFlags().Uint32Var(&rawArgs.quota, "quota-gb", 0, "specifies the maximum size of the share in gigabytes, 0 means you accept the file service's default quota.")
```
zezha-msft: Clarify that it's GiB?

testSuite/scripts/test_upload_page_blob.py
```diff
@@ -66,7 +66,7 @@ def test_page_blob_upload_partial_sparse_file(self):
         # execute azcopy pageblob upload.
         destination_sas = util.get_resource_sas(file_name)
         result = util.Command("copy").add_arguments(file_path).add_arguments(destination_sas).add_flags("log-level", "info"). \
-            add_flags("block-size", "4194304").add_flags("blob-type","PageBlob").execute_azcopy_copy_command()
```
zezha-msft: wow that's a lot of manual changes.. we should definitely improve this later

cmd/make.go
JohnRusk: Sound OK if I clarify only in the comments?  Alternative, naming it quota-gib seems odd to me somehow. (Not 100% sure why).

testSuite/scripts/test_upload_page_blob.py
JohnRusk: Agreed!  But not doing anything about it now ;-)

ChangeLog.md
```diff
@@ -1,5 +1,34 @@
 # Change Log
 
+## Version 10.0.10 (GA)
```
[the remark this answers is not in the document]
JohnRusk: Done

cmd/make.go
zezha-msft: it could be in the flag description, right?

testSuite/scripts/test_upload_page_blob.py
zezha-msft: :shipit:

cmd/make.go
JohnRusk: Yes. I've put it in the description.  Which was what I meant, when I foolishly said "comments"

zezha-msft: Sounds good to me!

Issue #317, opened 2019-04-04
Include file size when listing


brettrowberry:
### Which version of the AzCopy was used? 
10.0.8

### What command did you run?
list

### What problem was encountered?
Did not include file size

### How can we reproduce the problem in the simplest way?
List files in any storage account and see that only filename is included.

### Have you found a mitigation/solution?
No

zezha-msft:
Added in 10.1.0.

PR #341, opened 2019-04-18
QoL change: List file sizes atop names

Virepri:
Tackles issue #317.

This solution isn't *perfect* but it more than gets the job done. Would it be preferable if an optional parameter was added to treat 1 GB as 1000MB as some OSes, hard drive manufacturers, etc. do?

Atop that, would it be preferable to enable the ability to keep a running tally on file size and file count?


cmd/list.go
```diff
@@ -163,3 +165,25 @@ func printListContainerResponse(lsResponse *common.ListContainerResponse) {
 		glcm.Info(lsResponse.Blobs[index])
 	}
 }
+
+func ByteSizeToString(size int) string {
```
zezha-msft: we can make it lower-case since it doesn't need to be exported

cmd/list.go
```diff
@@ -163,3 +165,25 @@ func printListContainerResponse(lsResponse *common.ListContainerResponse) {
...
+	unit := 0
+	floatSize := float64(size)
+	gigSize := 1024
+
+	for floatSize/float64(gigSize) >= 1 {
+		unit++
+		floatSize /= float64(gigSize)
+	}
+
+	return strconv.FormatFloat(floatSize, 'f', 2, 64) + " " + units[unit]
```
zezha-msft: please consider adding some unit tests for validation

cmd/list.go
```diff
@@ -163,3 +165,25 @@ func printListContainerResponse(lsResponse *common.ListContainerResponse) {
 		glcm.Info(lsResponse.Blobs[index])
 	}
 }
+
+func ByteSizeToString(size int) string {
+	units := []string{
+		"bytes",
```
zezha-msft: abbreviations would read better, what do you think?

cmd/list.go
Virepri: No longer exported as of [9cc9d04](https://github.com/Azure/azure-storage-azcopy/pull/341/commits/9cc9d04ee8a8bc98aee541a17413c42e0f359cfe)

cmd/list.go
Virepri: Abbreviations inserted in [9cc9d04](https://github.com/Azure/azure-storage-azcopy/pull/341/commits/9cc9d04ee8a8bc98aee541a17413c42e0f359cfe)

cmd/list.go
Virepri: Unit tests added in [9cc9d04](https://github.com/Azure/azure-storage-azcopy/pull/341/commits/9cc9d04ee8a8bc98aee541a17413c42e0f359cfe)

cmd/bytesizetostring_test.go
```diff
@@ -0,0 +1,103 @@
+package cmd
+
+import (
+	"testing"
+)
+
+func TestBToString(t *testing.T) {
```
zezha-msft: Thanks for the tests! Please consider using the `check` package for validation. Refer to other test files in the cmd module.

seguler:
Wohoo! 

Some feedback: Can we return sizes in bytes by default so it is easy to do sum using PowerShell or bash ?

There could still be an option such as --human-readable to show bytes with KB/MB/GB units.

Thoughts ?

seguler:
Also any chance we can add a counter that we could print total size at the end of the listing ?

Virepri:
Absolutely. I added those as potential changes I'd be willing to make in the PR comment in the first place. I'll get to it once I find a good time.

seguler:
thanks @Virepri !
@zezha-msft do you agree with my proposal ?

zezha-msft:
@seguler :shipit: actually I think a `machine-readable` flag might be better. In other words we display human readable sizes by default, and avoid doing that if the flag is true.

zezha-msft:
@Virepri you should have write access now, please push to a branch on the repo to trigger CI.

Virepri:
@seguler's suggestion has been added, as well as feedback from @JohnRusk has been taken.

I think this is about as good as it's gonna get

zezha-msft:
tracking with #342

PR #343, opened 2019-04-20
Fixed clean local path for root drive

zezha-msft:


cmd/zc_traverser_local_test.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (s *localTraverserTestSuite) TestCleanLocalPathForWindows(c *chk.C) {
+	// ignore these tests when not running on Windows
+	// as the path cleaning behavior depends on the platform
+	if os.PathSeparator != '\\' {
+		c.SucceedNow()
+	}
+
+	testCases := map[string]string{
```
adreed-msft: I noticed the use of a raw string in zc_traverser_local.go line 146 to handle a backslash (perhaps to eliminate the confusion when using escape characters). Why not continue that trend in your tests (especially considering there are actual double slash tests in the non-exclusive path cleaning tests)?

cmd/zc_traverser_local_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	for orig, expected := range testCases {
+		c.Assert(cleanLocalPath(orig), chk.Equals, expected)
+	}
+}
+
+func (s *localTraverserTestSuite) TestCleanLocalPathForWindows(c *chk.C) {
+	// ignore these tests when not running on Windows
+	// as the path cleaning behavior depends on the platform
+	if os.PathSeparator != '\\' {
```
adreed-msft: I doubt there _is_ something akin to the `_windows`/`_unix`/etc. suffix for testing, but this feels a bit _off_ as a way to handle Windows only tests, even if there _isn't_ a file suffix like those.

I'd suggest using [`runtime.GOOS`](https://golang.org/pkg/runtime/#pkg-constants) instead of `os.PathSeparator` instead if you plan to make these tests Windows-exclusive.

zezha-msft: I agree with you that runtime.GOOS is more explicit, but I think keeping it like this is better since the same logic is used in the code that cleans the path. I'm not sure about the WSL behavior, but I'd imagine that the os.PathSeparator is more pertinent to what we are tryiing to do

cmd/zc_traverser_local_test.go
zezha-msft: good point!

cmd/zc_traverser_local_test.go
```diff
@@ -0,0 +1,49 @@
...
+		c.SucceedNow()
```
adreed-msft: Another quick note: Use `c.Skip(reason string)` instead. It'll give a more clear indication of why this test was not run.

cmd/zc_traverser_local_test.go
adreed-msft: So, if you're aiming to run these tasks on any OS that uses `\` as a path separator, perhaps it shouldn't be called `*ForWindows`? I'm really trying to think of a naming scheme that would be more fitting.

zezha-msft: Wait, this test is for Windows though, so the name is fine, right?

We have two options for detecting whether we should skip the test:

1. We ask runtime.GOOS to see if we are not on Windows
2. We check if os.PathSeparator is not '\\'

Both options seem equal to me, and option 2 is more consistent with the rest of the code base so I picked it. I don't feel strongly about this though.

zezha-msft: I need to merge now to start the change log update. We can chat about this tomorrow. 😄

PR #347, opened 2019-04-22
SAS Token auth on DFS endpoints

adreed-msft:
The following 3 commits enable the use of a SAS token to authenticate uploads and downloads on `dfs.core.windows.net` endpoints.

In order to facilitate this addition, some core files were shared to azbfs (namely relating to SAS tokens). Namely, `zc_sas_account.go` and `zc_sas_query_params.go`.

As usual, some changes are as a result of the automatic `go fmt` execution configured in my Goland setup.

### DIFF
--- a/azbfs/zc_sas_account.go
+++ b/azbfs/zc_sas_account.go
@@ -1,0 +1,218 @@
+package azbfs
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+)
+
+// AccountSASSignatureValues is used to generate a Shared Access Signature (SAS) for an Azure Storage account.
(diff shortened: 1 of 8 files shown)


azbfs/zc_sas_query_params.go
```diff
@@ -0,0 +1,261 @@
+package azbfs
```
zezha-msft: not reviewing these zc files and assuming they are from azblob

azbfs/zc_sas_account.go
```diff
@@ -0,0 +1,218 @@
+package azbfs
```
zezha-msft: minor: consider adding a sas unit test to azbfs

adreed-msft:
Uploaded to azure:SASauth, replacing this PR

PR #354, opened 2019-04-26
Fixed memory alignment issue for atomic functions on 32 bit system

zezha-msft:


cmd/sync.go
```diff
@@ -150,6 +150,23 @@ func (raw *rawSyncCmdArgs) cook() (cookedSyncCmdArgs, error) {
 }
 
 type cookedSyncCmdArgs struct {
+	// NOTE: for the 64 bit atomic functions to work on a 32 bit system, we have to guarantee the right 64-bit alignment
+	// so the 64 bit integers are placed first in the struct to avoid future breaks
+	// refer to: https://golang.org/pkg/sync/atomic/#pkg-note-BUG
+	// defines the number of files listed at the source and compared.
+	atomicSourceFilesScanned uint64
+	// defines the number of files listed at the destination and compared.
+	atomicDestinationFilesScanned uint64
```
adreed-msft: During our discussion on Friday you suggested we did function signatures (destination, source). If you were to construct a literal cookedSyncCmdArgs w/out variable names, this would be out of order.

Of course, this is a super minor nitpick that does not actually matter (especially because it's not exported)-- Follow your own discretion on this one.

zezha-msft: Since it's not exported (like the SDK), I'm not sure what we could do here.

adreed-msft: That's about what I expected. Nothing needed to be changed in the first place.

Issue #199, opened 2019-02-05
ADLS Gen2 - B2B User : provide relevant error message when tenantid not provided


kvaes:
### What command did you run?
(login with B2B user, where the tenant is different from the one of the owner of the storage account)
azcopy login
azcopy cp https://testaccount.dfs.core.windows.net/testcontainer/testfile. /test
=> "failed to perform copy command due to error: cannot start job due to error: the destination must be an existing directory in this download scenario."

(login with B2B user, whilst specifying the tenant of the one of the owner of the storage account)
azcopy login --tenant-id *insert tenant id of storage account/subscription owner*
azcopy cp https://testaccount.dfs.core.windows.net/testcontainer/testfile. /test
=> Scanning...
Using OAuth token for authentication.

Job eb7e7868-f41e-ad44-78ee-ea6a922a8015 has started

Log file is located at: /home/kvaes/.azcopy/eb7e7868-f41e-ad44-78ee-ea6a922a8015.log
0 Done, 0 Failed, 1 Pending, 0 Skipped, 1 Total , 2-sec Throughput (Mb/s): 81.0663

Job eb7e7868-f41e-ad44-78ee-ea6a922a8015 summary
Elapsed Time (Minutes): 0.0669
Total Number Of Transfers: 1
Number of Transfers Completed: 1
Number of Transfers Failed: 0
Number of Transfers Skipped: 0
TotalBytesTransferred: 21326192
Final Job Status: Completed

### Expected behaviour? Provide an error message that states that there was an issue with the authentication part.

kvaes:
Update : It seems that the above error message is also shown if the permissions (POSIX ACLs) are not correct. So it's not limited to the the tenant-id.

jiacfan:
@kvaes Thanks for reaching us!  We'll discuss this internal, and update the thread later.

Best Regards,
Jiachen

adreed-msft:
Hi, kvaes, could you provide the complete log for your test with no specified Tenant ID? I suspect there's something important logged that you have omitted. Before AzCopy will reach the error that you had, it will output a graceful error along the lines of `INFO: source is not validated as a single file: <Error>`.

I believe that will help us target what kind of error is the culprit and thus provide a meaningful error message.

kvaes:
Sent the log offline.

adreed-msft:
Hi, this was patched in #358 and has been released as a part of AzCopy 10.2.x. Enjoy!

PR #358, opened 2019-05-01
Tackling incorrect error presented in #199

adreed-msft:
These commits change how error handling works in download scenarios in the following ways:
- Enables a way for certain errors to stop their host enumerator in a clean way (Because if we don't stop it, the command continues even though we already know auth won't work)
- Adds catches for InvalidAuthenticationInfo ServiceCodes, generated primarily by errors related to an incorrect token issuer, etc.

Furthermore, it adds a warning to `azcopy login` in order to warn users in advance that they are logging in under the common tenant.

As a result, this solves issue #199 by providing relevant errors and warnings when a B2B user attempts to authenticate a storage account from the wrong tenant (as well as when a user attempts to log in from the common tenant).

### DIFF
--- a/cmd/copyDownloadBlobEnumerator.go
+++ b/cmd/copyDownloadBlobEnumerator.go
@@ -84,7 +84,9 @@ func (e *copyDownloadBlobEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 			}
 			return nil
 		} else {
-			handleSingleFileValidationErrorForBlob(err)
+			if handleSingleFileValidationErrorForBlob(err) {
+				return err
+			}
 		}
 	}
 
--- a/cmd/copyDownloadBlobEnumerator.go
+++ b/cmd/copyDownloadBlobEnumerator.go
@@ -84,7 +84,7 @@ func (e *copyDownloadBlobEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 			}
 			return nil
 		} else {
-			if handleSingleFileValidationErrorForBlob(err) {
+			if isFatal := handleSingleFileValidationErrorForBlob(err); isFatal {
 				return err
 			}
 		}
--- a/cmd/copyEnumeratorHelper.go
+++ b/cmd/copyEnumeratorHelper.go
@@ -84,46 +84,73 @@ var infoCopyFromBucketDirectoryListOfFiles = "trying to copy the source as bucke
 var infoCopyFromDirectoryListOfFiles = "trying to copy the source as directory/list of files"
 var infoCopyFromAccount = "trying to copy the source account"
 
-func handleSingleFileValidationErrorForBlob(err error) {
+func handleSingleFileValidationErrorForBlob(err error) (stop bool) {
 	errRootCause := err.Error()
 	stgErr, isStgErr := err.(azblob.StorageError)
 	if isStgErr {
-		if stgErr.ServiceCode() == azblob.ServiceCodeAuthenticationFailed {
+		if stgErr.ServiceCode() == azblob.ServiceCodeInvalidAuthenticationInfo {
+			errRootCause = fmt.Sprintf("%s, Please ensure all login details are correct (or that you are signed into the correct tenant for this blob)", stgErr.ServiceCode())
+			stop = true
+		} else if stgErr.ServiceCode() == azblob.ServiceCodeAuthenticationFailed {
 			errRootCause = fmt.Sprintf("%s, please check if SAS or OAuth is used properly, or source is a public blob", stgErr.ServiceCode())
+			stop = true
 		} else {
 			errRootCause = string(stgErr.ServiceCode())
 		}
 	}
 
-	glcm.Info(fmt.Sprintf(handleSingleFileValidationErrStr, errRootCause))
+	if stop {
+		glcm.Info(errRootCause)
+	} else {
+		glcm.Info(fmt.Sprintf(handleSingleFileValidationErrStr, errRootCause))
+	}
+	return
 }
 
-func handleSingleFileValidationErrorForAzureFile(err error) {
+func handleSingleFileValidationErrorForAzureFile(err error) (stop bool) {
 	errRootCause := err.Error()
 	stgErr, isStgErr := err.(azfile.StorageError)
 	if isStgErr {
-		if stgErr.ServiceCode() == azfile.ServiceCodeAuthenticationFailed {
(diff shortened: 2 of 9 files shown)


cmd/copyEnumeratorHelper.go
```diff
@@ -84,46 +84,73 @@ var infoCopyFromBucketDirectoryListOfFiles = "trying to copy the source as bucke
...
+		} else if stgErr.ServiceCode() == azblob.ServiceCodeAuthenticationFailed {
 			errRootCause = fmt.Sprintf("%s, please check if SAS or OAuth is used properly, or source is a public blob", stgErr.ServiceCode())
+			stop = true
 		} else {
 			errRootCause = string(stgErr.ServiceCode())
 		}
 	}
 
-	glcm.Info(fmt.Sprintf(handleSingleFileValidationErrStr, errRootCause))
+	if stop {
```
zezha-msft: a bit confused here, why the different format depending on `stop`?

cmd/login.go
```diff
@@ -103,6 +103,11 @@ func (lca loginCmdArgs) process() error {
 		return err
 	}
 
+	if lca.tenantID == "" || lca.tenantID == "common" {
+		glcm.Info("WARNING: Logging in under the \"Common\" tenant. This will log the account in under its' home tenant, NOT under your target tenant.")
```
zezha-msft: typos here:

glcm.Info("WARNING: Logging in under the \"Common\" tenant. This will log the account in under **its** home tenant, NOT under your target tenant.")
		glcm.Info("If you plan to use **AzCopy** for B2B **Storage Accounts**, please sign in under the target tenant with --tenant-id")

cmd/copyEnumeratorHelper.go
adreed-msft: So, the way that AzCopy checks if something is a file or folder is that it tries to error out when attempting to download a file first. If it errored out, it would give a strange error formatting in my opinion. I don't need to know that it isn't a file if I know that there's an authentication error, right?

cmd/login.go
adreed-msft: Will fix in a moment, thanks for catching me.

cmd/copyEnumeratorHelper.go
```diff
@@ -84,46 +84,73 @@ var infoCopyFromBucketDirectoryListOfFiles = "trying to copy the source as bucke
 var infoCopyFromDirectoryListOfFiles = "trying to copy the source as directory/list of files"
 var infoCopyFromAccount = "trying to copy the source account"
```
JohnRusk: Sorry to mention it, but there seem to be some cases that are not getting caught by this code.  The ones I've spotted are:

1. We also enumerate blob storage in copyS2SMigrationBlobEnumerator.go. But that code does not call this routine, so the changes in this PR currently are not helping with the enumeration in copyS2SMigrationBlobEnumerator.go.
2. Sometimes, depending on how AzCopy is called, the code here does not execute.  E.g. when I sign in with the wrong tenant, this code in this file here executes if I run `azcopy.exe cp "https://<account>.dfs.core.windows.net/test1/foo.txt" d:\testdata\foo --recursive` but it does _not_ execute if I run `azcopy.exe cp "https://<account>.dfs.core.windows.net/test1" d:\testdata\foo --recursive`

sooo…. (and this is just a half-baked idea) how about making the "handleSingleFileValidation*" methods so they _just_ focus on returning "stop".  But move the new handling for outputting an informative message out of this file, and put it at line 896 of copy.go.  There, I think it will catch every error. (You may or may not have to take nesting of error objects into account.  If you're lucky, the simple cast to azBlob/File/bfs.StorageError will take care of grabbing the right nested error for you).  If that works, it will be guaranteed to catch everything.

cmd/copyEnumeratorHelper.go
```diff
@@ -84,46 +84,73 @@ var infoCopyFromBucketDirectoryListOfFiles = "trying to copy the source as bucke
...
```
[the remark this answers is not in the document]
adreed-msft: That outputs a proper authentication failed message currently. These changes were intended to resolve an incorrect error message (couldn't verify whether the source was a file or a folder, or something of that variety).

cmd/copyDownloadBlobEnumerator.go
```diff
@@ -84,7 +84,9 @@ func (e *copyDownloadBlobEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 			}
 			return nil
 		} else {
-			handleSingleFileValidationErrorForBlob(err)
+			if handleSingleFileValidationErrorForBlob(err) {
```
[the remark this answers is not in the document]
adreed-msft: That actually is a good readability suggestion, I appreciate it.

cmd/login.go
```diff
@@ -103,6 +103,11 @@ func (lca loginCmdArgs) process() error {
 		return err
 	}
```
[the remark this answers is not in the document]
adreed-msft: This *is* possible, but this leads to code duplication due to the number of login types we have and the way things currently work. I will remove the warning, though.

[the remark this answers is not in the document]
adreed-msft: Specifically, that message is either going to be output during the login process or after it. If we do it during the process, then, code duplication ensues. If we do it afterwards, I feel as though warning them after the problem has occurred is a bit counter-productive.

cmd/copyEnumeratorHelper.go
```diff
@@ -84,46 +84,73 @@ var infoCopyFromBucketDirectoryListOfFiles = "trying to copy the source as bucke
...
+	return
 }
 
-func handleSingleFileValidationErrorForADLSGen2(err error) {
+func handleSingleFileValidationErrorForADLSGen2(err error) (stop bool) {
 	errRootCause := err.Error()
 	stgErr, isStgErr := err.(azbfs.StorageError)
 	if isStgErr {
-		if stgErr.ServiceCode() == azbfs.ServiceCodeAuthenticationFailed {
+		if stgErr.ServiceCode() == azbfs.ServiceCodeInvalidAuthenticationInfo {
```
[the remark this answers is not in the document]
adreed-msft: Thanks!

cmd/copyEnumeratorHelper.go
adreed-msft: This seems to be totally resolved in the latest commit.

zezha-msft:
@adreed-msft please use "squash and merge", so that the result commit shows up with your name.

PR #363, opened 2019-05-02
Output the end-of-job summary to the log file

adreed-msft:
Where's my staples button saying "that was easy"?

I would've __preferred__ to add this to `&lifecycleMgr.Exit()`, though I didn't want to change the function signature because there are other functions that don't have a summary or log. If you'd like me to move it over to `&lifecycleMgr.Exit()`, however, I'm more than happy to.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -25,6 +25,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


JohnRusk:
Looks good.

adreed-msft:
Merging to dev since all reviews passed.

PR #364, opened 2019-05-02
Drop logging the stack on HTTP errors

adreed-msft:
As discussed with Ze and John, this is an expensive operation that fills the log file far too much. PR coming to azure-pipeline-go soon with more log reductions.

### DIFF
--- a/ste/xferLogPolicy.go
+++ b/ste/xferLogPolicy.go
@@ -98,9 +98,6 @@ func NewRequestLogPolicyFactory(o RequestLogOptions) pipeline.Factory {
 				}
 
 				pipeline.WriteRequestWithResponse(b, prepareRequestForLogging(request), response.Response(), err)
-				if logLevel <= pipeline.LogError {
-					b.Write(stack()) // For errors (or lower levels), we append the stack trace (an expensive operation)
-				}
 				msg := b.String()
 
 				if forceLog {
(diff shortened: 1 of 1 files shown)


ste/xferLogPolicy.go
```diff
@@ -98,9 +98,6 @@ func NewRequestLogPolicyFactory(o RequestLogOptions) pipeline.Factory {
 				}
 
 				pipeline.WriteRequestWithResponse(b, prepareRequestForLogging(request), response.Response(), err)
-				if logLevel <= pipeline.LogError {
```
zezha-msft: We only want to remove the stack trace for HTTP errors. This is removing it altogether.

adreed-msft: Pardon me, misunderstood the code, thought it was exclusively for HTTP related errors.

ste/xferLogPolicy.go
```diff
@@ -98,8 +98,8 @@ func NewRequestLogPolicyFactory(o RequestLogOptions) pipeline.Factory {
...
-					b.Write(stack()) // For errors (or lower levels), we append the stack trace (an expensive operation)
+				if logLevel <= pipeline.LogError && !httpError {
```
JohnRusk: Can you whack a brief comment in here, to explain why we are not logging call stacks for HTTP errors. Namely they don't add much value (if any) but they do add a lot of bulk to the log in cases such as lots of 503 responses.

adreed-msft: Sure thing!

zezha-msft:
@adreed-msft please use "squash and merge", so that the result commit shows up with your name.

PR #365, opened 2019-05-02
Enabled obtaining a proxy from Edge/IE

adreed-msft:
This was MUCH easier than I would've thought. Thanks @JohnRusk for pointing me to the right place.

### DIFF
--- a/ste/getProxy_windows.go
+++ b/ste/getProxy_windows.go
@@ -8,12 +8,11 @@ import (
 )
 
 func getProxy() func(*http.Request) (*url.URL, error) {
-	if prox := httpproxy.FromEnvironment(); (prox.HTTPProxy != "" || prox.HTTPSProxy != "") && prox.NoProxy == "" {
+	if prox := httpproxy.FromEnvironment(); prox.HTTPProxy != "" || prox.HTTPSProxy != "" || prox.NoProxy != "" {
 		return http.ProxyFromEnvironment
-		//Option to default to env vars, with a fallback to env vars if grabbing from IE/Edge fails
 	}
 
-	key, err := registry.OpenKey(registry.LOCAL_MACHINE, `Software\Microsoft\Windows\CurrentVersion\Internet Settings`, registry.QUERY_VALUE)
+	key, err := registry.OpenKey(registry.CURRENT_USER, `Software\Microsoft\Windows\CurrentVersion\Internet Settings`, registry.QUERY_VALUE)
 	if err != nil {
 		return http.ProxyFromEnvironment
 	}
--- a/ste/getProxy_windows.go
+++ b/ste/getProxy_windows.go
@@ -1,0 +1,35 @@
+package ste
+
+import (
+	"golang.org/x/net/http/httpproxy"
+	"golang.org/x/sys/windows/registry"
+	"net/http"
+	"net/url"
+)
+
+func getProxy() func(*http.Request) (*url.URL, error) {
+	if prox := httpproxy.FromEnvironment(); (prox.HTTPProxy != "" || prox.HTTPSProxy != "") && prox.NoProxy == "" {
+		return http.ProxyFromEnvironment
+		//Option to default to env vars, with a fallback to env vars if grabbing from IE/Edge fails
+	}
+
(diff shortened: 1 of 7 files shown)


ste/mgr-JobPartMgr.go
```diff
@@ -90,7 +90,7 @@ const AzCopyMaxIdleConnsPerHost = 500
 func NewAzcopyHTTPClient() *http.Client {
 	return &http.Client{
 		Transport: &http.Transport{
-			Proxy: http.ProxyFromEnvironment,
+			Proxy: getProxy(),
```
JohnRusk: This is not the only place in our codebase where HTTP clients are created.  Need to find the others, and fix them too.

adreed-msft: The only other places are for the internal RPC (which I don't think _needs_ a proxy necessarily?), unless said clients are created inside the blob and file SDKs.

ste/getProxy_windows.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"golang.org/x/net/http/httpproxy"
+	"golang.org/x/sys/windows/registry"
+	"net/http"
+	"net/url"
+)
+
+func getProxy() func(*http.Request) (*url.URL, error) {
+	if prox := httpproxy.FromEnvironment(); (prox.HTTPProxy != "" || prox.HTTPSProxy != "") && prox.NoProxy == "" {
```
JohnRusk: Why does this require that NoProxy == "".  That seems like a bug to me. 

NoProxy is a list of addresses for which no proxy should be used.  If that list is populated, that doesn't appear to be a reason to disregard the "from environment" proxy.  IMHO we should use the environment-based proxy if either HTTPProxy or HTTPSProxy is set.

ste/getProxy_windows.go
```diff
@@ -0,0 +1,35 @@
...
+		return http.ProxyFromEnvironment
+		//Option to default to env vars, with a fallback to env vars if grabbing from IE/Edge fails
+	}
+
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, `Software\Microsoft\Windows\CurrentVersion\Internet Settings`, registry.QUERY_VALUE)
```
JohnRusk: @adreed-msft It looks like, as is common for many things in the registry, the settings that are actually used are per user. I.e. the ones that get used are in registry.CURRENT_USER, not registry.LOCAL_MACHINE. I verified this just now by changing my own settings, and observing that it was the CURRENT_USER section of the registry that changed, not LOCAL_MACHINE.

ste/getProxy_windows.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		return http.ProxyFromEnvironment
+	}
+	defer key.Close()
+
+	strURL, _, err := key.GetStringValue("ProxyServer")
```
JohnRusk: When I was looking at the IE proxy settings window on my machine just now, I realized we are not covering all the cases.  
1. The most minor (in terms of consequence) is that we are not covering the exclusion list that can be storage in ProxyOverride registry value.
2. More importantly, we are not respecting the ProxyEnable flag. That's important, because its possible for ProxyServer to be populated, but ProxyEnable set to false.
3. But potentially the biggest issue of all is that we can't easily support the auto-configuration option. (which I suspect is actually probably more commonly used that the ProxyServer value).

So.... to address the above, could you have a look at https://github.com/mattn/go-ieproxy and (a) see if it looks like a good way to address 2 and 3 above (and maybe 1). and (b) do a quick web search to see if that particular project looks like the best one to use, or if there are other strong contenders.

cc @zezha-msft in case he has any suggestions.

ste/getProxy_windows.go
adreed-msft: This would actually make sense-- I was following Microsoft documentation at the time and this was the location they were described to be in; will change this momentarily.

ste/getProxy_windows.go
adreed-msft: This was a misunderstanding of the environment variable's purpose.

adreed-msft:
Abandoning PR to change to mattn/go-ieproxy

PR #372, opened 2019-05-08
Enabled copying to/from various blob types S2S

adreed-msft:
This commit enables various copy types (such as block -> append/page, etc.) and adds tests for each type of blob copy direction.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -368,12 +368,6 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 			cooked.pageBlobTier != common.EPageBlobTier.None() {
 			return cooked, fmt.Errorf("blob-tier is not supported while copying from sevice to service")
 		}
-		// Disabling blob type override.
-		// i.e. not support block -> append/page, append -> block/page, page -> append/block,
-		// and when file and s3 is source, only block blob destination is supported.
-		if cooked.blobType != common.EBlobType.None() {
-			return cooked, fmt.Errorf("blob-type is not supported while coping from service to service")
-		}
 		if cooked.noGuessMimeType {
 			return cooked, fmt.Errorf("no-guess-mime-type is not supported while copying from service to service")
 		}
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -364,6 +364,12 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 		}
 		// Disabling blob tier override, when copying block -> block blob or page -> page blob, blob tier will be kept,
 		// For s3 and file, only hot block blob tier is supported.
+		if cooked.fromTo == common.EFromTo.FileBlob() && cooked.blobType != common.EBlobType.None() {
+			return cooked, fmt.Errorf("file -> blob is not a supported direction while copying from service to service")
+		}
+		if cooked.fromTo == common.EFromTo.S3Blob() && cooked.blobType != common.EBlobType.None() {
+			return cooked, fmt.Errorf("custom blob type is not supported while copying from s3 to blob")
+		}
 		if cooked.blockBlobTier != common.EBlockBlobTier.None() ||
 			cooked.pageBlobTier != common.EPageBlobTier.None() {
 			return cooked, fmt.Errorf("blob-tier is not supported while copying from sevice to service")
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -365,10 +365,10 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 		// Disabling blob tier override, when copying block -> block blob or page -> page blob, blob tier will be kept,
 		// For s3 and file, only hot block blob tier is supported.
 		if cooked.fromTo == common.EFromTo.FileBlob() && cooked.blobType != common.EBlobType.None() {
-			return cooked, fmt.Errorf("file -> blob is not a supported direction while copying from service to service")
+			return cooked, fmt.Errorf("blob-type is not supported while copying from file to blob")
 		}
 		if cooked.fromTo == common.EFromTo.S3Blob() && cooked.blobType != common.EBlobType.None() {
-			return cooked, fmt.Errorf("custom blob type is not supported while copying from s3 to blob")
+			return cooked, fmt.Errorf("blob-type is not supported while copying from s3 to blob")
 		}
 		if cooked.blockBlobTier != common.EBlockBlobTier.None() ||
 			cooked.pageBlobTier != common.EPageBlobTier.None() {
--- a/cmd/copyS2SMigrationBlobEnumerator.go
+++ b/cmd/copyS2SMigrationBlobEnumerator.go
@@ -68,7 +68,7 @@ func (e *copyS2SMigrationBlobEnumerator) enumerate(cca *cookedCopyCmdArgs) error
 				*e.destURL = urlExtension{*e.destURL}.generateObjectPath(fileName)
 			}
 			if cca.blobType == cca.blobType.PageBlob() && blobProperties.ContentLength()%512 != 0 {
-				return errors.New("page blobs must be a modulus of 512 bytes exactly")
+				return errors.New("page blobs must be a multiple of 512 bytes in length")
 			}
 			err := e.createDestBucket(ctx, *e.destURL, nil)
 			if err != nil {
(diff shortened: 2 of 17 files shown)


cmd/copyS2SMigrationBlobEnumerator.go
```diff
@@ -86,6 +90,9 @@ func (e *copyS2SMigrationBlobEnumerator) enumerate(cca *cookedCopyCmdArgs) error
 	// b: https://<blob-service>/containerprefix*/vd/blob
 	if isAccountLevel, containerPrefix := e.srcBlobURLPartExtension.isBlobAccountLevelSearch(); isAccountLevel {
 		glcm.Info(infoCopyFromAccount)
+		if cca.blobType == cca.blobType.PageBlob() {
+			glcm.Info("Copying entire account to page blobs will fail on files without a content length of modulus 512 bytes")
```
JohnRusk: Can we tweak this slightly to say "Copying entire account to page blobs will fail on any files where the content length is not a multiple of 512 bytes."

(I.e. 
1. add the word "any" before "files".  Otherwise user may feel we are _telling_ them that their files _are_ mis-sized. (In reality, we are displaying this message without any knowledge of the file sizes).
2. tweak wording to use multiple instead of modulus
)

cmd/copyS2SMigrationBlobEnumerator.go
```diff
@@ -104,6 +111,9 @@ func (e *copyS2SMigrationBlobEnumerator) enumerate(cca *cookedCopyCmdArgs) error
 		}
 	} else { // Case-3: Source is a blob container or directory
 		glcm.Info(infoCopyFromContainerDirectoryListOfFiles)
+		if cca.blobType == cca.blobType.PageBlob() {
+			glcm.Info("Copying directory to page blobs will fail on files without a content length of modulus 512 bytes")
```
JohnRusk: Same suggested pair of tweaks as above

cmd/copy.go
```diff
@@ -369,12 +369,6 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 			cooked.pageBlobTier != common.EPageBlobTier.None() {
 			return cooked, fmt.Errorf("blob-tier is not supported while copying from sevice to service")
 		}
-		// Disabling blob type override.
-		// i.e. not support block -> append/page, append -> block/page, page -> append/block,
-		// and when file and s3 is source, only block blob destination is supported.
-		if cooked.blobType != common.EBlobType.None() {
```
JohnRusk: Removing this enables Azure Files->Blob with a blob type override.  We don't have tests for that. So we need to either add tests for it, or else have this code fail the job `if cooked.FromTo == common.EFromTo.FileBlob() && cooked.blobType != common.EBlobType.None() `.
IMHO it's fine to just prevent it with the above test, rather than enabling it, since the requirement at present is only Page <-> Block

JohnRusk: Likewise, and perhaps more significantly, removing this enables S3 -> Page blob and S3 -> Append blob. So need to either add tests for that to our existing test suite (possibly in the tests that are implemented in Python, if that happens to be the only place where ew test from S3) OR have this code still fail the job if FromTo == S3Blob
IMHO it's fine to just prevent it with the above test, rather than enabling it, since the requirement at present is only Page <-> Block

JohnRusk: If you resolve the above by actually supporting the combinations (i.e. the "add tests" option), you'll need to remove the word Block from these two lines in cmd/helpMessages.go

```
  - Azure File (SAS) -> Azure Block Blob (SAS or OAuth authentication)
  - AWS S3 (Access Key) -> Azure Block Blob (SAS or OAuth authentication)
```

cmd/zt_copy_s2smigration_test.go
```diff
@@ -578,6 +578,282 @@ func (s *cmdIntegrationSuite) TestS2SCopyFromContainerToContainerNoPreserveBlobT
 	})
 }
 
+func (s *cmdIntegrationSuite) TestS2SCopyFromPageToBlockBlob(c *chk.C) {
```
JohnRusk: IIRC, these tests are effectively unit tests of the enumeration code (i.e. they only hit code under /cmd. They don't hit code under /ste). 
Let's discuss whether to also all end-to-end (integration style) tests in test_service_to_service_copy.py.

cmd/copy.go
zezha-msft: Agreed. We should be careful and not enable options that we haven't tested yet. Given our scale, users will find out the unintended features within a week.

zezha-msft:
This should probably be a new item we need add to the backlog: if the user does an account -> account transfer, shouldn't we tolerate blobs of different types and transfer them to the same type that they already were?

cmd/zt_copy_s2smigration_test.go
adreed-msft: I'm not particularly good at python but I'll take a stab at it.

adreed-msft: Tests *have* been added.

JohnRusk:
> This should probably be a new item we need add to the backlog: if the user does an account -> account transfer, shouldn't we tolerate blobs of different types and transfer them to the same type that they already were?

I've added this to our project backlog @zezha-msft

adreed-msft:
Force-pushed an upgrade to latest `dev` passing CI

cmd/copy.go
```diff
@@ -376,16 +376,16 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 		}
 		// Disabling blob tier override, when copying block -> block blob or page -> page blob, blob tier will be kept,
 		// For s3 and file, only hot block blob tier is supported.
+		if cooked.fromTo == common.EFromTo.FileBlob() && cooked.blobType != common.EBlobType.None() {
+			return cooked, fmt.Errorf("file -> blob is not a supported direction while copying from service to service")
```
JohnRusk: >rf("file -> blob i [](start = 26, length = 18)

I think this message is wrong.  Shouldn't it be similar to the "custom blob type is not supported"  message that's a few lines below?  (I.e. currently its saying you can't do file ->  blob; but IIRC its supposed to say you can't do file-> blob  WITH a custom blob type

cmd/copy.go
```diff
@@ -376,16 +376,16 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
...
+		}
+		if cooked.fromTo == common.EFromTo.S3Blob() && cooked.blobType != common.EBlobType.None() {
+			return cooked, fmt.Errorf("custom blob type is not supported while copying from s3 to blob")
```
JohnRusk: >custom blob type [](start = 30, length = 16)

Our convention s to use the actual parameter names in messages like this.  So the message would be "blob-type is not supported while..." (instead of using the phrase "custom blob type" to mean the blob-type param)

cmd/copy.go
adreed-msft: Yes, you're right, that's a incorrect message.

JohnRusk:
Looks good. Are there new Python tests, yet?  (re comment about the go ones hitting mocks/fakes).

adreed-msft:
Yes. Check https://github.com/Azure/azure-storage-azcopy/pull/372/commits/17c3d7979ec143073b73d9e44becda06146bf6b5 to see the smoke tests.

JohnRusk:
> Yes. Check 17c3d79 to see the smoke tests.

Oh, yeah. Those.  Not sure what I was thinking re missing them!

cmd/zt_scenario_helpers_for_test.go
```diff
@@ -175,6 +175,104 @@ func (scenarioHelper) generateBlobsFromList(c *chk.C, containerURL azblob.Contai
 	time.Sleep(time.Millisecond * 1500)
 }
 
+func (scenarioHelper) generateBlobsFromListForPageCopy(c *chk.C, containerURL azblob.ContainerURL, blobList []string) {
```
zezha-msft: I thought there's already a method that does this? e.g. generateBlobsFromList. You could rename it to generateBlockBlobsFromList, and take in the data to upload (`ForPageCopy` because of the data size right?).

cmd/zt_scenario_helpers_for_test.go
```diff
@@ -175,6 +175,104 @@ func (scenarioHelper) generateBlobsFromList(c *chk.C, containerURL azblob.Contai
...
+			nil)
+		c.Assert(err, chk.IsNil)
+		c.Assert(uResp.StatusCode(), chk.Equals, 201)
+	}
+
+	// sleep a bit so that the blobs' lmts are guaranteed to be in the past
+	time.Sleep(time.Millisecond * 1500)
+}
+
+func (scenarioHelper) generateAppendBlobsFromListForPageCopy(c *chk.C, containerURL azblob.ContainerURL, blobList []string) {
```
zezha-msft: Same here, could leverage the generic method instead.

cmd/zt_scenario_helpers_for_test.go
adreed-msft: Good point. I think I was just in the wrong mindset when writing these tests and ended up duplicating a lot of code.

zezha-msft:
@adreed-msft maybe rebase and try again, the build should be a bit faster now.

:shipit:

adreed-msft:
Rebased

adreed-msft:
Squashing and merging as CI is passing.

Issue #374, opened 2019-05-08
AzCopy fails to download gzip-ed blob that has the ContentEncoding property set to gzip


dterziev:
### Which version of the AzCopy was used? 
``` cmd
> azcopy_windows_amd64.exe --version
azcopy version 10.1.0
```

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.

This command is actually issued by Storage Explorer 1.8.0 when using the bundled version of AzCopy.

```
azcopy_windows_amd64.exe copy "source blob url with SAS" "local path" --overwrite=true --recursive --from-to=BlobLocal --check-md5 "FailIfDifferent"
```

### What problem was encountered?
Download fails the MD5 check with this error in the log file:

``` text
2019/05/08 21:21:47 ERR: [P#0-T#215] DOWNLOADFAILED: <url of the blob here>  : 000 : the MD5 hash of the data, as we received it, did not match the expected value, as found in the Blob/File Service. This means that either there is a data integrity error OR another tool has failed to keep the stored hash up to date. When Checking MD5 hash. X-Ms-Request-Id: 
```

### How can we reproduce the problem in the simplest way?
1. Upload a gzip-ed blob and set the ContentEncoding=gzip property and ContentMD5 to the hash of the compressed blob, this is a C# example:

``` c#
            CloudBlockBlob blob = (get blob reference);
            blob.Properties.ContentType = "application/xml";
            blob.Properties.ContentEncoding = "gzip";

            using (var outputStream = await blob.OpenWriteAsync().ConfigureAwait(false))
            using (var gzipStream = new GZipStream(outputStream, CompressionLevel.Optimal))
            {
                // write the bytes to gzipStream here
            }
```

2. Try downloading the blob from Storage Explorer (v1.8.0) with AzCopy and that will fail.

I suspect that the MD5 validation is performed after the content is decompressed because the HTTP response has the `Content-Encoding: gzip` header. This is, of course, different from the MD5 on the compressed blob.

### Have you found a mitigation/solution?

* Download with `--check-md5 "NoCheck"` - this skips the validation but then the blob is actualy decompressed which is a side-effect that may not be desirable.
* Remove the ContentEncoding property on the blob but then if a normal HTTP client downloads the content it does not decompress it which in my particular case leads to invalid XML file content.

[...]

PR #378, opened 2019-05-12
Prevent Go SDK from automatically decompressing gzipped files

JohnRusk:
Fix for issue #374


JohnRusk:
The test failures were auth failures in the smoke tests, and have been resolved.

PR #382, opened 2019-05-15
Transition azcopy to using go-ieproxy to support Internet Explorer/Edge Proxy pulling

adreed-msft:
This has a related issue under azure-pipeline-go, which should eliminate the original http.GetProxyFromEnvironment.


adreed-msft:
Closing to rebase

PR #399, opened 2019-05-21
Support for custom headers on ADLS Gen 2

adreed-msft:
Enables the support for custom headers on ADLS Gen 2.

The following custom headers are supported:
- content-language
- content-type
- content-disposition
- content-encoding
- cache-control

Manually tested with the following command:
`azcopy cp "C:\Users\v-adreed\foobar" "https://adeledevtestadls1.dfs.core.windows.net/sasauthtest/test" --content-language=fr --content-type=text/random --content-disposition=why --content-encoding=utf8 --cache-control=butwhy`

### DIFF
--- a/azbfs/url_file.go
+++ b/azbfs/url_file.go
@@ -128,16 +128,6 @@ func (f FileURL) GetProperties(ctx context.Context) (*PathGetPropertiesResponse,
 		nil, nil, nil, nil, nil)
 }
 
-func (f FileURL) SetProperties(ctx context.Context, headers BlobFSHTTPHeaders) (*PathUpdateResponse, error) {
-	overrideHttpVerb := "PATCH"
-
-	return f.fileClient.Update(ctx, PathUpdateActionSetProperties, f.fileSystemName, f.path, nil,
-		nil, nil, nil, nil,
-		&headers.CacheControl, &headers.ContentType, &headers.ContentDisposition, &headers.ContentEncoding, &headers.ContentLanguage,
-		nil, nil, nil, nil, nil, nil, nil,
-		nil, nil, nil, &overrideHttpVerb, nil, nil, nil, nil)
-}
-
 // UploadRange writes bytes to a file.
 // offset indiciates the offset at which to begin writing, in bytes.
 // custom headers are not valid on this operation
--- a/azbfs/url_file.go
+++ b/azbfs/url_file.go
@@ -17,9 +17,8 @@ type FileURL struct {
 	path           string
 }
 
-//rename me?
-// BlobFSFileHTTPHeaders represents the set of custom headers available for defining information about the content.
-type BlobFSFileHTTPHeaders struct {
+// BlobFSHTTPHeaders represents the set of custom headers available for defining information about the content.
+type BlobFSHTTPHeaders struct {
 	ContentType        string
 	ContentEncoding    string
 	ContentLanguage    string
@@ -56,7 +55,7 @@ func (f FileURL) WithPipeline(p pipeline.Pipeline) FileURL {
 
 // Create creates a new file or replaces a file. Note that this method only initializes the file.
 // For more information, see https://docs.microsoft.com/en-us/rest/api/storageservices/create-file.
-func (f FileURL) Create(ctx context.Context, headers BlobFSFileHTTPHeaders) (*PathCreateResponse, error) {
+func (f FileURL) Create(ctx context.Context, headers BlobFSHTTPHeaders) (*PathCreateResponse, error) {
 	return f.fileClient.Create(ctx, f.fileSystemName, f.path, PathResourceFile,
 		nil, PathRenameModeNone, nil, nil, nil, nil,
 		&headers.CacheControl, &headers.ContentType, &headers.ContentEncoding, &headers.ContentLanguage, &headers.ContentDisposition,
@@ -129,7 +128,7 @@ func (f FileURL) GetProperties(ctx context.Context) (*PathGetPropertiesResponse,
 		nil, nil, nil, nil, nil)
 }
 
-func (f FileURL) SetProperties(ctx context.Context, headers BlobFSFileHTTPHeaders) (*PathUpdateResponse, error) {
+func (f FileURL) SetProperties(ctx context.Context, headers BlobFSHTTPHeaders) (*PathUpdateResponse, error) {
 	overrideHttpVerb := "PATCH"
 
 	return f.fileClient.Update(ctx, PathUpdateActionSetProperties, f.fileSystemName, f.path, nil,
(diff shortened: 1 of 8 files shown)


azbfs/url_file.go
```diff
@@ -17,6 +17,16 @@ type FileURL struct {
 	path           string
 }
 
+//rename me?
```
JohnRusk: Minor: Our convention is to write things-to-do in the form "TODO" followed by action. In cases like this, can use the word "consider" in the text. E.g.:
// TODO: consider renaming me

azbfs/url_file.go
```diff
@@ -17,6 +17,16 @@ type FileURL struct {
...
+// BlobFSFileHTTPHeaders represents the set of custom headers available for defining information about the content.
+type BlobFSFileHTTPHeaders struct {
```
JohnRusk: If you'd like alternative name suggestions, then looking at the corresponding types in the File and Blob APIs, I'd suggest this one be named BlobFsHTTPHeaders.  (Technically, BlobFSHTTPHeaders would the "correct" capitalization, but that's very hard to read).

cmd/copy.go
```diff
@@ -298,10 +298,6 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	// Example1: for Local to Blob, preserve-last-modified-time flag should not be set to true
 	// Example2: for Blob to Local, follow-symlinks, blob-tier flags should not be provided with values.
 	switch cooked.fromTo {
-	case common.EFromTo.LocalBlobFS():
```
JohnRusk: There were 6 things in this check that's been removed, and only 5 of them relate to this PR. Shouldn't the check remain just for the other one (namely BlobType)?

ste/mgr-JobPartMgr.go
```diff
@@ -538,6 +547,13 @@ func (jpm *jobPartMgr) fileDstData(fullFilePath string, dataFileToXfer []byte) (
 	return azfile.FileHTTPHeaders{ContentType: jpm.inferContentType(fullFilePath, dataFileToXfer), ContentLanguage: jpm.fileHTTPHeaders.ContentLanguage, ContentEncoding: jpm.fileHTTPHeaders.ContentEncoding, ContentDisposition: jpm.fileHTTPHeaders.ContentDisposition, CacheControl: jpm.fileHTTPHeaders.CacheControl}, jpm.fileMetadata
 }
 
+func (jpm *jobPartMgr) bfsDstData(fullFilePath string, dataFileToXfer []byte) (headers azbfs.BlobFSFileHTTPHeaders) {
+	if jpm.planMMF.Plan().DstBlobData.NoGuessMimeType || dataFileToXfer == nil {
+		return jpm.bfsFileHTTPHeaders
+	} //TODO: come back!
```
JohnRusk: What's this comment for?

ste/uploader-blobFS.go
```diff
@@ -171,7 +173,8 @@ func (u *blobFSUploader) Epilogue() {
 	if jptm.TransferStatus() > 0 {
 		md5Hash, ok := <-u.md5Channel
 		if ok {
-			_, err := u.fileURL.FlushData(jptm.Context(), jptm.Info().SourceSize, md5Hash)
+			//Type assertions aren't my favorite thing to do but it does work when you need it to.
+			_, err := u.fileURL.FlushData(jptm.Context(), jptm.Info().SourceSize, md5Hash, jptm.(*jobPartTransferMgr).jobPartMgr.(*jobPartMgr).bfsFileHTTPHeaders)
 			if err != nil {
```
JohnRusk: Looks like this was written before jptm.BfsDstData was added. Now that that function exists, I think this comment can be removed, and the code can be simplified to call jptm.BfsDstData instead of the chain of properties and type assertions. Is that right?

ste/mgr-JobPartMgr.go
adreed-msft: I was going to come back there if something I did elsewhere didn't manage to get HTTP headers through for BlobFS. It seems I have a strong tendency to leave debug info leftover. Should probably be more careful about that and use safer practices to keep track of what I'm doing.

ste/uploader-blobFS.go
adreed-msft: *actually* I was trying to do that. I couldn't get it to work, so I resorted to type assertions. I'll take another shot at it.

cmd/copy.go
JohnRusk: This point appears to be still unresolved?

adreed-msft: Only one should remain, you're right.

ste/uploader-blobFS.go
adreed-msft: OK, so the reason for that pain is that we can't actually obtain the headers because we don't have access to the leading bytes of the local file.

Believe me, I'd love to make this cleaner. But that means we have to have Epilogue actually get the leading bytes in one way or another, which would end up touching a lot more code. BlobFS is the only one that actually *needs* this info, as FlushData overwrites the headers.

adreed-msft: Correction: FlushData overwrites x-ms-content-type. If it were anything else, it wouldn't be as concerning since that's actually what the leading bytes manipulate.

adreed-msft: So, @JohnRusk my question to you is: Should I go through the effort to actually get those leading bytes to jptm.BfsDstData, or should we continue with this one-off hack?

azbfs/url_file.go
```diff
@@ -119,8 +128,19 @@ func (f FileURL) GetProperties(ctx context.Context) (*PathGetPropertiesResponse,
 		nil, nil, nil, nil, nil)
 }
 
+func (f FileURL) SetProperties(ctx context.Context, headers BlobFSHTTPHeaders) (*PathUpdateResponse, error) {
```
JohnRusk: It looks like SetProperties was added because maybe it was going to be used, but in the end its not used. (No usages found at present).  Can we remove it?  (i.e. reverse the addition of it, since its effectively untested code, due to it not being used).

ste/uploader-blobFS.go
JohnRusk: Good point. Yes, we should use BfsDstData because otherwise we won't be consistent with the rest of the codebase. Specifically, FileDstData and BlobDst data respect the `NoGuessMimeType` flag, but the current type-assertion approach does not respect that flag (if I understand correctly).

Anyway, good news: you've already got and used the result of `jptm.BfsDstData` in the Prologue method, right?  And that reflects the leading bytes.  So there are two choices for how to handle this code in the epilogue:

1. Test without setting any headers here at all. (I.e. leave the Flush method passing nils for the things that correspond to header contents).  Maybe, as long as _all 5 of them_ are nil, it will not set any of them, and the values set in the prologue will persist. (IIRC that's consistent with how blob storage would approach the same thing - i.e. if you set one you should set them all, but if you set none, that's fine). OR
1. If the above doesn't work, then use the same approach that uploader-azureFiles uses.  It captures headers that that are used in the prologue (when we do have the leading bytes) and uses them again at the end. (uploader-azureFiles does't have the choice of doing option 1, above, due to a different API signature on the method called in the epilogue).

adreed-msft: Ah, good thought, cache the response. I don't know why that didn't occur to me.

azbfs/url_file.go
adreed-msft: I actually added it to just have parity with the API. Certainly can remove it.

Issue #122, opened 2018-11-14
azcopy sync fails with "AuthenticationErrorDetail: se is mandatory", however the se is definitely included


angrygreenfrogs:
AzCopy 10.0.4-Preview on Ubuntu 14.04

Command:
`azcopy sync 'https://production8858.blob.core.windows.net/wad-iis-logfiles?se=2030-01-01&sp=rl&sv=2018-03-28&sr=c&sig=REDACTED' /backup --recursive`

SAS was generated using:
```
az storage container generate-sas \
    --account-name production8858 \
    --account-key 'redacted' \
    --name wad-iis-logfiles \
    --permissions rl \
    --expiry 2030-01-01
```

The "se" value is definitely included, but azcopy fails.

Result:
```
Failed with error error starting the sync between source https://production8858.blob.core.windows.net/wad-iis-logfiles and destination /backup. 

Failed with error cannot list blobs for download. 

Failed with error -> github.com/Azure/azure-storage-azcopy/vendor/github.com/Azure/azure-storage-blob-go/azblob.NewResponseError, /go/src/github.com/Azure/azure-storage-azcopy/vendor/github.com/Azure/azure-storage-blob-go/azblob/zz_generated_response_error.go:28

===== RESPONSE ERROR (ServiceCode=AuthenticationFailed) =====
Description=Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:c18abdb2-801e-0049-2f14-7c2fba000000
Time:2018-11-14T12:19:11.8402297Z, Details: 
   AuthenticationErrorDetail: se is mandatory. Cannot be empty
   Code: AuthenticationFailed
   GET https://production8858.blob.core.windows.net/wad-iis-logfiles?comp=list&restype=container&sig=REDACTED&sp=rl&sr=c&sv=2018-03-28&timeout=901
   User-Agent: [AzCopy/10.0.4-Preview Azure-Storage/0.3 (go1.10.3; linux)]
   X-Ms-Client-Request-Id: [41df3c4f-57ec-4ac7-7e35-7af1f4d305f0]
   X-Ms-Version: [2018-03-28]
   --------------------------------------------------------------------------------
   RESPONSE Status: 403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
   Access-Control-Allow-Origin: [*]
   Access-Control-Expose-Headers: [Access-Control-Allow-Origin]
   Content-Length: [407]
[...]

PR #401, opened 2019-05-21
SAS Authentication on ADLS Gen 2

adreed-msft:
zc files are from azblob. Enables SAS Authentication on ADLS Gen 2 / dfs.core.windows.net endpoints.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,7 +8,7 @@
 
 ### New features
 
-1. SAS tokens are supported on ADLS Gen 2/dfs.core.windows.net endpoints
+1. SAS tokens are supported on HNS (Hierarchical NameSpace/Azure Data Lakes Generation 2) Storage Accounts
 
 ## Version 10.1.2
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,7 +8,7 @@
 
 ### New features
 
-1. SAS tokens are supported on HNS (Hierarchical NameSpace/Azure Data Lakes Generation 2) Storage Accounts
+1. SAS tokens are supported on HNS (Hierarchical Namespace/Azure Data Lake Generation 2) Storage Accounts
 
 ## Version 10.1.2
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,10 @@
 
 1. AzCopy v10 now outputs a sensible error & warning when attempting to authenticate a storage account business-to-business
 
+### New features
+
+1. SAS tokens are supported on ADLS Gen 2/dfs.core.windows.net endpoints
+
 ## Version 10.1.2
 
 ### Breaking change
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ AzCopy supports two types of authentication. See the table below to see which ty
 | ------------- | ------------- |
 | Blob storage  | SAS tokens OR Azure Active Directory Authentication  |
 | File storage  | SAS tokens  |
-| ADLS Gen2 | Azure Active Directory Authentication |
+| ADLS Gen2 | SAS tokens OR Azure Active Directory Authentication |
 
 > :exclamation::exclamation::exclamation:Note a [SAS token](https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1) is *NOT* an account key. SAS tokens are limited in scope and validity, and starts with a question mark which can be appended to a Blob URL. Here is an example: `?sv=2017-11-09&ss=bf&srt=co&sp=rwac&se=2018-11-16T03:59:09Z&st=2018-11-15T19:59:09Z&sip=10.102.166.17&spr=https,http&sig=k8xSm2K3crBbtNpfoxyvh9n%2BMjDTqRk2XpY8JYIAMaA%3D`.
 
--- a/cmd/copyDownloadBlobFSEnumerator.go
+++ b/cmd/copyDownloadBlobFSEnumerator.go
@@ -30,6 +30,7 @@ func (e *copyDownloadBlobFSEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 
 	// attempt to parse the source url
 	sourceURL, err := url.Parse(cca.source)
+	sourceURL = gCopyUtil.appendQueryParamToUrl(sourceURL, cca.sourceSAS)
 	if err != nil {
 		return errors.New("cannot parse source URL")
 	}
--- a/cmd/copyDownloadBlobFSEnumerator.go
+++ b/cmd/copyDownloadBlobFSEnumerator.go
@@ -30,10 +30,12 @@ func (e *copyDownloadBlobFSEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 
 	// attempt to parse the source url
 	sourceURL, err := url.Parse(cca.source)
-	sourceURL = gCopyUtil.appendQueryParamToUrl(sourceURL, cca.sourceSAS)
 	if err != nil {
 		return errors.New("cannot parse source URL")
 	}
+
+	// append the SAS query to the newly parsed URL
+	sourceURL = util.appendQueryParamToUrl(sourceURL, cca.sourceSAS)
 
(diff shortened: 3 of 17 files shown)


ChangeLog.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## Version XX.XX.XX
+
+### New features
+
+1. SAS tokens are supported on HNS (Hierarchical NameSpace/Azure Data Lakes Generation 2) Storage Accounts
```
JohnRusk: Namespace is normally written with lower cases. 

Lake is singular in Data Lake.  (Better fix, to get product names correct ;-)

cmd/copyDownloadBlobFSEnumerator.go
```diff
@@ -30,6 +30,7 @@ func (e *copyDownloadBlobFSEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 
 	// attempt to parse the source url
 	sourceURL, err := url.Parse(cca.source)
+	sourceURL = gCopyUtil.appendQueryParamToUrl(sourceURL, cca.sourceSAS)
```
JohnRusk: Looks odd with this new line inserted between the Parse call and the if statement that checks the parse result.  Better to put this after the if that checks the parse error.  Oh, and I see there's a comment on gCopyUtil deprecating its use. Maybe use the local util (which is already defined above, and does the same thing).

cmd/credentialUtil.go
```diff
@@ -139,17 +139,28 @@ func getBlobCredentialType(ctx context.Context, blobResourceURL string, canBePub
 // 1. If there is cached session OAuth token, indicating using token credential.
 // 2. If there is OAuth token info passed from env var, indicating using token credential. (Note: this is only for testing)
 // 3. Otherwise use shared key.
```
JohnRusk: Minor: need to either bring this comment into line with the revised code, or remove the comment. Given that BlobFS is more complex for auth options than anything else, retaining and updating the comment seems to make some sense. (Even tho normally we don't want to just restate the code in comment form).

cmd/credentialUtil.go
```diff
@@ -139,17 +139,28 @@ func getBlobCredentialType(ctx context.Context, blobResourceURL string, canBePub
...
+		return common.ECredentialType.Anonymous(), nil
+	}
+
 	name := os.Getenv("ACCOUNT_NAME")
 	key := os.Getenv("ACCOUNT_KEY")
 	if name != "" && key != "" { // TODO: To remove, use for internal testing, SharedKey should not be supported from commandline
 		return common.ECredentialType.SharedKey(), nil
 	} else {
-		return common.ECredentialType.Unknown(), errors.New("OAuth token or shared key should be provided for Blob FS")
+		return common.ECredentialType.Unknown(), errors.New("OAuth token or shared key should be provided for Blob FS (gbfsct)")
```
JohnRusk: This looks like a typo at the end of this line. (Or else a long acronym that I'm not familiar with). Also, the main body of the message is no longer correct. It should cover all three options that are now supporter, OAuth token, SAS Url, or shared key.

cmd/credentialUtil.go
```diff
@@ -139,17 +139,28 @@ func getBlobCredentialType(ctx context.Context, blobResourceURL string, canBePub
...
 		return common.ECredentialType.OAuthToken(), nil
 	}
 
+	sas := azbfs.NewBfsURLParts(*resourceURL).SAS
+
+	if isSASExisted := sas.Signature() != ""; isSASExisted || standaloneSAS {
+		return common.ECredentialType.Anonymous(), nil
+	}
+
 	name := os.Getenv("ACCOUNT_NAME")
```
JohnRusk: The blob code (above) gives preference to SAS token in the cases where both SAS token and OAuth token exists. I think we should do the same here, both for consistency and because if a SAS token is specified, the user (by specifying it) is explicitly opting in to use it.

JohnRusk:
Can we add two smoke tests (in python) that use the SAS tokens? One for upload and one for download.

README.md
```diff
@@ -61,7 +61,7 @@ AzCopy supports two types of authentication. See the table below to see which ty
 | ------------- | ------------- |
 | Blob storage  | SAS tokens OR Azure Active Directory Authentication  |
 | File storage  | SAS tokens  |
-| ADLS Gen2 | Azure Active Directory Authentication |
+| ADLS Gen2 | SAS tokens OR Azure Active Directory Authentication |
```
JohnRusk: Can you please also make the corresponding change (adding SAS token as on option) to line 25 of helpMessages.go

JohnRusk:
Please also add "SAS token" to the auth options listed on line 25 of cmd\helpmessages.go

cmd/credentialUtil.go
adreed-msft: Gah! You caught a leftover debug message again. gbfsct stood for getBlobFSCredentialType and it was to indicate which of the messages I was actually receiving the response from.

I've found myself doing these kinds of things less often since I started, but, it's still a bad habit to debug this way for me.

JohnRusk:
@adreed-msft Did you add smoke tests in Python that use the SAS tokens (one for upload, one for down?)

cmd/credentialUtil.go
JohnRusk: Code has the correct order of precendence (SAS before OAuth) but comment has it around the wrong wany.

---
In reply to: [287934924](https://github.com/Azure/azure-storage-azcopy/pull/401#discussion_r287934924) [](ancestors = 287934924)

adreed-msft:
No, I have not added smoke tests to this yet. Let me get on that.

adreed-msft:
Hi, @JohnRusk, the smoke tests have been added. Please take a look at them, they're pretty simple to understand even without understanding the actual framework.

JohnRusk:
Minor: the comment on getBlobFSCredentialType is inconsistent with the code. (the code is correct, since you corrected it in an earlier commit. Specifically, the code correctly favours SAS tokens over OAuth. But the comment still says the opposite.  Probably worth fixing the comment so it won't confuse folks.

testSuite/scripts/run.py
```diff
@@ -96,7 +96,7 @@ def parse_config_file_set_env():
 
 def check_env_not_exist(key):
     if os.environ.get(key, '-1') == '-1':
-        #print('Environment variable: ' + key + ' not set.')
+        print('Environment variable: ' + key + ' not set.')
```
zezha-msft: LOL 😸

zezha-msft:
@adreed-msft Could you please merge this as soon as possible? I need this change for testing purposes. :shipit:

adreed-msft:
Yeah, I will as soon as I get tests passing. It's being really uncooperative at the moment.

adreed-msft:
Squash and merge, squash and merge

nathankitchen:
I believe there's an issue with this code. The changes to the way that SAS URLs are handled assumes that the date fields st and se are both in the format "yyyy-MM-ddThh:MM.ssZ". However, it's possible to generate valid SAS tokens through supported Azure API's that don't follow that date format.

Inadvertently changing the date format in this way invalidates the SAS signature, making AzCopy fail.

Specifically: https://docs.microsoft.com/en-us/rest/api/storagerp/storageaccounts/listaccountsas

This is used by [ARM templates](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-functions-resource), and I suspect it's true of AZ CLI and possibly PowerShell too.

This API will generate valid SAS tokens using a longer date format:

`sv=2015-04-05&ss=b&srt=o&sp=w&st=2019-01-01T11%3A11%3A11.0000000Z&se=2029-01-01T11%3A11%3A11.0000000Z&spr=https&sig=faXg9BfjJNM6W%2FuanxZ862A6SZoit1CzJVwfDNvDEuw%3D`

Changing the date format to drop the millisecond values changes the signature hash, invalidating the SAS token and resulting in "Authentication Failed".

I think the only way to address this is to back out the work done as part of this branch that normalises the dates, instead handling them just as strings.

This is tracked in open bug #122, and I've added a comment there with the same explanation.

PR #406, opened 2019-05-22
Pull proxy details from internet explorer/edge

adreed-msft:
Making use of `mattn/go-ieproxy`, proxy details (aside from authentication) were pulled from internet explorer/edge.

To achieve this and enable the use of pacfiles, a middleman function was introduced to translate each of the interfaces to the standard Go proxy interface.


JohnRusk:
Currently, the CI build is still on "dep" for dependencies.  So gopkg.toml needs to be updated when we add a new dependency. However, since we are also moving towards module-based builds, go.mod also needs the corresponding update.  Ideally, we'd also update both of the derived files (gopkg.lock and go.sum respectively) and check this in too. I think you can do that  - i.e. update BOTH derived files - if you're running Go 1.12x.  Give me a yell if you want to discuss.

common/proxyMiddleman_windows.go
```diff
@@ -0,0 +1,30 @@
...
+			NoProxy:    conf.Static.NoProxy,
+		}
+
+		return func(req *http.Request) (i *url.URL, e error) {
+			return prox.ProxyFunc()(req.URL)
+		}
+	} else {
+		return http.ProxyFromEnvironment
+	}
+}
```
JohnRusk: Nice!

common/proxyMiddleman_windows.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/mattn/go-ieproxy"
+	"golang.org/x/net/http/httpproxy"
+	"net/http"
+	"net/url"
+)
+
+func proxyMiddleman() func(req *http.Request) (i *url.URL, e error) {
+	conf := ieproxy.GetConf()
```
JohnRusk: I've just realize something: GetConf only seems to cover two of the three cases that Windows supports. Windows supports naming a pac file (which is supported here by conf.Automatic in the return value), by static values (supported by conf.Static) AND by telling Windows "I'm not telling you anything. Just find the proxy for me all by yourself (using DNS and/or DHCP).".   You can see all three options in Window's Proxy Settings window.  But... the last one, of totally automatic config, is not covered by the existing code from the mattn library!  I previously assumed it was, but when I read the code in pac_windows.go I see that its only doing the option when it already has the URL of a pac script. It's not doing the DNS/DHCP option.  I think we need to cover that, otherwise we are not doing everythign that Windows does.

BUT, before going any further on this specific issue, we should figure out a way to TEST with that totally automatic config - since only with some tests setup will we be able to confirm whether we have a problem (I'm fairly sure we do) and whether we can fix it (which is what we most need testability for).

main_windows.go
```diff
@@ -59,3 +61,7 @@ func GetAzCopyAppPath() string {
 	}
 	return azcopyAppDataFolder
 }
+
+func init() {
+	http.DefaultTransport.(*http.Transport).Proxy = common.ProxyMiddleman()
```
JohnRusk: This is just a backup, or cases that actually use the DefaultTransport, right? In general AzCopy creates new Transports, instead of using the default one, and in those sites where it makes new transports, you've put a call to ProxyMiddleman() right?
If that's correct, can you pop a comment here, to explain that?  Since otherwise this line could confuse future maintainers (they may think it accomplishes more than it actually does).

NOTICE
```diff
@@ -0,0 +1,25 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
JohnRusk: We don't do separate notices for each component. There's a thirdPartyNotice.txt file fin the root directory. There will be an existing MIT section there, and we just need to add mattn/go-ieproxy (and, I think, the two copyright lines) to that section, following the existsing pattern that's already in that file. then this file NOTICE should be deleted.

JohnRusk:
@adreed-msft Please also update thirdpartynotice.txt with acknowledgement of adreed-mstf/go-ieproxy (even if the project is by a MSFT employee, it still needs to get acknowledged in there). And because its based on the mattn code, please leave in place the mattn acknowledgement that you've added.

JohnRusk:
> Looking good!
> 
> I wondered whether it was best to keep referencing the adreed/go-ieproxy fork, or to switch all our references back to mattn/go-ieproxy fork (now that the merge from the fork is completed), but I think I'm leaning towards staying on the fork. E.g. maybe we'll want to do something in the fork for authentication one day.

On second thoughts, the purpose of the fork was to test and stage the enhancement. That's done now.  So maybe better to go back to directly using the mattn original.

adreed-msft:
> > Looking good!
> > I wondered whether it was best to keep referencing the adreed/go-ieproxy fork, or to switch all our references back to mattn/go-ieproxy fork (now that the merge from the fork is completed), but I think I'm leaning towards staying on the fork. E.g. maybe we'll want to do something in the fork for authentication one day.
> 
> On second thoughts, the purpose of the fork was to test and stage the enhancement. That's done now. So maybe better to go back to directly using the mattn original.

Yeah, that's why I shifted back.

main_windows.go
```diff
@@ -59,3 +61,8 @@ func GetAzCopyAppPath() string {
 	}
 	return azcopyAppDataFolder
 }
+
+func init() {
+	//Catch everything that uses http.DefaultTransport with ieproxy.GetProxyFunc()
+	http.DefaultTransport.(*http.Transport).Proxy = ieproxy.GetProxyFunc()
```
zezha-msft: nice

adreed-msft:
Squashing and merging.

PR #417, opened 2019-05-27
Update README.md

thomasmaurer:
fixed installation path. Before it was fixed on the version folder, now it takes everything higher than 10.*


JohnRusk:
I changed the base branch to dev, to be consistent with our usual flow of changes.  It will hit master when we next do a release. (I know that, since this is documentation only, we could have safely pushed it into master "early". Just seemed cleaner to follow the same flow for everything).

Issue #398, opened 2019-05-21
Slice bounds out of range when writing to drive root


agentmorris:
### Which version of the AzCopy was used? 

10.1.2

### Which platform are you using? (ex: Windows, Mac, Linux)

Windows 

### What command did you run?

azcopy cp "https://blob.url" ".\destination_file"

### What problem was encountered?

I am running azcopy from g:\, so ".\" refers to "g:\".  I have write permissions to this drive.  When running it this way, I get:

Job 03660234-d484-ef41-560b-b18a7c3fe358 has started
Log file is located at: C:\Users\dmorris/.azcopy/03660234-d484-ef41-560b-b18a7c3fe358.log

0 Done, 0 Failed, 1 Pending, 0 Skipped, 1 Total, panic: runtime error: slice bounds out of range

goroutine 233 [running]:
github.com/Azure/azure-storage-azcopy/common.CreateParentDirectoryIfNotExist(0xc042966000, 0x11, 0xc0421b6710, 0xc0425a9ad0)
        /go/src/github.com/Azure/azure-storage-azcopy/common/writeThoughFile.go:30 +0x128
github.com/Azure/azure-storage-azcopy/common.CreateFileOfSizeWithWriteThroughOption(0xc042966000, 0x11, 0x1a33f3c800, 0x0, 0xc35200, 0x0, 0x0)
        /go/src/github.com/Azure/azure-storage-azcopy/common/writeThoughFile_windows.go:34 +0x40
github.com/Azure/azure-storage-azcopy/ste.remoteToLocal(0xcadf80, 0xc0429ab720, 0xc94c60, 0xc0421b66c0, 0xc04201a280, 0xc35208)
        /go/src/github.com/Azure/azure-storage-azcopy/ste/xfer-remoteToLocal.go:107 +0xca9
github.com/Azure/azure-storage-azcopy/ste.parameterizeDownload.func1(0xcadf80, 0xc0429ab720, 0xc94c60, 0xc0421b66c0, 0xc04201a280)
        /go/src/github.com/Azure/azure-storage-azcopy/ste/xfer.go:60 +0x6a
github.com/Azure/azure-storage-azcopy/ste.(*jobPartMgr).StartJobXfer(0xc0431e8000, 0xcadf80, 0xc0429ab720)
        /go/src/github.com/Azure/azure-storage-azcopy/ste/mgr-JobPartMgr.go:519 +0x6c
[...]

Issue #376, opened 2019-05-10
UNC path longer than 260 characters


Eric-ITaaS:
### Which version of the AzCopy was used?
10.1.1 
### Which platform are you using? (ex: Windows, Mac, Linux)
Windows 2008 R2 server

### What command did you run?
azcopy "\\\\server\share" "https://blobinfo" --recursive=true
### What problem was encountered?
chokes on UNC paths greater than 260 characters.

### How can we reproduce the problem in the simplest way?
Try with a UNC path longer than 260 characters

### Have you found a mitigation/solution?
No. I tried to prepend "\\\\?\UNC\" to the path but got this error
Cannot perform sync due to error: illegal local path, no pattern matching allowed for sync command

JohnRusk:
The issue is not unique to UNC paths.  We are actively working on long path support.

Eric-ITaaS:
Okay, thanks. Any idea when this will get fixed?

JohnRusk:
A wild guess would be that it's a few weeks away.  But don't count on that guess being accurate.  If we find that it's easier than anticipated, and that it makes sense to release it "early" rather than wait and release it with a bunch of other changes, we may release it earlier.  If we have significant difficulties in getting it working, it may take longer. Sorry about the uncertainty!

Eric-ITaaS:
No problem, I saw that 10.1.2 was out but I couldn't find any notes as to what was fixed/changed from 10.1.1.

JohnRusk:
Actually, it looks like the issue probably is unique to UNC paths. From my reading of the Go SDK source code, it should work fine if you just map a drive. (ie same path length, but not unc)

Eric-ITaaS:
Yes, works fine from a mapped drive. However I need to script this to run via a scheduled task. There isn't a way to expose a mapped drive in a script when not logged in, hence the use of a UNC path.

JohnRusk:
Oh, I see.  

BTW, I do wonder if there may be some way to expose a mapped drive to a script that is not logged in.  I've seen it done with a call to `New-PSDrive` in the script. (Probably need the `-Persist` parameter to make it work with AzCopy, and probably need to clean up/remove any old mapping to the same letter right at the very start of the script using, `Remove-PSDrive -name "..." -Force`)

I suspect it can also be done with "net use" (maybe) and I've also heard people saying they'd done it by just creating a symbolic link: https://superuser.com/a/651015/   The symbolic link option is probably easiest, if it works for you.

[...]

PR #422, opened 2019-05-30
Enable long file paths usage on Windows & patch issues with using . shortcut for current working directory

### LINKED ISSUE #104: On Windows, AzCopy Cannot Handle Long File Paths
10.0.3-preview

Windows

`.\azcopy.exe copy "..\..\1234567891234567981234567981321321654684354645313216843513213223111141445
45474567271271727231731324175417234173241732417324173aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaabbb" "https://REDACTED.blob.core.windows.net/uploadtohere?REDACTED"`

On upload, AzCopy failed with the error: failed to perform copy command due to error: cannot start job due to error: cannot find source to upload.
For download, it seems like AzCopy can handle slightly longer paths, but there is still a point where it fails.
Create a file, whose full path is at least 250ish characters long. Try to transfer it with AzCopy.
No.

adreed-msft:
Manually tested for all upload & download scenarios. Mixes with wildcard characters.

Fixes issues #104 #376 and #398.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/Azure/azure-pipeline-go/pipeline"
+	"path/filepath"
 
 	"io"
 	"net/url"
@@ -148,14 +149,49 @@ func (raw rawCopyCmdArgs) blockSizeInBytes() uint32 {
 func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	cooked := cookedCopyCmdArgs{}
 
-	fromTo, err := validateFromTo(raw.src, raw.dst, raw.fromTo) // TODO: src/dst
+	var err error
+	cooked.fromTo, err = validateFromTo(raw.src, raw.dst, raw.fromTo) // TODO: src/dst
 	if err != nil {
 		return cooked, err
 	}
+
+	//Convert local paths to long filepaths on windows, get absolute filepath everywhere.
+	//This ensures that long file paths work on Windows and that relative filepaths using . work everywhere.
+	if cooked.fromTo.From() == common.ELocation.Local() {
+		raw.src, err = filepath.Abs(raw.src)
+		if err != nil {
+			return cooked, err
+		}
+
+		if common.OS_PATH_SEPARATOR == `\` {
+			raw.src = common.ToLongPath(raw.src)
+			common.AZCOPY_PATH_SEPARATOR_STRING, common.AZCOPY_PATH_SEPARATOR_CHAR = `\`, '\\'
+			if fi, err := os.Stat(raw.src); err == nil {
+				if fi.IsDir() && !strings.HasSuffix(raw.src, `\`) {
+					raw.src += `\`
+				}
+			}
+		}
+	}
+	if cooked.fromTo.To() == common.ELocation.Local() {
+		raw.dst, err = filepath.Abs(raw.dst)
+		if err != nil {
+			return cooked, err
+		}
+
+		if common.OS_PATH_SEPARATOR == `\` {
+			raw.dst = common.ToLongPath(raw.dst)
+			common.AZCOPY_PATH_SEPARATOR_STRING, common.AZCOPY_PATH_SEPARATOR_CHAR = `\`, '\\'
+			if fi, err := os.Stat(raw.dst); err == nil {
+				if fi.IsDir() && !strings.HasSuffix(raw.dst, `\`) {
+					raw.dst += `\`
+				}
+			}
+		}
+	}
+
 	cooked.source = raw.src
 	cooked.destination = raw.dst
-
-	cooked.fromTo = fromTo
 
 	// copy&transform flags to type-safety
 	cooked.recursive = raw.recursive
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -26,8 +26,6 @@ import (
 	"errors"
 	"fmt"
 	"github.com/Azure/azure-pipeline-go/pipeline"
-	"path/filepath"
-
 	"io"
 	"net/url"
 	"os"
@@ -158,35 +156,15 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	//Convert local paths to long filepaths on windows, get absolute filepath everywhere.
 	//This ensures that long file paths work on Windows and that relative filepaths using . work everywhere.
 	if cooked.fromTo.From() == common.ELocation.Local() {
-		raw.src, err = filepath.Abs(raw.src)
+		raw.src, err = common.PreparePath(raw.src)
 		if err != nil {
 			return cooked, err
-		}
-
-		if common.OS_PATH_SEPARATOR == `\` {
-			raw.src = common.ToLongPath(raw.src)
-			common.AZCOPY_PATH_SEPARATOR_STRING, common.AZCOPY_PATH_SEPARATOR_CHAR = `\`, '\\'
-			if fi, err := os.Stat(raw.src); err == nil {
-				if fi.IsDir() && !strings.HasSuffix(raw.src, `\`) {
-					raw.src += `\`
-				}
-			}
 		}
 	}
 	if cooked.fromTo.To() == common.ELocation.Local() {
-		raw.dst, err = filepath.Abs(raw.dst)
+		raw.dst, err = common.PreparePath(raw.dst)
 		if err != nil {
 			return cooked, err
-		}
-
-		if common.OS_PATH_SEPARATOR == `\` {
-			raw.dst = common.ToLongPath(raw.dst)
-			common.AZCOPY_PATH_SEPARATOR_STRING, common.AZCOPY_PATH_SEPARATOR_CHAR = `\`, '\\'
-			if fi, err := os.Stat(raw.dst); err == nil {
-				if fi.IsDir() && !strings.HasSuffix(raw.dst, `\`) {
-					raw.dst += `\`
-				}
-			}
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


cmd/copy.go
```diff
@@ -148,15 +149,50 @@ func (raw rawCopyCmdArgs) blockSizeInBytes() uint32 {
...
+
+		if common.OS_PATH_SEPARATOR == `\` {
+			raw.src = common.ToLongPath(raw.src)
+			common.AZCOPY_PATH_SEPARATOR_STRING, common.AZCOPY_PATH_SEPARATOR_CHAR = `\`, '\\'
+			if fi, err := os.Stat(raw.src); err == nil {
+				if fi.IsDir() && !strings.HasSuffix(raw.src, `\`) {
+					raw.src += `\`
+				}
+			}
+		}
```
JohnRusk: How about factoring this out into a "preparePath" method, so that it doesn't need to be duplicated below?

cmd/copy.go
```diff
@@ -148,15 +149,50 @@ func (raw rawCopyCmdArgs) blockSizeInBytes() uint32 {
...
+
+		if common.OS_PATH_SEPARATOR == `\` {
+			raw.src = common.ToLongPath(raw.src)
+			common.AZCOPY_PATH_SEPARATOR_STRING, common.AZCOPY_PATH_SEPARATOR_CHAR = `\`, '\\'
+			if fi, err := os.Stat(raw.src); err == nil {
+				if fi.IsDir() && !strings.HasSuffix(raw.src, `\`) {
+					raw.src += `\`
+				}
+			}
```
JohnRusk: Nice.

cmd/copy.go
```diff
@@ -148,15 +149,50 @@ func (raw rawCopyCmdArgs) blockSizeInBytes() uint32 {
...
+	if cooked.fromTo.From() == common.ELocation.Local() {
+		raw.src, err = filepath.Abs(raw.src)
+		if err != nil {
+			return cooked, err
+		}
+
+		if common.OS_PATH_SEPARATOR == `\` {
+			raw.src = common.ToLongPath(raw.src)
+			common.AZCOPY_PATH_SEPARATOR_STRING, common.AZCOPY_PATH_SEPARATOR_CHAR = `\`, '\\'
+			if fi, err := os.Stat(raw.src); err == nil {
```
JohnRusk: I'm nervous about changing these.  I don't fully understand what Ze had in mind when he set these constants up; but I do seem to recall an assumption that they would in fact be constant and unchanging.  I.e. they didn't mean "what is the path separator that the current OS uses". They specifically meant "what is the path separator that AzCopy uses (regardless of the underyling OS)."

The difference may be significant because on the cloud side the separator is always / (i.e;. the old, hard-code value)

cmd/copyDownloadBlobEnumerator.go
```diff
@@ -154,7 +154,10 @@ func (e *copyDownloadBlobEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 				}
 				// check for the special character in blob relative path and get path without special character.
 				blobRelativePath = util.blobPathWOSpecialCharacters(blobRelativePath)
-
+				// force path to be AZCOPY_PATH_SEPARATOR_STRING, only really important on Windows.
+				if common.AZCOPY_PATH_SEPARATOR_STRING == `\` {
+					blobRelativePath = strings.Replace(blobRelativePath, "/", common.AZCOPY_PATH_SEPARATOR_STRING, -1)
+				}
```
JohnRusk: As noted elsewhere, I feel uneasy about changing AZCOPY_PATH_SEPARATOR_*.  Can you discuss with Ze please?

cmd/copyUploadEnumerator.go
```diff
@@ -25,11 +25,19 @@ func (e *copyUploadEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 	common.PanicIfErr(err)
 
 	// list the source files and directories
-	listOfFilesAndDirectories, err := filepath.Glob(cca.source)
+	// filepath.Glob incorrectly handles the long file paths indicator as a meta character, so trim it...
+	listOfFilesAndDirectories, err := filepath.Glob(common.ToShortPath(cca.source))
 	if err != nil || len(listOfFilesAndDirectories) == 0 {
 		return fmt.Errorf("cannot find source to upload")
 	}
```
JohnRusk: Does this mean that the code will fail here if the root path (cca.source) is itself a long path?

It looks like we can do long paths that are inside cca.source, but we can't do a long cca.source itself.  Is that right?

cmd/copyUtil.go
```diff
@@ -324,6 +324,12 @@ func (util copyHandlerUtil) isPathALocalDirectory(pathString string) bool {
 
 func (util copyHandlerUtil) generateLocalPath(directoryPath, fileName string) string {
 	var result string
+
+	//Because \ and / can't be used interchangeably with the long file path discriminator (\\?\), filter forward slash out
+	if strings.HasPrefix(directoryPath, `\\?\`) {
```
JohnRusk: >\\?\` [](start = 38, length = 5)

Can we have a constant defined for this, and the use that constant everywhere instead of the literal string?  Then, we can put a comment on the constant explaining it to folks who aren't familiar with Window's approach to long paths.

cmd/sync.go
```diff
@@ -100,6 +102,44 @@ func (raw *rawSyncCmdArgs) cook() (cookedSyncCmdArgs, error) {
...
+		}
+
+		if common.OS_PATH_SEPARATOR == `\` {
+			raw.src = common.ToLongPath(raw.src)
+			common.AZCOPY_PATH_SEPARATOR_STRING, common.AZCOPY_PATH_SEPARATOR_CHAR = `\`, '\\'
+			if fi, err := os.Stat(raw.src); err == nil {
+				if fi.IsDir() && !strings.HasSuffix(raw.src, `\`) {
+					raw.src += `\`
+				}
+			}
```
JohnRusk: Needs to call perparePath function suggested elsewhere

common/LongPathHandler.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func ToShortPath(long string) string {
+	if os.PathSeparator == '\\' {
+		if strings.HasPrefix(long, `\\?\UNC`) {
+			return strings.Replace(`\`+long[7:], `\`, `/`, -1)
+		} else if strings.HasPrefix(long, `\\?\`) {
+			return strings.Replace(long[4:], `\`, `/`, -1)
+		}
+	}
```
JohnRusk: This is changing not just the prefx, but also the direction the /, right?  I find that a bit confusing.  It might be just a namming thing.  E.g. maybe call ToLongPath something like "ToWindowsLongPath" and toShortPath something like "ToAzCopyShortPath" or "ToUnixShortPath" or soemthing. That might make it more understandable. Or alternatively, separate the concepts of long/short (i.e;. prefixed/unprefixed) vs / and \.   But i think the naming change will be the easier fix.

common/LongPathHandler_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+func (p *pathHandlerSuite) TestShortToLong(c *chk.C) {
+	if OS_PATH_SEPARATOR == `\` {
+		c.Assert(ToLongPath(`C:\myPath`), chk.Equals, `\\?\C:\myPath`)
+		c.Assert(ToLongPath(`\\myHost\myPath`), chk.Equals, `\\?\UNC\myHost\myPath`)
+		c.Assert(ToLongPath(`\\?\C:\myPath`), chk.Equals, `\\?\C:\myPath`)
+		c.Assert(ToLongPath(`\\?\UNC\myHost\myPath`), chk.Equals, `\\?\UNC\myHost\myPath`)
+	} else {
+		c.Skip("Test only pertains to Windows.")
+	}
```
JohnRusk: If LongPathHandler continues to also do replacements between / and \, then these tests should cover that.

cmd/copy.go
adreed-msft: Actually, Ze originally suggested shifting to this and seeing if there were any regressions. There were not.

cmd/copyDownloadBlobEnumerator.go
adreed-msft: As noted previously, Ze originally suggested this. It reduced a significant amount of code duplication, and reduced the amount of "hacks" I had to do in order to make it work.

cmd/copyUploadEnumerator.go
adreed-msft: Yes, filepath.Glob reads the long file path indicator's `?` as a meta character and fails to glob.

common/LongPathHandler.go
```diff
@@ -0,0 +1,56 @@
+package common
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+//ToExtendedPath converts short paths to an extended path.
```
JohnRusk: Probably worth a comment here to explain that, as per  [https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#maximum-path-length-limitation](https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#maximum-path-length-limitation), "File I/O functions in the Windows API convert "/" to "\" as part of converting the name to an NT-style name, except when using the "\\?\" prefix as detailed in the following sections." THerefore, when converting to an extended path we MUST convert to use standard Windows path separators

common/LongPathHandler.go
```diff
@@ -0,0 +1,56 @@
...
+func ToExtendedPath(short string) string {
```
JohnRusk: Can you add  general comment please, to explain why we are not using the built-in Go long path support. E.g. "We are not using the built-in Go long path support (that was introduced to the Go SDK in commit 231aa9d6d7) for two reasons:
(a) the main reason is that it does not support UNC paths, and we want to support those as best we can. (we have a known limitation when globbing, see globWrapper(), but apart from that, we believe we fully support long UNC paths with the approach we have taken)
(b) the secondary reason is that the Go SDK fix did not cover the lower-level OS calls that we open files with. "

JohnRusk:
IMPORTANT: If we do end up changing the slash format that is persistent into plan files, we will have to change the version number for plan files.

common/LongPathHandler.go
adreed-msft: I brain farted on both of these entirely. I don't need to fix the direction of the `/`, and I also need to return a `\` to the UNC path.

adreed-msft:
Closing PR to redo implementation in just-in-time conversion style

PR #426, opened 2019-05-31
AzCopy Readme restructure

normesta:
Hello - We now have a rich set of AzCopy v10 content on docs. We are working hard to keep that up to date. This PR removes duplicate help content that is starting to fall out of date with the core content. Having two sources of truth will start causing confusion with our customers. John Rusk is in support of a restructure. This version of the readme drives visibility to docs and strips the content down to just feature declarations and FAQ. The readme might also be a great place to keep release notes and temporary bugs and workarounds that our dev team can directly add and update without needing a writer involved.


README.md
```diff
@@ -1,229 +1,67 @@
...
-```
-./azcopy cp /data/directory1 "https://myaccount.blob.core.windows.net/mycontainer/directory1?sastokenhere" --recursive=true
-```
+:white_check_mark: Upload files and folders.
 
-The following command will upload all files directly under `directory1` without recursing into sub-directories, to the path at `https://myaccount.blob.core.windows.net/mycontainer/directory1`.
-```
-./azcopy cp /data/directory1/* "https://myaccount.blob.core.windows.net/mycontainer/directory1?sastokenhere"
-```
+:white_check_mark: Download files and folders.
```
zezha-msft: minor: consider using "directory" instead of "folder"

README.md
```diff
@@ -1,229 +1,67 @@
...
+For complete guidance, visit any of these articles on the docs.microsoft.com website.
 
-Copying data between two Azure Storage accounts make use of the PutBlockFromURL API, and does not use the client machine's network bandwidth. Data is copied between two Azure Storage servers. AzCopy simply orchestrates the copy operation.
-```
-./azcopy cp "https://myaccount.blob.core.windows.net/?sastokenhere" "https://myotheraccount.blob.core.windows.net/?sastokenhere" --recursive=true
-```
+:eight_spoked_asterisk: [Get started with AzCopy](https://docs.microsoft.com/azure/storage/common/storage-use-azcopy-v10)
 
-### Advanced Use Cases
+:eight_spoked_asterisk: [Transfer data with AzCopy and blob storage](storage-use-azcopy-blobs.md)
```
zezha-msft: these don't appear to be valid links yet, are these supposed to point to the docs website?

README.md
```diff
@@ -1,229 +1,67 @@
...
 
-```
-export https_proxy=<user>:<pass>@<proxy IP>:<proxy port>
-# or with a domain:
-export https_proxy=<domain>%5C<user>:<pass>@<proxy IP>:<proxy port>
-```
+To see a list of commands, type `azcopy` and then press the ENTER key.
 
-### Configure log location
+To learn about a specific command, type `azCopy` followed by the name of the command.
```
zezha-msft: typo: `azCopy` should be `azcopy`

zezha-msft: should be `azcopy cp --help` or `azcopy cp -h` instead. If `azcopy command` is called without any args, an error message is printed.

README.md
```diff
@@ -1,229 +1,67 @@
...
 
-Set the environment variable 'AZCOPY_LOG_LOCATION' to a directory of your choice where there is plenty of disk space as logs for large data transfers may use up Gigabytes of space depending on the chosen log level.
+![AzCopy command help example](readme-command-prompt.png)
```
zezha-msft: please consider using `azcopy list -h` instead

README.md
```diff
@@ -1,229 +1,67 @@
...
 
-## Troubleshooting and Reporting Issues
+## Frequently asked questions
 
-### Check Logs for errors
+### What is the difference between `sync` and `copy`?
 
-AzCopy creates a log file for all the jobs. Look for clues in the logs to understand the problem. AzCopy will print UPLOADFAILED, COPYFAILED, and DOWNLOADFAILED strings for failures with the paths along with the error reason.
+The `copy` command is a simple transferring operation, it scans the source and attempts to transfer every single file/blob. The supported source/destination pairs are listed in the help message of the tool. On the other hand, `sync` makes sure that whatever is present in the source will be replicated to the destination, and also whatever is not at the source will be deleted from the destination. If your goal is to simply move some files, then `copy` is definitely the right command, since it offers much better performance.
```
zezha-msft: Perhaps it's better to clarify that by default sync leaves the extra files at the destination alone. The user can choose to delete them.

README.md
```diff
@@ -1,229 +1,67 @@
...
 
-     cat 04dc9ca9-158f-7945-5933-564021086c79.log | grep -i UPLOADFAILED
+For `sync`, last modified times are used to determine whether to transfer the same file present at both the source and the destination. If the use case is to incrementally transfer data
```
zezha-msft: nice!

README.md
normesta: Checking on this one as far as our style guide.

normesta: Ok - changed from "folder" to "directory".

normesta:
Thanks for all the feedback @zezha-msft ! If these changes address your concerns, can you please merge? Thx!

zezha-msft:
I'll let @JohnRusk take a look too. We generally do 2 reviews per change. 😄

JohnRusk:
FYI @normesta, we normally put release notes into the ChangeLog.md file, so those won't come into the Readme. At least I don't think they will.

README.md
```diff
@@ -1,229 +1,67 @@
...
-set https_proxy=<proxy IP>:<proxy port>
-# For Linux:
-export https_proxy=<proxy IP>:<proxy port>
-# For MacOS
-export https_proxy=<proxy IP>:<proxy port>
-```
+### Find help from your command prompt
 
-For proxy authentication, use the following format:
+Be sure to add AzCopy to your system path. Then, open a command prompt window.
```
JohnRusk: I'm in two minds about this. Technically, adding it to the path is not strictly necessary. But, in PowerShell at least, if its not in the path you need to invoke it as .\azcopy instead of just azcopy.

As a user, I never like being told I _have_ to add something to my path. But I can see how it simplifies the rest of the doc if they do add it. What do you think?

README.md
JohnRusk: Yes. Maybe use a followup section for this (similar to the existing one about overwrite).  the sync one could be title "Will sync delete files from the destination?" and would explain the relevant paramter.

---
In reply to: [289576263](https://github.com/Azure/azure-storage-azcopy/pull/426#discussion_r289576263) [](ancestors = 289576263)

README.md
normesta: Thanks John. Perhaps we can tell them to "optionally" add this to their path. With a path variable in place, the use of "azcopy" at the command prompt becomes universal. ".\azcopy" is Windows PowerShell specific. This won't work for a standard command prompt and I'm not sure about Linux environments. Perhaps we can add a note to the core docs content that if they don't add this to their path, and they are in a PowerShell environment, they will have to preface with ".\"?

README.md
normesta: Ok Hows this?

README.md
normesta: Ok - I added some text for the optional system path and for PowerShell prompts.

normesta:
@zezha-msft I think we are just waiting on your approval

zezha-msft:
Thanks @normesta!!

Issue #221, opened 2019-02-20
Do we need to write logs to `~/.azcopy` directory?


ppanyukov:
### Which version of the AzCopy was used? 
/azcopy_linux_amd64_10.0.7/azcopy

### Which platform are you using? (ex: Windows, Mac, Linux)
Linux

### What problem was encountered?
Logs are written to `~/.azcopy` directory as per:

```
Log file is located at: /home/ppanyukov/.azcopy/99599e85-8ef6-d844-595e-ca9c12a8f762.log
```

I would say this is not what programs on Linux normally do. Certainly writing logs to a dot directory in user's home feels very unorthodox.

My preference would be to write to `stdout` and `stderr` as appropriate by default, with a possible option to specify `--log` parameter or similar to override this (not even sure this is necessary as I can easily redirect `stdout/stderr` to wherever if I need to).

What are people's thoughts on this?

----

Possible hacky workaround for this: run `azcopy` in a docker container, this will make sure logs don't get stored anywhere locally for a long time. On the other hand, if logs *are* required, this running in docker surfaces other issues.

zezha-msft:
Hi @ppanyukov, thanks for reaching out!

I understand and appreciate your perspective. However, we'd prefer to save the log as a file, since there could be a lot of useful information, if any transfer ever goes wrong. If we accepted a --log-file-location flag, what if the user doesn't provide one? One option would be to save the log file in the current directory, from where the user is running the command; but this is not a great practice either since the user could be in the directory were they are trying to upload data from, and putting the log file there is undesirable.

You could customize the log location. Please refer to `./azcopy env`.

ppanyukov:
Well, I'm sure this is done out of best intentions and I agree it's a good thing to have logs, but also programs for Linux should probably strive to fit into what is customary for Linux.  As I pointed out, I think it is more standard to write to `stdout/stderr` rather than in dot directories in user's home such as `~/.azcopy`.

Why not take `rsync` as a model for a tool that does pretty much the same kind of job and is well established?  That is not to say `rsync` is ideal, it's what people probably used and are likely to be familiar with.

BTW, I just discovered there is also `~/.azcopy/plans` directory which keeps growing.  Is this something that needs to be kept after the job finished?

Oh, and if we had `--log-file` parameter or env var `AZCOPY_LOG_FILE_LOCATION` (rather than directory), we could give it `/dev/stderr` and that might kinda solve this.

zezha-msft:
[...]

PR #433, opened 2019-06-07
Clean up plans & logs on success w/ flag & fix --log-level none

adreed-msft:
I know I technically wasn't scheduled to handle this, but it was genuinely trivial, I'm waiting on CR updates, and only took me about an hour.

This handles #221. `--log-level none` drops no logs whatsoever, and there is an option provided to clean up both log and plan files upon a fully successful transfer.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 1. AzCopy v10 now outputs a sensible error & warning when attempting to authenticate a storage account business-to-business
+1. `--log-level=none` now drops no logs, and has a listing in `--help`
 
 ## Version 10.1.2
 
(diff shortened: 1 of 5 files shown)


Virepri:
Oh hey, this is probably going to need a data schema version update. And I probably should set the cleanup settings to carry over when a job gets revived if it fails.

JohnRusk:
I vaguely recall that there were some good reasons (I can't recall the details) why cleanup immediately after success was not something that we have wanted to do so far. We were leaning towards a different approach that didn't do the cleanup _immediately_.

How about you simplify this PR to just focus on log level NONE (to make it work) and to update the in-app help for the log-level command to actually list the valid values?  Then, we can tackle cleanup in a separate PR, after discussions with @zezha-msft .

adreed-msft:
Sounds good to me. Would it be optimal to handle moving the plans along with the cleanup?

JohnRusk:
I think we're already tracking "plan _move_" as a separate issue (separate to logs and separate to cleanup)

PR #434, opened 2019-06-10
Fix --log-level=none, provide documentation

adreed-msft:
As requested by @JohnRusk, this PR was slimmed down to only include the fix for `--log-level=none`

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 1. AzCopy v10 now outputs a sensible error & warning when attempting to authenticate a storage account business-to-business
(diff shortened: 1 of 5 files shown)


ChangeLog.md
```diff
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 1. AzCopy v10 now outputs a sensible error & warning when attempting to authenticate a storage account business-to-business
+1. `--log-level=none` now drops no logs, and has a listing in `--help`
```
zezha-msft: minor: consider rephrasing `drops no logs` to  something like `does not create any log file`.

PR #438, opened 2019-06-12
Flush as we go on ADLS Gen 2

adreed-msft:
**NOTE** this is not a finished PR. This is purely here to gather thoughts.

@JohnRusk I know you "discovered" this-- wanted to see what you think of this implementation. I know it eats a goroutine but I attempted to make it sleep as much as possible.

@zezha-msft I'd like to see what you think as well.

Here are my concerns:
1. We were suggested to flush either after a certain data size or a certain number of blocks. I'm happy to switch between them.
2. How should we init bfsTrackerTodo? Check the comment below it for details.
3. Any other concerns from you guys?

I don't intend this to make it into the next release, so get to this at your leisure.

### DIFF
--- a/azbfs/url_file.go
+++ b/azbfs/url_file.go
@@ -165,14 +165,10 @@ func (f FileURL) AppendData(ctx context.Context, offset int64, body io.ReadSeeke
 
 // flushes writes previously uploaded data to a file
 // The contentMd5 parameter, if not nil, should represent the MD5 hash that has been computed for the file as whole
-func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byte, headers BlobFSHTTPHeaders) (*PathUpdateResponse, error) {
+func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byte, headers BlobFSHTTPHeaders, retainUncommited bool, closeParameter bool) (*PathUpdateResponse, error) {
 	if fileSize < 0 {
 		panic("fileSize must be >= 0")
 	}
-
-	// hardcoded to be false for the moment
-	// azcopy does not need this
-	retainUncommittedData := false
 
 	var md5InBase64 *string = nil
 	if len(contentMd5) > 0 {
@@ -185,14 +181,9 @@ func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byt
 	// See similar todo, with larger comments, in AppendData
 	overrideHttpVerb := "PATCH"
 
-	// We only call FlushData once the whole body is appended, so close can be safely set to true all the time
-	// It sends a notification on the service end, which doesn't concern AzCopy, so no need to make it configurable
-	// note that we should make it configurable if this SDK is ever split off and become its own thing
-	closeParameter := true
-
 	// TransactionalContentMD5 isn't supported currently.
 	return f.fileClient.Update(ctx, PathUpdateActionFlush, f.fileSystemName, f.path, &fileSize,
-		&retainUncommittedData, &closeParameter, nil, nil,
+		&retainUncommited, &closeParameter, nil, nil,
 		&headers.CacheControl, &headers.ContentType, &headers.ContentDisposition, &headers.ContentEncoding, &headers.ContentLanguage,
 		md5InBase64, nil, nil, nil, nil, nil, nil, nil,
 		nil, nil, &overrideHttpVerb, nil, nil, nil, nil)
--- a/azbfs/url_file.go
+++ b/azbfs/url_file.go
@@ -165,7 +165,7 @@ func (f FileURL) AppendData(ctx context.Context, offset int64, body io.ReadSeeke
 
 // flushes writes previously uploaded data to a file
 // The contentMd5 parameter, if not nil, should represent the MD5 hash that has been computed for the file as whole
-func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byte, headers BlobFSHTTPHeaders, retainUncommited bool, closeParameter bool) (*PathUpdateResponse, error) {
+func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byte, headers BlobFSHTTPHeaders, retainUncommittedData bool, closeFile bool) (*PathUpdateResponse, error) {
 	if fileSize < 0 {
 		panic("fileSize must be >= 0")
 	}
@@ -183,7 +183,7 @@ func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byt
 
 	// TransactionalContentMD5 isn't supported currently.
 	return f.fileClient.Update(ctx, PathUpdateActionFlush, f.fileSystemName, f.path, &fileSize,
-		&retainUncommited, &closeParameter, nil, nil,
+		&retainUncommittedData, &closeFile, nil, nil,
 		&headers.CacheControl, &headers.ContentType, &headers.ContentDisposition, &headers.ContentEncoding, &headers.ContentLanguage,
 		md5InBase64, nil, nil, nil, nil, nil, nil, nil,
 		nil, nil, &overrideHttpVerb, nil, nil, nil, nil)
--- a/ste/uploader-blobFS.go
+++ b/ste/uploader-blobFS.go
@@ -23,6 +23,7 @@ package ste
 import (
 	"context"
 	"fmt"
+	"math"
 	"net/url"
 	"os"
 	"time"
@@ -169,18 +170,26 @@ func (u *blobFSUploader) GenerateUploadFunc(id common.ChunkID, blockIndex int32,
 	})
 }
 
+var flushThreshold int64 = (1024 ^ 3) * 20 //20 GB flush threshold
+
 func (u *blobFSUploader) Epilogue() {
 	jptm := u.jptm
 
 	// flush
 	if jptm.TransferStatus() > 0 {
 		md5Hash, ok := <-u.md5Channel
+		ss := jptm.Info().SourceSize
 		if ok {
-			//Type assertions aren't my favorite thing to do but it does work when you need it to.
-			_, err := u.fileURL.FlushData(jptm.Context(), jptm.Info().SourceSize, md5Hash, *u.creationTimeHeaders)
-			if err != nil {
-				jptm.FailActiveUpload("Flushing data", err)
-				// don't return, since need cleanup below
+			for i := int64(math.Min(float64(ss), float64(flushThreshold))); i <= ss; i = int64(math.Min(float64(ss), float64(i+flushThreshold))) {
+				_, err := u.fileURL.FlushData(jptm.Context(), i, md5Hash, *u.creationTimeHeaders)
+				if err != nil {
+					jptm.FailActiveUpload("Flushing data", err)
+					// don't return, since need cleanup below
+				}
+
+				if i == ss {
+					break
+				}
 			}
 		} else {
 			jptm.FailActiveUpload("Getting hash", errNoHash)
--- a/ste/uploader-blobFS.go
+++ b/ste/uploader-blobFS.go
@@ -140,7 +140,7 @@ func (u *blobFSUploader) RemoteFileExists() (bool, error) {
 func (u *blobFSUploader) Prologue(state common.PrologueState) {
 	jptm := u.jptm
 
-	u.flushThreshold = int64(u.chunkSize * 7500)
+	u.flushThreshold = int64(u.chunkSize * ADLSFlushThreshold)
 
 	h := jptm.BfsDstData(state.LeadingBytes)
 	u.creationTimeHeaders = &h
@@ -182,7 +182,7 @@ func (u *blobFSUploader) Epilogue() {
 		md5Hash, ok := <-u.md5Channel
 		if ok {
 			// Flush incrementally to avoid timeouts on a full flush
-			for i := int64(math.Min(float64(ss), float64(u.flushThreshold))); i <= ss; i = int64(math.Min(float64(ss), float64(i+u.flushThreshold))) {
+			for i := int64(math.Min(float64(ss), float64(u.flushThreshold))); ; i = int64(math.Min(float64(ss), float64(i+u.flushThreshold))) {
 				// Close only at the end of the file, keep all uncommitted data before then.
 				_, err := u.fileURL.FlushData(jptm.Context(), i, md5Hash, *u.creationTimeHeaders, i != ss, i == ss)
 				if err != nil {
--- a/ste/uploader-blobFS.go
+++ b/ste/uploader-blobFS.go
@@ -139,9 +139,8 @@ func (u *blobFSUploader) RemoteFileExists() (bool, error) {
 func (u *blobFSUploader) Prologue(state common.PrologueState) {
 	jptm := u.jptm
 
-	if !thresholdSetup {
-		setupThreshold(int64(u.chunkSize))
-		thresholdSetup = true
+	if flushThreshold == 0 {
+		flushThreshold = int64(u.chunkSize * 7500)
 	}
 
 	h := jptm.BfsDstData(state.LeadingBytes)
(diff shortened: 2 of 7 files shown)


zezha-msft:
Hi, thanks for taking a stab at this!

My two concerns:

1. Spending one extra go routine per file is probably not a good idea, as we are already limiting number of workers due to CPU usage. 
2. Per the latest communication, doing all the flushes at the end is also acceptable. 

The alternative approach is much simpler to implement and cheaper in resource usage. So I don't recommend the current approach.

adreed-msft:
>Spending one extra go routine per file

Nope, it's just one total. It tracks all transfers. See `sync.Once`.

adreed-msft:
Also

> Doing all the flushes at the end is also acceptable


I feel... Uncomfortable about that. I know it's alright but you really have to wonder how large the actual buffer is. Are we going to get a customer transferring a file large enough to overflow it?

It's ridiculous to ask this question but I am a bit concerned about it.

ste/uploader-blobFS.go
```diff
@@ -25,13 +25,56 @@ import (
...
+
+	Currently, 5000 tracker entries may fill up and cause transfer upload funcs to end up blocking until they can push.
+	This *may* not actually degrade perf because we're trying to upload multiple chunks at a time
+	However, I also know that we do a limited number of goroutines (I think).
+*/
+
+func bfsTrackerRoutine() {
+	for bfsu := range bfsTrackerTodo { //The routine will sleep when we try to pull from the range, letting others do work
+		for ; bfsu.ti < len(bfsu.tp) && bfsu.tp[bfsu.ti].Completed; bfsu.ti++ {
+			if bfsu.ti == -1 {
```
JohnRusk: Must check jptm status for the bfsu here, since status may have changed to cancelled since the value was pushed to the channel.

ste/uploader-blobFS.go
```diff
@@ -136,6 +185,9 @@ func (u *blobFSUploader) RemoteFileExists() (bool, error) {
 }
 
 func (u *blobFSUploader) Prologue(state common.PrologueState) {
+	//Start bfsTracker before anything else, but only once.
+	go bfsTrackerOnce.Do(bfsTrackerRoutine)
```
JohnRusk: I have the opposite view to Ze on this one ;-)  I'm not comfortable with us just running one Go routine for all uploaders. It's architecturally different from everything else that we do, seems less understandable, and harder to reason about failure scenarios.  For downloads, we already run one extra GR per file; I don't see why we couldn't do the same for uploads.

ste/uploader-blobFS.go
```diff
@@ -166,14 +218,17 @@ func (u *blobFSUploader) GenerateUploadFunc(id common.ChunkID, blockIndex int32,
 			jptm.FailActiveUpload("Uploading range", err)
 			return
 		}
+
+		u.tp[blockIndex].Completed = true
```
JohnRusk: I'm not particularly comfortable with one GR writing Completed and another reading it.   It's really straying outside of what's consider safe multi-threading (multi GR) practice. In this case, I think we'd probably get away with it, because the reader would eventually read the right thing, even if it does't read it quite as soon as we expect.  But... I'm be a lot more comfortable if we had 1 GR per file; and what we pushed to the channel was the  id.OffsetInFile, and then the GR that reads the channnel would maintain sate however it whished (e.g. probabl3y with an array that it has as a local variable).

ste/uploader-blobFS.go
```diff
@@ -25,13 +25,56 @@ import (
...
 	"os"
+	"sync"
 	"time"
 
 	"github.com/Azure/azure-pipeline-go/pipeline"
 	"github.com/Azure/azure-storage-azcopy/azbfs"
 	"github.com/Azure/azure-storage-azcopy/common"
 )
 
+const bfsTrackerThreshold = (1024 ^ 3) * 10 // 10 GB
```
JohnRusk: Let's go for 20 GB. (10 GB is only 10 seconds worth or so, at our target speeds).

ste/uploader-blobFS.go
```diff
@@ -166,14 +218,17 @@ func (u *blobFSUploader) GenerateUploadFunc(id common.ChunkID, blockIndex int32,
...
+		bfsTrackerTodo <- u //ping bfsTracker about progress
 	})
 }
 
 func (u *blobFSUploader) Epilogue() {
 	jptm := u.jptm
 
 	// flush
-	if jptm.TransferStatus() > 0 {
+	/*if jptm.TransferStatus() > 0 {
```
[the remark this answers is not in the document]
Virepri: Actually, the tracker already does the final flush.

It would be a double flush if we did this.

ste/uploader-blobFS.go
```diff
@@ -25,13 +25,56 @@ import (
...
+		for ; bfsu.ti < len(bfsu.tp) && bfsu.tp[bfsu.ti].Completed; bfsu.ti++ {
+			if bfsu.ti == -1 {
+				continue
+			}
+
+			id := bfsu.tp[bfsu.ti].ID
+			if id.OffsetInFile-bfsu.lf >= bfsTrackerThreshold || bfsu.ti+1 == len(bfsu.tp) {
+				bfsu.lf = id.OffsetInFile //update last flush position
+				md5, ok := <-bfsu.md5Channel
+				if ok {
```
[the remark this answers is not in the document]
Virepri: Oh shoot! OK, I thought the md5 came before the file and I guess I missed returning it to the channel.

I'll go ahead and adjust that.

zezha-msft:
> Nope, it's just one total. It tracks all transfers. See sync.Once.

Oops, sorry 😜 I read too fast.

zezha-msft:
> I feel... Uncomfortable about that. I know it's alright but you really have to wonder how large the actual buffer is. Are we going to get a customer transferring a file large enough to overflow it?

I'm a bit confused here. 

My point is that the logic/code change is simpler if we flushed at the end. Since we know the total size, we can issue flush calls one after the other until the entire file range is flushed. The amount of time it takes to flush does increase, but its proportion to the entire file transfer time is still extremely small (given how large the files are).

The other thing is that the ADLS Gen2 code is bound to be short-lived. Once the inter-op is GA, we'd simply delete all the ADLS Gen2 code from our repo, and talk to the blob endpoint instead. This is the other reason why I do not want to take on any additional complexity.

ste/uploader-blobFS.go
```diff
@@ -25,13 +25,56 @@ import (
...
+
+func bfsTrackerRoutine() {
+	for bfsu := range bfsTrackerTodo { //The routine will sleep when we try to pull from the range, letting others do work
+		for ; bfsu.ti < len(bfsu.tp) && bfsu.tp[bfsu.ti].Completed; bfsu.ti++ {
+			if bfsu.ti == -1 {
+				continue
+			}
+
+			id := bfsu.tp[bfsu.ti].ID
+			if id.OffsetInFile-bfsu.lf >= bfsTrackerThreshold || bfsu.ti+1 == len(bfsu.tp) {
```
JohnRusk: As far as I can tell, this doesn't handle the case where the last chunk completes before, say, the second-to-last.  But... if we reinstate the code that runs in the Epilogue, then that's OK, because the epilogue can do a final flush

JohnRusk: Oops, no I was mistaken on the previous comment re last chunk completing before second to last.. Will delete it .

ste/uploader-blobFS.go
JohnRusk: Correction, I think we would be OK, but only because we'd be relying on a side effect of channel operations that they would, in effect, guarantee the thread safety of .Completed. But, the Go documentation on that page suggests that, if you have to rely on that, you probably shouldn't.

adreed-msft:
@JohnRusk @zezha-msft per both of you, the entire file is now flushed at the end, divided into 20GB chunks.

ste/uploader-blobFS.go
```diff
@@ -169,18 +170,26 @@ func (u *blobFSUploader) GenerateUploadFunc(id common.ChunkID, blockIndex int32,
...
 		md5Hash, ok := <-u.md5Channel
+		ss := jptm.Info().SourceSize
 		if ok {
-			//Type assertions aren't my favorite thing to do but it does work when you need it to.
-			_, err := u.fileURL.FlushData(jptm.Context(), jptm.Info().SourceSize, md5Hash, *u.creationTimeHeaders)
-			if err != nil {
-				jptm.FailActiveUpload("Flushing data", err)
-				// don't return, since need cleanup below
+			for i := int64(math.Min(float64(ss), float64(flushThreshold))); i <= ss; i = int64(math.Min(float64(ss), float64(i+flushThreshold))) {
+				_, err := u.fileURL.FlushData(jptm.Context(), i, md5Hash, *u.creationTimeHeaders)
```
zezha-msft: the md5 hash shouldn't be sent with every request, right? since it's an overall hash of the entire file

JohnRusk: It's probably not doing any harm, because its put into the x-ms-content-md5 field (which is the whole of file one).   However, there's no benefit in setting it (or the other headers) on any call except the last.

ste/uploader-blobFS.go
```diff
@@ -169,18 +170,26 @@ func (u *blobFSUploader) GenerateUploadFunc(id common.ChunkID, blockIndex int32,
...
 	if jptm.TransferStatus() > 0 {
 		md5Hash, ok := <-u.md5Channel
+		ss := jptm.Info().SourceSize
 		if ok {
-			//Type assertions aren't my favorite thing to do but it does work when you need it to.
-			_, err := u.fileURL.FlushData(jptm.Context(), jptm.Info().SourceSize, md5Hash, *u.creationTimeHeaders)
-			if err != nil {
-				jptm.FailActiveUpload("Flushing data", err)
-				// don't return, since need cleanup below
+			for i := int64(math.Min(float64(ss), float64(flushThreshold))); i <= ss; i = int64(math.Min(float64(ss), float64(i+flushThreshold))) {
```
JohnRusk: Can you add a comment to say something like "Flush incrementally to avoid timeouts that can happen if flush big file all in one go"

JohnRusk: + follow up on impact analysis as discussed offline.

adreed-msft:
@JohnRusk @zezha-msft could you guys shoot me another review on this?

ste/uploader-blobFS.go
```diff
@@ -138,6 +139,11 @@ func (u *blobFSUploader) RemoteFileExists() (bool, error) {
 func (u *blobFSUploader) Prologue(state common.PrologueState) {
 	jptm := u.jptm
 
+	if !thresholdSetup {
```
JohnRusk: Prologue only runs once.  So, unless I'm missing something, you don't need `thresholdSetup`  Does that answer your question in the comment below?

ste/uploader-blobFS.go
```diff
@@ -169,22 +175,44 @@ func (u *blobFSUploader) GenerateUploadFunc(id common.ChunkID, blockIndex int32,
...
 
+var flushThreshold int64 = (1024 * 1024 * 1024) * 20 //20GB
+//TODO: Should a flag be present for the threshold?
+var thresholdSetup bool
+
+func setupThreshold(blockSize int64) {
+	//Find the closest multiple of the block size to the flush threshold
+	m := flushThreshold % blockSize
+
+	if math.Abs(float64(m)) < math.Abs(float64((flushThreshold-m+blockSize)-flushThreshold)) {
```
JohnRusk: This line looks odd. Is it just saying if m < blocksize/2?

ste/uploader-blobFS.go
```diff
@@ -169,22 +175,44 @@ func (u *blobFSUploader) GenerateUploadFunc(id common.ChunkID, blockIndex int32,
...
+	}
+}
+
 func (u *blobFSUploader) Epilogue() {
 	jptm := u.jptm
 
 	// flush
 	if jptm.TransferStatus() > 0 {
 		md5Hash, ok := <-u.md5Channel
+		ss := jptm.Info().SourceSize
```
JohnRusk: Minor (and probably don't bother fixing it in this case). Stylistically, we'd normally reverse the order of the md5Hash and ss := lines, so that the if ok test was immediately after the thing that sets ok.

ste/uploader-blobFS.go
```diff
@@ -169,22 +175,44 @@ func (u *blobFSUploader) GenerateUploadFunc(id common.ChunkID, blockIndex int32,
...
+			// Flush incrementally to avoid timeouts on a full flush
+			for i := int64(math.Min(float64(ss), float64(flushThreshold))); i <= ss; i = int64(math.Min(float64(ss), float64(i+flushThreshold))) {
+				// Close only at the end of the file, keep all uncommitted data before then.
+				_, err := u.fileURL.FlushData(jptm.Context(), i, md5Hash, *u.creationTimeHeaders, i != ss, i == ss)
+				if err != nil {
+					jptm.FailActiveUpload("Flushing data", err)
+					break // don't return, since need cleanup below
+				}
+
+				if i == ss {
```
JohnRusk: Do you need this. Doesn't the for loop already take care of termination condition?

ste/uploader-blobFS.go
```diff
@@ -169,22 +175,44 @@ func (u *blobFSUploader) GenerateUploadFunc(id common.ChunkID, blockIndex int32,
...
 }
 
+var flushThreshold int64 = (1024 * 1024 * 1024) * 20 //20GB
+//TODO: Should a flag be present for the threshold?
+var thresholdSetup bool
+
+func setupThreshold(blockSize int64) {
+	//Find the closest multiple of the block size to the flush threshold
+	m := flushThreshold % blockSize
+
```
JohnRusk: Actually... I just realized something. If you check the most recent email we got from the ADLS Gen 2 team, you'll see that they it depends more on the block count, that the size of data.  So, it would be more correct to set flush threshold to something like blocksize * 7500.   That will have the good side effect of giving us much larger flush sizes on large files (since we use larger block sizes there).  I calculate that this may reduce our number of flush calls by a factor of 35 in some cases!  So probably worth doing.  Sorry I didn't think of it in the earlier review!

azbfs/url_file.go
```diff
@@ -165,15 +165,11 @@ func (f FileURL) AppendData(ctx context.Context, offset int64, body io.ReadSeeke
 
 // flushes writes previously uploaded data to a file
 // The contentMd5 parameter, if not nil, should represent the MD5 hash that has been computed for the file as whole
-func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byte, headers BlobFSHTTPHeaders) (*PathUpdateResponse, error) {
+func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byte, headers BlobFSHTTPHeaders, retainUncommited bool, closeParameter bool) (*PathUpdateResponse, error) {
```
zezha-msft: Minor: retainUncommited could named retainUncommittedData to clarify meaning

ste/uploader-blobFS.go
```diff
@@ -169,22 +175,44 @@ func (u *blobFSUploader) GenerateUploadFunc(id common.ChunkID, blockIndex int32,
 	})
 }
 
+var flushThreshold int64 = (1024 * 1024 * 1024) * 20 //20GB
```
zezha-msft: block size might vary for each file, right? so this should perhaps not be a global variable

ste/uploader-blobFS.go
zezha-msft: Good point.

Block size might vary for each file, right? so if we rely on block size to figure out the threshold, then `flushThreshold` should perhaps not be a global variable

ste/uploader-blobFS.go
adreed-msft: Oh, I thought it was once per file. That's fair.

ste/uploader-blobFS.go
adreed-msft: Actually... That's a really good point and I don't know what I was thinking.

adreed-msft: I was trying to find which block the flushThreshold was closest to, but, you just pointed out a way better solution.

ste/uploader-blobFS.go
adreed-msft: That's fair. I'll rework some of this to flush on a multiple of blockSize (or the end of the file, whichever comes sooner)

ste/uploader-blobFS.go
adreed-msft: Yes, it is actually necessary. I wrote this to exit the for loop only *after* firing one last time at the end of the file. If I dropped this, the loop would never exit, as `i` can never be more than `ss`. As such, we can have this or minor code duplication.

ste/uploader-blobFS.go
```diff
@@ -138,6 +139,10 @@ func (u *blobFSUploader) RemoteFileExists() (bool, error) {
 func (u *blobFSUploader) Prologue(state common.PrologueState) {
 	jptm := u.jptm
 
+	if flushThreshold == 0 {
+		flushThreshold = int64(u.chunkSize * 7500)
```
zezha-msft: chunkSize can be different for each file, right? flushThreshold should not be a global variable.

zezha-msft:
Please make sure the existing CI tests are indeed covering the change.

ste/uploader-blobFS.go
JohnRusk: Oh, I see. 
Could the for loop have therefore been written like this?  (note the "; ;" in the middle)

`for i := int64(math.Min(float64(ss), float64(u.flushThreshold))); ; i = int64(math.Min(float64(ss), float64(i+u.flushThreshold))) `

I'd suggest that such a formulation would show your intent more clearly, and be more maintainable.

JohnRusk: BTW, I do like the fact that this test here, at the end of the loop, makes it very clear that we are only exiting exactly when i == ss.  That makes it easier to read and understand that we really do always reach the exact end of the file.

adreed-msft:
Squashing and merging.

PR #439, opened 2019-06-13
Allow fractional block size for copy and sync

JohnRusk:
For the (rare) cases where folks need sizes smaller than 1 MiB


adreed-msft:
(FYI: Corrected the conflict between this & dev)

cmd/zt_block_size_to_bytes_test.go
```diff
@@ -0,0 +1,54 @@
...
+	chk "gopkg.in/check.v1"
+)
+
+type blockSizeFilterSuite struct{}
+
+var _ = chk.Suite(&blockSizeFilterSuite{})
+
+func (s *genericFilterSuite) TestConversions(c *chk.C) {
+
+	testData := []struct {
```
adreed-msft: Minor nitpick: Inline struct definition.

Doesn't super duper matter since it's never used again but I'm also pretty sure this isn't idiomatic

cmd/copy.go
```diff
@@ -140,8 +141,27 @@ func (raw *rawCopyCmdArgs) parsePatterns(pattern string) (cookedPatterns []strin
...
+		return 0, errors.New("negative block size not allowed")
+	}
+	rawSizeInBytes := rawBlockSizeInMiB * 1024 * 1024 // internally we use bytes, but users' convenience the command line uses MiB
+	if rawSizeInBytes > math.MaxUint32 {
+		return 0, errors.New("block size too big for uint32")
+	}
+	const epsilon = 0.001 // arbitrarily using a tolerance of 1000th of a byte
+	_, frac := math.Modf(rawSizeInBytes)
+	isWholeNumber := frac < epsilon || frac > 1.0-epsilon // frac is very close to 0 or 1, so rawSizeInBytes is (very close to) an integer
+	if !isWholeNumber {
```
adreed-msft: Minor: This is a good check to have, but I think it'd also be good to have a flag to override it.

cmd/zt_block_size_to_bytes_test.go
JohnRusk: I copied it out of Dave Cheney's blog post on table driven tests in Go, so I guess that qualifies it as idiomatic ;-)   Will leave it here, but I don't think I use it outside of tests tho.

cmd/copy.go
JohnRusk: Override what, specifically? Checking that the resulting number of bytes is a whole number?  What would be an example of a time when someone would want to override that?

cmd/sync.go
```diff
@@ -507,7 +508,7 @@ func init() {
 
 	rootCmd.AddCommand(syncCmd)
 	syncCmd.PersistentFlags().BoolVar(&raw.recursive, "recursive", true, "true by default, look into sub-directories recursively when syncing between directories.")
-	syncCmd.PersistentFlags().Uint32Var(&raw.blockSizeMB, "block-size-mb", 0, "use this block size (specified in MiB) when uploading to/downloading from Azure Storage. Default is automatically calculated based on file size.")
+	syncCmd.PersistentFlags().Float64Var(&raw.blockSizeMB, "block-size-mb", 0, "use this block size (specified in MiB) when uploading to/downloading from Azure Storage. Default is automatically calculated based on file size.")
```
zezha-msft: missing "Decimal fractions are allowed - e.g. 0.25")"

cmd/zt_block_size_to_bytes_test.go
```diff
@@ -0,0 +1,54 @@
+// Copyright © 2017 Microsoft <wastore@microsoft.com>
```
zezha-msft: Minor: just started noticing this, should we change the year to 2019 for new files?

The year says 2017 for the oldest files generated back in December 2017 (with Cobra).

cmd/zt_block_size_to_bytes_test.go
```diff
@@ -0,0 +1,54 @@
...
+		floatMiB         float64
+		expectedBytes    uint32
+		expectedErrorMsg string
+	}{
+		{100, 100 * 1024 * 1024, ""},
+		{1, 1024 * 1024, ""},
+		{0.25, 256 * 1024, ""},
+		{0.000030517578125, 32, ""}, // 32 bytes, extremely small case
+		{-1, 0, "negative block size not allowed"},
+		{0.333, 0, "while fractional numbers of MiB are allowed as the block size, the fraction must result to a whole number of bytes. 0.333000000000 MiB resolves to 349175.808 bytes"},
```
zezha-msft: nice test.

cmd/copy.go
adreed-msft: Eh, fair point.

cmd/zt_block_size_to_bytes_test.go
JohnRusk: It's not clear that updates are strictly necessary. E.g. https://softwareengineering.stackexchange.com/questions/130478/what-copyright-date-for-an-update-to-an-open-source-project-from-last-year . I'll email with more details offline.

PR #440, opened 2019-06-13
Fixed bug where piping was not picking up the service version override

zezha-msft:


cmd/root.go
```diff
@@ -137,7 +129,7 @@ func detectNewVersion() {
 		return
 	}
 
-	blobBody := blobStream.Body(azblob.RetryReaderOptions{MaxRetryRequests: downloadMaxTries})
+	blobBody := blobStream.Body(azblob.RetryReaderOptions{MaxRetryRequests: ste.MaxRetryPerDownloadBody})
```
adreed-msft: Minor: Weren't we trying to get rid of the download options (per ste/xfer.go:40)?

zezha-msft: hmmm I don't have any context on that comment unfortunately

PR #441, opened 2019-06-14
Set up CI with Azure Pipelines

zezha-msft:


azure-pipelines.yml
```diff
@@ -0,0 +1,36 @@
...
+          GOARCH=386 GOOS=windows go build -o "$(Build.ArtifactStagingDirectory)/azcopy_windows_386.exe"
+        displayName: 'Generate builds'
+
+      - task: PublishBuildArtifacts@1
+        displayName: 'Publish Artifacts'
+        condition: succeededOrFailed()
+
+  - job: MacOS_Build
+    pool:
+      vmImage: 'xcode9-macos10.13'
```
adreed-msft: Wait, can't we safely compile this without a macos VM?

Unless `common/credCache_darwin.go`'s dependencies require it. I recall that being an issue.

Feel free to disregard if that's the case.

azure-pipelines.yml
```diff
@@ -0,0 +1,36 @@
+trigger:
```
adreed-msft: Is there no way to trigger tests on PRs?

Or are we going to continue using Travis for that?

azure-pipelines.yml
zezha-msft: Indeed unfortunately.

azure-pipelines.yml
zezha-msft: This triggers for master and all PRs targeting it. I'll see if we should enable dev too.

zezha-msft:
@adreed-msft  @JohnRusk This changed is ready to be merged, please take a look! 😃

adreed-msft:
Thanks for pinging me, I'll take a look Monday morning.

cmd/zt_test.go
```diff
@@ -73,15 +73,20 @@ const (
...
-	for i := 0; !strings.Contains(name, "Suite"); i++ { // The tests are all scoped to the suite, so this ensures getting the actual test name
-		f = runtime.FuncForPC(pc[i])
-		name = f.Name()
+	// The following lines step up the stack find the name of the test method
+	// Note: the way to do this changed in go 1.12, refer to release notes for more info
+	var pcs [10]uintptr
+	n := runtime.Callers(1, pcs[:])
+	frames := runtime.CallersFrames(pcs[:n])
+	name := "TestFoo" // default stub "Foo" is used if anything goes wrong with this procedure
+	for {
```
adreed-msft: Yeah breaking is probably a better way to do this than what we've got in azure-storage-blob-go. We should get that changed over there when we find a convenient time.

zezha-msft: Sounds good. The next change we do there can include this.

adreed-msft: Actually, this is really silly in comparison to what we've got in azblob.

```for f, next := frames.Next(); next; f, next = frames.Next() {
		name = f.Function
		if strings.Contains(name, "Suite") {
			break
		}
	}```

Consider this version of the loop over in azblob.

adreed-msft: We're doing roughly the exact same thing, just this one has an exit condition inside the loop itself, rather than in the initialization of the for loop.

PR #447, opened 2019-06-17
Convert to extended paths just in time

adreed-msft:
Take 3. This touched a non-trivial amount of code and I'm not especially proud of it.

I literally coated as much as I feasibly could in `common.ToExtendedPath`.

I've tested relative paths, wildcards, the combination of the two, extended file paths, UNC, and the combination of any number of these things.

This is probably the worst PR you'll ever have to go through.


zezha-msft:
Closing it as it will be address post-refactoring. 😄

Issue #410, opened 2019-05-23
Can't copy a single file to a sub folder in Azure Files


maroony:
### Which version of the AzCopy was used? 
10.1.2

### Which platform are you using? (ex: Windows, Mac, Linux)
Linux, WSL

### What command did you run?
`azcopy cp file.txt "${fileStorageUrl}${storage}/folder/subfolder/?${sasToken}"`

### What problem was encountered?
log output: `RESPONSE Status: 400 The specifed resource name contains invalid characters.`

### How can we reproduce the problem in the simplest way?
see the command above

### Have you found a mitigation/solution?
no

JohnRusk:
Possibly related to (but not identical to) #397

JohnRusk:
@maroony Thanks for logging this.  One possible workaround you might try (maybe a long shot) but worth trying, is to try the same command but without the "/" at the end of "subfolder". I.e. like this:

```
azcopy cp file.txt "${fileStorageUrl}${storage}/folder/subfolder?${sasToken}"
```
I don't know yet exactly where the bug is, so I don't know if that may avoid it or not, but its worth trying if you have time.

maroony:
@JohnRusk If you remove the last "/", the `file.txt` will copied to the folder `folder` *and* renamed to `subfolder`. So a possible workarround would be

```
azcopy cp file.txt "${fileStorageUrl}${storage}/folder/subfolder/file.txt?${sasToken}"
```

In my scenario I can't you wildcards.

JohnRusk:
Good point. Thanks!

adreed-msft:
Hi there, this was patched in #451 and is now present in AzCopy v10.2.x. Enjoy!

Issue #397, opened 2019-05-20
File with <no name> name is created when copying local folder with only one file inside


mkozioro:
### What command did you run?
azcopy copy \
    "/some_dir/*.gz" \
    "https://${ACCOUNT_NAME}.blob.core.windows.net/${CONTAINER_NAME}/some_dir/?${STORAGE_ACCOUNT_SAS}"

When /some_dir/ has only one file with .gz extension, uploaded file has <no name> name. I expect file name to be name of .gz file. When multiple .gz files exists in folder, names are correct.

![image](https://user-images.githubusercontent.com/8501148/58033640-1e11cc00-7b25-11e9-9744-79c8c9bf9d82.png)

adreed-msft:
Hi there, this was patched in #451 and has made it into AzCopy v10.2.x. Enjoy!

elsing:
When using the older service API for use with Azure Stack Hub - this bug once again appears. Is this something that can be fixed easily? 

<img width="264" alt="Screenshot 2021-12-24 at 08 27 23" src="https://user-images.githubusercontent.com/5830702/147334553-ba9341c2-eae9-45b9-8af4-91debd45d806.png">

https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-storage-transfer?view=azs-2108&tabs=az1#azcopy

PR #451, opened 2019-06-18
Fix single wildcard match uploads

adreed-msft:
We were initially going to just create a failing test for this and get to it in the refactor. *HOWEVER*, I discovered it was such a trivial fix that I just went right through with it. Best to have it in the June release, anyway.

If you're wondering, no, this change actually doesn't break plan files. It indirectly FIXES this functionality on past versions when resuming.

If you're curious as to _what_ was even happening before, consider the following request:

`PUT https://<account>.blob.core.windows.net/<container>/test/?timeout=901`

The service was inferring our file name as `test`, and thus plopping it under `test/test`. With this change, the request then becomes

`PUT https://<account>.blob.core.windows.net/<container>/test/<file>?timeout=901`

Fixes the following issues: #410, #397

### DIFF
--- a/cmd/copyUploadEnumerator.go
+++ b/cmd/copyUploadEnumerator.go
@@ -49,7 +49,7 @@ func (e *copyUploadEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 				return fmt.Errorf("for the use of include flag, source needs to be a directory")
 			}
 			// append file name as blob name in case the given URL is a container
-			if (e.FromTo == common.EFromTo.LocalBlob() && util.urlIsContainerOrShare(destinationURL)) ||
+			if (e.FromTo == common.EFromTo.LocalBlob() && util.urlIsContainer(destinationURL)) ||
 				(e.FromTo == common.EFromTo.LocalFile() && util.urlIsAzureFileDirectory(ctx, destinationURL)) {
 				destinationURL.Path = util.generateObjectPath(destinationURL.Path, f.Name())
 			}
--- a/cmd/copyUploadEnumerator.go
+++ b/cmd/copyUploadEnumerator.go
@@ -30,9 +30,8 @@ func (e *copyUploadEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 		return fmt.Errorf("cannot find source to upload")
 	}
 
-	// if the user specifies a virtual directory ex: /container_name/extra_path
-	// then we should extra_path as a prefix while uploading
-	// temporarily save the path of the container
+	// If the user specifies a virtual directory, (ex: /container/extra_path)
+	// we should add the extra path as a prefix to the destination.
 	cleanContainerPath := destinationURL.Path
 
 	// when a single file is being uploaded, we need to treat this case differently, as the destinationURL might be a blob
@@ -69,9 +68,6 @@ func (e *copyUploadEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 			}
 
 			cleanedSource := strings.Replace(listOfFilesAndDirectories[0], common.OS_PATH_SEPARATOR, common.AZCOPY_PATH_SEPARATOR_STRING, -1)
-			if destinationURL.Path[len(destinationURL.Path)-1:] == `/` {
-				destinationURL.Path = util.generateObjectPath(cleanContainerPath, f.Name())
-			}
 			err = e.addTransfer(common.CopyTransfer{
 				Source:           cleanedSource,
 				Destination:      destinationURL.String(),
--- a/cmd/copyUploadEnumerator.go
+++ b/cmd/copyUploadEnumerator.go
@@ -49,7 +49,7 @@ func (e *copyUploadEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 				return fmt.Errorf("for the use of include flag, source needs to be a directory")
 			}
 			// append file name as blob name in case the given URL is a container
-			if (e.FromTo == common.EFromTo.LocalBlob() && util.urlIsContainer(destinationURL)) ||
+			if (e.FromTo == common.EFromTo.LocalBlob() && util.urlIsContainerOrVirtualDirectory(destinationURL)) ||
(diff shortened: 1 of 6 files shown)


cmd/copyUploadEnumerator.go
```diff
@@ -30,6 +30,11 @@ func (e *copyUploadEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 		return fmt.Errorf("cannot find source to upload")
 	}
 
+	// if the user specifies a virtual directory ex: /container_name/extra_path
+	// then we should extra_path as a prefix while uploading
```
zezha-msft: typo, missing verb

cmd/copyUploadEnumerator.go
```diff
@@ -64,6 +69,9 @@ func (e *copyUploadEnumerator) enumerate(cca *cookedCopyCmdArgs) error {
 			}
 
 			cleanedSource := strings.Replace(listOfFilesAndDirectories[0], common.OS_PATH_SEPARATOR, common.AZCOPY_PATH_SEPARATOR_STRING, -1)
+			if destinationURL.Path[len(destinationURL.Path)-1:] == `/` {
```
zezha-msft: hmmm I'm a bit confused, why isn't line 53-55 covering this case already?

cmd/zt_copy_blob_upload_test.go
```diff
@@ -28,6 +28,65 @@ import (
 	"strings"
 )
 
+// regular local file->blob upload
+func (s *cmdIntegrationSuite) TestUploadSingleFileToBlobFolder(c *chk.C) {
```
zezha-msft: Folder => virtual directory

cmd/copyUploadEnumerator.go
adreed-msft: I literally just copied the majority of a working test 👀, fixing the wording

adreed-msft: oh yikes

cmd/copyUploadEnumerator.go
adreed-msft: urlIsContainerOrShare is... a really odd function to start with. It doesn't even do the job that the rules listed on top of the function even describe in the first place. I'm rewriting it.

adreed-msft:
@zezha-msft Could you hit this up again? I've rewritten the relevant function to well, actually function.

JohnRusk:
Thanks Adele for the description of the change.  It helped me get my head around it.

cmd/copyUtil.go
```diff
@@ -55,26 +54,21 @@ func (copyHandlerUtil) numOfWildcardInURL(url url.URL) int {
...
-	} else if ((!isIPEndpointStyle && numOfSlashes == 1) || (isIPEndpointStyle && numOfSlashes == 2)) && strings.HasSuffix(url.Path, "/") { // this checks if container_name/ was given
-		return true
+	if azblob.NewBlobURLParts(*url).IPEndpointStyleInfo.AccountName == "" {
+		//Typical endpoint style
+		//If there's no slashes after the first, it's a container.
+		//If there's a slash on the end, it's a virtual directory/container.
+		//Otherwise, it's just a blob.
+		return strings.HasSuffix(url.Path, "/") || strings.Count(url.Path[1:], "/") == 0
+	} else {
+		//URL endpoint style: https://IP:port/accountname/container
```
JohnRusk: This is an example of IP endpoint style, but its labelled URL style.  I think its just the comment that's wrong. Should say `// IP endpoint style: https://IP:port/accountname/container`. Do you agree?

cmd/copyUtil.go
```diff
@@ -55,26 +54,21 @@ func (copyHandlerUtil) numOfWildcardInURL(url url.URL) int {
...
+		//If there's 2 or less slashes after the first, it's a container.
```
JohnRusk: I'm not getting why the constant here is 2 (rather than 1).  Can you add tests to copyUtil_tests.go that construct an actual IP style endpoint, and exercise this branch of the code.  That will (a) give test coverage to this branch and (b) prove whether it should be 1 or 2.

adreed-msft: So, IP endpoints don't contain the account name in their host. Instead, they contain the account name in their path.

As such, the path takes 2 slashes or less after the first to reach the container.

`/adeledevtest1/proxysync/` for instance, would be the path to my `proxysync` container on an IP endpoint. In comparison, the URL endpoint would be `adeledevtest1.blob.core.windows.net/proxysync/`.

cmd/copyUtil.go
adreed-msft: Yeah, my mistake when writing that

cmd/copyUtil.go
adreed-msft: If the constant were just 1, then `/adeledevtest1/proxysync` would be considered a blob by AzCopy, and thus we would not provide the service a file name.

cmd/copyUtil_test.go
```diff
@@ -29,27 +29,39 @@ type copyUtilTestSuite struct{}
...
-	isContainerOrShare = util.urlIsContainerOrShare(&testUrl)
-	c.Assert(isContainerOrShare, chk.Equals, true)
+	isContainer = util.urlIsContainer(&testUrl)
+	c.Assert(isContainer, chk.Equals, true)
+}
+
+func (s *copyUtilTestSuite) TestIPIsContainerOrBlob(c *chk.C) {
+	util := copyHandlerUtil{}
+
+	testIP := url.URL{Host: "127.0.0.1:8256", Path: "/account/container"}
```
JohnRusk: Need to also test one of the form /account/container/virtualdir  Why, since without that test we have not tested the boundary condition of where (with how many slashes) the IP-form of the logic transitions from returning true to returning false.

cmd/copyUtil.go
JohnRusk: So it should have one more slash than a URL endpoint, right?  But the constant is not one greater than what is used in URL endpoints, it is currently _two_ greater.  That doesn't smell right.

adreed-msft: Good point.

adreed-msft:
Squashing and merging

PR #455, opened 2019-06-21
Feature/command line bandwidth cap

JohnRusk:
Allows capping of throughput for upload, download and S2S.

The S2S implementation is relatively coarse-grained (at block level) while the upload and download implementations are finer-grained (because we can see the actual data).

As discussed, there's no dynamic rate-seeking here. It just takes and optional mbps cap on the command line.


adreed-msft:
How is this cap ingested? I only see changes inside ste. Is this an environment variable or something?

main.go
```diff
@@ -58,10 +58,19 @@ func main() {
 	concurrentConnections := common.ComputeConcurrencyValue(runtime.NumCPU())
 	concurrentFilesLimit := computeConcurrentFilesLimit(maxFileAndSocketHandles, concurrentConnections)
 
-	err = ste.MainSTE(concurrentConnections, concurrentFilesLimit, 2400, azcopyAppPathFolder, azcopyLogPathFolder)
-	common.PanicIfErr(err)
+	// now we know how we want to start the STE, but create func to defer startup until after cmd line params are parsed
+	// TODO: reviewers, I'm not totally happy with the amount of indirection created by the use of this function.
+	//    Alternatives I thought of were:
+	//    1. Move more of the code that's here into root.go. Maybe that's an option...
+	//    2. Don't use a command line parameter for the cap on transfer rate, do it with an environment variable instead.
```
adreed-msft: I feel like an environment variable may actually be preferable to some users. I recall that the original intent of this was to "be a good neighbor" network wise. In a way, that sounded to me like users wanted a set-and-forget solution. This may actually also be preferable scripting wise.

I can see why it feels "wrong" though, since this may just be a single-transfer thing that the user wants.

Maybe the best solution is a mix of both. Infer from the environment variable unless the command line parameter is explicitly specified.

cmd/root.go
```diff
@@ -57,18 +70,19 @@ var rootCmd = &cobra.Command{
...
 var glcm = common.GetLifecycleMgr()
 
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
-func Execute(azsAppPathFolder, logPathFolder string) {
+func Execute(azsAppPathFolder, logPathFolder string, steStartup steStartupFunc) {
 	azcopyAppPathFolder = azsAppPathFolder
 	azcopyLogPathFolder = logPathFolder
 
+	rootCmd.PersistentPreRunE = createPersistentPreRunE(steStartup)
```
adreed-msft: So, my understanding of what's going on here is that we're feeding our steStartup function from main.go into the root cmd, baking it into a pre-run function, and then letting cobra ingest it here. Is it feasible to just cut this down to the last step? That might proactively clean up some of the wild-goose-chaseyness of our codebase.

adreed-msft:
> 		case <-p.done:

Which of these suicide channels goes unused? Is it possible we could cut one and convert everything to the other?

---
Refers to: ste/pacer-tokenBucketPacer.go:138 in d20ee4a. [](commit_id = d20ee4a89e4b4674c0571b7a5f374cca49becf25, deletion_comment = False)

ste/pacer-tokenBucketPacer.go
```diff
@@ -66,23 +85,44 @@ func newTokenBucketPacer(ctx context.Context, bytesPerSecond int64, expectedByte
 	return p
 }
 
-// RequestRightToSend function is called by goroutines to request right to send a certain amount of bytes.
+// RequestTrafficAllocation function is called by goroutines to request right to send a certain amount of bytes.
 // It controls their rate by blocking until they are allowed to proceed
-func (p *tokenBucketPacer) RequestRightToSend(ctx context.Context, bytesToSend int64) error {
-	for atomic.AddInt64(&p.atomicTokenBucket, -bytesToSend) < 0 {
+func (p *tokenBucketPacer) RequestTrafficAllocation(ctx context.Context, byteCount int64) error {
```
adreed-msft: I'll admit, this entire setup is really clever, basically using max throughput as a currency of sorts. I've got to wonder though, would it be theoretically better to pace disk usage alongside pacing network usage? I could be dead wrong on how we handle disk usage. My thinking is that disk-usage wise, we could limit reads to a buffer slightly larger than the actual network output (so that we have a buffer zone for disk latency), and thus be a good neighbor on the local machine.

adreed-msft:
Note to @zezha-msft when you get around to reviewing this, it'll look like a scary amount of code touched. However, the bulk of this PR actually lives inside of the following few files. Everything else is *mostly* function signature changes:

- `main.go`
- `cmd/root.go`
- `ste/jobsAdmin.go`
- `ste/pacedReadSeeker.go`
- `ste/pacer-tokenBucketPacer.go`

ste/pacer-tokenBucketPacer.go
```diff
@@ -66,23 +85,44 @@ func newTokenBucketPacer(ctx context.Context, bytesPerSecond int64, expectedByte
...
 		// by taking our desired count we've moved below zero, which means our allocation is not available
-		// right now, so put back what we asked for, and wait
-		atomic.AddInt64(&p.atomicTokenBucket, bytesToSend)
+		// right now, so put back what we asked for, and then wait
+		atomic.AddInt64(&p.atomicTokenBucket, byteCount)
+
+		// vary the wait amount, to reduce risk of any kind of pulsing or synchronization effect, without the perf and
+		// and threadsafety issues of actual random numbers
+		totalWaitsSoFar := atomic.AddInt64(&p.atomicWaitCount, 1)
+		modifiedSleepDuration := time.Duration(float32(bucketDrainSleepDuration) * (float32(totalWaitsSoFar%10) + 5) / 10) // 50 to 150% of bucketDrainSleepDuration
```
adreed-msft: Another case where I could be dead wrong, but if I'm making enough transfers at once, is it possible that I could stack the layers too high and get more throughput than the limit?

If so, this method of limiting throughput may actually cause issues with larger transfers.

My suggestion for this is to handle transfers as a queue-- if something goes over the maximum amount of tokens in the bucket, wait until the bucket is full. If less, wait until the bucket has _just_ enough to push it through.

adreed-msft:
More info for @zezha-msft to have a bit speedier a review: Roughly the way that I can tell this works is that an STE startup function gets passed from `main.go` to `cmd/root.go`, gets baked into the root pre-run function w/ the new parameter in `cmd/root.go`.

Then, in STE, if the maximum megabits per second is above 0, it uses the new `tokenBucketPacer`. If not, it uses `nullAutoPacer`, which doesn't limit throughput. `tokenBucketPacer` basically treats throughput as a replenishing currency, and semi-queues requests by delaying them up to 150% of the maximum wait time, depending on how many waits are currently happening (modulo 10). on

adreed-msft:
I say semi-queues because it doesn't *actually* create a queue, but it more or less just layers transfers X seconds into the future.

main.go
JohnRusk: Good question. Let's see what @zezha-msft says. Also, its worth noting that when we are called from tools like Storage Explorer is probably (slightly) easier for them to set a command line parameter.

cmd/root.go
JohnRusk: Yes, we could possibly move it all to there.  Interested in what @zezha-msft thinks....

JohnRusk:
> case <-p.done:
> 
> 
> Which of these suicide channels goes unused? Is it possible we could cut one and convert everything to the other?

Actually, both are used.  The Close based one (p.done) is used when this code is used by the page blob pacing code (file-specific pacing, has been in place since before GA).  I left the context-based one in place because we don't currenly have a place to call Close for JobsAdmin, so the global (command-line cap) pacer won't have Close called on it.

One alternative would be to simply get rid of the context-based cancellation, and rely on the fact that JobsAdmin runs until the app exits anyway, and so it doesn't really matter if this routine is still running at that time.  I.e. we simplify the code as Adele suggests, and just never shutdown the global pacer. What do you think, @adreed-msft  and @zezha-msft ?

ste/pacer-tokenBucketPacer.go
JohnRusk: Do you mean, "We should limit disk usage to avoid conflicting with other things that also need _disk_" or "We should limit disk usage as a preferrable implementation of controlling _network_ throughput" or "We should do both, because doing just one won't be enough"?   

I must admit I never thought of using a disk throughput control as the sole method of controlling network throughput. I think it has a theoretical disadvantage from the network perpective, which is that, as you say, there would have to be a buffer. In the cases where that buffer is non-empty, our control over network throughput would be much looser, and so network throughput could be more spikey.  My feeling is that smooth network throughput (at the expense of spikey disk) is better than smooth disk throughput (at the expense of spikey network).  For the reason outlined below, I don't think its necessary to do both.

I guess the other question is whether a user concerned about disk throughput, could reduce AzCopy's disk throughput by using the _network_ cap as currently implement.  And the answer is, yes they could.  If you set a low network cap, our disk buffers will soon reach their limit, and your disk throughput will slow to whatever you set as the network cap.

ste/pacer-tokenBucketPacer.go
JohnRusk: I don't currently believe it's possible to get more throughput than the limit.  The logic being that tokens are only put into the bucket at a certain rate.  The net rate(*) at which we remove them is always <= the rate at at which they are added, so we can't go faster than the add rate.

I.e. if we remove too many, on line 93, then we immediately put back everything we removed, so our net change for that pass through the loop 0. The invariant maintained by all callers of RequestTrafficAllocation is that, for their final "take" from the bucket (the one that they do not immediately undo it) the remaining token ballance in the bucket was not negative after they removed their allocation. I.e. every single caller to RequestTrafficAllocation can sap "_I_ did not leave the bucket with a negative count". Since that applies to all callers, we know the bucket never is left with a negative count, so we know that the rate of taking tokens out is always <= the rate at which pacerBody is putting them in.

So I don't belive there's a bug here.  The token bucket implementation is a very standard solution to these types of problem, so I'm reluctant to change it to a queue.  Also, I've used variants of this code in two other apps and have never seen them go over their limits.  But, if there is indeed a bug, please talk me through it and help me see it! :-)

JohnRusk:
>delaying them up to 150% of the maximum wait time, depending on how many waits are currently happening (modulo 10). on

There is no limit to how long it will make them wait. (The 150% limit applies only to a single iteration; but there is no limit to how many iterations there may be).

JohnRusk:
> A rather important one is the comment on ste/pacer-tokenBucketPacer.go:102. In theory, AzCopy could bust past the user-suggested limit with enough transfers.

It can't bust past. (If I'm wrong, please give me a little more help to see that ;-)

> Moreso code cleanliness oriented, a comment unrelated to your changes about having two suicide channels for ste/tokenBucketPacer.go:136.

Awaiting reply from you and Ze to comment above.


> Finally, another code cleanliness oriented suggested change would be to try to move the ste initialization into cmd/root.go rather than having part of it live in main.go, and then just go along a sort of assembly-line for it to finally be placed into the root pre-run function.

Awaiting reply from you and Ze to comment above.

zezha-msft:
@JohnRusk Sorry for the delay, I'll review on Monday (Redmond time).

ste/pacer-tokenBucketPacer.go
adreed-msft: I meant the former, rather than the latter.

ste/pacer-tokenBucketPacer.go
adreed-msft: Oh, yikes, for some reason my brain read "for" as "if"-- you're right, it can't bust past.

Had to give the code a good re-read until I found what I was missing.

ste/pacer-tokenBucketPacer.go
JohnRusk: I think we can let folks do that just by using the existing flag as implemented. (As per final paragraph of my comment above). I think we don't want to have _two_ flags - that would just be confusing because only the lower of the two would actually have any meaningful effect.  On that basis I'm marking this thread as resolved, but if that hasn't adequately answered your question please unresolved it and elaborate in a comment.

JohnRusk:
Ping @zezha-msft  for Wednesday

adreed-msft:
> One alternative would be to simply get rid of the context-based cancellation, and rely on the fact that JobsAdmin runs until the app exits anyway, and so it doesn't really matter if this routine is still running at that time. I.e. we simplify the code as Adele suggests, and just never shutdown the global pacer. What do you think, @adreed-msft and @zezha-msft ?

This would probably work just fine. I don't think there's anything _too_ important perf wise after a transfer is over, so it doesn't particularly hurt to keep the routine alive here.

main.go
zezha-msft: Good point @adreed-msft, but from my experience there are many users who launch the tool in simultaneous processes (not advised of course), so keeping it as a command line flag might be easier to use.

JohnRusk: I'm also leaning towards staying with the command line parameter. In my experience with other tooling, the first thing folks want to do with a feature like this is try it about 5 times with different speeds, and that's easier to do from the command line.  

Also, our current env vars are basically things which either we don't want in command line history (i.e. secrets) or things which we really wish the customer didn't need to think of at all.

Given the discussion so far, and the fact that having it on the command line worked well in FDT, I'm going to mark this conversation resolved.

cmd/root.go
zezha-msft: Agreed. Having the initializing code all in one place would help a lot with the readability.

zezha-msft:
> One alternative would be to simply get rid of the context-based cancellation, and rely on the fact that JobsAdmin runs until the app exits anyway, and so it doesn't really matter if this routine is still running at that time. I.e. we simplify the code as Adele suggests, and just never shutdown the global pacer. 

Sounds good to me, the global pacer and live/die with the whole application.

zezha-msft:
@adreed-msft thanks a lot for the insights, they made it super easy to CR this change.

@JohnRusk really sorry for the delay in review! 😢

JohnRusk:
No prob, @zezha-msft  :-)

adreed-msft:
@zezha-msft thoughts on the way we're baking a lambda from main.go into the root pre-run function? It feels like a bit of a wild goose chase to find the actual function there when it could've just as easily been a part of `root.go`.

zezha-msft:
@JohnRusk I'm merging now to get a full build into perf testing. Please address the comments on your Monday. 😄

JohnRusk:
Lambda is fixed, can ctx removed. Both changes are here: #472

PR #460, opened 2019-06-25
Carry secondary host improvements from azure-storage-blob-go to azure-storage-azcopy

adreed-msft:


JohnRusk:
Also ping @zezha-msft  for Wednesday

Virepri:
Yes, this has been applied to Azure Files as well.

adreed-msft:
Yes, this did make it to Azure Files as well.

PR #461, opened 2019-06-25
Experimental, unused ADLSG2 traverser

adreed-msft:
Hi there, folks.

Again, this is not intended to be merged in this release. Feel free to ignore this until you genuinely have spare time (or the next release has been pushed out).

However, @zezha-msft does want to make use of this in remove, and it will likely be useful when we start the copy refactor. Atop that, I initially did this to get an idea of how the traversers work in a more hands-on way.

I *suppose* the next logical move after getting this in would be to implement the copy enumerators with it (despite the lack of support for wildcard chars). It actually wouldn't be a huge amount of work past this to support wildcards, I think. I would like to discuss implementing that sometime soon (tm)


cmd/zc_traverser_blobfs.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	if pgr.XMsResourceType() == "directory" {
+		return pgr, false
+	}
+
+	return pgr, true
+}
+
+func (_ *blobFSTraverser) parseLMT(t string) time.Time {
```
zezha-msft: this should be in the SDK itself, but let's tolerate it since we'll remove it after interop

cmd/zt_blobfs_traverser_test.go
```diff
@@ -0,0 +1,182 @@
...
+// validate traversing a single Blob, a single Azure File, and a single local file
+// compare that the traversers get consistent results
+func (s *bfsTraverserSuite) TestBfsTraverserWithSingleObject(c *chk.C) {
+	bfsu := getBFSU()
+	filesystemURL, _ := createNewFilesystem(c, bfsu)
+	defer deleteFilesystem(c, filesystemURL)
+
+	// test two scenarios, either blob is at the root virtual dir, or inside sub virtual dirs
+	for _, storedObjectName := range []string{"sub1/sub2/singleblobisbest", "nosubsingleblob", "满汉全席.txt"} {
+		// set up the container with a single blob
```
zezha-msft: fix comment

cmd/zt_blobfs_traverser_test.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// validate traversing a container, a share, and a local directory containing the same objects
+// compare that traversers get consistent results
+func (s *bfsTraverserSuite) TestBfsTraverserContainerAndLocalDirectory(c *chk.C) {
+	bfsu := getBFSU()
+	filesystemURL, _ := createNewFilesystem(c, bfsu)
+	defer deleteFilesystem(c, filesystemURL)
+
+	// set up a filesystem with the same files
```
zezha-msft: fix comment, what same files?

cmd/zt_blobfs_traverser_test.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// validate traversing a virtual and a local directory containing the same objects
+// compare that blob and local traversers get consistent results
+func (s *bfsTraverserSuite) TestBfsTraverserWithVirtualAndLocalDirectory(c *chk.C) {
+	bfsu := getBFSU()
+	filesystemURL, _ := createNewFilesystem(c, bfsu)
+	defer deleteFilesystem(c, filesystemURL)
+
+	// set up the container with numerous blobs
```
zezha-msft: fix comments

cmd/zt_blobfs_traverser_test.go
```diff
@@ -0,0 +1,182 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package cmd
+
+import (
```
zezha-msft: These tests are duplicated in cmd/zt_generic_traverser_test.go, right?

azbfs/zz_generated_models.go
```diff
@@ -407,6 +408,31 @@ type Path struct {
 	Permissions *string `json:"permissions,omitempty"`
 }
 
+func (p Path) LastModifiedTime() time.Time {
```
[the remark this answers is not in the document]
JohnRusk: How and why would it have been generated, when the swagger file hasn't changed?  @adreed-msft ?

cmd/zc_traverser_blobfs.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (t *blobFSTraverser) traverse(processor objectProcessor, filters []objectFilter) (err error) {
+	bfsURLParts := azbfs.NewBfsURLParts(*t.rawURL)
+
+	pathProperties, isFile := t.getPropertiesIfSingleFile()
+	if isFile {
+		storedObject := newStoredObject(
+			getObjectNameOnly(bfsURLParts.DirectoryOrFilePath),
+			"",
```
JohnRusk: Minor: it would help with understanding to copy here, the comment which Ze already has on the similar case in blob: // relative path makes no sense when the full path already points to the file

(Without that comment here, I had to find usages of newStorageObject, to understand what the second param was and why it should be empty here)

cmd/zc_traverser_blobfs.go
```diff
@@ -0,0 +1,124 @@
...
+	for {
+		dlr, err := dirUrl.ListDirectorySegment(t.ctx, &marker, t.recursive)
+
+		if err != nil {
+			return fmt.Errorf("could not list files. Failed with error %s", err.Error())
+		}
+
+		for _, v := range dlr.Paths {
+			if v.IsDirectory == nil {
+				// TODO: Index file.
```
JohnRusk: What's this TODO?

cmd/zc_traverser_blobfs.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+
+		return processIfPassedFilters(filters, storedObject, processor)
+	}
+
+	dirUrl := azbfs.NewDirectoryURL(*t.rawURL, t.p)
+	marker := ""
+	searchPrefix := bfsURLParts.DirectoryOrFilePath
+
+	if !strings.HasSuffix(searchPrefix, "/") {
```
JohnRusk: For consistency with rest of the current codebase, should be using AZCOPY_PATH_SEPARATOR_STRING on both of these.

azbfs/zz_generated_models.go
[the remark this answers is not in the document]
adreed-msft: This wasn't generated, but you're right, it should probably go elsewhere.

cmd/copyUtil.go
```diff
@@ -372,6 +372,11 @@ func (util copyHandlerUtil) getContainerUrl(blobParts azblob.BlobURLParts) url.U
 	return blobParts.URL()
 }
 
+func (util copyHandlerUtil) getFilesystemUrl(parts azbfs.BfsURLParts) url.URL {
```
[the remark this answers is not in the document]
adreed-msft: Actually uncertain this was even needed in the first place. It doesn't seem to be used anywhere. Removing it. I think I'd just added it to have feature parity with Blob and File.

cmd/zt_blobfs_traverser_test.go
adreed-msft: Fair point.

cmd/zc_traverser_blobfs.go
adreed-msft: Unnecessary. I forgot to remove it after I'd handled the case.

azbfs/zz_generated_models.go
```diff
@@ -407,6 +407,18 @@ type Path struct {
 	Permissions *string `json:"permissions,omitempty"`
 }
 
+func (p Path) ContentMD5() []byte {
```
JohnRusk: This is also not generated, right? And so could go into path extensions file?

JohnRusk:
BTW what does this mean in the commit message: "shift ADLS rm CI tools over to traverser CI tools"?

adreed-msft:
Squashing and Merging.

(FYI, already answered John's question in teams. The short and skinny of it is that me and Ze stepped over each other's toes on this and the ADLSG2 rm command. I just cut the fat and made it all converge to a single set of tools to create ADLSG2 objects in CI)

PR #462, opened 2019-06-26
Added ADLS Gen2 support for rm command

zezha-msft:
There are still some work left to do (docs and more tests). I'm posting it as daft to get some feedbacks.

I looked at the code path for remove transfers, and was really hesitant to add a bunch of code only to throw them away after inter-op. Since the feature ask was to just enable the remove scenarios, I think it might be better to implement it 'the lazy way' and simply invoke the azbfs SDK to delete files/directories directly, instead of doing a full traversal. This does mean that include/exclude flags would not be supported when deleting adls gen2 resources, but one of the main advantages of this service is to delete/rename directories atomically, so I'm not sure how useful these filters would be even if they are implemented. 

Please let me know your thoughts.

### DIFF
--- a/cmd/credentialUtil.go
+++ b/cmd/credentialUtil.go
@@ -246,6 +246,10 @@ func getCredentialType(ctx context.Context, raw rawFromToInfo) (credentialType c
 		if credentialType, err = getBlobCredentialType(ctx, raw.source, false, raw.sourceSAS != ""); err != nil {
 			return common.ECredentialType.Unknown(), err
 		}
+	case common.EFromTo.BlobFSTrash():
+		if credentialType, err = getBlobFSCredentialType(ctx, raw.source, raw.sourceSAS != ""); err != nil {
+			return common.ECredentialType.Unknown(), err
+		}
 	case common.EFromTo.BlobLocal(), common.EFromTo.BlobPipe():
 		if credentialType, err = getBlobCredentialType(ctx, raw.source, true, raw.sourceSAS != ""); err != nil {
 			return common.ECredentialType.Unknown(), err
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -53,8 +53,10 @@ func init() {
 				raw.fromTo = common.EFromTo.BlobTrash().String()
 			} else if srcLocationType == common.ELocation.File() {
 				raw.fromTo = common.EFromTo.FileTrash().String()
+			} else if srcLocationType == common.ELocation.BlobFS() {
+				raw.fromTo = common.EFromTo.BlobFSTrash().String()
 			} else {
-				return fmt.Errorf("invalid source type %s pased to delete. azcopy support removing blobs and files only", srcLocationType.String())
+				return fmt.Errorf("invalid source type %s to delete. azcopy support removing blobs/files/adls gen2", srcLocationType.String())
 			}
 
 			// Since remove uses the copy command arguments cook, set the blobType to None and validation option
@@ -67,12 +69,12 @@ func init() {
 		Run: func(cmd *cobra.Command, args []string) {
 			cooked, err := raw.cook()
 			if err != nil {
-				glcm.Error("failed to parse user input due to error " + err.Error())
+				glcm.Error("failed to parse user input due to error: " + err.Error())
 			}
 			cooked.commandString = copyHandlerUtil{}.ConstructCommandStringFromArgs()
 			err = cooked.process()
 			if err != nil {
-				glcm.Error("failed to perform copy command due to error " + err.Error())
+				glcm.Error("failed to perform remove command due to error: " + err.Error())
 			}
 
 			glcm.SurrenderControl()
--- a/cmd/removeEnumerator.go
+++ b/cmd/removeEnumerator.go
@@ -20,7 +20,15 @@
 
 package cmd
 
-import "github.com/Azure/azure-storage-file-go/azfile"
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/Azure/azure-storage-azcopy/azbfs"
+	"github.com/Azure/azure-storage-file-go/azfile"
+	"net/url"
+	"strings"
+)
 
 // provide an enumerator that lists a given blob resource (could be a blob or virtual dir)
 // and schedule delete transfers to remove them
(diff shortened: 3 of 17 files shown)


cmd/credentialUtil.go
```diff
@@ -140,9 +140,10 @@ func getBlobCredentialType(ctx context.Context, blobResourceURL string, canBePub
 // 2. If there is OAuth token info passed from env var, indicating using token credential. (Note: this is only for testing)
 // 3. Otherwise use shared key.
 func getBlobFSCredentialType() (common.CredentialType, error) {
-	if oAuthTokenExists() {
-		return common.ECredentialType.OAuthToken(), nil
-	}
+	// TODO uncomment before merging
```
JohnRusk: Just leaving a comment here to make sure this doesn't get forgotten -;)

cmd/removeEnumerator.go
```diff
@@ -109,3 +117,74 @@ func (s *directoryStack) Pop() (*azfile.DirectoryURL, bool) {
 		return &e, true
 	}
 }
+
+// TODO move after ADLS/Blob interop goes public
+// TODO this simple remove command is only here to support the scenario temporarily
+// Ultimately, this code can be merged into the newRemoveBlobEnumerator
+func removeBfsResource(cca *cookedCopyCmdArgs) (successMessage string, err error) {
+	ctx := context.Background()
+
```
JohnRusk: We don't support wildcards, right? e.g. * at the end of the URL (or inside the filename at the end of the URL). Is there an error message somewhere, if one is provided, to say that we don't support them?

cmd/remove.go
```diff
@@ -53,8 +53,10 @@ func init() {
 				raw.fromTo = common.EFromTo.BlobTrash().String()
 			} else if srcLocationType == common.ELocation.File() {
 				raw.fromTo = common.EFromTo.FileTrash().String()
+			} else if srcLocationType == common.ELocation.BlobFS() {
+				raw.fromTo = common.EFromTo.BlobFSTrash().String()
```
JohnRusk: (Not sure where to leave this comment). You mentioned that include/exclude will not be supported for ADLS Gen 2 remove, but I don't see any error message that's displayed if they are provided when doing an ADLS Gen 2 remove.

cmd/removeEnumerator.go
```diff
@@ -109,3 +117,74 @@ func (s *directoryStack) Pop() (*azfile.DirectoryURL, bool) {
...
+	if urlParts.DirectoryOrFilePath == "" {
+		return "", errors.New("cannot delete filesystem")
+	}
+
+	// we do not know if the source is a file or a directory
+	// we assume it is a directory and get its properties
+	directoryURL := azbfs.NewDirectoryURL(*sourceURL, p)
+	props, err := directoryURL.GetProperties(ctx)
+	if err != nil {
+		return "", fmt.Errorf("cannot verify ressource due to error: %s", err)
```
JohnRusk: minor: typo

cmd/credentialUtil.go
adreed-msft: Also leaving a comment here to remind you to cherrypick in the SAS ADLSG2 commit and make use of it in tests

cmd/zt_remove_adls_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+package cmd
+
+import (
+	"context"
+	"github.com/Azure/azure-storage-azcopy/azbfs"
+	chk "gopkg.in/check.v1"
+)
+
+func (s *cmdIntegrationSuite) TestRemoveDirectory(c *chk.C) {
```
adreed-msft: Good tests.

cmd/removeEnumerator.go
zezha-msft: Added a check for better error message, thanks

cmd/removeEnumerator.go
zezha-msft: which word?

zezha-msft: nvm it was fixed locally XD

zezha-msft:
> Requesting changes WRT two things:

> Restrictions around incorrect setups (--include and --exclude, wildcards)
You said you were going to make use of ADLSG2 SAS authentication to write proper tests. This is mostly either a "do it or say you won't since we already have some enumerator tests for remove."

Both are addressed. 😄

cmd/removeEnumerator.go
```diff
@@ -109,3 +117,89 @@ func (s *directoryStack) Pop() (*azfile.DirectoryURL, bool) {
...
 }
+
+// TODO move after ADLS/Blob interop goes public
+// TODO this simple remove command is only here to support the scenario temporarily
+// Ultimately, this code can be merged into the newRemoveBlobEnumerator
+func removeBfsResource(cca *cookedCopyCmdArgs) (successMessage string, err error) {
+	ctx := context.Background()
+
+	// return an error if the unsupported options are passed in
+	if len(cca.includePatterns)+len(cca.excludePatterns) > 0 {
```
adreed-msft: ...Shouldn't this be in cca.cook()?

Though, then again, this code is meant to be removed. May as well just leave this here so it's easier to get rid of when the time comes.

cmd/removeEnumerator.go
```diff
@@ -109,3 +117,89 @@ func (s *directoryStack) Pop() (*azfile.DirectoryURL, bool) {
...
+	}
+
+	// append the SAS query to the newly parsed URL
+	sourceURL = gCopyUtil.appendQueryParamToUrl(sourceURL, cca.sourceSAS)
+
+	// parse the given source URL into parts, which separates the filesystem name and directory/file path
+	urlParts := azbfs.NewBfsURLParts(*sourceURL)
+
+	// patterns are not supported
+	if strings.Contains(urlParts.DirectoryOrFilePath, "*") {
```
adreed-msft: same to here

cmd/removeEnumerator.go
zezha-msft: that's what I thought too, it'd be good to put these things all together

zezha-msft: the things to be removed later

PR #465, opened 2019-06-27
Removed dep-related files as they are no longer used/maintained

zezha-msft:


JohnRusk:
Will the Travis build still build (e.g. did you update it to later go version?)

adreed-msft:
Seems fine to me. Question is when do we want to push this through?

zezha-msft:
@JohnRusk yes! The CI builds all pass.

@adreed-msft I'm thinking we should merge it now. Since AzCopy is not an SDK, and the dep file is literally just for the project itself.

PR #466, opened 2019-06-27
Fix/add catch all log redaction

JohnRusk:
Reviewers: please look at two things in particular (in addition to all the usual stuff)

1. Does this cover everywhere where we log? (It does pipeline.ForceLog and the job logger, and we don't have an applogger any more. Are there any other loggers Ze? I'm not aware of any, but want your double check on my thinking pls).

2. Does it search for enough strings?  Basically its just doing "sig" and "signature" and (for good measure) "token".  But can you think of any other key name that would be associated with sensitive content?


common/logSanitizer.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (s *azCopyLogSanitizer) redact(msg, key string) string {
+
+	// The leading and trailing '-' chars are not in our older redaction code (which will continue to run,
+	// at the point of logging requests etc.)
+	// By including the dashes here we can examine logs (if we ever want to)
+	// to see how many redactions are from here (with the -'s) and how many
+	// are from the old code (without -'s).
+	const redacted = "-REDACTED-"
```
zezha-msft: nice

common/logSanitizer.go
```diff
@@ -0,0 +1,103 @@
...
+// logged, but that's less maintainable in the long term.
+type azCopyLogSanitizer struct {
+}
+
+func NewAzCopyLogSanitizer() pipeline.LogSanitizer {
+	return &azCopyLogSanitizer{}
+}
+
+var sensitiveQueryStringKeys = []string{
+	strings.ToLower(SigAzure),
```
zezha-msft: should a "=" follow the key? If we look for all "sig"s, we might be redacting things other than secrets, for example, if the user's account is called `signaturevent` or something like that.

common/zt_logSanitizer_test.go
```diff
@@ -0,0 +1,68 @@
...
+		{"Foo=x;Signature=bar", "Foo=x;Signature=-REDACTED-"},                                                     // not in a query string
+		{"Signature = bar;Foo = x", "Signature = -REDACTED-;Foo = x"},                                             // not in a query string, with spaces
+		{"Foo=x;Signature=bar", "Foo=x;Signature=-REDACTED-"},                                                     // not in a query string
+		{"Foo : x, Signature : bar, Other: z", "Foo : x, Signature : -REDACTED-, Other: z"},                       // not in a query string, with commas and spaces
+
+		// two replacements in same string
+		{"http://foo?sig=somevalue and http://bar?sig=othervalue BlahBlah", "http://foo?sig=-REDACTED- and http://bar?sig=-REDACTED- BlahBlah"},
+
+		// word "sig" inside the signature
+		{"http://foo?sig=sigvalue BlahBlah", "http://foo?sig=-REDACTED- BlahBlah"},
```
zezha-msft: consider adding cases where `sig` or `signature` is not part of the query and shouldn't be redacted.

zezha-msft:
We might print enumeration errors on the command line, right? Should we be safe and also redact the error message in lifecyclemanager.Error?

common/logSanitizer.go
JohnRusk: If I add a test, as per your other comment (and ensure the test passes) does that cover this comment too?

JohnRusk:
> Should we be safe and also redact the error message in lifecyclemanager.Error?

Yes I think we should. Great idea. I'll add that now.

common/logSanitizer.go
JohnRusk: Actually, it will already pass. (And it does, I just added it). Reason is that the `Contains` test will think its a match, but the regex test will not, so nothing will be replaced.  So only downside of not having the = in the "key string" is the extra perf hit of running the regex -but that only happens if the `Contains` gets a match - and no most strings it won't so we won't have the perf hit.

The upside of _not_ having the = here is that it gives us flexibility to successfully redact things that are not as strictly formatted as query strings. Specifically, things where there is as space before the = or there's a : instead of a =.  I'm not sure that we realistically expect to encounter any of those in the near future; but doing it the way its don't now at least does offer that flexibility (and I can't really see a downside, since the perf hit, as above, will be small and very rare).

JohnRusk: I propose to mark this conversation resolved, but if I'm missing something please reopen it.

common/zt_logSanitizer_test.go
JohnRusk: Done

JohnRusk:
All changes are done, including the lifecycle manager one.

JohnRusk:
So waiting on a final comment from you Ze

common/lifecyleMgr.go
```diff
@@ -128,6 +131,9 @@ func (lcm *lifecycleMgr) Progress(o OutputBuilder) {
 }
 
 func (lcm *lifecycleMgr) Info(msg string) {
+
+	msg = lcm.logSanitizer.SanitizeLogMessage(msg) // sometimes error-like text comes through Info, before the final "we've failed, please stop now" signal comes to Error. So we sanitize in both places.
```
zezha-msft: When do we output errors as INFO? Example?

common/logSanitizer.go
zezha-msft: Ok, the reasoning makes sense to me.

common/logSanitizer.go
```diff
@@ -0,0 +1,103 @@
...
+	for _, key := range sensitiveQueryStringKeys {
+		// We don't care what's before the key (in a query string it will always be ? or &, but that's not
+		// the case in say, an auth header).
+		// Also, for flexibility and robustness we allow : or = as the delimiter, and allow space around it.
+		// We do ASSUME that the value to be redacted will never contain a &. We _think_
+		// that assumption is reasonably for all anticipated values. (Without that
+		// assumption, we'd have to redact query strings all the way to the end of the whole query string.)
+		// Regex has two groups: first gets key and delimiter.
+		// Second group gets as many chars as possible that do not terminate the value.
+		sensitiveRegexMap[key] = regexp.MustCompile("(?i)(?P<key>" + key + "[ \t]*[:=][ \t]*)(?P<value>[^& ,;\t\n\r]+)")
```
zezha-msft: This is pretty awesome, I definitely need to be less ignorant about regexes.

common/lifecyleMgr.go
JohnRusk: In the case I tested it on!  Which was the enumerator being unable to connect. (I had to turn the retry count down to make the test run in reasonable time).  There were multiple lines of output describing the error.  Basically a bunch of stuff was logged with Info, and then the actual error was logged.  Basically is if Info was being used as lcm.Println, and Error was being used as lcm.PrintlnAndTerminate.

common/logSanitizer.go
JohnRusk: There's an art to it. If you don't watch out, this happens: http://regex.info/blog/2006-09-15/247 

Basically, practice, reading the docs, and unit tests!

common/lifecyleMgr.go
zezha-msft: Ok! Sounds good

adreed-msft:
I'll solve your conflicts here quickly because they're pretty trivial and I'm pretty sure any of us would come to the same solution-- In summary:

- `main.go`: Import ordering & 1 new import
- `cmd/copy.go`: ADLSG2 SAS support has been added, so gating it off is not necessary.

ste/xferLogPolicy.go
```diff
@@ -18,6 +18,8 @@ import (
 // Because V10 SDK supports flexibility for injecting customized logging policy,
 // and considering redact x-amz-signature's request header for logging is not a general demand for Azure Storage Blob Go SDK.
 // TODO: Further discuss whether to add callback into RequestLogOptions for Azure Storage Blob Go SDK.
+// TODO: (new) consider also the relationship between the above comment and todos, and the new LogSanitizer
+//    Do we really need this copied version of the blob Storage SDK file now?
```
adreed-msft: I don't think we need this change copied over, but adding the constants over there may not be a bad idea either.

common/logSanitizer.go
```diff
@@ -0,0 +1,103 @@
...
+// safe for concurrent reads
+var sensitiveRegexMap = make(map[string]*regexp.Regexp)
+
+// init a map of pre-prepared regexes, one for each key
+func init() {
+	for _, key := range sensitiveQueryStringKeys {
+		// We don't care what's before the key (in a query string it will always be ? or &, but that's not
+		// the case in say, an auth header).
+		// Also, for flexibility and robustness we allow : or = as the delimiter, and allow space around it.
+		// We do ASSUME that the value to be redacted will never contain a &. We _think_
```
adreed-msft: That would be correct. & is an escaped character in queries.

azbfs/zc_policy_request_log.go
```diff
@@ -111,17 +115,17 @@ func NewRequestLogPolicyFactory(o RequestLogOptions) pipeline.Factory {
 
 func redactSigQueryParam(rawQuery string) (bool, string) {
 	rawQuery = strings.ToLower(rawQuery) // lowercase the string so we can look for ?sig= and &sig=
-	sigFound := strings.Contains(rawQuery, "?sig=")
+	sigFound := strings.Contains(rawQuery, "?"+SigAzure+"=")
 	if !sigFound {
-		sigFound = strings.Contains(rawQuery, "&sig=")
+		sigFound = strings.Contains(rawQuery, "&"+SigAzure+"=")
```
adreed-msft: Why don't we just search for `sig=`?

This seems like a roundabout way of finding it. I also don't believe "=" is a valid character in a URL outside of the query anyway. Even there, it's escaped if it's in a key (if it even can be in one) or a value.

common/zt_logSanitizer_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (s *logSanitizerSuite) TestLogSanitizer(c *chk.C) {
+
+	cases := []struct {
+		raw               string
+		expectedSanitized string
+	}{
+		{"string with no secrets", "string with no secrets"},
+
+		// DON'T redact these
```
adreed-msft: I'd recommend adding a test where `sig` is a value, actually.

I don't *think* it will fail, but, it's best to just make sure that it won't.

common/zt_logSanitizer_test.go
```diff
@@ -0,0 +1,72 @@
...
+		{"This is the sig that I have and x=y", "This is the sig that I have and x=y"},     // sig not in a URL
+		{"http://foo/path/with/sig/in/it?x=y", "http://foo/path/with/sig/in/it?x=y"},       // match is not in the QUERY part of a URL
+		{"http://www.signature.example.com/blah", "http://www.signature.example.com/blah"}, // another with match that's not in QUERY part the URL
+		{"http://foo?signatureevent=123&x=y", "http://foo?signatureevent=123&x=y"},         // our keyword (signature) is not the END of the key-value key name
+
+		// DO redact all of the following
+		{"http://foo?sig=somevalue&x=y", "http://foo?sig=-REDACTED-&x=y"},                                         // remainder of query string is preserved
+		{"http://foo?x=y&" + SigAzure + "=somevalue", "http://foo?x=y&sig=-REDACTED-"},                            // basic case
+		{"http://foo?x=y&sig=somevalue\r\nBlah", "http://foo?x=y&sig=-REDACTED-\r\nBlah"},                         // newline after, case preserved in other text
```
adreed-msft: Another case of "I don't think this will fail", but it'd be best to include a case with weird capitalization on "sig" specifically.

adreed-msft:
I feel like if you just add the two tests I suggested and maybe at least try the change I suggested in `azbfs/zc_policy_request_log.go` this would be good to go.

zezha-msft:
@JohnRusk I'm merging now to get a full build into perf testing. Please address the comments on your Monday. 😄

ste/xferLogPolicy.go
JohnRusk: The question was a TODO about whether we really need xferLogPolicy in AzCopy, not about whether we copy the recent changes to the SDK.  I don't think either of those questions is urgent. And both would be resolve if/when we one day rationalize the policy implementations.  So I propose to mark this resolved, with no further action. Lemme know if you think we should re-open it.

common/logSanitizer.go
JohnRusk: Yes, its correct for queries. I've tried to avoid assumptions that would lead to the sanitizer ONLY working on queries. I.e. for the sake of future proofing (for unknown chnges that may happen to the app in the future(*))  it's desirable if it also redacts name/value pairs that are not in query strings. In this latter case, with the name/value pair not being in a query string, we can't assume any particular escaping behaviour - which is why we implicitly have an assumption that the values to be redacted don't contain '&'.  It's possibly a minor thing - since the non-query string case is one that we can't really even test for in realistic scearios at present, because it doens't exist there. (Except possibly for Auth headers in some S2S calls to AWS).

azbfs/zc_policy_request_log.go
JohnRusk: Do you mean why do the two stage search for "?sig=" and then for "&sig="?   If so, that's not something I'm going to change in this PR. It's existing behaviour, and I'm not going to alter it in a PR that is fundamentally unrelated to that that part of the code.  Too much risk of changing behaviour in an unanticipated way.  

(And, now that I think about it, searching for just "sig=" does change behaviour - in theory at least, if not in practice).

common/zt_logSanitizer_test.go
JohnRusk: Noted. Will hopefuly get to it later this week.

common/zt_logSanitizer_test.go
JohnRusk: Noted, hope to get to it tomorrow, if not earlier.

JohnRusk:
@adreed-msft Tests are added here: #471 

As noted above, I don't propose to change zc_policy_reqruest_log.go.

PR #468, opened 2019-06-28
Use different log output for skipped files 

JohnRusk:
So they don't look like failures

### DIFF
--- a/ste/xfer-anyToRemote.go
+++ b/ste/xfer-anyToRemote.go
@@ -81,7 +81,8 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer *pacer, se
 			return
 		}
 		if exists {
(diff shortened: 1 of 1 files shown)


ste/xfer-anyToRemote.go
```diff
@@ -81,7 +81,8 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer *pacer, se
 			return
 		}
 		if exists {
-			jptm.LogSendError(info.Source, info.Destination, "File already exists", 0)
+			// logging as Warning so that it turns up even in compact logs, and because previously we use Error here
+			jptm.LogAtLevelForCurrentTransfer(pipeline.LogWarning, "File already exists, so will be skipped")
```
adreed-msft: Thought it was a severity level issue. That was easy.

PR #472, opened 2019-07-01
Fix/refactor unclear pacer code

JohnRusk:
Refactor app startup and pacer shutdown


adreed-msft:
Much nicer now.

zezha-msft:
:shipit:

PR #478, opened 2019-07-02
Post-transfer S2S length check

adreed-msft:
This adds a post-transfer length check for S2S (`--s2s-check-length`), as we'd been asked for recently.

This doesn't _guarantee_ file integrity, but it sure is better than nothing.

## Manual testing scenario:

Add a breakpoint to `ste/xfer-anyToRemote.go:294`.

Run in debug mode. When it hits the breakpoint, open storage explorer and upload a new file over the destination of a different length.

Resume execution, and watch the transfer fail.

Test without intentional failure afterwards.


ste/xfer-anyToRemote.go
```diff
@@ -289,7 +290,24 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s ISenderBase, si
 		}
 	}
 
-	s.Epilogue()
+	s.Epilogue() // Perform service-specific cleanup before jptm cleanup. Some services may actually require setup to make the file actually appear.
+
+	// In theory, due to the limits imposed in copy.go, this should never be true unless it's a S2S transfer.
```
zezha-msft: hmmm there is no such limit though

testSuite/scripts/utility.py
```diff
@@ -566,7 +566,7 @@ def verify_operation(command):
             command, stderr=subprocess.STDOUT, shell=True, timeout=600,
             universal_newlines=True)
     except subprocess.CalledProcessError as exec:
-        print("command failed with error code ", exec.returncode, " and message " + exec.output)
+        # print("command failed with error code ", exec.returncode, " and message " + exec.output)
```
zezha-msft: uncomment?

ste/xfer-anyToRemote.go
```diff
@@ -289,7 +290,24 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s ISenderBase, si
...
+	if info.S2SDestLengthValidation {
+		if s2sc, isS2SCopier := s.(s2sCopier); isS2SCopier { // TODO: Implement this for upload and download?
```
zezha-msft: I'm ok with putting this code here since we might enable this check for upload (any to remote doesn't include download).

testSuite/scripts/utility.py
adreed-msft: This is an inconsistent uncomment if we do. Furthermore, we need to sanitize the test framework properly as I've seen SAS tokens leak out.

ste/xfer-anyToRemote.go
adreed-msft: Ah, yeah. I had a limit implemented a while back (https://github.com/Azure/azure-storage-azcopy/pull/478/commits/5902a7527a177033c27107b300bcf3c1218279db).

I'll cut out this comment.

cmd/copy.go
```diff
@@ -302,6 +303,11 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 		return cooked, err
 	}
 
+	// Do not check for an error on this. Leaving this as true doesn't hurt, as the event never triggers unless it's an s2s copy anyway.
+	// This is thanks to a type assertion attempt.
+	// Atop this, leaving this on board for up/downloads in the future would be useful.
```
JohnRusk: Good point. Do you think we should consider enabling it right now for all transfers, not just S2S?

ste/xfer-anyToRemote.go
```diff
@@ -289,7 +290,23 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s ISenderBase, si
 		}
 	}
 
-	s.Epilogue()
+	s.Epilogue() // Perform service-specific cleanup before jptm cleanup. Some services may actually require setup to make the file actually appear.
```
JohnRusk: Nice

ste/sender-blockBlobFromURL.go
```diff
@@ -146,3 +147,14 @@ func (c *urlToBlockBlobCopier) generatePutBlockFromURL(id common.ChunkID, blockI
 		}
 	})
 }
+
+// GetDestinationLength gets the destination length.
+func (c *urlToBlockBlobCopier) GetDestinationLength() (int64, error) {
+	ctxWithLatestServiceVersion := context.WithValue(c.jptm.Context(), ServiceAPIVersionOverride, azblob.ServiceVersion)
```
JohnRusk: Do we need this special versioning for getting the length?  Just asking because we don't have it for some other blob types (e.g append) and I would imagine that getting the length is something that exist in all API versions... but maybe there's some reason I haven't thought of why we need it here.

ste/xfer-anyToRemote.go
```diff
@@ -289,7 +290,23 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s ISenderBase, si
...
+	
+	if info.S2SDestLengthValidation {
+		if s2sc, isS2SCopier := s.(s2sCopier); isS2SCopier { // TODO: Implement this for upload and download?
+			destLength, err := s2sc.GetDestinationLength()
+
+			if err != nil {
+				jptm.FailActiveSend("epilogueWithCleanupSendToRemote", err)
```
JohnRusk: Minor: I think the first parameter here is "when" and we normally would put something like "getting destination length" in there, so that if the error is ever reported, it can be narrowed down to the particular operation in question, rather than just to the method.

ste/sender-blockBlobFromURL.go
zezha-msft: This can be future proofing for Azure Stack. They may support 2018-03 version later, which will allow users to import data into Azure Stack from public cloud. In that case, all calls should get the default version, and anything beyond 2018-03 would make them throw an error.

cmd/copy.go
zezha-msft: I'm ok with it, up to @adreed-msft

ste/sender-blockBlobFromURL.go
JohnRusk: I'm not sure I understand that @zezha-msft. Do you mean that the code, as it stands, should or shouldn't be changed?

zezha-msft: @JohnRusk we could keep this code, as Azure Stack might support 2018-03 one day, and cx might want to import data from public Azure into Azure Stack.

zezha-msft:
@adreed-msft :shipit:

PR #483, opened 2019-07-03
Emphasize download links in README

zezha-msft:


zezha-msft:
@normesta I don't want to mess up your nice formatting, so I'll merge like this. 😉

PR #490, opened 2019-07-08
Upgrade go-ieproxy version

adreed-msft:
As discussed in the triage earlier, our original patch to go-ieproxy made it so that it and any dependents could no longer build on OSes that weren't Windows, Linux, or OSX. This upgrades azcopy's target version to support other OSes *unofficially*.


adreed-msft:
Linked to https://github.com/Azure/azure-pipeline-go/pull/24

PR #492, opened 2019-07-09
Platform independent DevNull

timblackwell:
When running on windows, if you use the window equivalent of `/dev/null/`, `NUL ` you get an error saying that the directory does not exist.

By using `os.DevNull` you get the correct string representing dev null for the target platform.

### DIFF
--- a/common/fe-ste-models.go
+++ b/common/fe-ste-models.go
@@ -43,7 +43,7 @@ const (
 	AZCOPY_PATH_SEPARATOR_STRING = "/"
 	AZCOPY_PATH_SEPARATOR_CHAR   = '/'
 	OS_PATH_SEPARATOR            = string(os.PathSeparator)
-	Dev_Null                     = "/dev/null"
+	Dev_Null                     = os.DevNull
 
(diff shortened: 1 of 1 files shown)


zezha-msft:
Thanks @timblackwell for your contribution!!

JohnRusk:
Did anyone test to see what happens now if you do use /dev/null on Windows? (Because we've been telling people to do that)  What happens?

JohnRusk:
I guess as long as it now gives a clear error, that's OK...

timblackwell:
Ah, no one told me and didn't think to try that.

I have just tested and can confirm the error:

`failed to perform copy command due to error: cannot start job due to error: the destination must be an existing directory in this download scenario.`

JohnRusk:
That sounds good enough. Nice and clear. Nothing weird or confusing. So I think that's OK. Thanks for testing to see what happens.

JohnRusk:
PS I've put a note in the wiki page, where we mentioned it, to say it will change on Windows to the Windows standard (NUL) after version 10.2.1

PR #493, opened 2019-07-10
Full Blob <-> Blob OAuth support

adreed-msft:
Should a user have access to Blob Data Storage Owner, they can generate User Delegation Keys.

What this means for the user is that they can perform service-to-service transfers without generating a SAS token (assuming it's just blob to blob.)

### DIFF
--- a/cmd/helpMessages.go
+++ b/cmd/helpMessages.go
@@ -23,7 +23,7 @@ Copies source data to a destination location. The supported directions are:
   - local <-> Azure Blob (SAS or OAuth authentication)
   - local <-> Azure File (Share/directory SAS authentication)
   - local <-> ADLS Gen 2 (SAS, OAuth, or SharedKey authentication)
-  - Azure Blob (SAS or public) <-> Azure Blob (SAS or OAuth authentication)
+  - Azure Blob (SAS, Public, or OAuth) <-> Azure Blob (SAS or OAuth authentication)
   - Azure File (SAS) -> Azure Block Blob (SAS or OAuth authentication)
   - AWS S3 (Access Key) -> Azure Block Blob (SAS or OAuth authentication)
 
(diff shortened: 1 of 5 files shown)


cmd/helpMessages.go
```diff
@@ -23,7 +23,7 @@ Copies source data to a destination location. The supported directions are:
   - local <-> Azure Blob (SAS or OAuth authentication)
   - local <-> Azure File (Share/directory SAS authentication)
   - local <-> ADLS Gen 2 (SAS, OAuth, or SharedKey authentication)
-  - Azure Blob (SAS or public) <-> Azure Blob (SAS or OAuth authentication)
+  - Azure Blob (SAS, Public, or OAuth) <-> Azure Blob (SAS or OAuth authentication)
```
mahilleb-msft: I guess this arrow shouldn't be bidirectional, independent of your change?

```suggestion
  - Azure Blob (SAS, Public, or OAuth) -> Azure Blob (SAS or OAuth authentication)
```

adreed-msft: Ah, yeah, that would make sense.

adreed-msft:
Note: This will have to be redone after the refactor, since it lives in the enumerators. Perhaps I should move these steps to cooking instead?

JohnRusk:
@adreed-msft I'm not close enough to the problem to comment on whether you should move these steps.  It sounds like we should defer further action on this PR until the refactor has reached a suitable point, and you have make the necessary changes here.  Is that correct?  I'm going to mark this with a pr-on-hold tag to represent that fact, but please remove that tag if I'm incorrect.

adreed-msft:
That is correct. We should hold this until the refactor reaches a suitable place to continue.

adreed-msft:
Abandoning this PR as a new one is in the works and this is based on old dead code.

Issue #168, opened 2019-01-14
Feature: Support filter on Attribute


jeffwmiles:
I'd like to request that azcopy v10 gain support for inclusion or exclusion on file Attributes, as the [general release of azcopy supports](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy#azcopy-parameters) with the /A, /IA, and /XA flags.

zezha-msft:
Hi @jeffwmiles, thanks for reaching out!

I've logged your feature request for consideration.

panfeng0007:
@jeffwmiles Hi I am interested in implementing this feature if no one has implemented it yet. Could you elaborate a little bit more on how this feature should work? By "inclusion or exclusion on file Attributes", do you mean you want to see certain Blob/File properties from the `azcopy list` output?

zezha-msft:
Hi @panfeng0007, FYI the docs for legacy AzCopy has moved to [here](https://docs.microsoft.com/en-us/previous-versions/azure/storage/storage-use-azcopy#a).

JohnRusk:
@panfeng0007  Thanks heaps for your offer to implement this feature.  I'm not sure whether you're thinking of implementing in the cmd/* files (e.g. copyUploadEnumerator) or in the ste/* files (e.g. xfer-anyToRemote).  I think it probably belongs best in the cmd ones.... but, we are about to do some major refactoing of those. Specifically, we are going to refactor the copy* implementations to re-use the types can concepts from the sync* implementations.  

So, anything that is written in today's implementation of the copy* types will need to be totally re-done after the refactoring.  Maybe, if you don't want to wait for the refactoring, you could implement the feature for sync (since if it works there, it should be easily portable to copy after the refactoring).

panfeng0007:
Hi @JohnRusk, Thanks for letting me know of the refactoring. Yes my approach would be to implement the feature in the `cmd/*` files and since the `sync` command works similarly with the `copy` command, I could start implementing it in `sync` first. I’ll also keep an eye on the upstream and keep my branch up to date with the dev or master branch.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

panfeng0007:
@zezha-msft Thanks for pointing out the documentation. The attributes indicated in the doc are for NTFS files only. So I assume for Linux, I should include the file attributes returned by `lsattr` and also on macOS by `xattr`?

zezha-msft:
[...]

PR #501, opened 2019-07-16
Feature: Filter on File Attributes (Windows)

panfeng0007:
This pull request implements the feature proposed in Issue #168. 
Edit: For now only the `sync` command uses this new type of filters.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,14 +1,7 @@
 trigger:
   branches:
     include:
-      - master
-      - dev
-
-pr:
-  branches:
-    include:
-      - master
-      - dev
+      - fp_filter_file_attributes
 
 jobs:
   - job: Linux_and_Windows_Builds
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,7 +1,14 @@
 trigger:
   branches:
     include:
-      - fp_filter_file_attributes
+      - dev
+      - master
+
+pr:
+  branches:
+    include:
+      - dev
+      - master
 
 jobs:
   - job: Linux_and_Windows_Builds
--- a/cmd/syncEnumerator.go
+++ b/cmd/syncEnumerator.go
@@ -113,10 +113,13 @@ func newSyncUploadEnumerator(cca *cookedSyncCmdArgs) (enumerator *syncEnumerator
 	includeFilters := buildIncludeFilters(cca.include)
 	excludeFilters := buildExcludeFilters(cca.exclude)
 
-	includeAttrFilters := buildIncludeAttrFilters(cca.includeFileAttributes, sourceTraverser.fullPath)
-	excludeAttrFilters := buildExcludeAttrFilters(cca.excludeFileAttributes, sourceTraverser.fullPath)
+	includeAttrFilters := buildAttrFilters(cca.includeFileAttributes, sourceTraverser.fullPath, true)
+	excludeAttrFilters := buildAttrFilters(cca.excludeFileAttributes, sourceTraverser.fullPath, false)
 
 	// set up the filters in the right order
+	// Note: includeFilters and includeAttrFilters are ANDed
+	// They must both pass to get the file included
+	// Same rule applies to excludeFilters and excludeAttrFilters
 	filters := append(includeFilters, includeAttrFilters...)
 	filters = append(filters, excludeFilters...)
 	filters = append(filters, excludeAttrFilters...)
--- a/cmd/syncEnumerator.go
+++ b/cmd/syncEnumerator.go
@@ -113,8 +113,13 @@ func newSyncUploadEnumerator(cca *cookedSyncCmdArgs) (enumerator *syncEnumerator
 	includeFilters := buildIncludeFilters(cca.include)
 	excludeFilters := buildExcludeFilters(cca.exclude)
 
+	includeAttrFilters := buildIncludeAttrFilters(cca.includeFileAttributes, sourceTraverser.fullPath)
+	excludeAttrFilters := buildExcludeAttrFilters(cca.excludeFileAttributes, sourceTraverser.fullPath)
+
 	// set up the filters in the right order
-	filters := append(includeFilters, excludeFilters...)
+	filters := append(includeFilters, includeAttrFilters...)
+	filters = append(filters, excludeFilters...)
+	filters = append(filters, excludeAttrFilters...)
 
 	// set up the comparator so that the source/destination can be compared
 	indexer := newObjectIndexer()
--- a/cmd/zc_attr_filter_unix.go
+++ b/cmd/zc_attr_filter_unix.go
@@ -1,0 +1,73 @@
+// +build linux darwin
+
+// Copyright © 2017 Microsoft <wastore@microsoft.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package cmd
+
+type excludeAttrFilter struct {
+	fileAttributes uint32
+	filePath       string
+}
+
+func (f *excludeAttrFilter) doesSupportThisOS() (msg string, supported bool) {
+	msg = "'exclude-attributes' is not supported on this OS. This option will be ignored."
+	supported = false
+	return
+}
+
+func (f *excludeAttrFilter) doesPass(storedObject storedObject) bool {
+	// ignore this option on Unix systems
+	return true
+}
+
+func buildExcludeAttrFilters(attributes []string, fullPath string) []objectFilter {
+	// ignore this option on Unix systems
+	filters := make([]objectFilter, 0)
+	if len(attributes) > 0 {
+		filters = append(filters, &excludeAttrFilter{fileAttributes: 0, filePath: fullPath})
+	}
+	return filters
+}
+
+type includeAttrFilter struct {
+	fileAttributes uint32
+	filePath       string
+}
+
+func (f *includeAttrFilter) doesSupportThisOS() (msg string, supported bool) {
+	msg = "'include-attributes' is not supported on this OS. This option will be ignored."
+	supported = false
+	return
+}
+
+func (f *includeAttrFilter) doesPass(storedObject storedObject) bool {
+	// ignore this option on Unix systems
+	return true
+}
+
+func buildIncludeAttrFilters(attributes []string, fullPath string) []objectFilter {
+	// ignore this option on Unix systems
+	filters := make([]objectFilter, 0)
+	if len(attributes) > 0 {
+		filters = append(filters, &includeAttrFilter{fileAttributes: 0, filePath: fullPath})
+	}
+	return filters
+}
--- a/cmd/zc_attr_filter_windows.go
+++ b/cmd/zc_attr_filter_windows.go
@@ -1,0 +1,165 @@
+// Copyright © 2017 Microsoft <wastore@microsoft.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package cmd
+
+import (
+	"path"
+	"syscall"
+	"strings"
+)
+
+type excludeAttrFilter struct {
+	fileAttributes uint32
+	filePath       string
+}
+
+func (f *excludeAttrFilter) doesSupportThisOS() (msg string, supported bool) {
+	msg = ""
+	supported = true
+	return
+}
+
+func (f *excludeAttrFilter) doesPass(storedObject storedObject) bool {
+	fileName := path.Join(f.filePath, storedObject.name)
+	lpFileName, err := syscall.UTF16PtrFromString(fileName)
+
+	// If it fails to retrive the pointer from file path, let the filter pass
+	if err != nil {
+		return true
+	}
+
+	attributes, err := syscall.GetFileAttributes(lpFileName)
+
+	if err != nil {
+		return true
+	}
+
+	// if a file shares one or more attributes with the ones provided by the filter,
+	// exclude this file
+	if attributes & f.fileAttributes > 0 {
+		return false
+	}
+
+	return true
+}
+
+func buildExcludeAttrFilters(attributes []string, fullPath string) []objectFilter {
+	var fileAttributes uint32
+	filters := make([]objectFilter, 0)
+	// Available attributes (NTFS) include:
+	// R = Read-only files
+	// A = Files ready for archiving
+	// S = System files
+	// H = Hidden files
+	// C = Compressed files
+	// N = Normal files
+	// E = Encrypted files
+	// T = Temporary files
+	// O = Offline files
+	// I = Non-indexed files
+	// Reference for File Attribute Constants:
+	// https://docs.microsoft.com/en-us/windows/win32/fileio/file-attribute-constants
+	fileAttributeMap := map[string]uint32 {
+		"R" : 1,
+		"A" : 32,
+		"S" : 4,
+		"H" : 2,
+		"C" : 2048,
+		"N" : 128,
+		"E" : 16384,
+		"T" : 256,
+		"O" : 4096,
+		"I" : 8192,
+	}
+
+	for _, attribute := range attributes {
+		fileAttributes |= fileAttributeMap[strings.ToUpper(attribute)]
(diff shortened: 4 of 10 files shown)


cmd/zc_attr_filter_unix.go
```diff
@@ -0,0 +1,73 @@
...
+	// ignore this option on Unix systems
+	filters := make([]objectFilter, 0)
+	if len(attributes) > 0 {
+		filters = append(filters, &excludeAttrFilter{fileAttributes: 0, filePath: fullPath})
+	}
+	return filters
+}
+
+type includeAttrFilter struct {
+	fileAttributes uint32
```
JohnRusk: Same comment as above, we don't need any fields in the non-Windows implementation, right?

cmd/zc_attr_filter_windows.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func (f *excludeAttrFilter) doesSupportThisOS() (msg string, supported bool) {
+	msg = ""
+	supported = true
+	return
+}
+
+func (f *excludeAttrFilter) doesPass(storedObject storedObject) bool {
+	fileName := filepath.Join(f.filePath, storedObject.name)
```
JohnRusk: Can you add a comment above this line, saying the following (to remind us later when we are working on long-path support)
```
TODO: review handling of filename for long path support
```

There's no point changing any code now, because we have not finalized exactly how we'll do the long path support (it's partially there already in the Go SDK, but there are unsupported edge cases)

cmd/zc_attr_filter_windows.go
```diff
@@ -0,0 +1,165 @@
...
+	lpFileName, err := syscall.UTF16PtrFromString(fileName)
+
+	// If it fails to retrive the pointer from file path, let the filter pass
```
JohnRusk: @zezha-msft Any thoughts on letting it pass if there's an error?  It's kind of necessary with the current method signatures, unless we panic instead.  Panicing here doesn't seem too crazy to me, because the conversion should "never" fail.

cmd/zc_attr_filter_windows.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	attributes, err := syscall.GetFileAttributes(lpFileName)
+
+	// If the storedObject has invalid file attributes, let the filter pass
+	if err != nil {
+		return true
+	}
+
+	// if a file shares one or more attributes with the ones provided by the filter,
```
JohnRusk: Am I correct in understanding that includeAttrFilter and excludeAttrFilter are identical expect that `return true` and `return false` here are around the other way?

If so, what about changing them so that there's only one type, called something like attrFilter, like this:
```
type attrFilter struct { 
	fileAttributes uint32 
	filePath       string 
        resultIfMatch bool
}
```

And then this code here would become
```
// if a file shares one or more attributes with the ones provided by the filter, it's a match. So return the 
// value that we are supposed to return for matches, else return the opposite
	if attributes & f.fileAttributes > 0 { 
		return f.resultIfMatch
	} 

	return !f.resultIfMatch
```

That would reduce a lot of code duplication.  BTW, maybe there is some better variation of what I've suggested that would also remove the duplication, but might not be done exactly as I've suggested. That would probably be fine too. The key thing is to minimize duplication.

(Sorry if this comment seems like nitpicking.  Based on experience with the codebase over the past 6 months, we've become very conscious of the need to control, and eventually reduce, duplication of code.  So it's something we look hard at now.)

cmd/zc_enumerator.go
```diff
@@ -180,6 +181,10 @@ func passedFilters(filters []objectFilter, storedObject storedObject) bool {
 	if filters != nil && len(filters) > 0 {
 		// loop through the filters, if any of them fail, then return false
 		for _, filter := range filters {
+			msg, supported := filter.doesSupportThisOS()
+			if !supported {
+				glcm.Info(msg)
```
JohnRusk: I'm not completely sure, but I _think_ I'd lean towards this being a glcm.Error call here, instead of Info. Why? Because any time where the user gives us some parameters, and we carry on without fully respecting them, makes me nervous. The worst-case scenario would be that a user fails to see the outputted "info" message, assumes their job worked as they expected, and then does something like deleting their source files - even though what was transferred may not have matched what they asked for. Admittedly that's not super-likely here, but I still lean towards the idea that if we can't do exactly what the user asked for, we fail hard with an error.

@zezha-msft what do you think?

cmd/zt_sync_upload_windows_test.go
```diff
@@ -0,0 +1,90 @@
...
+func (s *cmdIntegrationSuite) TestSyncUploadWithIncludeAttrFlag(c *chk.C) {
+	bsu := getBSU()
+
+	srcDirName := scenarioHelper{}.generateLocalDirectory(c)
+	scenarioHelper{}.generateCommonRemoteScenarioForLocal(c, srcDirName, "")
+
+	// add special files with attributes that we wish to include
+	filesToInclude := []string{"file1.txt", "file2.pdf", "file3.pdf"}
+	scenarioHelper{}.generateLocalFilesFromList(c, srcDirName, filesToInclude)
+	includeAttrsStr := "H;I"
```
JohnRusk: Minor: As you've already done for Exclude, it might be nice if includeAttrsStr was a set of attrs that overlaps with attrList, rather than an exact match. I.e. each has at least one that's not in the other.  Then the test would demonstrate that we are not requiring exact match, only overlap.

cmd/zt_sync_upload_windows_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	rawContainerURLWithSAS := scenarioHelper{}.getRawContainerURLWithSAS(c, containerName)
+	raw := getDefaultSyncRawInput(srcDirName, rawContainerURLWithSAS.String())
+	raw.includeFileAttributes = includeAttrsStr
+
+	runSyncAndVerify(c, raw, func(err error) {
+		c.Assert(err, chk.IsNil)
+		validateUploadTransfersAreScheduled(c, "", "", filesToInclude, mockedRPC)
+	})
+}
```
JohnRusk: I wonder if we should also have a test where we are asserting that --include (the name filter) is AND'd with --include-attributes

I.e. use both in the test, and have some files that match only the name filter, some that match only the attribute filter, and some that match both.

That would define the behaviour when combining filters, and alert us if there was ever any accidental change in it.

cmd/syncEnumerator.go
```diff
@@ -113,8 +113,13 @@ func newSyncUploadEnumerator(cca *cookedSyncCmdArgs) (enumerator *syncEnumerator
 	includeFilters := buildIncludeFilters(cca.include)
 	excludeFilters := buildExcludeFilters(cca.exclude)
 
+	includeAttrFilters := buildIncludeAttrFilters(cca.includeFileAttributes, sourceTraverser.fullPath)
+	excludeAttrFilters := buildExcludeAttrFilters(cca.excludeFileAttributes, sourceTraverser.fullPath)
+
 	// set up the filters in the right order
-	filters := append(includeFilters, excludeFilters...)
+	filters := append(includeFilters, includeAttrFilters...)
```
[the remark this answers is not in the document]
panfeng0007: Comment added in my next commit. Actually all filters appended in `filters` are ANDed. This behavior is what we expected, right?

cmd/zc_attr_filter_unix.go
```diff
@@ -0,0 +1,73 @@
+// +build linux darwin
```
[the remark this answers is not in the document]
panfeng0007: Done.

cmd/zc_attr_filter_unix.go
```diff
@@ -0,0 +1,73 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package cmd
+
+type excludeAttrFilter struct {
+	fileAttributes uint32
```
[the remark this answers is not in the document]
panfeng0007: You are right. We don't need the fields in the non-Windows implementation. However the function prototype for building the attribute filters has to be the same as the Windows implementation, i.e. the function has to pass the same parameters even if the function body doesn't consume them.

cmd/zc_attr_filter_unix.go
panfeng0007: Same as above

cmd/zc_attr_filter_windows.go
panfeng0007: Done.

cmd/zc_attr_filter_windows.go
panfeng0007: I modified the code like you suggested and it did reduce a lot of duplication!

cmd/zt_sync_upload_windows_test.go
panfeng0007: Done.

cmd/zc_enumerator.go
panfeng0007: I agree with the "failing hard if can't fully deliver" approach. I wasn't sure if I should use `glcm.Error` so I used `glcm.Info` as a placeholder.

cmd/zt_sync_upload_windows_test.go
panfeng0007: I have a question about these tests. I ran it on my own Windows computer with my own Azure Blob account but the `TestUploadWithIncludeAttrFlag` kept failing because the interceptor obtained more files than expected (number of files created by `generateCommonRemoteScenarioForLocal` + number of files in `filesToInclude`). I then used "H" and "I" flags, which are not default attributes in most Windows files, but it still got more files than what's in the `filesToInclude` list. Any idea why?

cmd/syncEnumerator.go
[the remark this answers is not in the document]
JohnRusk: Yes.

cmd/zc_enumerator.go
JohnRusk: No reply from Ze yet, but given that you and I both agree @panfeng0007, let's go with glcm.Error. I checked, and we are already using that like that in other places, so its fine for you to do so here too.

cmd/zt_sync_upload_windows_test.go
JohnRusk: Hmm. That's weird. Do you know what the extra files were?

cmd/syncEnumerator.go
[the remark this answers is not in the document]
JohnRusk: PS. Just checking, I don't see the new commit here yet.  That's still coming, right?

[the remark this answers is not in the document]
panfeng0007: Yes I'll push the commit once I finished all the change requests from you.

cmd/zt_sync_upload_windows_test.go
panfeng0007: Not really but they have to come from the 50 files created by `generateCommonRemoteScenarioForLocal`. Last time I ran the test the interceptor caught 33 transfers but only 3 of them were expected, i.e. files from `filesToInclude`. Is there a way to print the list of files from `mockedRPC`?

cmd/syncEnumerator.go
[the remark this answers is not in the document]
panfeng0007: I just pushed my latest commits.

cmd/zt_sync_upload_windows_test.go
```diff
@@ -88,3 +87,35 @@ func (s *cmdIntegrationSuite) TestSyncUploadWithIncludeAttrFlag(c *chk.C) {
...
+
+	srcDirName := scenarioHelper{}.generateLocalDirectory(c)
+	scenarioHelper{}.generateCommonRemoteScenarioForLocal(c, srcDirName, "")
+
+	filesToInclude := []string{"file1.txt", "file2.txt", "file3.pdf"}
+	scenarioHelper{}.generateLocalFilesFromList(c, srcDirName, filesToInclude)
+	includeString := "*.pdf;file*.txt"
+	includeAttrsStr := "H;I;S"
+	attrList := []string{"H", "I", "C"}
+	scenarioHelper{}.setAttributesForLocalFiles(c, srcDirName, filesToInclude, attrList)
```
JohnRusk: I'm not sure if this test is actually making any assertion that would distinguish between ANDing and ORing - since all filesToInclude match both the include string and the attrs string.  It would be nice to have at least 3 files in the file set, with the following characteristics: 
* One that only matches includeString, but not the attributes
* One that only matches an attribute, but not the include string
* And one that matches both (and the test would assert that only this one gets transferred)

To setup, you've had to maybe change the include string to be just *.txt, and then pass filesToInclude[1:] to setAttributesForLocalFiles.  The expected transfer would be just filesToInclude[1:2], since only that matches both.

(And I guess "filesToInclude" is not really the right variable name anymore ;-)

cmd/zt_sync_upload_windows_test.go
JohnRusk: I think the mockedRPC must store the list somewhere (otherwise validateUploadTransfersAreScheduled wouldn't work).  I'd suggest you look at where it stores them, and then temporarily add your own code to just call fmt.PrintLn to dump them out.

cmd/zc_attr_filter_windows.go
```diff
@@ -0,0 +1,105 @@
...
+	supported = true
+	return
+}
+
+// TODO: review handling of filename for long path support
+func (f *attrFilter) doesPass(storedObject storedObject) bool {
+	fileName := filepath.Join(f.filePath, storedObject.name)
+	lpFileName, err := syscall.UTF16PtrFromString(fileName)
+
+	// If it fails to retrive the pointer from file path, let the filter pass
```
adreed-msft: I feel as though it should probably panic if we hit an error here. This conversion has no reason to fail.

cmd/zc_attr_filter_windows.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return true
+	}
+
+	attributes, err := syscall.GetFileAttributes(lpFileName)
+
+	if err != nil {
+		return true
```
adreed-msft: I'm a bit averse to passing a filter if there's an error while testing it; it may let unintended files through.

Personally, I think that we should just return a fail on the filter and perhaps log something about the error.

cmd/zc_attr_filter_windows.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"path/filepath"
+	"syscall"
+	"strings"
+)
+
+type attrFilter struct {
+	fileAttributes uint32
+	filePath       string
+	resultIfMatch  bool
```
adreed-msft: For clarity's sake, I'd recommend calling this "include". I felt initially a bit confused by the actual behavior of this flag until I did a little reading.

adreed-msft: That being said, there's also totally an argument to be had that `resultIfMatch` makes more sense-- as this is more literally what it does. My view is just that `resultIfMatch` is just a bit of an obtuse way to say "include these attributes"

---
In reply to: [305002730](https://github.com/Azure/azure-storage-azcopy/pull/501#discussion_r305002730) [](ancestors = 305002730)

cmd/zt_sync_upload_windows_test.go
adreed-msft: Ditto, this only creates one set of files with both sets of includable items.

---
In reply to: [304708701](https://github.com/Azure/azure-storage-azcopy/pull/501#discussion_r304708701) [](ancestors = 304708701)

cmd/zt_sync_upload_windows_test.go
```diff
@@ -0,0 +1,125 @@
...
+		validateUploadTransfersAreScheduled(c, "", "", filesToInclude, mockedRPC)
+	})
+}
+
+// Asserting that name filter and attribute filter are ANDed
+// Create one file that matches only the name filter
+// Create one file that matches only the attribute filter
+// Create one file that matches both
+// Only the last file should be transferred
+func (s *cmdIntegrationSuite) TestSyncUploadWithIncludeAndIncludeAttrFlags(c *chk.C) {
```
adreed-msft: You have an (intended) AND case for includes, could you do the same for excludes?

cmd/zc_attr_filter_windows.go
JohnRusk: Good question. But is it better to transfer too much, or too little?  (I.e. to fall back to passing, or to not passing).  E.g. maybe better to transfer too much?  (Especially if, say, after the transfer, the user is about to delete the source!)

I'd be very interested to hear what @zezha-msft thinks, as original architect of the filter interface.  Should there be an error return added to the function signature.

Also, we need to think about the case where the file has been deleted or renamed, between the time the enumeration process saw it, and the time it got filtered.  (I'm not sure how big that time gap may be).  

In any case, once we've thought about all of the above, it's possible the right Boolean to return isn't a hard coded true or false, but rather `resultIfMatch` or `!resultIfMatch` - i.e. the right response to a failure maybe opposite for include vs exclude filters.

Thanks @adreed-msft for seeing this issue.

cmd/zc_attr_filter_windows.go
JohnRusk: What about `isIncludeFilter` - since the word `include` on its own may also initially confuse some folks.  FWIW, I like `isIncludeFilter` better than `resultIfMatch`

cmd/zt_sync_upload_windows_test.go
panfeng0007: Done. By ANDing exclude filter and exclude attribute filter, **none** of the 3 special files (one matches name filter, one matches attribute filter and one matches both) should be transferred.

cmd/zc_attr_filter_windows.go
panfeng0007: Yeah. I like `isIncludeFilter` better, too. I don't care too much about what exactly this Boolean should be called, as long as the name clarifies what the variable is meant to do.

cmd/zc_attr_filter_windows.go
panfeng0007: I added a "TODO" comment. 😉

cmd/zc_attr_filter_windows.go
panfeng0007: Sounds good. I removed the `if err != nil` block.

panfeng0007:
Just rebased my commits on top of `dev` branch of the upstream.

JohnRusk:
@adreed-msft can you take an look a finish off this PR please? Are there any other changes required?  Once we have two approvals, you or I can merge it. 

(@panfeng0007 I don't think you'll have rights to merge)

adreed-msft:
LGTM. Still awaiting @zezha-msft's view on how we should handle GetFileAttributes failing-- once that's done, this should be good to merge.

panfeng0007:
@adreed-msft Sounds good! I’ll update the code accordingly then. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

cmd/zc_attr_filter_windows.go
zezha-msft: That's a great question. I think the current behavior is fine: if the file was deleted (or changed somehow so that getting attributes fails), this filter would not be the right place to throw that error and say the transfer should not proceed.

From the design perspective, if a filter fails to perform its check, it should act like it didn't exist in the first place. In general, the strategy is to be as fault tolerant as possible, and avoid unnecessary support escalations.

zezha-msft:
@panfeng0007 how was your experience with ramping up on the project? please don't hesitate if you have any feedback concerning the architecture/design. 😄

azure-pipelines.yml
```diff
@@ -1,14 +1,14 @@
 trigger:
   branches:
     include:
-      - master
       - dev
+      - master
```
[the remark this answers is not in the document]
panfeng0007: I created my own branch and my own Azure DevOps account so I only wanted this branch to trigger a build.

panfeng0007:
> @panfeng0007 how was your experience with ramping up on the project? please don't hesitate if you have any feedback concerning the architecture/design. 😄

I just pushed my final changes and rebased one more time to be on top of the `dev` branch. It was really a great pleasure working on this project with you people and the architecture/design that you have allowed me to add features in a minimal intrusive way (meaning I was able to make less change to the current codebase). I really enjoyed developing in Golang and I look forward to contributing more on this project in the future! 😃

zezha-msft:
@panfeng0007 that's great to hear! We'd gladly take more contributions from you in the future!

cmd/zt_sync_upload_windows_test.go
zezha-msft: There was a minor bug, I've fixed it [here](https://github.com/Azure/azure-storage-azcopy/pull/527). We need to use the relative path to construct the full URL, otherwise the get attribute syscall fails, and the filter passes the file, so the all the files inside folders failed, that's where the 30 files came from. I added a INFO output if any error gets returned from the get attribute call, so that it's more visible.

PR #509, opened 2019-07-18
Added list-of-files support for rm command

zezha-msft:
This PR adds support for:
- list-of-files for rm command, including Blob, File, ADLS Gen2

This PR enables future work (as part of refactoring):
- list-of-files for File
- list-of-files for S2S (Blob)

Note:
- Travis seems to be acting up again. I'll rerun builds later.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -209,21 +209,27 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	if err != nil {
 		return cooked, err
 	}
-	// User can provide either listOfFilesToCopy or include since listOFFiles mentions
-	// file names to include explicitly and include file may mention the pattern.
-	// This could conflict enumerating the files to queue up for transfer.
-	if len(raw.listOfFilesToCopy) > 0 && len(raw.legacyInclude) > 0 {
-		return cooked, fmt.Errorf("both list-of-files and include flag were provided." +
-			"Only one of them is allowed at a time")
-	}
 
-	// new implementation of list-of-files only works for remove command for now
-	if raw.fromTo == common.EFromTo.BlobTrash().String() || raw.fromTo == common.EFromTo.FileTrash().String() ||
-		raw.fromTo == common.EFromTo.BlobFSTrash().String() {
+	// Note: new implementation of list-of-files only works for remove command for now
+	if raw.fromTo == common.EFromTo.BlobTrash().String() || raw.fromTo == common.EFromTo.FileTrash().String() {
+		cooked.listOfFilesLocation = raw.listOfFilesToCopy
+	} else if raw.fromTo == common.EFromTo.BlobFSTrash().String() {
+		if len(raw.listOfFilesToCopy) > 0 && (len(raw.include) > 0 || len(raw.exclude) > 0) {
+			return cooked, fmt.Errorf("include/exclude flags are not supported for this destination")
+		}
+
 		cooked.listOfFilesLocation = raw.listOfFilesToCopy
 	} else if len(raw.listOfFilesToCopy) > 0 {
+		// TODO remove this legacy implementation after copy enumerator refactoring
 
-		// TODO remove this legacy implementation
+		// User can provide either listOfFilesToCopy or include since listOFFiles mentions
+		// file names to include explicitly and include file may mention the pattern.
+		// This could conflict enumerating the files to queue up for transfer.
+		if len(raw.listOfFilesToCopy) > 0 && len(raw.legacyInclude) > 0 {
+			return cooked, fmt.Errorf("both list-of-files and include flag were provided." +
+				"Only one of them is allowed at a time")
+		}
+
 		// If the user provided the list of files explicitly to be copied, then parse the argument
 		// The user passes the location of json file which will have the list of files to be copied.
 		// The "json file" is chosen as input because there is limit on the number of characters that
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -213,15 +213,23 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	// file names to include explicitly and include file may mention the pattern.
 	// This could conflict enumerating the files to queue up for transfer.
 	if len(raw.listOfFilesToCopy) > 0 && len(raw.legacyInclude) > 0 {
-		return cooked, fmt.Errorf("user provided argument with both listOfFilesToCopy and include flag. Only one should be provided")
+		return cooked, fmt.Errorf("both list-of-files and include flag were provided." +
+			"Only one of them is allowed at a time")
 	}
 
-	// If the user provided the list of files explicitly to be copied, then parse the argument
-	// The user passes the location of json file which will have the list of files to be copied.
-	// The "json file" is chosen as input because there is limit on the number of characters that
-	// can be supplied with the argument, but Storage Explorer folks requirements was not to impose
-	// any limit on the number of files that can be copied.
-	if len(raw.listOfFilesToCopy) > 0 {
+	// new implementation of list-of-files only works for remove command for now
+	if raw.fromTo == common.EFromTo.BlobTrash().String() || raw.fromTo == common.EFromTo.FileTrash().String() ||
+		raw.fromTo == common.EFromTo.BlobFSTrash().String() {
+		cooked.listOfFilesLocation = raw.listOfFilesToCopy
+	} else if len(raw.listOfFilesToCopy) > 0 {
+
+		// TODO remove this legacy implementation
+		// If the user provided the list of files explicitly to be copied, then parse the argument
+		// The user passes the location of json file which will have the list of files to be copied.
+		// The "json file" is chosen as input because there is limit on the number of characters that
+		// can be supplied with the argument, but Storage Explorer folks requirements was not to impose
+		// any limit on the number of files that can be copied.
+
 		jsonFile, err := os.Open(raw.listOfFilesToCopy)
 		if err != nil {
 			return cooked, fmt.Errorf("cannot open %s file passed with the list-of-file flag", raw.listOfFilesToCopy)
@@ -486,11 +494,12 @@ type cookedCopyCmdArgs struct {
 	excludePatterns []string
 
 	// filters from flags
-	listOfFilesToCopy []string
-	recursive         bool
-	followSymlinks    bool
-	withSnapshots     bool
-	forceWrite        bool
+	listOfFilesToCopy   []string
+	listOfFilesLocation string
+	recursive           bool
+	followSymlinks      bool
+	withSnapshots       bool
+	forceWrite          bool
 
 	// options from flags
 	blockSize uint32
@@ -874,21 +883,18 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 		e := copyDownloadBlobFSEnumerator(jobPartOrder)
 		err = e.enumerate(cca)
 	case common.EFromTo.BlobTrash():
-		e, createErr := newRemoveBlobEnumerator(cca)
-		if createErr != nil {
-			return createErr
-		}
-
-		err = e.enumerate()
+		fallthrough
 	case common.EFromTo.FileTrash():
-		e, createErr := newRemoveFileEnumerator(cca)
+		e, createErr := newRemoveEnumerator(cca)
 		if createErr != nil {
 			return createErr
 		}
 
 		err = e.enumerate()
+
 	case common.EFromTo.BlobFSTrash():
-		msg, err := removeBfsResource(cca)
+		// TODO merge with BlobTrash case
+		msg, err := removeBfsResources(cca)
 		if err == nil {
 			glcm.Exit(func(format common.OutputFormat) string {
 				return msg
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -211,10 +211,14 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	}
 
 	// Note: new implementation of list-of-files only works for remove command for now
-	if raw.fromTo == common.EFromTo.BlobTrash().String() || raw.fromTo == common.EFromTo.FileTrash().String() {
+	if cooked.fromTo == common.EFromTo.BlobTrash() || cooked.fromTo == common.EFromTo.FileTrash() {
 		cooked.listOfFilesLocation = raw.listOfFilesToCopy
-	} else if raw.fromTo == common.EFromTo.BlobFSTrash().String() {
-		if len(raw.listOfFilesToCopy) > 0 && (len(raw.include) > 0 || len(raw.exclude) > 0) {
+	} else if cooked.fromTo == common.EFromTo.BlobFSTrash() {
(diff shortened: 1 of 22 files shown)


adreed-msft:
FYI, just taking a look at zc_traverser_list.go's implementation like you'd suggested the other day.

cmd/zc_traverser_list.go
```diff
@@ -0,0 +1,127 @@
...
+		// fetch an appropriate traverser, and go through the child path, which could be
+		//   1. a single entity
+		//   2. a directory entity that needs to be scanned
+		childTraverser, err := l.childTraverserGenerator(childPath)
+		if err != nil {
+			glcm.Info(fmt.Sprintf("Skipping %s due to error %s", childPath, err))
+			continue
+		}
+
+		err = childTraverser.traverse(processor, filters)
```
adreed-msft: Quick note: Traversers return relative paths from the origin. This means that if I supply you a list akin to the following:

FolderA/
FolderB/
FolderC/file.txt

and a source with the following structure:

FolderA/file.txt
FolderA/abc.txt
FolderB/file.txt
FolderB/def.txt
FolderC/file.txt

You will process with the following relative paths:

"/file.txt"
"/abc.txt"
"/file.txt"
"/def.txt"
""

To solve this in my implementation of a glob traverser, I simply added a middleman processor that'd create the ACTUAL relative path using the original base path.

My implementation was roughly as follows. t.GetBasePath() just stripped the wildcards from the base path and used that to construct the relative path from the base.

```go
nt := newLocalTraverser(v, t.recursive, t.incrementEnumerationCounter)
  injectedProcessor := func(object storedObject) error {
   relativePath := strings.TrimPrefix(v, t.getBasePath())
   object.relativePath = relativePath

   return processor(object)
  }

  err = nt.traverse(injectedProcessor, filters)
  if err != nil {
   return err
  }```

File: cmd/zc_traverser_list.go [hunk 2]
```diff
@@ -0,0 +1,127 @@
...
+
+	// in case the list of entities was not read properly, we can no longer continue enumeration
+	if err := scanner.Err(); err != nil {
+		return fmt.Errorf("unable to scan %s to retrieve the required list of entities to transfer", l.fileLocation)
+	}
+
+	return nil
+}
+
+func (l *listTraverser) setPreProcessingCallback(callback preProcessingCallback) {
```
adreed-msft: Is it possible we could just hand this off to all sub-traversers?

cmd/zc_traverser_list.go
```diff
@@ -0,0 +1,127 @@
...
+	"fmt"
+	"github.com/Azure/azure-storage-azcopy/common"
+	"net/url"
+	"os"
+)
+
+// a meta traverser that goes through a list of paths (potentially directory entities) and scans them one by one
+// behaves like a single traverser (basically a "traverser of traverser")
+type listTraverser struct {
+	fileLocation            string
```
adreed-msft: Suggestion: make this a io.Reader or a channel so that we can use it internally. I'd like to use this to do a filepath.Glob and feed it into the listTraverser.

cmd/zc_traverser_list.go
```diff
@@ -0,0 +1,127 @@
...
+			return traverser, nil
+		}
+	case common.ELocation.File():
+		traverserGenerator = func(relativeChildPath string) (resourceTraverser, error) {
+			// assume child path is not URL-encoded yet, this is consistent with the behavior of previous implementation
+			childURL, _ := url.Parse(parent)
+			childURL.Path = common.GenerateFullPath(childURL.Path, relativeChildPath)
+
+			// construct traverser that goes through child
+			traverser, err := newFileTraverserForCopy(childURL.String(), parentSAS, credential, recursive)
```
adreed-msft: Can we stop making these highly situational constructors and just have a _global_ traverser generator? Ask to take a look at my initResourceTraverser function sometime today, since it currently only lives on a local repo since the refactor is still very much a WIP.

adreed-msft: The reason I'm asking is because all this is doing is creating a lot of pointless code duplication that we're inevitably going to have a pain cleaning up.

---
In reply to: [305028492](https://github.com/Azure/azure-storage-azcopy/pull/509#discussion_r305028492) [](ancestors = 305028492)

cmd/zc_traverser_list.go
adreed-msft: Pardon me, it seems as though I missed line 72, which seams to clean this up.

cmd/zc_traverser_list.go
adreed-msft: (alternatively, execute the specified callback before we execute our specified callback)

cmd/zc_traverser_list.go
zezha-msft: No worries! I thought of doing the same thing, but I wanted the full path to be present before the filters operate on the stored objects.

I should adopt your implementation though, so that we can better merge our code.

cmd/zc_traverser_list.go
zezha-msft: Yes, but the drawback is that filters see the incomplete relative path. I think we can live with it ok.

cmd/zc_traverser_list.go
zezha-msft: Ok!

cmd/zc_traverser_list.go
zezha-msft: Yes, we can certainly roll them into one constructor.

cmd/copy.go
```diff
@@ -209,19 +209,33 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
...
-		return cooked, fmt.Errorf("user provided argument with both listOfFilesToCopy and include flag. Only one should be provided")
-	}
 
-	// If the user provided the list of files explicitly to be copied, then parse the argument
-	// The user passes the location of json file which will have the list of files to be copied.
-	// The "json file" is chosen as input because there is limit on the number of characters that
-	// can be supplied with the argument, but Storage Explorer folks requirements was not to impose
-	// any limit on the number of files that can be copied.
-	if len(raw.listOfFilesToCopy) > 0 {
+	// Note: new implementation of list-of-files only works for remove command for now
```
adreed-msft: Note-to-self: Remove this check during refactoring and test list-of-files on my own.

cmd/copy.go
```diff
@@ -209,19 +209,33 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
...
+	if raw.fromTo == common.EFromTo.BlobTrash().String() || raw.fromTo == common.EFromTo.FileTrash().String() {
```
adreed-msft: Minor: cook the fromTo first before testing against it. This _works_, but it's a bit hacky in the sense that it jumps ahead of the line.

cmd/copy.go
```diff
@@ -874,21 +889,18 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 		e := copyDownloadBlobFSEnumerator(jobPartOrder)
 		err = e.enumerate(cca)
 	case common.EFromTo.BlobTrash():
-		e, createErr := newRemoveBlobEnumerator(cca)
-		if createErr != nil {
-			return createErr
-		}
-
-		err = e.enumerate()
+		fallthrough
```
adreed-msft: We can use commas instead of fallthrough statements, ex:

```go
switch myString {
case "xyz", "abc", "def":
    // do thing
case "ghi":
    // do thing
}```

File: cmd/copyTraverser.go [hunk 8]
```diff
@@ -23,55 +23,47 @@ package cmd
...
+	"fmt"
+	"github.com/Azure/azure-storage-azcopy/common"
 	"github.com/Azure/azure-storage-azcopy/ste"
 	"net/url"
 	"strings"
 )
 
-func newBlobTraverserForRemove(cca *cookedCopyCmdArgs) (t *blobTraverser, err error) {
+// TODO implement for local and ADLS Gen2
+func newTraverserForCopy(targetURL string, targetSAS string, targetType common.Location,
```
adreed-msft: This is better than before, but we should REALLY have a global newTraverser function.

I already have this in the refactor. Perhaps my PR after the refactor will be cleaning all of these up.

adreed-msft:
> var specialNames = []string{

NOT CR STUFF: This list is something special. I think it needs more memes though.

---
Refers to: cmd/zt_scenario_helpers_for_test.go:46 in 36ad21e. [](commit_id = 36ad21e4242acf41c00a4c5f4347a10940d82560, deletion_comment = False)

adreed-msft:
Just minor suggestions and notes to self, really.

cmd/copy.go
```diff
@@ -209,19 +209,33 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
...
-	// The "json file" is chosen as input because there is limit on the number of characters that
-	// can be supplied with the argument, but Storage Explorer folks requirements was not to impose
-	// any limit on the number of files that can be copied.
-	if len(raw.listOfFilesToCopy) > 0 {
+	// Note: new implementation of list-of-files only works for remove command for now
+	if raw.fromTo == common.EFromTo.BlobTrash().String() || raw.fromTo == common.EFromTo.FileTrash().String() {
+		cooked.listOfFilesLocation = raw.listOfFilesToCopy
+	} else if raw.fromTo == common.EFromTo.BlobFSTrash().String() {
+		if len(raw.listOfFilesToCopy) > 0 && (len(raw.include) > 0 || len(raw.exclude) > 0) {
+			return cooked, fmt.Errorf("include/exclude flags are not supported for this destination")
```
JohnRusk: Are they "not supported for this destination" in all cases, or just "not supported for this destinatino when list-of-files is supplied?" The code implies the lattre, but the error message says the former

cmd/remove.go
```diff
@@ -86,4 +86,5 @@ func init() {
 	deleteCmd.PersistentFlags().StringVar(&raw.logVerbosity, "log-level", "INFO", "define the log verbosity for the log file, available levels: INFO(all requests/responses), WARNING(slow responses), ERROR(only failed requests), and NONE(no output logs).")
 	deleteCmd.PersistentFlags().StringVar(&raw.include, "include", "", "only include files whose name matches the pattern list. Example: *.jpg;*.pdf;exactName")
 	deleteCmd.PersistentFlags().StringVar(&raw.exclude, "exclude", "", "exclude files whose name matches the pattern list. Example: *.jpg;*.pdf;exactName")
+	deleteCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "defines the location of a file which contains the list of files/directories to be deleted")
```
JohnRusk: Maybe add something like "One file per line, delimited by line breaks"

cmd/removeEnumerator.go
```diff
@@ -21,63 +21,43 @@
...
-	transferScheduler := newRemoveTransferProcessor(cca, NumOfFilesPerDispatchJobPart, isSingleBlob)
-	includeFilters := buildIncludeFilters(cca.includePatterns)
-	excludeFilters := buildExcludeFilters(cca.excludePatterns)
-
-	// set up the filters in the right order
-	filters := append(includeFilters, excludeFilters...)
-
-	finalize := func() error {
-		jobInitiated, err := transferScheduler.dispatchFinalPart()
+	if len(cca.listOfFilesLocation) > 0 {
```
JohnRusk: Can we detect, and throw a clear error, if it is a JSON file. Maybe just checking extension would do....

cmd/removeEnumerator.go
```diff
@@ -147,21 +132,56 @@ func removeBfsResource(cca *cookedCopyCmdArgs) (successMessage string, err error
...
+
+	// in case the list of entities was not read properly, we can no longer continue enumeration
+	if err := scanner.Err(); err != nil {
+		return "", fmt.Errorf("unable to scan %s to retrieve the required list of entities to transfer", cca.listOfFilesLocation)
+	}
+
+	return fmt.Sprintf("Successfully removed %v entities.", successCount), nil
+}
+
+// TODO move after ADLS/Blob interop goes public
```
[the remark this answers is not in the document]
JohnRusk: Or is the plan to keep the single object support for BFS (so you can delete a whole tree if you want) but convert just _list of files_ to use the blob API?

cmd/copy.go
zezha-msft: Oops, yes you are right, I'll fix it.

cmd/copyTraverser.go
zezha-msft: ok!

cmd/zc_traverser_list.go
```diff
@@ -0,0 +1,91 @@
...
+		// the results are indeed starting right under the parent
+		// ex: parent = /usr/foo
+		// case 1: child1 is a file under the parent
+		//         the relative path returned by the child traverser would be ""
+		//         it should be "child1" instead
+		// case 2: child2 is a directory, and it has items under it such as child2/grandchild1
+		//         the relative path returned by the child traverser would be "grandchild1"
+		//         it should be "child2/grandchild1" instead
+		preProcessor := func(object storedObject) error {
+			if object.relativePath == "" {
```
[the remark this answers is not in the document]
zezha-msft: Sorry for the confusion! A better way to think of it (I'll update the comments in zc_enumerator too) is that the rootDir is given by the user, sometimes it turns out the rootDir is actually pointing to a file. So in that case, the relative path is empty since rootDir already points to it, relatively speaking the path is nothing.

rootDir=/var/a/b/c/d/e/f.pdf fullPath=/var/a/b/c/d/e/f.pdf => relativePath=""

cmd/zc_traverser_list.go
```diff
@@ -0,0 +1,91 @@
...
+				// if the child is a single file
+				object.relativePath = childPath
```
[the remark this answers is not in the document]
zezha-msft: Ah that's right, I'll fix it

cmd/removeEnumerator.go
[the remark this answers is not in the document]
zezha-msft: The current expectation is that the rename and delete operations will both be added to the blob endpoint, and will work even when HNS is not enabled (a marker will be returned). So our current implementation for azbfs might actually become the blob implementation. Chances are the API interfaces will be redesigned to fit the "blob standard".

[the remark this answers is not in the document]
zezha-msft: @adreed-msft FYI

cmd/copy.go
zezha-msft: Good catch, they are never supported!

This makes me think though, when the ADLS Gen2 interop happens eventually, maybe we have to keep the current blob implementation so that users can have this kind of functionality, and we can shift to the new directory delete API if no include/exclude is supplied. I'll add a comment.

cmd/remove.go
zezha-msft: I'll make the flag public, since it seems to be a sensible feature for a lot of users.

cmd/removeEnumerator.go
zezha-msft: I'm hesitant to rely on the extension as the file content might be in the right format despite the name. May be validate the first line is not "{"? I'm not sure.

zezha-msft:
Had to rebase to resolve conflicts with dev branch.

PR #510, opened 2019-07-21
Fix init of log sanitizer, so can use AZCOPY_PROFILE_CPU again

JohnRusk:
Had to fix this bug as part of the highconcurrency fixes (coming soon in separate PR)

### DIFF
--- a/common/logSanitizer.go
+++ b/common/logSanitizer.go
@@ -43,7 +43,7 @@ func NewAzCopyLogSanitizer() pipeline.LogSanitizer {
 }
 
 var sensitiveQueryStringKeys = []string{
-	strings.ToLower(SigAzure),
+	"sig", // was strings.ToLower(SigAzure), but that isn't fully init-order-safe, see init() method below
 	//omitted because covered by "signature" below strings.ToLower(SigXAmzForAws),
 	"signature", // covers both "signature" and x-amz-signature. The former may be used in AWS authorization headers. Not sure if we ever log those, but may as well redact them if we do
 	"token",     // seems worth removing in case something uses it one day (e.g. if we support a new backend)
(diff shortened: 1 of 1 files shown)


common/logSanitizer.go
```diff
@@ -99,5 +100,19 @@ func init() {
...
+		// see comment below
+		if key == strings.ToLower(SigAzure) {
+			mapContainsAzureSig = true
+		}
+	}
+
+	// Double check.
+	// We can't directly do strings.ToLower(SigAzure) in the declaration of
+	// sensitiveRegexMap itself, since it won't be ready if something else logs before that
+	// initialization has happened (as can be the case e.g. init of lifecylemanager).
```
zezha-msft: a bit confused by this, what's the bug and how is this fixing it?

JohnRusk: Ha ha, I was confused too when I saw it. The bug goes like this:

1. lifecycleManager has a global singleton called lcm that is initialized by an in-line func.  var lcm =  func(){…}()
2. lifecycleManager.go comes alphabetically before logSanitizer.go, so its variable initialization happens before any initialization in logSanitizer.go
3. The in-line func that initialize the variable in lifecylceManager triggers a glcm Info message if it is enabling perf profiling.
4. That hits the logSanitizer … _before_ the variable init in logSanitizer has run.
5. So logSanitizer's `sensitiveQueryStringKeys`  is not fully initialized at that point. Specifically, it has its correct size (3 elements), and any elements that are constants have their correct values. BUT any element that requires code to initizalize it (e.g. `strings.ToLower(SigAzure)`) has value "" because the code that initializes it _has not yet run_ (as per 2).
6. So the glcm.Info call, into logSanitizer, fails, when it hits that "" value.

Another possible fix would have been to move the lcm init to an actual 'init' function (since those run after all variable initialization) but that only solves this one instance of the issue, and won't protect against any other recurrences if other in-line variable initialization functions trigger sanitization.  So if figured it was better just to make everything constants.

JohnRusk: PS. I don't actually recall seeing any documentation that says that the constants will all "be there" in a variable like `sensitiveQueryStrings` even before any `code` for it's initialization has run.  However, that's the behavior I observe, and it does seem to make sense logically (e.g. compiler pops the constants in)

zezha-msft: Ahhh I see, that's certainly not a trivial bug to figure out. Glad you got it.

adreed-msft:
lgtm

PR #511, opened 2019-07-22
Experimental/high concurrency

JohnRusk:
Addresses a number of things needed for high concurrency:

1. Briefer logging by default (don't log request and response headers for successful operations)
2. Solve dialer-related issues that were creating too many threads. Now we don't get the explosion to 10,000 threads that we sometimes got before, and even our steady-state ("not exploding") case has a thread count about 10x lower than before.  (Most of those old threads were mostly unused, but its still tidier not to have them)
3. Enhance chunk status logger to show epilogues and local file creation. (Since those factors are significant when processing large numbers of small files with high concurrency).

This PR does NOT automatically tune the number of Goroutines, nor does it change the default number.  But it does allow users to manually configure high counts (e.g. 900 on a two-core machine; 2000 on a bigger machine) for beneficial performance improvements in "small files" S2S scenarios.


ste/concurrency.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// If we get a negative or ridiculously low value, bring it up to some kind of sensible floor
+	// (and take our chances of running out of total handles - which is effectively a bet that
+	// we were too conservative earlier)
+	if concurrentFilesLimit < defaultTransferInitiationPoolSize {
+		concurrentFilesLimit = defaultTransferInitiationPoolSize
+	}
+	return concurrentFilesLimit
+
+	// TODO: consider adding environment var to optionally allow bringing concurrentFilesLimit down lower than what we calculate here
```
zezha-msft: maybe the user request on limiting simultaneous file uploads can be added

cmd/credentialUtil.go
```diff
@@ -299,9 +300,11 @@ func createBlobPipeline(ctx context.Context, credInfo common.CredentialInfo) (pi
 			MaxRetryDelay: ste.UploadMaxRetryDelay,
 		},
 		nil,
-		ste.NewAzcopyHTTPClient()), nil
+		ste.NewAzcopyHTTPClient(frontEndMaxIdleConnectionsPerHost)), nil
 }
 
+const frontEndMaxIdleConnectionsPerHost = http.DefaultMaxIdleConnsPerHost
```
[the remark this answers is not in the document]
JohnRusk: It was just my way of documenting that in the front end we deliberately don't use the large value that we use in the STE.  I propose to leave it as-is.

ste/concurrency.go
JohnRusk: @zezha-msft I've added this to the PR. It was more code change than I expected, in order to avoid duplication.  But good to have that setting configurable.  Interested to know if you can think of a better name for the environment variable, and if you think its OK that it doesn't _really_ configure the number of files _exactly_.

It doesn't really configure number of files _exactly_ because number of files != size of transfer initiation pool, in cases where the files are very small and we get unprocessed files queueing up in our chunkfunc channels.  But naming it something like "transfer initiation pool size" in the environment variable would be confusing; and making it _really_ control the number of files (accurately in all cases) would be a bigger code change for little if any benefit.

adreed-msft:
Looks good to me as far as I can tell

ste/concurrency.go
JohnRusk: Just waiting on your re-review @zezha-msft , for the changes added in response to this comment

common/environment.go
```diff
@@ -61,7 +62,14 @@ func (EnvironmentVariable) CertificatePassword() EnvironmentVariable {
...
 	return EnvironmentVariable{
 		Name:        "AZCOPY_CONCURRENCY_VALUE",
-		Description: "Overrides how many Go Routines work on transfers. By default, this number is determined based on the number of logical cores on the machine.",
+		Description: "Overrides how many HTTP connections work on transfers. By default, this number is determined based on the number of logical cores on the machine.",
+	}
+}
+
+func (EnvironmentVariable) TransferInitiationPoolSize() EnvironmentVariable {
+	return EnvironmentVariable{
+		Name:        "AZCOPY_CONCURRENT_FILES",
```
zezha-msft: Looks good!

PR #512, opened 2019-07-22
Implement a S3 traverser

adreed-msft:
S3 Traverser to add to S2S functionality in the refactor.

This is a draft until we decide upon how to implement service-level traversal. At that point, this will contain S3 service level traversal.

### DIFF
--- a/cmd/zc_traverser_s3.go
+++ b/cmd/zc_traverser_s3.go
@@ -1,0 +1,128 @@
+package cmd
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/minio/minio-go"
+
+	"github.com/Azure/azure-storage-azcopy/common"
+)
+
+type s3Traverser struct {
+	rawURL    *url.URL // No pipeline needed for S3
+	ctx       context.Context
+	recursive bool
+
+	s3URLParts s3URLPartsExtension
+	s3Client   *minio.Client
+
+	// A generic function to notify that a new stored object has been enumerated
+	incrementEnumerationCounter func()
+}
+
+func (t *s3Traverser) traverse(processor objectProcessor, filters []objectFilter) (err error) {
+	// Check if resource is a single object.
+	if t.s3URLParts.IsObjectSyntactically() && !t.s3URLParts.IsDirectorySyntactically() && !t.s3URLParts.IsBucketSyntactically() {
+		objectPath := strings.Split(t.s3URLParts.ObjectKey, "/")
+		objectName := objectPath[len(objectPath)-1]
+
+		oi, err := t.s3Client.StatObject(t.s3URLParts.BucketName, t.s3URLParts.ObjectKey, minio.StatObjectOptions{})
+
+		if err != nil {
+			return err
+		}
+
+		err = processIfPassedFilters(filters, newStoredObject(
+			objectName,
+			"", // No need for relative path when we already know the object location
+			oi.LastModified,
+			oi.Size,
+			nil,
+			blobTypeNA,
+		), processor)
+
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	// Check if the resource is a service level URL
+	if t.s3URLParts.IsServiceSyntactically() {
+		// TODO: Let's just figure this one out later.
+		return errors.New("service-level enumeration is not allowed")
+	}
+
+	searchPrefix, _, wildcard := t.s3URLParts.searchObjectPrefixAndPatternFromS3URL()
+
+	if wildcard {
+		return fmt.Errorf("cannot traverse s3 with wildcard")
+	}
+
+	// It's a bucket or virtual directory.
+	for objectInfo := range t.s3Client.ListObjectsV2(t.s3URLParts.BucketName, searchPrefix, t.recursive, t.ctx.Done()) {
+		if objectInfo.Err != nil {
+			return fmt.Errorf("cannot list objects, %v", objectInfo.Err)
+		}
+
+		if objectInfo.StorageClass == "" {
+			// Directories are the only objects without storage classes.
+			continue
+		}
+
+		objectPath := strings.Split(objectInfo.Key, "/")
+		objectName := objectPath[len(objectPath)-1]
+
+		relativePath := strings.TrimPrefix(objectInfo.Key, searchPrefix)
+
+		err = processIfPassedFilters(filters,
+			newStoredObject(
+				objectName,
+				relativePath,
+				objectInfo.LastModified,
+				objectInfo.Size,
+				nil,
+				blobTypeNA,
+			),
+			processor)
+
+		if err != nil {
+			return
+		}
+	}
+	return
+}
+
+func newS3Traverser(rawURL *url.URL, ctx context.Context, recursive bool, incrementEnumerationCounter func()) (t *s3Traverser, err error) {
+	t = &s3Traverser{rawURL: rawURL, ctx: ctx, recursive: recursive, incrementEnumerationCounter: incrementEnumerationCounter}
+
+	// initialize S3 client and URL parts
+	var s3URLParts common.S3URLParts
+	s3URLParts, err = common.NewS3URLParts(*t.rawURL)
+
+	if err != nil {
+		return
+	} else {
+		t.s3URLParts = s3URLPartsExtension{s3URLParts}
+	}
+
+	t.s3Client, err = common.CreateS3Client(
+		t.ctx,
+		common.CredentialInfo{
+			CredentialType: common.ECredentialType.S3AccessKey(),
+			S3CredentialInfo: common.S3CredentialInfo{
+				Endpoint: t.s3URLParts.Endpoint,
+				Region:   t.s3URLParts.Region,
+			},
+		},
+		common.CredentialOpOptions{
+			LogError: glcm.Error,
+		})
+
+	return
+}
--- a/cmd/zt_generic_traverser_test.go
+++ b/cmd/zt_generic_traverser_test.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/Azure/azure-storage-file-go/azfile"
@@ -52,6 +53,11 @@ func (s *genericTraverserSuite) TestTraverserWithSingleObject(c *chk.C) {
 	bfsu := GetBFSSU()
 	filesystemURL, _ := createNewFilesystem(c, bfsu)
 	defer deleteFilesystem(c, filesystemURL)
+
+	s3Client, err := createS3ClientWithMinio(createS3ResOptions{})
+	c.Assert(err, chk.IsNil)
+	bucketName := createNewBucket(c, s3Client, createS3ResOptions{})
+	defer deleteBucket(c, s3Client, bucketName)
 
 	// test two scenarios, either blob is at the root virtual dir, or inside sub virtual dirs
 	for _, storedObjectName := range []string{"sub1/sub2/singleblobisbest", "nosubsingleblob", "满汉全席.txt"} {
@@ -128,6 +134,23 @@ func (s *genericTraverserSuite) TestTraverserWithSingleObject(c *chk.C) {
 
 		c.Assert(localDummyProcessor.record[0].relativePath, chk.Equals, bfsDummyProcessor.record[0].relativePath)
 		c.Assert(localDummyProcessor.record[0].name, chk.Equals, bfsDummyProcessor.record[0].name)
+
+		// set up the bucket with a single file
+		s3List := []string{storedObjectName}
+		scenarioHelper{}.generateObjects(c, s3Client, bucketName, s3List)
+
+		// construct a s3 traverser
+		url := scenarioHelper{}.getRawS3ObjectURL(c, "", bucketName, storedObjectName)
+		S3Traverser, err := newS3Traverser(&url, ctx, false, func() {})
+		c.Assert(err, chk.IsNil)
+
+		s3DummyProcessor := dummyProcessor{}
+		err = S3Traverser.traverse(s3DummyProcessor.process, nil)
+		c.Assert(err, chk.IsNil)
+		c.Assert(len(s3DummyProcessor.record), chk.Equals, 1)
+
+		c.Assert(localDummyProcessor.record[0].relativePath, chk.Equals, s3DummyProcessor.record[0].relativePath)
+		c.Assert(localDummyProcessor.record[0].name, chk.Equals, s3DummyProcessor.record[0].name)
 	}
 }
 
@@ -146,6 +169,11 @@ func (s *genericTraverserSuite) TestTraverserContainerAndLocalDirectory(c *chk.C
 	filesystemURL, _ := createNewFilesystem(c, bfsu)
 	defer deleteFilesystem(c, filesystemURL)
 
+	s3Client, err := createS3ClientWithMinio(createS3ResOptions{})
+	c.Assert(err, chk.IsNil)
+	bucketName := createNewBucket(c, s3Client, createS3ResOptions{})
+	defer deleteBucket(c, s3Client, bucketName)
+
 	// set up the container with numerous blobs
 	fileList := scenarioHelper{}.generateCommonRemoteScenarioForBlob(c, containerURL, "")
 	c.Assert(containerURL, chk.NotNil)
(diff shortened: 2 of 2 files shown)


adreed-msft:
Requesting review in advance, but feel free to submit feedback in the meantime

adreed-msft:
@zezha-msft and @JohnRusk this is good to go for review now.

cmd/zt_generic_service_traverser_test.go
```diff
@@ -0,0 +1,100 @@
...
+	c.Assert(len(fileDummyProcessor.record), chk.Equals, len(localIndexer.indexMap)*len(containerList))
+	c.Assert(len(s3DummyProcessor.record), chk.Equals, len(localIndexer.indexMap)*len(containerList))
+	for _, storedObject := range append(append(blobDummyProcessor.record, fileDummyProcessor.record...), s3DummyProcessor.record...) {
+		correspondingLocalFile, present := localIndexer.indexMap[storedObject.relativePath]
+		_, cnamePresent := cnames[storedObject.containerName]
+
+		c.Assert(present, chk.Equals, true)
+		c.Assert(cnamePresent, chk.Equals, true)
+		c.Assert(correspondingLocalFile.name, chk.Equals, storedObject.name)
+	}
```
zezha-msft: 👍 I like good tests

zezha-msft:
@adreed-msft to confirm, should we review this PR now? In other words, can it be merged right after approvals are given?

adreed-msft:
@zezha-msft yes, this PR is good for review.

cmd/zc_traverser_blob_account.go
```diff
@@ -0,0 +1,93 @@
...
+	accountURL       azblob.ServiceURL
+	p                pipeline.Pipeline
+	ctx              context.Context
+	containerPattern string
+
+	// a generic function to notify that a new stored object has been enumerated
+	incrementEnumerationCounter func()
+}
+
+func (t *blobAccountTraverser) traverse(processor objectProcessor, filters []objectFilter) error {
```
JohnRusk: I like the way this is so clear and concise.  Definitely showing the benefit of the refactoring!

cmd/zc_enumerator.go
```diff
@@ -65,14 +67,15 @@ func (storedObject *storedObject) isMoreRecentThan(storedObject2 storedObject) b
 }
 
 // a constructor is used so that in case the storedObject has to change, the callers would get a compilation error
-func newStoredObject(name string, relativePath string, lmt time.Time, size int64, md5 []byte, blobType azblob.BlobType) storedObject {
+func newStoredObject(name string, relativePath string, lmt time.Time, size int64, md5 []byte, blobType azblob.BlobType, containerName string) storedObject {
```
JohnRusk: I notice that this parameter is always "" in all calls. It's only the middleman processor/decorator that ever sets the value.  So what about removing the param here, and all the "// not needed" comments at the call sites?  Just let this field be something that's only set, in all the cases where its needed, by the middleman/decorator.

cmd/zc_traverser_blobfs_account.go
```diff
@@ -0,0 +1,102 @@
+// Copyright © 2019 Microsoft <wastore@microsoft.com>
```
JohnRusk: Please omit year in new files

cmd/zc_traverser_blobfs_account.go
```diff
@@ -0,0 +1,102 @@
...
+	"errors"
+	"net/url"
+	"path/filepath"
+
+	"github.com/Azure/azure-pipeline-go/pipeline"
+
+	"github.com/Azure/azure-storage-azcopy/azbfs"
+)
+
+// We don't allow S2S from BlobFS, but what this gives us is the ability for users to download entire accounts at once.
```
JohnRusk: Good thinking to make things consistent in this way.  Is there a smoke test case written for this?  (There needs to be, esp since its close enough to a niche usage that it won't naturally get covered in load tests, manual tests etc)

cmd/zc_traverser_blobfs_account.go
```diff
@@ -0,0 +1,102 @@
...
+		for _, v := range resp.Filesystems {
+			var fsName string
+			if v.Name != nil {
+				fsName = *v.Name
+			} else {
+				return errors.New("filesystem listing returned nil container names")
+			}
+
+			if t.fileSystemPattern != "" {
+				if ok, err := filepath.Match(t.fileSystemPattern, fsName); err != nil {
```
JohnRusk: Same comment as I left on the corresponding blob enumeration code.  And, same suggested solution, create a sepanrate methohd for this, and _all_ the account tranversers can call it.(I suggested naming it containerMatchesPattern in previous comment. THat name is probably OK even if re-used for Blob FS, AWS etc... unless you can think of a better word for container that means "top level entity that's immediately under the account level".    e.g. maybe rootNameMatchesPattern????  Not  sure I like that last suggestion any better than containerNameMatchesPattern tho....

cmd/zc_traverser_blobfs_account.go
```diff
@@ -0,0 +1,102 @@
...
+
+			err = fileSystemTraverser.traverse(middlemanProcessor, filters)
+
+			if err != nil {
+				return err
+			}
+		}
+
+		if marker == "" {
+			break
```
JohnRusk: I notice that this marker handling code contains no `marker := nextMarker`.  This implies that either we have a bug, or the way that `ListFilesystemsSegment` handles the marker is a lot cleverer than what the blob code does.  I just took a quick look.  I don't think it is that clever. (It looks like it might be, because the type is a pointer, but I don't believe it ever updates it).

cmd/zc_traverser_blob_account.go
```diff
@@ -0,0 +1,93 @@
...
+			}
+		}
+
+		marker = resp.NextMarker
+	}
+
+	return nil
+}
+
+func newBlobAccountTraverser(rawURL *url.URL, p pipeline.Pipeline, ctx context.Context, incrementEnumerationCounter func()) (t *blobAccountTraverser) {
```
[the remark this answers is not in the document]
JohnRusk: If you do add this param, then for consistency it should be added (with the same panic if true) behavior, to the other account traversers that are added in this PR

cmd/zc_traverser_file_account.go
```diff
@@ -0,0 +1,91 @@
+// Copyright © 2019 Microsoft <wastore@microsoft.com>
```
JohnRusk: Ditto

cmd/zc_traverser_file_account.go
```diff
@@ -0,0 +1,91 @@
...
+		resp, err := t.accountURL.ListSharesSegment(t.ctx, marker, azfile.ListSharesOptions{})
+
+		if err != nil {
+			return err
+		}
+
+		for _, v := range resp.ShareItems {
+			// Match a pattern for the share name and the share name only
+			if t.sharePattern != "" {
+				if ok, err := filepath.Match(t.sharePattern, v.Name); err != nil {
```
JohnRusk: Ditto

cmd/zc_traverser_list.go
```diff
@@ -1,3 +1,23 @@
+// Copyright © 2019 Microsoft <wastore@microsoft.com>
```
JohnRusk: Ditto. And good catch to add the missing header here.

cmd/zc_traverser_s3.go
```diff
@@ -0,0 +1,146 @@
+// Copyright © 2019 Microsoft <wastore@microsoft.com>
```
JohnRusk: Ditto

cmd/zc_traverser_s3.go
```diff
@@ -0,0 +1,146 @@
...
+			return nil
+		}
+	}
+
+	// Append a trailing slash if it is missing.
+	if !strings.HasSuffix(t.s3URLParts.ObjectKey, "/") && t.s3URLParts.ObjectKey != "" {
+		t.s3URLParts.ObjectKey += "/"
+	}
+
+	// Ignore *s in URLs and treat them as normal characters
```
JohnRusk: Worth adding a "why comment" here to say, "... because * can be a real URL character and, after discussion, we decided to treat * in URL as an ordinary character for that reason - i.e. not a wildcard".   Generally, if was significant or non-obvious enough that we had to discuss the reason why, then its worth including the reason why in a comment.

cmd/zc_traverser_s3_service.go
```diff
@@ -0,0 +1,116 @@
...
+	// a generic function to notify that a new stored object has been enumerated
+	incrementEnumerationCounter func()
+}
+
+func (t *s3ServiceTraverser) traverse(processor objectProcessor, filters []objectFilter) error {
+	if bucketInfo, err := t.s3Client.ListBuckets(); err == nil {
+		for _, v := range bucketInfo {
+			// Match a pattern for the bucket name and the bucket name only
+			if t.bucketPattern != "" {
+				if ok, err := filepath.Match(t.bucketPattern, v.Name); err != nil {
```
JohnRusk: Ditto

azbfs/url_service.go
```diff
@@ -1,25 +1,30 @@
...
 func NewServiceURL(url url.URL, p pipeline.Pipeline) ServiceURL {
 	if p == nil {
 		panic("p can't be nil")
 	}
-	client := newManagementClient(url, p)
+	client := filesystemClient{newManagementClient(url, p)}
 	return ServiceURL{client: client}
 }
 
+func (s ServiceURL) ListFilesystemsSegment(ctx context.Context, marker *string) (*FilesystemList, error) {
```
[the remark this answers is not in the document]
adreed-msft: FileSystemURL is specific to a single filesystem. In my opinion, it would not make sense to have to specify a filesystem before we list other filesystems.

cmd/zc_traverser_blob_account.go
```diff
@@ -0,0 +1,93 @@
...
+		resp, err := t.accountURL.ListContainersSegment(t.ctx, marker, azblob.ListContainersSegmentOptions{})
+
+		if err != nil {
+			return err
+		}
+
+		for _, v := range resp.ContainerItems {
+			// Match a pattern for the container name and the container name only.
+			if t.containerPattern != "" {
+				if ok, err := filepath.Match(t.containerPattern, v.Name); err != nil {
```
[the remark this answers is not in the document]
adreed-msft: It's only intended to match one single portion of the URL (the container name). The container name cannot contain slashes, anyway.

We do have a test for this exact functionality, `TestServiceTraverserWithWildcards`. This tests the container name matching.

cmd/zc_traverser_blob_account.go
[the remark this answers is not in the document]
adreed-msft: On one hand, I'd agree that this is a good idea frontend-facing.

On the other hand, I'd argue this is a poor idea code-facing. Why add a seemingly pointless parameter to a function when all we do is error out on the only _other_ possible option? It _really_ defeats the purpose of a parameter, imo. We should be adding this error check to initResourceTraverser when we do the S2S refactor.

cmd/zc_traverser_blob_account.go
```diff
@@ -0,0 +1,93 @@
...
+				} else if !ok {
+					// Ignore the container if it doesn't match the pattern.
+					continue
+				}
+			}
+
+			containerURL := t.accountURL.NewContainerURL(v.Name).URL()
+			containerTraverser := newBlobTraverser(&containerURL, t.p, t.ctx, true, t.incrementEnumerationCounter)
+
+			middlemanProcessor := initAccountMiddlemanProcessor(v.Name, processor)
```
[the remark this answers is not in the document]
adreed-msft: Yeah, I'd agree with this. Done.

azbfs/url_service.go
```diff
@@ -1,25 +1,30 @@
...
+	"context"
 	"net/url"
 
 	"github.com/Azure/azure-pipeline-go/pipeline"
 )
 
 // A ServiceURL represents a URL to the Azure Storage File service allowing you to manipulate file shares.
 type ServiceURL struct {
-	client managementClient
+	client filesystemClient
```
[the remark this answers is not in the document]
adreed-msft: I couldn't get the effect of ListFilesystems here without doing this change, and see the above response.

cmd/zc_traverser_blobfs_account.go
adreed-msft: As these will get integrated into initResourceTraverser during the S2S refactor, I'll include smoke testing for that. As is, there is unit testing for the traversers.

cmd/zc_traverser_blobfs_account.go
adreed-msft: So, azbfs doesn't have a marker struct.

Instead, it just uses a ptr to a string.

adreed-msft: All the marker struct _is_ anyway is just a pointer to a string. Which is odd because it just allocates a new string, and thus we have to update the marker in our code.

cmd/zc_enumerator.go
adreed-msft: Fair point.

cmd/zc_traverser_blobfs_account.go
adreed-msft: Should we change this to be consistent with our other SDKs? I do feel like just using a struct that contains a pointer to another, new string just walks around what's _actually_ happening.

azbfs/url_service.go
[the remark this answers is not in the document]
JohnRusk: I get it, that does make sense.

azbfs/url_service.go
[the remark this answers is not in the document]
JohnRusk: Resolved as per above conversation.

cmd/zc_traverser_blobfs_account.go
JohnRusk: I wouldn't bother making the BlobFS one consistent, because it will be phased out mostly (or entirely) once interop goes GA.  But I do suspect there's a bug in the current zc_traverser_blobfs_account, because I don't think anything is updating the marker to the next marker.  Our code isn't and, contrary to the impression given by the API's use of a pointer, I don't believe the azbfs code is updating it either.

cmd/zc_traverser_blob_account.go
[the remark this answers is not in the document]
JohnRusk: What kinds of wildcards do we currently support in container names in 10.2.1?  Is it just *, or is it everything that filepath.Match supports?

cmd/zc_traverser_blob.go
```diff
@@ -76,14 +76,7 @@ func (t *blobTraverser) traverse(processor objectProcessor, filters []objectFilt
...
 	if isBlob {
-		storedObject := newStoredObject(
-			getObjectNameOnly(blobUrlParts.BlobName),
-			"", // We already know the object's location -- no need.
-			blobProperties.LastModified(),
-			blobProperties.ContentLength(),
-			blobProperties.ContentMD5(),
-			blobProperties.BlobType(),
-			"") // We already know the container -- no need.
+		storedObject := newStoredObject(getObjectNameOnly(blobUrlParts.BlobName), "", blobProperties.LastModified(), blobProperties.ContentLength(), blobProperties.ContentMD5(), blobProperties.BlobType())
```
JohnRusk: Minor: Why the formatting change? The formatting change to a single line makes it harder to read. (And it makes the diffs harder to read.  I'd suggest that generally, when actioning PR comments, best to avoid unnecessary format changes, since any format change makes the diff harder for reviewers to read.  And if fact it will also adversely affect all historical diffs (version-to-version) and use of annotate/blame.  Basically, unnecessary formatting changes make it harder to read code history, since they make diff tools overstate the extent of the change.)

cmd/zc_traverser_blob_account.go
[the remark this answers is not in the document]
adreed-msft: Everything filepath.Match supports.

I gotta wonder if there's a way to _just_ match against *.

cmd/zc_traverser_blob.go
adreed-msft: Oh, it was just accidental. IDE didn't retain formatting when I auto-refactored

cmd/zc_traverser_blobfs_account.go
adreed-msft: I'll take a look.

cmd/zc_traverser_blob_account.go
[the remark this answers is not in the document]
JohnRusk: I guess, given we are already supporting everything that filepath.Match does,  and given your comment above about there being no slashes, lets just continue with it as it is in the PR right now.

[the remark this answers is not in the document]
JohnRusk: I.e. let's not change what you have.

cmd/zc_traverser_blob.go
JohnRusk: Ew, that was nasty on the part of the IDE.

adreed-msft: Resolved.

cmd/zc_traverser_blobfs_account.go
adreed-msft: Turns out this bug was in theory present in both the container and account traversers.

PR #514, opened 2019-07-25
Added doc command to generate documentation in Markdown format

zezha-msft:
Design consideration: in order to generate the doc tree, we must be able to import the root command and invoke the appropriate Cobra API. The easiest way to do that is to add a new command, since it would be in the same `cmd` package and have visibility on the root command.

In addition, it is best to bundle this functionality with the tool itself, to avoid inconsistency problems. In other words, the tool is truly "self-documenting".

Open question:
- Should we hide the `doc` command? Would it be a useful command for some users?

In the future, the tool can also generate man pages, or more interestingly [bash completions](https://github.com/spf13/cobra/blob/master/bash_completions.md).

### DIFF
--- a/ThirdPartyNotice.txt
+++ b/ThirdPartyNotice.txt
@@ -38,6 +38,7 @@ jiacfan/keychain, Copyright (c) 2015 Keybase
 mattn/go-ieproxy, Copyright (c) 2014 mattn
 mattn/go-ieproxy, Copyright (c) 2017 oliverpool
 mattn/go-ieproxy, Copyright (c) 2019 Adele Reed
+cpuguy83/go-md2man, Copyright (c) 2014 Brian Goff
 
 
 The MIT License (MIT)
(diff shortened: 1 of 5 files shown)


adreed-msft:
Generally speaking lgtm

ThirdPartyNotice.txt
```diff
@@ -38,6 +38,7 @@ jiacfan/keychain, Copyright (c) 2015 Keybase
 mattn/go-ieproxy, Copyright (c) 2014 mattn
 mattn/go-ieproxy, Copyright (c) 2017 oliverpool
 mattn/go-ieproxy, Copyright (c) 2019 Adele Reed
+cpuguy83/go-md2man, Copyright (c) 2014 Brian Goff
```
adreed-msft: Out of curiosity, are these dependencies of cobra/doc?

zezha-msft: Yes indeed

Issue #356, opened 2019-04-28
Exclude a directory path and all files under it


benc-uk:
I wish to exclude a entire file path from the sync, e.g. `node_modules/` folder which I do not need and can be very large.

I have tried every combination in the exclude that I can think of and none work, e.g.
```
--exclude="node_modules"
--exclude="*node_modules*"
--exclude="node_modules/*"
--exclude="node_modules/**"
--exclude="**/node_modules/"
etc etc.
```
None work!

Problem only happens with `sync`, when running `copy --recursive true` the exclude string is honoured and works

### Which version of the AzCopy was used? 
azcopy version 10.1.0

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows 10

### What command did you run?
```
azcopy sync nodejs-demoapp https://bcmisc.blob.core.windows.net/backup/dev1\?sv=REMOVED --exclude="*node_modules*"
```

### What problem was encountered?
Exclude doesn't work

### How can we reproduce the problem in the simplest way?
Create a directory with a subdirectory inside it (and some test files inside that), and try to sync the main directory while exluding the sub-directory

### Have you found a mitigation/solution?
No

zezha-msft:
Hi @benc-uk, thanks for reaching out! And sorry for the inconvenience. 

The difference in `sync`'s behavior (compared to `cp`) was very much intentional for the sake of simplifying the user experience. Both `./azcopy sync --help` and the [wiki](https://github.com/Azure/azure-storage-azcopy/wiki/Sync-command-guide) explain the meaning of the exclude flag:

> exclude files whose name matches the pattern list. Example: *.jpg;*.pdf;exactName

It's good to hear your feedback though. We can perhaps rename the current flag to `--exclude-files`, and add a new flag `--exclude-directories` to address your needs. What do you think?

zezha-msft:
Or alternatively, we could do `--exclude-pattern` which is a list of patterns that match to the file names, and `--exclude-path` which is a list of literal paths.

The reason why I don't think we should mix up the two lists, is because it might be confusing as to what's the expected behavior. Ex: if the user is working on MacOS and wishes to exclude all .DS_Store files, and if we only have 1 flag, would all .DS_Store files be ignored, or just the one under the given path?

benc-uk:
Hi @zezha-msft, ok this makes some sense now.
I think the `--exclude-directories` flag would be a great option to add

jesseik:
[...]

PR #516, opened 2019-07-25
Refactor part 1: Upload enumerators replacement.

adreed-msft:
This contains the initial refactoring for upload. There are a couple of workarounds that just leave old functionality in place for un-refactored parts.

There is also some reworking to the list traverser, namely in how it initializes traversers under it.


cmd/syncTraverser.go
```diff
@@ -29,6 +29,11 @@ import (
 	"sync/atomic"
 )
 
+// TODO: Get rid of highly situational constructors.
```
zezha-msft: Agreed. To say that it's equivalent to the mistakes of copy might be an exaggeration though LOL. 

Keep in mind that this was also a refactoring effort, and as with all refactoring efforts, it's a step-by-step process.

cmd/syncTraverser.go
```diff
@@ -54,7 +59,10 @@ func newLocalTraverserForSync(cca *cookedSyncCmdArgs, isSource bool) (*localTrav
 		atomic.AddUint64(counterAddr, 1)
 	}
 
-	traverser := newLocalTraverser(fullPath, cca.recursive, incrementEnumerationCounter)
+	// TODO: Implement this flag (followSymlinks).
+	// It's extra work and would require testing at the moment, hence why I didn't do it.
```
zezha-msft: You can leave it for the moment as there's a separate item for this.

cmd/copy.go
```diff
@@ -876,13 +901,45 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
-		err = e.enumerate(cca)
+	case common.EFromTo.LocalBlob(),
+		common.EFromTo.LocalBlobFS(),
+		common.EFromTo.LocalFile():
+		// TODO: Count files enumerated during scanning.
+		// ^ Will probably get to this in a later PR, would be best to not do it now as other scenarios would falsely output 0 objects found.
+		transfers := 0
+		var traverser resourceTraverser
+
+		traverser, err = initResourceTraverser(cca.source, cca.fromTo.From(), nil, nil, &cca.followSymlinks, &cca.listOfFilesLocation, cca.recursive, func() { transfers++ })
```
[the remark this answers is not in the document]
adreed-msft: Good point, it seems as though I've added it twice.

cmd/copy.go
```diff
@@ -876,13 +901,45 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+		if err != nil {
+			return err
+		}
+
+		filters := cca.initModularFilters()
+		processor := func(object storedObject) error {
+			transfers++
```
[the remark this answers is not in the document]
adreed-msft: Accidental. Only wanted to count the processed files to ensure that at least one transfer occurred.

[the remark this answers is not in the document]
adreed-msft: Though, upon looking at the original source code it just goes ahead and uses copyJobPartOrder.Transfers to check that, so no need to be counting in the first place.

cmd/copy.go
```diff
@@ -876,13 +901,45 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
-		fallthrough
-	case common.EFromTo.LocalBlobFS():
-		fallthrough
-	case common.EFromTo.LocalFile():
-		e := copyUploadEnumerator(jobPartOrder)
-		err = e.enumerate(cca)
+	case common.EFromTo.LocalBlob(),
+		common.EFromTo.LocalBlobFS(),
+		common.EFromTo.LocalFile():
+		// TODO: Count files enumerated during scanning.
```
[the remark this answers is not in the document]
adreed-msft: I meant in the enumeration progress indicator like sync does.

cmd/copyUtil.go
```diff
@@ -134,6 +134,28 @@ func (util copyHandlerUtil) ConstructCommandStringFromArgs() string {
 	return s.String()
 }
 
+func (util copyHandlerUtil) objectIsContainerOrDirectory(object string, location common.Location) bool {
```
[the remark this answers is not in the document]
adreed-msft: Oh right, this isn't even used in the DL refactor anyway, so I can avoid adding it all together. I'll just cherrypick some changes

cmd/copyUploadEnumerator.go
```diff
@@ -1,453 +0,0 @@
-package cmd
-
-import (
```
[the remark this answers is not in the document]
adreed-msft: 💯💯💯💯💯👌👌👌👌👌👌👌👌 mmmmMMMMMmmmm yeah that GOOD refactor removal 👌👌👌👌👌👌👌👌👌 💯💯💯💯💯💯💯💯💯

cmd/copy.go
```diff
@@ -948,6 +1005,114 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+		loc := common.ELocation.Unknown()
+
+		if source {
+			loc = cca.fromTo.From()
+		} else {
+			loc = cca.fromTo.To()
+		}
+		pathParts := strings.Split(path, "/")
+
+		// Encode disallowed characters on windows. Retain this functionality on other OSes.
```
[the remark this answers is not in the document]
adreed-msft: Consistent functionality.

cmd/zc_traverser_blob.go
```diff
@@ -42,6 +42,39 @@ type blobTraverser struct {
...
+	containerURL := azblob.NewContainerURL(rawContainerURL, t.p)
+	resp, err := containerURL.ListBlobsFlatSegment(t.ctx, azblob.Marker{}, azblob.ListBlobsSegmentOptions{})
+
+	if err != nil {
+		return false
+	}
+
+	// If items exist in the search, this IS a virtual directory.
+	// Virtual directories aren't otherwise possible in blob storage.
+	if len(resp.Segment.BlobItems) > 0 {
```
zezha-msft: to clarify, we are verifying whether the path represents an **existing** virtual directory, right? if yes, we should update the function name accordingly

cmd/zc_traverser_local.go
```diff
@@ -56,28 +67,69 @@ func (t *localTraverser) traverse(processor objectProcessor, filters []objectFil
...
+
+							if err != nil {
+								glcm.Info(fmt.Sprintf("Failed to open symlink %s: %s", filePath, err))
+								return nil
+							}
+
+							result, err = filepath.Abs(result)
+
+							if err != nil {
+								fmt.Println(fmt.Sprintf("Failed to resolve symlink %s: %s", filePath, err))
```
zezha-msft: please use glcm for all outputs, otherwise the text might get intertwined with other stuffs

cmd/zc_traverser_local.go
```diff
@@ -56,28 +67,69 @@ func (t *localTraverser) traverse(processor objectProcessor, filters []objectFil
...
+					if fileError != nil {
+						glcm.Info(fmt.Sprintf("Accessing %s failed with error: %s", filePath, fileError))
+						return nil
+					}
+
+					computedRelativePath := strings.TrimPrefix(cleanLocalPath(filePath), cleanLocalPath(queueItem.fullPath))
+					computedRelativePath = cleanLocalPath(filepath.Join(queueItem.relativeBase, computedRelativePath))
+					computedRelativePath = strings.TrimPrefix(computedRelativePath, common.AZCOPY_PATH_SEPARATOR_STRING)
+
+					if fileInfo.Mode()&os.ModeSymlink != 0 {
```
zezha-msft: nice, consider adding integration tests for symlink support

zezha-msft: I don't know if smoke tests actually have good enough coverage

cmd/zc_filter.go
```diff
@@ -20,14 +20,46 @@
...
 
 type excludeFilter struct {
 	pattern string
 }
 
 func (f *excludeFilter) doesPass(storedObject storedObject) bool {
-	matched, err := path.Match(f.pattern, storedObject.name)
+	checkItem := storedObject.name
+	if strings.Contains(f.pattern, "/") {
+		checkItem = storedObject.relativePath
```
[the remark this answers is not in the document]
adreed-msft: oh, I forgot to go put that in. I prefer the alternate flags, personally.

cmd/copy.go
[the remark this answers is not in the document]
adreed-msft: Fair, won't do it then.

cmd/copy.go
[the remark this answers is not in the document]
adreed-msft: That's fair. I've limited the functionality to Windows.

cmd/zc_enumerator.go
```diff
@@ -72,6 +82,112 @@ func newStoredObject(name string, relativePath string, lmt time.Time, size int64
...
+
+			if err != nil {
+				return nil, fmt.Errorf("failed to glob: %s", err)
+			}
+
+			// A channel really would be preferable for this kind of behaviour but whatever.
+			strToRead := ""
+
+			for _, v := range matches {
+				strToRead += strings.TrimPrefix(strings.ReplaceAll(v, common.OS_PATH_SEPARATOR, common.AZCOPY_PATH_SEPARATOR_STRING), basePath) + "\n"
```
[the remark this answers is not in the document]
adreed-msft: Fair. Changed to bytes.Buffer.

cmd/zc_enumerator.go
```diff
@@ -72,6 +82,112 @@ func newStoredObject(name string, relativePath string, lmt time.Time, size int64
...
+
+		if strings.Index(source, "*") != -1 {
+			basePath := trimWildcards(source)
+			matches, err := filepath.Glob(source)
+
+			if err != nil {
+				return nil, fmt.Errorf("failed to glob: %s", err)
+			}
+
+			// A channel really would be preferable for this kind of behaviour but whatever.
```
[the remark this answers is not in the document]
adreed-msft: bytes.Buffer achieves the functionality I was aiming for.

Unless you'd like me to adjust it so that the list traverser reads until a channel closes. The upside of this is that we could spin up a routine to read a file/glob instead of letting the traverser read line-by-line, so that these would occur side-by-side.

cmd/zc_traverser_blob.go
adreed-msft: Not anymore. This function is intended to check whether something is intended to be a directory on the destination, and whether something is an existing directory on the source.

cmd/zc_traverser_local.go
adreed-msft: Whoops.

cmd/zc_traverser_local.go
adreed-msft: Will do

cmd/zc_enumerator.go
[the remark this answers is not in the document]
zezha-msft: Hmm a channel would indeed be much better. If you are ok with it I'd suggest to adopt the channel approach.

cmd/zc_traverser_blob.go
zezha-msft: Any blob path can be a virtual directory though, they are called virtual because they don't really exist. So for blob, it should always return true then?

adreed-msft: Well, no. If a user wants to upload `filexyz` under `container/filexyz` and `container/filexyz/filexyz` (for whatever reason), in the past of azcopy, the command

`azcopy copy filexyz <URL>/container/filexyz` would treat the result as an immediate destination

and

`azcopy copy filexyz <URL>/container/filexyz/` would treat it as a virtual directory.

For uploads, this information is important because it determines where we place things.
For downloads, this information is important because it determines if we need to be recursive or not.

adreed-msft: I've yet to push my changes to github yet, still working on them locally, but you'll see what I mean after they come along.

zezha-msft: Ok, I see what you mean now, it matters when uploading a single file, because we need to know whether the blob should be put "under" the given dst path, or "at" the given path.

cmd/syncTraverser.go
adreed-msft: Wasn't necessarily meaning it was _equivalent_, more or less just meant we were going back down a very similar path. Codebase PTSD, if you would.

cmd/copy.go
```diff
@@ -211,7 +215,7 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	}
 
 	// Note: new implementation of list-of-files only works for remove command for now
-	if cooked.fromTo == common.EFromTo.BlobTrash() || cooked.fromTo == common.EFromTo.FileTrash() {
+	if cooked.fromTo == common.EFromTo.BlobTrash() || cooked.fromTo == common.EFromTo.FileTrash() || cooked.fromTo.From() == common.ELocation.Local() {
```
JohnRusk: This change invalidates the comment immediately above it. Please update the comment, becuase its too confusing if comment and code are not consistent.

cmd/copy.go
```diff
@@ -755,6 +770,14 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 	// Strip the SAS from the source and destination whenever there is SAS exists in URL.
 	// Note: SAS could exists in source of S2S copy, even if the credential type is OAuth for destination.
 	switch from {
+	case common.ELocation.Local():
+		tmpSrc, err := filepath.Abs(cca.source)
+		if err != nil {
+			return fmt.Errorf("couldn't get absolute path of the source location %s. Failed with errror %s", cca.source, err.Error())
+		}
+
+		jobPartOrder.SourceRoot = cleanLocalPath(trimWildcards(tmpSrc))
```
JohnRusk: What is trimWildcards here? Does it remove them? I thought our plan was to _fail with an error_ if there were any wildcards in places where we did not expect them.  Otherwise, don't we run the risk that the user thinks we have done one thing, but really we have done another?

cmd/copy.go
```diff
@@ -876,13 +899,67 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+			destURL = copyHandlerUtil{}.appendQueryParamToUrl(destURL, cca.destinationSAS)
+			dst = destURL.String()
+		}
+
+		traverser, err = initResourceTraverser(cca.source, cca.fromTo.From(), nil, nil, &cca.followSymlinks, &cca.listOfFilesLocation, cca.recursive, func() {})
+		if err != nil {
+			return err
+		}
+
+		destTraverser, err = initResourceTraverser(dst, cca.fromTo.To(), &ctx, &cca.credentialInfo, &cca.followSymlinks, nil, cca.recursive, func() {})
```
JohnRusk: I initially felt confused why there was a destination traverser for a _copy_ operation. Maybe could consider make a routine called something like isDir(), which constructs this traverser, calls isDirectory on it (which is all we need) and throws it away.  To move that stuff out of this routine.

cmd/copy.go
```diff
@@ -876,13 +899,67 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+		if err != nil {
+			return err
+		}
+
+		isDir := traverser.isDirectory(false)
+		isDestDir := destTraverser.isDirectory(true)
+
+		if isDir && !cca.recursive {
+			return errors.New("cannot copy from container or directory without --recursive or trailing wildcard (/*)")
+		}
```
JohnRusk: What if isDir == true and isDestDir==false?  On the face of it, that seems to me like it should be an error...

cmd/copy.go
```diff
@@ -876,13 +899,67 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+		destTraverser, err = initResourceTraverser(dst, cca.fromTo.To(), &ctx, &cca.credentialInfo, &cca.followSymlinks, nil, cca.recursive, func() {})
+		if err != nil {
+			return err
+		}
+
+		isDir := traverser.isDirectory(false)
+		isDestDir := destTraverser.isDirectory(true)
+
+		if isDir && !cca.recursive {
+			return errors.New("cannot copy from container or directory without --recursive or trailing wildcard (/*)")
```
JohnRusk: Nice.

cmd/copy.go
```diff
@@ -948,6 +1025,136 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+
+	firstWCIndex := strings.Index(path, "*")
+	result := replacePathSeparators(path[:firstWCIndex])
+	lastSepIndex := strings.LastIndex(result, "/")
+	result = result[:lastSepIndex+1]
+
+	return result
+}
+
+func (cca *cookedCopyCmdArgs) findObject(source bool, dstIsDir bool, object storedObject) (relativePath string) {
```
JohnRusk: This feels out of place as a method of the cooked command _args_.  I think if you follow Ze's suggestion above, about moving code out of the big switch statement, you could move it into somewhere/someplace where this could be move to too. (e.g. some kind of copyConfigurator/copyInitiator or something. Can't quite think what the best name would be)

cmd/copy.go
```diff
@@ -876,13 +899,67 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+		isDestDir := destTraverser.isDirectory(true)
+
+		if isDir && !cca.recursive {
+			return errors.New("cannot copy from container or directory without --recursive or trailing wildcard (/*)")
+		}
+
+		filters := cca.initModularFilters()
+		processor := func(object storedObject) error {
+			src := cca.findObject(true, isDestDir, object)
+			// Why hand in the source on the destination? Because we need to adjust the path for the source if there's no wildcard.
```
JohnRusk: I struggeld to understand what this comment is trying to say, or which of the two lines it refers to.  I think "object" = "sourceObject" and this is referring to the dst line... but it took me a while to figure that out and I'm not 100% sure I've got it right.  Maybe whitespace before the comment would help, and maybe rename object to sourceObject?

cmd/copy.go
JohnRusk: The name "find" implies to me that it will do a lookup - e.g. hit a remote resource, do a disk operation locally.  All it's really doing is constructing the object's full relative path, right?  If so, can it be named something like makeRelativePath?

cmd/copy.go
```diff
@@ -876,13 +899,67 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+
+		isDir := traverser.isDirectory(false)
+		isDestDir := destTraverser.isDirectory(true)
+
+		if isDir && !cca.recursive {
+			return errors.New("cannot copy from container or directory without --recursive or trailing wildcard (/*)")
+		}
+
+		filters := cca.initModularFilters()
+		processor := func(object storedObject) error {
```
JohnRusk: I _love_ how compact this code as turned out to be, especially with the helper methods you've created.

cmd/copy.go
```diff
@@ -1166,6 +1373,12 @@ func init() {
 	cpCmd.PersistentFlags().BoolVar(&raw.withSnapshots, "with-snapshots", false, "include the snapshots. Only valid when the source is blobs.")
 	cpCmd.PersistentFlags().StringVar(&raw.legacyInclude, "include", "", "only include these files when copying. "+
 		"Support use of *. Files should be separated with ';'.")
+	cpCmd.PersistentFlags().StringVar(&raw.includePath, "include-path", "", "only include these paths when copying. "+
+		"Supports use of * considering the relative path of items (from the root of the search.) ex. myFolder/*.txt\n;*/subDirName/*.pdf"+
```
zezha-msft: Ah I see that you are also supporting patterns in the path flags, I thought we wouldn't, but it actually looks good, since we enable more specific targeting.

cmd/copy.go
```diff
@@ -1166,6 +1373,12 @@ func init() {
...
+		"Note: This considers the _entire_ relative path, file name included.")
+	cpCmd.PersistentFlags().StringVar(&raw.excludePath, "exclude-path", "", "only exclude these paths when copying. "+
+		"Supports use of * considering the relative path of items (from the root of the search.) ex. myFolder/*.txt\n;*/subDirName/*.pdf"+
```
zezha-msft: The new line \n might make the help message look weird with a different resolution (on other OS where it can be adjusted easily), I suggest removing it.

cmd/copy.go
```diff
@@ -1166,6 +1373,12 @@ func init() {
 	cpCmd.PersistentFlags().BoolVar(&raw.withSnapshots, "with-snapshots", false, "include the snapshots. Only valid when the source is blobs.")
 	cpCmd.PersistentFlags().StringVar(&raw.legacyInclude, "include", "", "only include these files when copying. "+
```
zezha-msft: TODO: rename the include/exclude flags to include-pattern and exclude-pattern, unhide the include one too

cmd/remove.go
```diff
@@ -86,5 +86,7 @@ func init() {
 	deleteCmd.PersistentFlags().StringVar(&raw.logVerbosity, "log-level", "INFO", "define the log verbosity for the log file, available levels: INFO(all requests/responses), WARNING(slow responses), ERROR(only failed requests), and NONE(no output logs).")
 	deleteCmd.PersistentFlags().StringVar(&raw.include, "include", "", "only include files whose name matches the pattern list. Example: *.jpg;*.pdf;exactName")
```
zezha-msft: renamed to include-name or include-pattern to be more specific?

cmd/sync.go
```diff
@@ -187,7 +191,9 @@ type cookedSyncCmdArgs struct {
 	recursive             bool
 	followSymlinks        bool
 	include               []string
+  includePath           []string
```
zezha-msft: minor: turn on file watcher to auto-format when you save

cmd/zc_enumerator.go
```diff
@@ -72,6 +82,128 @@ func newStoredObject(name string, relativePath string, lmt time.Time, size int64
...
```
zezha-msft: rm outdated comment

cmd/zc_traverser_file.go
```diff
@@ -41,6 +41,10 @@ type fileTraverser struct {
 	incrementEnumerationCounter func()
 }
 
+func (t *fileTraverser) isDirectory(isDest bool) bool {
+	return copyHandlerUtil{}.urlIsAzureFileDirectory(t.ctx, t.rawURL, t.p) // This handles all of the fanciness for us.
```
zezha-msft: it's best to pull the code out of copyHandlerUtil, which we should try to remove towards the end of refactoring. It's a dumping ground for unrelated (sometimes untested) code, best to rid of it for our own good. 😈

cmd/zc_enumerator.go
```diff
@@ -72,6 +82,128 @@ func newStoredObject(name string, relativePath string, lmt time.Time, size int64
 // pass each storedObject to the given objectProcessor if it passes all the filters
 type resourceTraverser interface {
 	traverse(processor objectProcessor, filters []objectFilter) error
+	isDirectory(isDest bool) bool // isDest is literally only useful on blob because we have seperate rules for blob on up/down.
```
zezha-msft: please clarify in the comment what's the special rule

cmd/zc_traverser_list.go
```diff
@@ -1,27 +1,32 @@
...
 // behaves like a single traverser (basically a "traverser of traverser")
 type listTraverser struct {
-	listReader              io.ReadCloser
+	listReader              chan string
 	childTraverserGenerator childTraverserGenerator
 }
 
 type childTraverserGenerator func(childPath string) (resourceTraverser, error)
 
+func (l *listTraverser) isDirectory(isDest bool) bool {
```
zezha-msft: there's no impact for this returning false, right? since the listTraverser is a meta-traverser and does not represent a single resource dir/file, please leave a comment to explain accordingly

cmd/zc_traverser_list.go
```diff
@@ -52,34 +57,37 @@ func (l *listTraverser) traverse(processor objectProcessor, filters []objectFilt
...
+		if parentType != common.ELocation.Local() {
+			// assume child path is not URL-encoded yet, this is consistent with the behavior of previous implementation
+			childURL, _ := url.Parse(parent)
+			childURL.Path = common.GenerateFullPath(childURL.Path, relativeChildPath)
+
+			// construct traverser that goes through child
+			source = copyHandlerUtil{}.appendQueryParamToUrl(childURL, parentSAS).String()
+		} else {
+			// is local, only generate the full path
+			source = common.GenerateFullPath(parent, relativeChildPath)
```
zezha-msft: ah yes, the tests (for rm) were only on remote resources. Thanks for fixing it.

cmd/copy.go
```diff
@@ -948,6 +1025,136 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+		}
+
+		return pathEncodeRules(relativePath)
+	}
+
+	// If it's out here, the source is a folder (or wildcard) of some kind.
+
+	relativePath = "/" + strings.Replace(object.relativePath, common.OS_PATH_SEPARATOR, common.AZCOPY_PATH_SEPARATOR_STRING, -1)
+
+	if !source && !strings.Contains(cca.source, "*") {
```
JohnRusk: Can you move this (i.e. `strings.Contains(cca.source, "*") ` to a function called something like `pathPointsToContents` so that this line would become
`if !source && !pathPointsToContents(cca.source) {`

and pathPointsToContents would be something like the following?  The reason is that our /* suffix logic is sufficiently noteworthy to be wrapped in a function, so that (a) it can be clearly documented there and (b) that function can be reused if needed to become the single source of truth about this behavior and (c) by doing find-usages on that function we should, in theory, be able to find all parts of the code that have conditional logic based on this special suffix.

```
\\ pathPointsToContents differentiates between the two types of source path that we support.  Ones like this:
\\ foo\bar
\\ and ones like this:
\\ foo\bar\*
\\ The former case says to copy the whole directory (i.e. replacate bar, including the name 'bar', to the destination) whereas the latter says to only copy the contents of bar, not bar itself.
\\ NOTE: it is assumed that the path has already been validated. E.g. we don't check here for what's in front 
\\ of the *, which should be a \ or a /
func pathPointsToContents(path string) {
   return strings.HasSuffix(path, "*")      
}
```

cmd/copy.go
```diff
@@ -948,6 +1025,136 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+		// Save to a directory
+		relativePath = "/" + filepath.Base(cca.source) + relativePath
+	}
+
+	return pathEncodeRules(relativePath)
+}
+
+// Initialize the modular filters outside of copy to increase readability.
+func (cca *cookedCopyCmdArgs) initModularFilters() []objectFilter {
+	filters := make([]objectFilter, 0) // same as []objectFilter{} under the hood
```
JohnRusk: What's the plan for incorporating the new filters from commit 9008a3b?  (Asking because we need to make sure we don't forget?  Don't want to have them get unhooked from the codepaths, just because they're new and we are refactoing)  Maybe could make a PBI to hook them in in next sprint?

cmd/copy.go
```diff
@@ -948,6 +1025,136 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+		// We ONLY need to do this adjustment to the destination.
+		// The source SAS has already been removed. No need to convert it to a URL or whatever.
+		// Save to a directory
+		relativePath = "/" + filepath.Base(cca.source) + relativePath
+	}
+
+	return pathEncodeRules(relativePath)
+}
+
+// Initialize the modular filters outside of copy to increase readability.
```
JohnRusk: Nice

cmd/copyUtil.go
```diff
@@ -143,14 +143,13 @@ func (util copyHandlerUtil) urlIsBFSFileSystemOrDirectory(ctx context.Context, u
 	return dirURL.IsDirectory(context.Background())
 }
 
-func (util copyHandlerUtil) urlIsAzureFileDirectory(ctx context.Context, url *url.URL) bool {
+func (util copyHandlerUtil) urlIsAzureFileDirectory(ctx context.Context, url *url.URL, p pipeline.Pipeline) bool {
```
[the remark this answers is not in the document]
adreed-msft: 👍

cmd/copy.go
```diff
@@ -1166,6 +1373,12 @@ func init() {
 	cpCmd.PersistentFlags().BoolVar(&raw.withSnapshots, "with-snapshots", false, "include the snapshots. Only valid when the source is blobs.")
 	cpCmd.PersistentFlags().StringVar(&raw.legacyInclude, "include", "", "only include these files when copying. "+
 		"Support use of *. Files should be separated with ';'.")
+	cpCmd.PersistentFlags().StringVar(&raw.includePath, "include-path", "", "only include these paths when copying. "+
```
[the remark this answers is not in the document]
adreed-msft: - We are going to have both variants. We opted to have both so the user could explicitly specify path. If we wanted, we could chop off the object name from the path calculation.
- Of course not.
- I didn't even realize include was hidden. @zezha-msft thoughts on making it not hidden? We have tests on it in our suite anyway.
- Currently it means that `foo/bar/eee` would get caught in the include filter.
- Hmmm, good question. I think a good one-sentence answer is "Wildcards cover a single path segment and (currently) must contain a file name filter, (ex. foo/*/bar/*.txt would match to foo/eee/bar/thing.txt but not foo/eee/fff/bar/thing.txt)". I almost feel as though dropping the name requirement would be _good_ but there may be a use case where somebody wants to include .txt files from one dir but not the other

cmd/copy.go
adreed-msft: trimWildcards is intended to trim the wildcards from a local path. We decided to enable wildcards for local usage as that comes effectively for free, but disallow them on non-local sources.

@zezha-msft correct me if I'm wrong.

cmd/copy.go
adreed-msft: I'm inclined to agree. I didn't necessarily _want_ to do this, but we also didn't have a unified way to tell if something was a directory or not. I'd tried to introduce one into copyHandlerUtil, but we all know that file is bloated to hell and back and doesn't need additional items.

cmd/copy.go
adreed-msft: We treat the destination as a directory in this case. This retains old functionality where I could safely do something like so:

```azcopy copy /my/local/directory https://myaccount.blob.core.windows.net/container/virdir```

File: cmd/copy.go (hunk 20 above)
Comment (adreed-msft) [reply cmd/copy.go]: My feeling is that a lot of these things feel out of place _everywhere_ as is. copyUtilHandler is bloated, cooked command args isn't the right place for these things, etc.

I also don't want to go causing another copyHandlerUtil style incident.

File: cmd/copy.go (hunk 21 above)
Comment (adreed-msft) [reply cmd/copy.go]: OK so this was a comment before cca.findObject had changed its signature and I need to remove it.

Initially, cca.findObject took in a path. I handed in the source path there.

File: cmd/copy.go (hunk 20 above)
Comment (adreed-msft) [reply cmd/copy.go]: Fair. I'll change it to makeEscapedRelativePath.

File: cmd/copy.go (hunk 22 above)
Comment (adreed-msft) [reply cmd/copy.go]: That was the goal here. I'm glad it came out rather clean as well. Should make copy much nicer to work with in the future, especially when we add new scenarios.

File: cmd/copy.go (hunk 23 above)
Comment (adreed-msft) [reply cmd/copy.go]: Oh! I must have misunderstood then. Glad it was a happy accident.

File: cmd/copy.go (hunk 24 above)
Comment (adreed-msft) [reply cmd/copy.go]: Will do.

File: cmd/sync.go (hunk 27 above)
Comment (adreed-msft) [reply cmd/sync.go]: Gah! How did that make it through? I've had it on for ages.
image

File: cmd/remove.go (hunk 26 above)
Comment (adreed-msft) [reply cmd/remove.go]: opted for include/exclude-pattern

File: cmd/copy.go [hunk 38]
```diff
@@ -948,6 +1025,136 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+	case common.ELocation.BlobFS():
+		p, err = createBlobFSPipeline(ctx, credential)
+	default:
+		err = fmt.Errorf("can't produce new pipeline for location %s", location)
+	}
+
+	return
+}
+
+func trimWildcards(path string) string {
```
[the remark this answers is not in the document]
adreed-msft: For remote sources, yeah, it will chop off the trailing /*. For local sources, we support wildcards in the path, so it will chop off ALL wildcards. It should *probably* be changed to getPathBeforeFirstWildcard.

cmd/copy.go
```diff
@@ -948,6 +1025,136 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+func (cca *cookedCopyCmdArgs) findObject(source bool, dstIsDir bool, object storedObject) (relativePath string) {
+	var pathEncodeRules = func(path string) string {
+		loc := common.ELocation.Unknown()
+
+		if source {
+			loc = cca.fromTo.From()
+		} else {
+			loc = cca.fromTo.To()
+		}
+		pathParts := strings.Split(path, "/")
```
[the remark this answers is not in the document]
adreed-msft: Right, pardon me.

cmd/copy.go
```diff
@@ -948,6 +1025,136 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+
+		// Encode disallowed characters on windows.
+		// TODO: Inform tests of this
+		if loc == common.ELocation.Local() && runtime.GOOS == "windows" {
+			invalidChars := `<>\/:"|?*` + string(0x00)
```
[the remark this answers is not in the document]
adreed-msft: Yes, you are correct. This is ONLY for downloads.

cmd/copy.go
```diff
@@ -948,6 +1025,136 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+				relativePath = "/" + object.name
+			} else {
+				relativePath = ""
+			}
+		}
+
+		return pathEncodeRules(relativePath)
+	}
+
+	// If it's out here, the source is a folder (or wildcard) of some kind.
```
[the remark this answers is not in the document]
adreed-msft: Will do.

cmd/copy.go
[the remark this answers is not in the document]
adreed-msft: Making it exclusive to downloads now. I'll need to add in a reversing operation on uploads.

cmd/copy.go
adreed-msft: Will do.

cmd/copy.go
adreed-msft: Ping me once they get merged in and I'll merge them into this branch.

adreed-msft: Good question. Let me merge them into this branch and include them now. No reason they _can't_ be included yet.

cmd/copy.go
zezha-msft: I just realized there might be an issue with including sub-directories:

ex: source is /usr/root, and the user wants to include a specific set of sub-directories:
1. dir1 (has files directly under it)
2. dir2 (has more sub-dirs under it)

for dir1, the user can do --include-path="dir1/*", and it will match to every file under it. However, for dir2, how does the customer specify their intention? "dir2/*" only matches to the files directly under it right? Missing out the other files like "dir2/subdir/bla.txt".

zezha-msft: To work around this issue, and make the flag simpler, perhaps we can disallow \*s in the path flags, and let users specify relative paths to files and directories. In our implementation, we use a listTraverser to go through each one.

cmd/copy.go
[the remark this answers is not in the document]
zezha-msft: I don't know for sure, but I believe that was the intention. In general it's good practice to avoid string literals which are annoying to find/replace in case of changes. 😄

cmd/copy.go
zezha-msft: @adreed-msft yes that's what we decided. We only support `/*` for remote sources.

SourceRoot is used to build the actually transfers, we have find it since in the actual job parts, we give relative paths for each transfer.

cmd/copy.go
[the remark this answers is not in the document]
zezha-msft: The original issue is [here](https://github.com/Azure/azure-storage-azcopy/issues/356#issuecomment-487778832). The intention is to allow users to clearly their intention to:

1. target file names only
2. target file's relative path

As we just realized, it's much easier to not allow \* in the path flags, and treat --include-path as the same as list-of-files, while --exclude-path is a filter that looks for the specified prefix.

cmd/sync.go
zezha-msft: hmmm that's strange indeed, maybe they have a bug. Hope an upgrade could fix it.

cmd/copy.go
```diff
@@ -479,6 +481,12 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	cooked.includePathPatterns = raw.parsePatterns(raw.includePath)
 	cooked.excludePathPatterns = raw.parsePatterns(raw.excludePath)
 
+	if (raw.includeFileAttributes != "" || raw.excludeFileAttributes != "") && fromTo.To() != common.ELocation.Local() {
```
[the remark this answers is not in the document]
adreed-msft: That would be what is called a brain fart. You're right, it should be checking against fromTo.From()

cmd/zc_traverser_local.go
```diff
@@ -32,13 +32,28 @@ import (
 )
 
 type localTraverser struct {
-	fullPath  string
-	recursive bool
+	fullPath       string
```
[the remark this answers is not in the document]
adreed-msft: Good catch.

cmd/zc_traverser_local.go
```diff
@@ -56,28 +71,69 @@ func (t *localTraverser) traverse(processor objectProcessor, filters []objectFil
...
-				}
-
-				t.incrementEnumerationCounter()
-
-				// the relative path needs to be computed from the full path
-				computedRelativePath := strings.TrimPrefix(cleanLocalPath(filePath), t.fullPath)
-
-				// leading path separators are trimmed away
-				computedRelativePath = strings.TrimPrefix(computedRelativePath, common.AZCOPY_PATH_SEPARATOR_STRING)
+			// We want to re-queue symlinks up in their evaluated form because filepath.Walk doesn't evaluate them for us.
```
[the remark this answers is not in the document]
adreed-msft: We do check both errs up there. The one from t.getInfoIfSingleFile and the one from processedIfPassedFilters in other branches is redundant.

cmd/zc_traverser_local.go
```diff
@@ -56,28 +71,69 @@ func (t *localTraverser) traverse(processor objectProcessor, filters []objectFil
...
+			// Furthermore, we use a map as a hashset to avoid re-walking any paths we already know.
+			type walkItem struct {
+				fullPath     string // We need the full, symlink-resolved path to walk against
+				relativeBase string // We also need the relative base path we found the symlink at.
+			}
 
-				return processIfPassedFilters(filters, newStoredObject(fileInfo.Name(), computedRelativePath,
-					fileInfo.ModTime(), fileInfo.Size(), nil, blobTypeNA), processor)
-			})
+			var walkQueue = []walkItem{{fullPath: t.fullPath, relativeBase: ""}}
```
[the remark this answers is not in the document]
adreed-msft: I have a feeling it's way too intertwined with existing code but I'll certainly take a stab at cutting it out into its own function for testing. I feel as though the code would be cleaner if it were living in its own function, anyway.

cmd/copy.go
```diff
@@ -217,18 +218,50 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
...
+
+			if err != nil {
+				return cooked, fmt.Errorf("cannot open %s file passed with the list-of-file flag", raw.listOfFilesToCopy)
+			}
+		}
+
+		go func() {
+			defer close(listChan)
+
+			if f != nil {
```
JohnRusk: Can the user specify both list-of-files and includePathList?  Maybe we should do something in validation to prevent that, since its kind of confusing?  (It's hard to think of a case where it makes sense, and if it did make sense, I think user would expect the two criteria to be AND-ed together - only transfer things that are covered by both - whereas there's no way we want to AND them. We are OR-ing them because AND-ing is just too complicated to implement).

cmd/removeEnumerator.go
```diff
@@ -52,17 +53,17 @@ func newRemoveEnumerator(cca *cookedCopyCmdArgs) (enumerator *copyEnumerator, er
...
 	if err != nil {
 		return nil, err
 	}
 
 	transferScheduler := newRemoveTransferProcessor(cca, NumOfFilesPerDispatchJobPart)
-	includeFilters := buildIncludeFilters(cca.includePatterns, false) // TODO: Add include-path in here?
+	includeFilters := buildIncludeFilters(cca.includePatterns) // TODO: Add include-path in here?
 	excludeFilters := buildExcludeFilters(cca.excludePatterns, false)
-	includePathFilters := buildIncludeFilters(cca.includePathPatterns, true)
+	includePathFilters := buildIncludeFilters(cca.includePathPatterns)
```
JohnRusk: Why is this line still here?  I was expecting it to go, with the change to use listOfFilesChannel

cmd/copy.go
```diff
@@ -529,12 +561,12 @@ type cookedCopyCmdArgs struct {
...
 
 	// filters from flags
-	listOfFilesToCopy   []string
-	listOfFilesLocation string
-	recursive           bool
-	followSymlinks      bool
-	withSnapshots       bool
-	forceWrite          bool
+	listOfFilesToCopy  []string
+	listOfFilesChannel *chan string // We make it a pointer so we can check if it exists w/o reading from it & tack things onto it if necessary.
```
JohnRusk: Minor: IIRC `chan string` is nillable (the non-pointer version).  For that reason its not normal, AFAICT, to use pointers to channels https://stackoverflow.com/a/44367183

cmd/syncEnumerator.go
```diff
@@ -51,8 +51,8 @@ func newSyncDownloadEnumerator(cca *cookedSyncCmdArgs) (enumerator *syncEnumerat
 	}
 
 	transferScheduler := newSyncTransferProcessor(cca, NumOfFilesPerDispatchJobPart)
-	includeFilters := buildIncludeFilters(cca.include, false)
-	includePathFilters := buildIncludeFilters(cca.includePath, true)
+	includeFilters := buildIncludeFilters(cca.include)
+	includePathFilters := buildIncludeFilters(cca.includePath)
```
JohnRusk: As per other comment, is this meant to be still here?

cmd/syncEnumerator.go
```diff
@@ -114,8 +114,8 @@ func newSyncUploadEnumerator(cca *cookedSyncCmdArgs) (enumerator *syncEnumerator
...
```
JohnRusk: Ditto

cmd/syncEnumerator.go
```diff
@@ -129,7 +129,7 @@ func newSyncUploadEnumerator(cca *cookedSyncCmdArgs) (enumerator *syncEnumerator
 	filters := append(includeFilters, includeAttrFilters...)
 	filters = append(filters, excludeFilters...)
 	filters = append(filters, excludeAttrFilters...)
-  filters = append(filters, includePathFilters...)
+	filters = append(filters, includePathFilters...)
```
JohnRusk: Ditto

cmd/zc_traverser_local.go
```diff
@@ -54,6 +54,98 @@ func (t *localTraverser) isDirectory(isDest bool) bool {
 	return props.IsDir()
 }
 
+// For some reason, Goland was refusing to believe this existed when placed below traverse(). Compiled just fine though.
```
JohnRusk: Can you remove this comment please.

cmd/zc_traverser_local.go
```diff
@@ -63,18 +155,53 @@ func (t *localTraverser) traverse(processor objectProcessor, filters []objectFil
...
+						fileInfo.Name(),
+						relPath,
+						fileInfo.ModTime(),
+						fileInfo.Size(),
+						nil,
+						blobTypeNA),
+					processor)
+			}
+
+			if t.followSymlinks {
```
JohnRusk: Nice. I hadn't realised this would be possible (to use different walker in each case). It works out well.

cmd/zc_traverser_local.go
```diff
@@ -63,18 +155,53 @@ func (t *localTraverser) traverse(processor objectProcessor, filters []objectFil
...
+				return WalkWithSymlinks(t.fullPath, processFile)
+			} else {
+				return filepath.Walk(t.fullPath, processFile)
+			}
+
+			/*// We want to re-queue symlinks up in their evaluated form because filepath.Walk doesn't evaluate them for us.
```
JohnRusk: Please remove this commented out block. We do not leave big commented out blocks like this in the code. (We just use file history in Git if we need the old version)

cmd/zc_traverser_local.go
```diff
@@ -63,18 +155,53 @@ func (t *localTraverser) traverse(processor objectProcessor, filters []objectFil
...
+				if fileInfo.IsDir() {
+					return nil
+				}
+
+				relPath := strings.TrimPrefix(strings.TrimPrefix(cleanLocalPath(filePath), cleanLocalPath(t.fullPath)), "/")
+				if !t.followSymlinks && fileInfo.Mode()&os.ModeSymlink != 0 {
+					glcm.Info(fmt.Sprintf("Skipping over symlink at %s because --follow-symlinks is false", filepath.Join(t.fullPath, relPath)))
+					return nil
+				}
+
```
JohnRusk: Need this here, right?

```
		if t.incrementEnumerationCounter != nil {
			t.incrementEnumerationCounter()
		}
```

cmd/zt_generic_filter_test.go
```diff
@@ -32,7 +32,7 @@ func (s *genericFilterSuite) TestIncludeFilter(c *chk.C) {
 	// set up the filters
 	raw := rawSyncCmdArgs{}
 	includePatternList := raw.parsePatterns("*.pdf;*.jpeg;exactName")
-	includeFilter := buildIncludeFilters(includePatternList, false)[0]
+	includeFilter := buildIncludeFilters(includePatternList)[0]
```
JohnRusk: Ditto

cmd/zt_generic_traverser_test.go
```diff
@@ -38,6 +40,43 @@ type genericTraverserSuite struct{}
 
 var _ = chk.Suite(&genericTraverserSuite{})
 
+// Test follow symlink functionality
+func (s *genericTraverserSuite) TestWalkWithSymlinks(c *chk.C) {
+	if runtime.GOOS == "windows" {
+		c.Skip("Windows does not allow true symlinks")
```
JohnRusk: This is not correct. It was correct many years ago, but now Windows has fine and real symlinks, see `mklink` command: https://www.howtogeek.com/howto/16226/complete-guide-to-symbolic-links-symlinks-on-windows-or-linux/

So I would expect `os.Symlink` to work on Windows.  Is that not the case?

cmd/zt_generic_traverser_test.go
```diff
@@ -38,6 +40,43 @@ type genericTraverserSuite struct{}
...
+	}
+
+	// You get a sense of pride and satisfaction if you know what the first one references.
```
JohnRusk: Usual advice, re keeping comments formal and professional.

cmd/zt_generic_traverser_test.go
```diff
@@ -38,6 +40,43 @@ type genericTraverserSuite struct{}
 
 var _ = chk.Suite(&genericTraverserSuite{})
 
+// Test follow symlink functionality
+func (s *genericTraverserSuite) TestWalkWithSymlinks(c *chk.C) {
```
JohnRusk: We also need approx two additional tests, to test the seenPaths functionality.

e.g. a test where tmpDir contains two symlinks to the same target (Test: can we reject duplicates when both are the target of symlinks?)
and a test where symlinkDir contains a link back to a file in tempDir (Test: can we reject duplicates where one is the target of an symlink and the other is not)

JohnRusk: And it wouldn't hurt to have an test that actually sets up an cycle - eg. tmpDir links to smlinkDir, and smlinxDir contains a link right back to tempDir itself.  If theory (I think) the other two tests I've suggested would cover that, but in practice it won't hurt to explicitly test the cycle case (since that would probably be the most embrassing bug if it went wrong in the wild)

cmd/copy.go
```diff
@@ -217,18 +218,50 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
...
-	if cooked.fromTo == common.EFromTo.BlobTrash() || cooked.fromTo == common.EFromTo.FileTrash() || cooked.fromTo.From() == common.ELocation.Local() {
-		cooked.listOfFilesLocation = raw.listOfFilesToCopy
-	} else if cooked.fromTo == common.EFromTo.BlobFSTrash() {
-		// Note: when the ADLS Gen2 interop happens eventually, we should keep the current blob implementation
-		// so that users can still use include/exclude flags,
-		// and we can shift to the new directory delete API (equivalent to the current bfs implementation)
-		// if no include/exclude is supplied.
-		if len(raw.include) > 0 || len(raw.exclude) > 0 {
+	if fromTo.To() == common.ELocation.Unknown() || cooked.fromTo.From() == common.ELocation.Local() {
+		// This handles both list-of-files and include-path as a list enumerator.
```
JohnRusk: BTW, all the in-app help comments (in sync, remove and copy) all still say that wildcards are supported in include-path and exclude-path. Can you correct that please?

cmd/sync.go
JohnRusk: I've had the same thing occasionally, where i'v had it turned on but it didn't process some things.

cmd/copy.go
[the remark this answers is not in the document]
zezha-msft: Yes, we had a discussion, and here is the [problem with allowing \*s ](https://github.com/Azure/azure-storage-azcopy/pull/516#discussion_r310300542)

cmd/copy.go
```diff
@@ -876,13 +893,58 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
-		fallthrough
-	case common.EFromTo.LocalFile():
-		e := copyUploadEnumerator(jobPartOrder)
-		err = e.enumerate(cca)
+	case common.EFromTo.LocalBlob(),
+		common.EFromTo.LocalBlobFS(),
+		common.EFromTo.LocalFile():
+		// TODO: Count files enumerated during scanning.
+		// ^ Will probably get to this in a later PR, would be best to not do it now as other scenarios would falsely output 0 objects found.
+		var traverser resourceTraverser
```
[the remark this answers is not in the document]
adreed-msft: 👍 Will be in my next push

cmd/copy.go
adreed-msft: Yes, already is removed.

cmd/copy.go
adreed-msft: Dropped wildcard implementation, correcting help text

cmd/sync.go
adreed-msft: Yeah, sometimes it'll process in the middle of my work and I'll have to control+Z a bit to get it back

cmd/zc_traverser_file.go
adreed-msft: Done.

cmd/zc_enumerator.go
adreed-msft: There actually never really was a special rule going on here. If I recall correctly, this was because of how I was trying to implement directory checks on source being a pain.

cmd/zc_traverser_list.go
adreed-msft: Nope, no impact, note has been made

cmd/copy.go
adreed-msft: Oh! I actually had no clue, I assumed it'd just initialize a bufferless channel.

cmd/copy.go
adreed-msft: Yes. That was the intent. I'll prevent the combo, I suppose.

cmd/removeEnumerator.go
adreed-msft: Good point, there was no need for this.

cmd/copy.go
adreed-msft: Corrected this behavior.

cmd/syncEnumerator.go
adreed-msft: Oh. Yikes. I'm gonna have to implement this in sync.

adreed-msft: As discussed, we're dropping include and exclude path from sync.

cmd/zt_generic_traverser_test.go
adreed-msft: Let's find out.

adreed-msft: Huh. It does work. Python on the other hand can't handle this behavior, so it won't persist into our smoke tests.

cmd/zt_generic_traverser_test.go
adreed-msft: Filled out these tests, and added another one actually. Had symlinks target both a root and child directory to ensure deduping occurred there.

cmd/copy.go
adreed-msft: Corrected.

cmd/removeEnumerator.go
```diff
@@ -39,30 +41,34 @@ import (
...
 
 	transferScheduler := newRemoveTransferProcessor(cca, NumOfFilesPerDispatchJobPart)
-	includeFilters := buildIncludeFilters(cca.includePatterns)
-	excludeFilters := buildExcludeFilters(cca.excludePatterns)
+	includeFilters := buildIncludeFilters(cca.includePatterns, false) // TODO: Add include-path in here?
+	excludeFilters := buildExcludeFilters(cca.excludePatterns, false)
+	includePathFilters := buildIncludeFilters(cca.includePathPatterns, true)
+	excludePathFilters := buildExcludeFilters(cca.excludePathPatterns, true)
 
 	// set up the filters in the right order
```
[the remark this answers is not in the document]
adreed-msft: This is resolved. All include filters are added before exclude filters. I don't believe this is necessary but better safe than sorry.

Include-path is handled as a part of ListOfFilesChannel if you're wondering where that disappeared to.

cmd/zc_enumerator.go
```diff
@@ -72,6 +82,127 @@ func newStoredObject(name string, relativePath string, lmt time.Time, size int64
...
+
+			go func() {
+				defer close(globChan)
+				for _, v := range matches {
+					globChan <- strings.TrimPrefix(strings.ReplaceAll(v, common.OS_PATH_SEPARATOR, common.AZCOPY_PATH_SEPARATOR_STRING), basePath)
+				}
+			}()
+
+			output = newListTraverser(cleanLocalPath(basePath), "", location, nil, nil, recursive, globChan)
+		} else {
```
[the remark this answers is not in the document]
adreed-msft: Good point. I'll make it take in those parameters.

testSuite/scripts/test_upload_block_blob.py
```diff
@@ -409,7 +409,7 @@ def test_upload_block_blob_include_flag(self):
         # uploading the directory with sub-dir in the include flag.
         result = util.Command("copy").add_arguments(dir_n_files_path).add_arguments(util.test_container_url). \
             add_flags("recursive", "true").add_flags("log-level", "info") \
-            .add_flags("include-path", "sub_dir_include_flag_set_upload/*").add_flags("output-type",
+            .add_flags("include-path", "sub_dir_include_flag_set_upload/").add_flags("output-type",
```
JohnRusk: Out of curiosity, why was it necesary to change the tests here?  I didn't _notice_ anything that would have changed behaviour in this regard, in the code.

cmd/zc_traverser_blob.go
```diff
@@ -33,46 +34,19 @@ import (
...
+	ctx                     context.Context
+	recursive               bool
+	errorOnDirWOutRecursive bool // Error out if this is true and pointing to a directory without recursive
 
 	// a generic function to notify that a new stored object has been enumerated
 	incrementEnumerationCounter func()
 }
 
-func (t *blobTraverser) isDirectory(isDest bool) bool {
+func (t *blobTraverser) isDirectory() bool {
```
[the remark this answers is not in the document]
adreed-msft: I think I'm going to return to the old implementation because I found a better way of handling it.

cmd/copyEnumeratorInit.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// TODO: in download refactor, handle trailing wildcard properly here.
+	// As is, we don't cut it off at the moment,
+	// and we also don't properly handle the "source points to contents" scenario aside from on local, which waives it through wildcard support.
+	traverser, err := initResourceTraverser(cca.source, cca.fromTo.From(), nil, nil, &cca.followSymlinks, cca.listOfFilesChannel, cca.recursive, true, func() {})
+	if err != nil {
+		return nil, err
+	}
+
+	// source directory check occurs within the traversers. This allows us to error out properly when pointing at a folder.
```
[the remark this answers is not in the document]
adreed-msft: The check on the source was moved to the traverser because I'd catch a lot of exceptions with when a transfer would actually succeed on the download refactor. This actually ensured the failure would occur on a directory.

Realistically, I feel as though it might be a little unsafe to implement it any other way. resourceTraversers still have .isDirectory, and we could pretty safely only use that on source with all checks needed, as we currently have the isDestDirectory function.

I did rethink it though, and I'm going to retry the isDirectory approach compared to this setup.

testSuite/scripts/test_upload_block_blob.py
adreed-msft: include-path does not use wildcards when present. Thus, this test would fail.

cmd/zt_generic_traverser_test.go
```diff
@@ -42,11 +41,103 @@ var _ = chk.Suite(&genericTraverserSuite{})
...
+		return nil
+	}), chk.IsNil)
+
+	// 3 files live in base, 3 files live in symlink
+	c.Assert(fileCount, chk.Equals, 6)
+}
+
+// Test cancel symlink loop functionality
+func (s *genericTraverserSuite) TestWalkWithSymlinksBreakLoop(c *chk.C) {
+	fileNames := []string{"thirty to fifty feral hogs.txt", "my feral hogs have unionized.txt", "theyre in the yard.txt"}
```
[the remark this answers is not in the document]
adreed-msft: Yeah, I'll change it. I can understand that reasoning.

cmd/zc_traverser_blob.go
```diff
@@ -22,31 +22,39 @@ package cmd
...
+func (t *blobTraverser) isDirectory(isSource bool) bool {
 	isDirDirect := copyHandlerUtil{}.urlIsContainerOrVirtualDirectory(t.rawURL)
-	return isDirDirect
+
+	// Skip the single blob check if we're checking a destination.
+	// This is an individual exception for blob because blob supports virtual directories and blobs sharing the same name.
+	if isDirDirect || !isSource {
+		return isDirDirect
+	}
+
```
JohnRusk: This new version worked out nice and tidy

cmd/copy.go
```diff
@@ -210,19 +216,57 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
...
+
+			if f != nil {
+				scanner := bufio.NewScanner(f)
+				for scanner.Scan() {
+					v := scanner.Text()
+					listChan <- v
+				}
+			}
+
+			// This occurs much earlier than the other include or exclude filters. I'd _like_ to move it closer in cook() once we get rid of the other bits of list-of-files.
```
zezha-msft: Avoid using 'I' in comments.

cmd/copy.go
```diff
@@ -876,13 +946,16 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
-	case common.EFromTo.LocalBlobFS():
-		fallthrough
-	case common.EFromTo.LocalFile():
-		e := copyUploadEnumerator(jobPartOrder)
-		err = e.enumerate(cca)
+	case common.EFromTo.LocalBlob(),
+		common.EFromTo.LocalBlobFS(),
+		common.EFromTo.LocalFile():
+		var e *copyEnumerator
+		e, err = cca.initEnumerator(jobPartOrder, ctx)
```
zezha-msft: Extremely happy to see this conciseness here.

cmd/zc_traverser_local.go
```diff
@@ -32,13 +32,137 @@ import (
...
+				slPath, err := filepath.Abs(filePath)
+				if err != nil {
+					glcm.Info(fmt.Sprintf("Failed to get absolute path of %s: %s", filePath, err))
+				}
+
+				if _, ok := seenPaths[result]; !ok {
+					seenPaths[result] = true
+					seenPaths[slPath] = true // Note we've seen the symlink as well. We shouldn't ever have issues if we _don't_ do this because
```
zezha-msft: finish the comment after 'because'

cmd/zt_generic_traverser_test.go
```diff
@@ -38,6 +39,175 @@ type genericTraverserSuite struct{}
...
+
+		fileCount++
+		return nil
+	}), chk.IsNil)
+
+	// 3 files live in base, 3 files live in first symlink, second & third symlink is ignored.
+	c.Assert(fileCount, chk.Equals, 6)
+}
+
+func (s *genericTraverserSuite) TestWalkWithSymlinksToParentAndChild(c *chk.C) {
```
zezha-msft: really like these tests 👍

zezha-msft:
I approve-approve, final approve. 😉

Issue #517, opened 2019-07-26
azcopy command hangs


imlight:
### Which version of the AzCopy was used? 
AzCopy 10.2.1

### Which platform are you using? (ex: Windows, Mac, Linux)
Linux Server release 6.10
### What command did you run?
azcopy ( without any argument) same thing happens for other success command , failed command errors out and doesnt get hung .

### What problem was encountered?
azcopy commands are getting hung .

### How can we reproduce the problem in the simplest way?
Download the mentioned version and run the command in Linux 6 or Linux 7

### Have you found a mitigation/solution?
NO

I have to press control C to cancel azcopy command to come in main prompt 

```
# azcopy
AzCopy 10.2.1
Project URL: github.com/Azure/azure-storage-azcopy

AzCopy is a command line tool that moves data into/out of Azure Storage.
To report issues or to learn more about the tool, go to github.com/Azure/azure-storage-azcopy

The general format of the commands is: 'azcopy [command] [arguments] --[flag-name]=[flag-value]'.

Usage:
  azcopy [command]

Available Commands:
  copy        Copies source data to a destination location
  env         Shows the environment variables that can configure AzCopy's behavior
  help        Help about any command
  jobs        Sub-commands related to managing jobs
  list        List the entities in a given resource
  login       Log in to Azure Active Directory to access Azure Storage resources.
  logout      Log out to terminate access to Azure Storage resources.
  make        Create a container/share/filesystem
  remove      Delete entities from Azure Storage Blob/File/ADLS Gen2
  sync        Replicate source to the destination location

Flags:
      --cap-mbps uint32      caps the transfer rate, in Mega bits per second. Moment-by-moment throughput may vary slightly from the cap. If zero or omitted, throughput is not capped.
  -h, --help                 help for azcopy
      --output-type string   format of the command's output, the choices include: text, json. (default "text")
      --version              version for azcopy

Use "azcopy [command] --help" for more information about a command.
^C

```

I am attaching strace output of the this command ( just removed domain names from the output) 
(thanks - I've downloaded and saved the strace - JohnRusk)

imlight:
There is nothing in logs

```
[root@dev-mado21 ~]# ls -al .azcopy/
total 12
drwxr-xr-x   3 root root 4096 Jul 25 10:15 .
dr-x------. 23 root root 4096 Jul 26 10:03 ..
drwxr-xr-x   2 root root 4096 Jul 25 10:15 plans
[...]

Issue #419, opened 2019-05-29
All commands hang with 100% CPU [if have only 1 CPU core]


dalee-bis:
##### Note: The version is visible when running AzCopy without any argument

AzCopy 10.1.2

### Which platform are you using? (ex: Windows, Mac, Linux)

* New Azure VM (nothing installed)
* Windows Server 2016 Datacenter
* Standard DS1 v2 (1 vcpus, 3.5 GiB memory)
* With an identity and permissions to access the storage account (trying to test permissions)

##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.

```
azcopy login
azcopy sync 'C:\Users\me\UploadTest\' https://uploadpermtest.blob.core.windows.net/test/ --recursive --log-level INFO
azcopy list https://uploadpermtest.blob.core.windows.net/test/
```

### What problem was encountered?

Other then when commands are incorrect and help text is displayed, all commands immediately hang with no output and use 100% CPU. Adding the log level argument has no effect.

The only output is a `C:\Users\me\.azcopy\plans` folder which is empty.

### How can we reproduce the problem in the simplest way?

Create an Azure VM matching the setup above, download AzCopy, execute `azcopy login`.

### Have you found a mitigation/solution?

No

zezha-msft:
Hi @dalee-bis, thanks for reaching out!

It is a known issue that AzCopy has trouble in low memory environments. We have it on our backlog to look into.

JohnRusk:
@zezha-msft I ran a few tests. It looks very much like the issues is not memory related, but rather is related to core count. Specifically:
* The failure can be observed with AzCopy login, which does not touch the parts of the code that we have already earmarked for memory usage improvement.
* When I reproduced it on the same VM size as @dalee-bis, Task Manager showed very high CPU usage, but very low RAM usage.
* I can also reproduce it on a much larger VM, by using a CPU affinity mask to restrict AzCopy to just one CPU core. Change the mask to allow the process to use two cores, and the problem goes away.

@dalee-bis, due to the above, I'm going to add an suffix about core count to the issue title.

JohnRusk:
PR #519 includes a change that fixes this.

JohnRusk:
Fixed in version 10.3, which is now released.

PR #519, opened 2019-07-28
Never hang if fail to resolve URL of version file

JohnRusk:
Prior to this fix, login and help commands would hang if aka.ms couldn't be found

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -31,6 +31,7 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"time"
 )
 
 var azcopyAppPathFolder string
@@ -65,8 +66,9 @@ var rootCmd = &cobra.Command{
 		// spawn a routine to fetch and compare the local application's version against the latest version available
 		// if there's a newer version that can be used, then write the suggestion to stderr
 		// however if this takes too long the message won't get printed
-		// Note: this function is only triggered for non-help commands
-		go detectNewVersion()
+		// Note: this function is neccessary for non-help, non-login commands, since they don't reach the corresponding
+		// beginDetectNewVersion call in Execute (below)
+		beginDetectNewVersion()
 
 		return nil
 	},
@@ -86,9 +88,13 @@ func Execute(azsAppPathFolder, logPathFolder string, maxFileAndSocketHandles int
 		glcm.Error(err.Error())
 	} else {
 		// our commands all control their own life explicitly with the lifecycle manager
-		// only help commands reach this point
-		// execute synchronously before exiting
-		detectNewVersion()
+		// only commands that don't explicitly exit actually reach this point (e.g. help commands and login commands)
+		select {
+		case <-beginDetectNewVersion():
+			// noop
+		case <-time.After(time.Second * 8):
+			// don't wait too long
+		}
 		glcm.Exit(nil, common.EExitCode.Success())
 	}
 }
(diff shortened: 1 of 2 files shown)


JohnRusk:
This is to fix #517

JohnRusk:
Also fixes #419

main.go
```diff
@@ -70,3 +73,13 @@ func configureGC() {
 		debug.SetGCPercent(20)       // activate more aggressive/frequent GC than the default
 	}()
 }
+
+// Ensure we always have more than 1 OS thread running goroutines, since there are issues with having just 1.
+// (E.g. version check doesn't happen at login time, if have only one go proc. Not sure why that happens if have only one
+// proc. Is presumably due to the high CPU usage we see on login if only 1 CPU, even tho can't see any busy-wait in that code)
```
zezha-msft: strange indeed.. why is it set to 1 btw? because the system has only 1 core?

cmd/root.go
```diff
@@ -115,63 +121,76 @@ func init() {
 	rootCmd.PersistentFlags().MarkHidden("send-random-data-ext")
 }
 
-func detectNewVersion() {
-	const versionMetadataUrl = "https://aka.ms/azcopyv10-version-metadata"
+// always spins up a new goroutine, because sometimes the aka.ms URL can't be reached (e.g. a constrained environment where
+// aka.ms is not resolvable to a reachable IP address). In such cases, this routine will run for ever, and the caller should
```
zezha-msft: If aka.ms is not resolvable, shouldn't there be a network error and we'd retry it? Why is it stuck forever?

cmd/root.go
```diff
@@ -115,63 +121,76 @@ func init() {
...
+// just give up on it.
+// We spin up the GR here, not in the caller, so that the need to use a separate GC can never be forgotten
+// (if do it synchronously, and can't resolve URL, this blocks caller for ever)
+func beginDetectNewVersion() chan struct{} {
+	completionChannel := make(chan struct{})
+	go func() {
+		const versionMetadataUrl = "https://aka.ms/azcopyv10-version-metadata"
+
+		// step 0: check the Stderr before checking version
```
zezha-msft: Stderr is no longer used.

zezha-msft:
Lgtm, thanks for fixing this!

cmd/root.go
JohnRusk: Good point. Maybe I just gave up before the finite timeout. But it was long enough to fool both me and the user who logged it into thinking it was "for ever"

main.go
JohnRusk: You mean, why would it be set to 1 before we change it?  Yes, because system has only 1 core.  Go runtime tries to match gomaxprocs to the core count, which makes sense given the way the scheduler works in Go.

GezHenry:
@JohnRusk Having same issue on Windows Server 2012 R2 (2 CPUs) on VMWare. Details as follows ...

  >SET AZCOPY_SPA_CLIENT_SECRET=
  >AZCOPY LOGIN --service-principal --application-id --tenant-id=
returns
  Failed to perform login command: adal: Failed to execute the refresh request. Error = 'Post https://login.microsoftonline.com/e1nnnn-nnnnn-nnnn-nnnn-nnnnn/oauth2/token?api-version=1.0: dial tcp 40.126.1.166:443: i/o timeout'

Windows Event Viewer ...
==> REQUEST/RESPONSE (Try=1/21.0252026s[SLOW >3s], OpTime=21.0272058s) -- REQUEST ERROR GET https://aka.ms/azcopyv10-version-metadata?timeout=901 User-Agent: [AzCopy/10.2.1 Azure-Storage/0.7 (go1.12; Windows_NT)] X-Ms-Client-Request-Id: [c7499e49-44c3-4766-7307-769f25b98352] X-Ms-Version:

JohnRusk:
@GezHenry The fix is coming in version 10.3, due to be released soon. (BTW, we only tag the issues, not the PRs, with release milestones, so you can't see the milestone here on this page.  Sorry about that not being clearer.)

PR #525, opened 2019-07-31
Feature/percent complete

JohnRusk:


ste/mgr-JobPartTransferMgr.go
```diff
@@ -107,6 +107,19 @@ type chunkFunc func(int)
...
+	// which are either completed or failed.
+	// NumberOfChunksDone determines the final cancellation or completion of a transfer
+	atomicChunksDone uint32
+
+	// used defensively to protect against accidental double counting
+	atomicCompletionIndicator uint32
+
+	// how many bytes have been successfully transferred
+	// (hard to infer from atomicChunksDone because that counts both successes and failures)
+	atomicSuccessfulBytes uint64
```
zezha-msft: maybe try a windows 32 bit build to make sure the memory alignment works

JohnRusk: I _believe_ that, because I've moved all the atomics to the start of the struct, they are guaranteed to have correct alignment now (whereas before they were not).  So, I propose to close this comment, and leave testing for our already-scheduled 32-bit pre-release testing.

ste/JobsAdmin.go
```diff
@@ -294,8 +300,9 @@ func (ja *jobsAdmin) transferProcessor(workerID int) {
 // There will be only 1 instance of the jobsAdmin type.
 // The coordinator uses this to manage all the running jobs and their job parts.
 type jobsAdmin struct {
-	logger        common.ILoggerCloser
-	jobIDToJobMgr jobIDToJobMgr // Thread-safe map from each JobID to its JobInfo
+	atomicSuccessfulBytesInActiveFiles int64
```
adreed-msft: I could be dead wrong, but I don't think I see anywhere that we clean this up in the event of a failed transfer.

adreed-msft:
Generally speaking looks good to me

ste/JobsAdmin.go
JohnRusk: @adreed-msft It's a bit subtle, because the cleanup is done by the exact same line of code in both the success and failure cases. Specifically, after all chunks in the file have completed, we deduct the transfer's "share" of this value.  We always make that deduction, regardless of whether the file failed or succeeded. I.e. we always deduct that file's successful bytes from this total since regardless of whether the file succeeded or failed, it is now no longer "active" so any bytes that were tracked here for it should now be removed.

So, unless you can see a hole in the above logic, I think this is all good.  I'm going to complete the PR now but, if there is a hole in my logic, please be sure to let me know and I'll make a followup PR.

PR #527, opened 2019-08-01
Fixed minor bug and test for attribute filter

zezha-msft:
- We need to use the relative path to construct the full path, otherwise the get attribute syscall fails.
- Added INFO output in case of error from get attribute call.


cmd/zc_attr_filter_windows.go
```diff
@@ -40,20 +41,22 @@ func (f *attrFilter) doesSupportThisOS() (msg string, supported bool) {
 
 // TODO: review handling of filename for long path support
 func (f *attrFilter) doesPass(storedObject storedObject) bool {
-	fileName := filepath.Join(f.filePath, storedObject.name)
+	fileName := filepath.Join(f.filePath, storedObject.relativePath)
```
JohnRusk: I thought there we cases where relativePath = "" (but this may be my on-going misunderstanding of that field ;-)  I must get I straight in my head sometime!)

cmd/zt_interceptors_for_test.go
```diff
@@ -83,7 +85,7 @@ type mockedLifecycleManager struct{}
 
 func (mockedLifecycleManager) Progress(common.OutputBuilder)                            {}
 func (mockedLifecycleManager) Init(common.OutputBuilder)                                {}
-func (mockedLifecycleManager) Info(string)                                              {}
+func (mockedLifecycleManager) Info(msg string)                                          { fmt.Println(msg) }
```
JohnRusk: Is this supposed to be here permanently? (Only asking because the other method bodies are all empty)

cmd/zc_attr_filter_windows.go
panfeng0007: I might be able to clarify here. Actually `f.filePath` came from `sourceTraverser.fullPath`, which has whatever root folder the users put in the command line when they use `sync`...That's why I used `f.filePath`/`storedObject.name` as a full path. Did the test with `storedObject.relativePath` work?

zezha-msft: @panfeng0007 yes, the test works now. The storedObject.name only has the file name, for example, if the file is located in a sub-dir inside the full path(/users/xxxx/dir), "subdir1/filename.txt", then storedObject.name is just "filename.txt", while storedObject.relativePath is "subdir1/filename.txt".

If the full path pointed to a file already, e.g. /users/xxxx/dir/subdir1/filename.txt, then the storedObject.relativePath is empty "".

cmd/zt_interceptors_for_test.go
zezha-msft: yah this is just for testing, as we often output messages when some folder is skipped

cmd/zc_attr_filter_windows.go
JohnRusk: Thanks @Ze.  I think I finally understand that field now too! :-)

adreed-msft:
Oh hey, I actually encountered this one in the refactor! We had the same patch.

PR #532, opened 2019-08-05
Move smoke tests to devops [READY]

zezha-msft:


zezha-msft:
@JohnRusk it's much faster too! Only 20min instead of 30min+ on Travis.

PR #533, opened 2019-08-06
Added overwrite prompt option for copy

zezha-msft:


ste/overwritePrompter.go
```diff
@@ -0,0 +1,93 @@
...
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/Azure/azure-storage-azcopy/common"
+)
+
+// asks questions to confirm overwrite, one routine at a time, i.e. only one active question at a time
+type overwritePrompter struct {
+	lock *sync.Mutex
```
adreed-msft: Good choice to use a mutex here

ste/xfer-anyToRemote.go
```diff
@@ -81,11 +82,22 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
...
+			if jptm.GetOverwriteOption() == common.EOverwriteOption.Prompt() {
+				// don't show the full URL, as it's too long and hard to read
+				parsed, _ := url.Parse(info.Destination)
+				shouldOverwrite = jptm.GetOverwritePrompter().shouldOverwrite(parsed.Path)
+			}
+
+			if !shouldOverwrite {
+				// logging as Warning so that it turns up even in compact logs, and because previously we use Error here
+				jptm.LogAtLevelForCurrentTransfer(pipeline.LogWarning, "File already exists, so will be skipped")
+				jptm.SetStatus(common.ETransferStatus.FileAlreadyExistsFailure())
```
adreed-msft: Alright here's the bit where I'm less familiar with-- Is this considered a failure or registered as a skip in azcopy's output? If it's considered a failure, I can see some confusion being generated.

ste/xfer-remoteToLocal.go
```diff
@@ -48,18 +48,27 @@ func remoteToLocal(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, d
...
+
+			// if necessary, prompt to confirm user's intent
+			if jptm.GetOverwriteOption() == common.EOverwriteOption.Prompt() {
+				shouldOverwrite = jptm.GetOverwritePrompter().shouldOverwrite(info.Destination)
+			}
+
+			if !shouldOverwrite {
+				// logging as Warning so that it turns up even in compact logs, and because previously we use Error here
+				jptm.LogAtLevelForCurrentTransfer(pipeline.LogWarning, "File already exists, so will be skipped")
+				jptm.SetStatus(common.ETransferStatus.FileAlreadyExistsFailure())
```
adreed-msft: same as xfer-anyToRemote.go:97

ste/overwritePrompter.go
```diff
@@ -0,0 +1,93 @@
...
+	// if false, the user has already specified a "for all" answer
+	shouldPromptUser bool
+
+	// if the user made a "for all" selection, save the response here
+	savedResponse bool
+}
+
+func (o *overwritePrompter) shouldOverwrite(objectPath string) (shouldOverwrite bool) {
+	// only one routine can ask the question or check the saved response at a time
+	o.lock.Lock()
```
adreed-msft: Actually, one thing that does mildly concern me. This is a pointer, shouldn't we be doing some nil checking?

ste/xfer-anyToRemote.go
zezha-msft: It's considered skipped, which I think John just changed to prevent such confusions.

adreed-msft: Ah. I feel like changing the name of things that would cause a skip but not a failure to have a suffix like Skipped instead of Failure (ex. `FileAlreadyExistsSkipped` instead of `FileAlreadyExistsFailure`) would be good for readability. Thoughts?

zezha-msft:
Will update this CR to make it easier for StgExp to parse out the options.

zezha-msft:
Example output:

```
https://zemaintest2.blob.core.windows.net/test/cmd/zc_traverser_local_test.go already exists at the destination. Do you wish to overwrite? Please confirm with: [Y] Yes  [N] No  [A] Yes for all  [L] No for all y
INFO: Confirmed. https://zemaintest2.blob.core.windows.net/test/cmd/zc_traverser_local_test.go will be overwritten.
https://zemaintest2.blob.core.windows.net/test/cmd/zt_copy_blob_download_test.go already exists at the destination. Do you wish to overwrite? Please confirm with: [Y] Yes  [N] No  [A] Yes for all  [L] No for all n
INFO: https://zemaintest2.blob.core.windows.net/test/cmd/zt_copy_blob_download_test.go will be skipped
https://zemaintest2.blob.core.windows.net/test/cmd/s3NameResolver.go already exists at the destination. Do you wish to overwrite? Please confirm with: [Y] Yes  [N] No  [A] Yes for all  [L] No for all l
INFO: No overwriting will happen from now onwards.                                       
0.3 %, 1 Done, 0 Failed, 75 Pending, 1 Skipped, 77 Total, 


Job efc9bcf6-17ee-474c-7635-8b68357e80b0 summary
Elapsed Time (Minutes): 0.2004
Total Number Of Transfers: 77
Number of Transfers Completed: 1
Number of Transfers Failed: 0
Number of Transfers Skipped: 76
TotalBytesTransferred: 1774
Final Job Status: CompletedWithSkipped

```

Originally I was displaying only the relative path from the container, but it looked a bit confusing.. So now I'm showing the full URL.

ste/overwritePrompter.go
zezha-msft: I think we shouldn't do any check here, and let it panic, as there's no sane way of handing the situation where the lock is absent, and we must learn of the programming error as soon as possible (instead of hiding it with a if-check).

ste/xfer-anyToRemote.go
zezha-msft: Renamed accordingly.

ste/overwritePrompter.go
adreed-msft: That's a fair analysis.

---
In reply to: [312795852](https://github.com/Azure/azure-storage-azcopy/pull/533#discussion_r312795852) [](ancestors = 312795852)

zezha-msft:
@JohnRusk Could you please take a look? The CI is failing due to other reason.

ste/init.go
```diff
@@ -432,9 +432,7 @@ func GetJobSummary(jobID common.JobID) common.ListJobSummaryResponse {
 						Dst:            dst,
 						TransferStatus: common.ETransferStatus.Failed(),
 						ErrorCode:      jppt.ErrorCode()}) // TODO: Optimize
-			case common.ETransferStatus.BlobAlreadyExistsFailure(),
-				common.ETransferStatus.FileAlreadyExistsFailure(),
-				common.ETransferStatus.ADLSGen2PathAlreadyExistsFailure():
+			case common.ETransferStatus.SkippedFileAlreadyExists():
```
JohnRusk: This is a good simplification. Is there anywhere (e.g. `jobs show` results, JSON-format progress messages) where the old enums get exposed outside the app? If so, we should note that is a change in the changelog.

cmd/copy.go
```diff
@@ -103,14 +103,13 @@ type rawCopyCmdArgs struct {
...
 	s2sCheckLength           bool
 	// defines the type of the blob at the destination in case of upload / account to account copy
-	blobType        string
-	blockBlobTier   string
-	pageBlobTier    string
-	background      bool
-	output          string
-	acl             string
-	logVerbosity    string
-	cancelFromStdin bool
```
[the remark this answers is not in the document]
zezha-msft: Yes, I verify with Stg Exp that we can safely send cancel signals instead (which is the right thing to do, much less hacky)

ste/init.go
zezha-msft: It doesn't appear that we've ever used `BlobAlreadyExistsFailure` or `ADLSGen2PathAlreadyExistsFailure`.

common/fe-ste-models.go
```diff
@@ -125,6 +125,66 @@ func (dd DeleteDestination) String() string {
...
 
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// represents one possible response
+var EResponseOption = ResponseOption{ResponseType: "", UserFriendlyResponseType: "", ResponseString: ""}
+
+type ResponseOption struct {
+	ResponseType             string // helps us clarify the user's intent and support partner team's localization
+	UserFriendlyResponseType string // text to print in interactive mode
+	ResponseString           string // customizable string to map to ResponseOption, future-proof
```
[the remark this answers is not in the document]
zezha-msft: clarified with better comment 👍

PR #536, opened 2019-08-07
Fixed minor bug in percent complete calculations (divide by 0)

zezha-msft:
When the job finishes, PercentComplete becomes NaN as TotalBytesExpected is 0. It fails the JSON marshaling when --output-type=json


JohnRusk:
@Ze, was this just for 1 summary? Does it also need to be fixed for the sync job summary in addition to copy (or vice versa)?

zezha-msft:
> @ze, was this just for 1 summary? Does it also need to be fixed for the sync job summary in addition to copy (or vice versa)?

Oh yes, I forgot that code is duplicated.. I'll fix it for sync too.

PR #539, opened 2019-08-08
Add distrbuted mutex

zezha-msft:
Final step in transferring our build to DevOps:

1. Lock down the live tests to a single build at a time
2. Fixed unit tests related to concurrency
3. Gave meaningful names to all steps

### DIFF
--- a/common/extensions_test.go
+++ b/common/extensions_test.go
@@ -4,7 +4,11 @@ import (
 	chk "gopkg.in/check.v1"
 )
 
-func (s *mainTestSuite) TestGenerateFullPath(c *chk.C) {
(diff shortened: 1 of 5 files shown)


common/extensions_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	chk "gopkg.in/check.v1"
 )
 
-func (s *mainTestSuite) TestGenerateFullPath(c *chk.C) {
+func (s *ste.mainTestSuite) TestGenerateFullPath(c *chk.C) {
```
zezha-msft: unintentional change

PR #544, opened 2019-08-11
Refactor computeJobXfer switch statement

JohnRusk:
To support new combinations more easily and clearly in the future (e.g. the three Benchmark FromTo's that I expect to add in the Benchmark PR)


zezha-msft:
:shipit:

PR #546, opened 2019-08-12
Download enumerator refactor

adreed-msft:
Pretty roughly based off the upload enumerator refactor. There's only a couple of places that really needed altering.


cmd/copy.go
```diff
@@ -1246,6 +1250,7 @@ func init() {
 	rootCmd.AddCommand(cpCmd)
 
 	// filters change which files get transferred
+	cpCmd.PersistentFlags().BoolVar(&raw.copyContents, "strip-top-dir", false, "copy everything under the source, not the source itself")
```
zezha-msft: >copy everything under the source, not the source itself [](start = 77, length = 55)

consider rephrase it be more user-friendly. "not the source itself" => something like "the top directory itself will not be replicated on the destination side"

cmd/copy.go
```diff
@@ -83,6 +83,7 @@ type rawCopyCmdArgs struct {
 	// filters from flags
 	listOfFilesToCopy string
 	recursive         bool
+	copyContents      bool
```
zezha-msft: >copyContents [](start = 1, length = 12)

rename to correspond to the flag name.

This whole command "copies contents" so the field name can be quite confusing.

cmd/copyDownloadBlobEnumerator.go
```diff
@@ -1,341 +0,0 @@
...
-
-	"strings"
-
-	"github.com/Azure/azure-storage-azcopy/common"
-	"github.com/Azure/azure-storage-azcopy/ste"
-	"github.com/Azure/azure-storage-blob-go/azblob"
-)
-
-type copyDownloadBlobEnumerator common.CopyJobPartOrderRequest
-
```
zezha-msft: aw yiiiis

zezha-msft:
Looking good! Now we are really seeing the advantage of the modular design:
image

zezha-msft:
Please consider adding the changelog entry now, to explain the change in behavior for download, to avoid forgetting the details at release time

cmd/copyDownloadBlobEnumerator.go
zezha-msft: image

testSuite/scripts/test_upload_block_blob.py
```diff
@@ -504,15 +504,15 @@ def test_download_blob_include_flag(self):
             x = json.loads(result, object_hook=lambda d: namedtuple('X', d.keys())(*d.values()))
         except:
             self.fail('error parsing the output in Json Format')
-        self.assertEquals(x.TransfersCompleted, 3)
+        self.assertEquals(x.TransfersCompleted, 6)
```
JohnRusk: It seems weird that we didn't change the test at all, and the test only includes 3 files with the include pattern, but now we are downloading 6 files instead of 3. What's up there?

testSuite/scripts/test_upload_block_blob.py
```diff
@@ -559,14 +559,14 @@ def test_download_blob_exclude_flag(self):
         except:
             self.fail('error parsing the output in JSON Format')
         # Number of expected successful transfer should be 18 since two files in directory are set to exclude
-        self.assertEquals(x.TransfersCompleted, 17)
+        self.assertEquals(x.TransfersCompleted, 14)
```
JohnRusk: Ditto, why the change in behaviour with so little change to the test?

cmd/copy.go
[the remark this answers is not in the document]
adreed-msft: stripTopDir is inferred when wildcarding a local path. Otherwise, yes it is supported with a local source. Because stripTopDir defaults to false, error checking an inconsistency isn't possible because we don't know whether it was _explicitly set_ to false or _defaulted_ to false.

If we wanted to (pointlessly in my opinion) turn it into an enum of (default, false, true), we could manipulate that functionality. However, if a user really wanted the result of --strip-top-dir=false with a wildcard, couldn't they just manually append the top directory to the destination themself?

cmd/copyEnumeratorInit.go
```diff
@@ -40,7 +52,7 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 
 	// Ensure we're only copying from a directory with a trailing wildcard or recursive.
 	isSourceDir := traverser.isDirectory(true)
-	if isSourceDir && !cca.recursive {
+	if isSourceDir && !cca.recursive && !cca.copyContents {
```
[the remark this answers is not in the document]
adreed-msft: So, back when this was `copy-contents` I'd imagined trying to use it _without_ recursive. In this case, it'd have non-recursive sync-like functionality (sans the LMT checking etc.) where it just copies the contents of the directory but not sub-directories.

cmd/zc_traverser_file.go
```diff
@@ -101,24 +101,22 @@ func (t *fileTraverser) traverse(processor objectProcessor, filters []objectFilt
 			for _, fileInfo := range lResp.FileItems {
 				f := currentDirURL.NewFileURL(fileInfo.Name)
 
-				//// TODO: the cost is high while otherwise we cannot get the last modified time. As Azure file's PM description, list might get more valuable file properties later, optimize the logic after the change...
-				//// TODO this traverser is only being used by rm at the moment, so we don't need the properties, uncomment in the future when this is no longer true
-				//fileProperties, err := f.GetProperties(t.ctx)
-				//if err != nil {
-				//	return err
-				//}
+				fileProperties, err := f.GetProperties(t.ctx)
```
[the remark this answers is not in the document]
adreed-msft: This is false. The file enumerator was only used for remove.

The problem I encountered is that with this disabled, I'd duplicate the source to the destination as empty files.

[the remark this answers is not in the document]
adreed-msft: Remove didn't actually need much info to remove the file, hence why we never supplied that information.

We _could_ get the info in STE, yes. Though, files sync is something we intend to do and this info is needed for sync regardless.

cmd/zc_traverser_file.go
```diff
@@ -101,24 +101,22 @@ func (t *fileTraverser) traverse(processor objectProcessor, filters []objectFilt
...
 				relativePath := strings.TrimPrefix(fileURLParts.DirectoryOrFilePath, targetURLParts.DirectoryOrFilePath)
 				relativePath = strings.TrimPrefix(relativePath, common.AZCOPY_PATH_SEPARATOR_STRING)
 
 				storedObject := storedObject{
-					name:         getObjectNameOnly(fileInfo.Name),
-					relativePath: relativePath,
-					//lastModifiedTime: fileProperties.LastModified(),
-					//md5:              fileProperties.ContentMD5(),
-					//size:             fileProperties.ContentLength(),
+					name:             getObjectNameOnly(fileInfo.Name),
```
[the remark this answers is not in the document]
adreed-msft: Changing to the newStoredObject func, I will not revert the getProperties call due to
- Potential future usage in sync
- Getting these properties in STE feels somewhat janky when all other cases get these properties in the enumerator
- The only reason we didn't get properties here is because it was only `rm` that was using it.

testSuite/scripts/test_upload_block_blob.py
adreed-msft: So, include and exclude previously only did things on the root directory (because they were checking the relative path, not the file name).

We ended up declaring this a bug and that the include and exclude filters should affect file names under all sub directories.

adreed-msft: As a result, there are 3 matching files under each directory. Therefore, 6 files get included.

testSuite/scripts/test_upload_block_blob.py
adreed-msft: Explained in above comment.

cmd/copyDownloadBlobEnumerator.go
adreed-msft: image

testSuite/scripts/test_blob_download.py
```diff
@@ -350,19 +357,21 @@ def test_blob_download_wildcard_recursive_true_1(self):
...
+            execute_azcopy_copy_command_get_output()
         # parse the result to get the last job progress summary
         result = util.parseAzcopyOutput(result)
         try:
             # parse the Json Output
             x = json.loads(result, object_hook=lambda d: namedtuple('X', d.keys())(*d.values()))
         except:
             self.fail('error parsing the output in Json Format')
-        # since the wildcards '*/*' exists at the end of dir_name in the sas
+        # since the wildcards '*' exists at the end of dir_name in the sas
```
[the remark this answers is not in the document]
adreed-msft: Done

cmd/copyEnumeratorInit.go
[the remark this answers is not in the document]
JohnRusk: So now that its strip-to-dir, what's the relationship with recursive?  E.g. only usable with recursive?

cmd/zc_traverser_file.go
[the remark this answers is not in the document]
JohnRusk: Ah, I get it now.  Yes, let's say with the properties call that you are making. I had forgotten it wasn't already used for everything.

cmd/copy.go
```diff
@@ -183,15 +183,17 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
...
 	// copy&transform flags to type-safety
-	cooked.copyContents = raw.copyContents
+	cooked.stripTopDir = raw.stripTopDir
 	cooked.recursive = raw.recursive
 	cooked.followSymlinks = raw.followSymlinks
 	cooked.withSnapshots = raw.withSnapshots
 	cooked.forceWrite = raw.forceWrite
 
-	// Local supports wildcards just fine.
+	// cooked.stripTopDir is effectively a workaround for the lack of wildcards in remote sources.
```
JohnRusk: Thanks for the comment. I get it now.

cmd/copy.go
```diff
@@ -957,7 +958,15 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
-		cca.destination = cleanLocalPath(cca.destination)
+		var result string
+		result, err = filepath.Abs(cca.destination)
+
+		if err != nil {
+			return err
+		}
+
+		cca.destination = cleanLocalPath(result)
+		jobPartOrder.DestinationRoot = cca.destination
```
JohnRusk: Minor: not needed because the same thing is done on the very next line (immediately after the switch ends)

cmd/copy.go
```diff
@@ -1250,7 +1259,7 @@ func init() {
 	rootCmd.AddCommand(cpCmd)
 
 	// filters change which files get transferred
-	cpCmd.PersistentFlags().BoolVar(&raw.copyContents, "strip-top-dir", false, "copy everything under the source, not the source itself")
+	cpCmd.PersistentFlags().BoolVar(&raw.stripTopDir, "strip-top-dir", false, "strip the source's root folder from the destination path, akin to \"cp dir/*\" (ex. sourcedir/subdir1/file1 copies to subdir1/file1 on destination)")
```
JohnRusk: Maybe add "whereas without strip-to-dir, it would copy to sourcedir/subdir1/file1 on the destination"

cmd/copyEnumeratorInit.go
```diff
@@ -52,7 +52,7 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 
 	// Ensure we're only copying from a directory with a trailing wildcard or recursive.
 	isSourceDir := traverser.isDirectory(true)
-	if isSourceDir && !cca.recursive && !cca.copyContents {
+	if isSourceDir && !cca.recursive && !cca.stripTopDir {
```
JohnRusk: Please see new question in my earlier thread about this (from first-pass review)

cmd/copy.go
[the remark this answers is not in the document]
JohnRusk: Good point re can't tell why its false.  I'm closing this comment thread.

cmd/copyEnumeratorInit.go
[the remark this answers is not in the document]
adreed-msft: Same relationship as before, I think. @zezha-msft thoughts on retaining this functionality or making it so that it must be used alongside recursive?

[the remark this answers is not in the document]
zezha-msft: `--strip-top-dir` has the meaning of targeting the given path's children instead of itself, in that sense I think the current behavior (target only the direct children but not sub-dirs).

However, the name of the flag may create confusion, so the behavior should be clearly documented and shown in examples.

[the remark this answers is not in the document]
adreed-msft: Alright, document it I will.

cmd/copy.go
```diff
@@ -1258,7 +1258,7 @@ func init() {
 	rootCmd.AddCommand(cpCmd)
 
 	// filters change which files get transferred
-	cpCmd.PersistentFlags().BoolVar(&raw.stripTopDir, "strip-top-dir", false, "strip the source's root folder from the destination path, akin to \"cp dir/*\" (ex. sourcedir/subdir1/file1 copies to subdir1/file1 on destination; whereas without --strip-top-dir, it copies to sourcedir/subdir1/file1)")
+	cpCmd.PersistentFlags().BoolVar(&raw.stripTopDir, "strip-top-dir", false, "strip the source's root folder from the destination path, akin to \"cp dir/*\". Without --recursive, just copies files under the folder but does not recurse into sub-directories akin to a non-recursive sync (ex. sourcedir/subdir1/file1 copies to subdir1/file1 on destination; whereas without --strip-top-dir, it copies to sourcedir/subdir1/file1)")
```
JohnRusk: Can you re-order and tweak wording slightly please, to:

> strip the source's root folder from the destination path, akin to \"cp dir/*\".  E.g. sourcedir/subdir1/file1 copies to subdir1/file1 on destination; whereas without --strip-top-dir, it copies to sourcedir/subdir1/file1.  May be used with and without --recursive. Without --recursive, just copies files under the folder but does not recurse into sub-directories

Reason for suggested change is that the recursive stuff is detail, and having it before the main example would confuse users.  And the allusion to "non-recursive sync" potentially added questions ("Oh, I need to learn what sync does before I can understand this flag" etc)

adreed-msft:
Squash and merge, squash and merge!

PR #548, opened 2019-08-12
Handling missed comments on previous PRs

adreed-msft:


cmd/copy.go
```diff
@@ -1284,7 +1281,7 @@ func init() {
 	cpCmd.PersistentFlags().BoolVar(&raw.background, "background-op", false, "true if user has to perform the operations as a background operation.")
 	cpCmd.PersistentFlags().StringVar(&raw.acl, "acl", "", "Access conditions to be used when uploading/downloading from Azure Storage.")
 
-	cpCmd.PersistentFlags().BoolVar(&raw.s2sCheckLength, "s2s-check-length", true, "Check the length of a file transferred S2S after the transfer. If there is a mismatch, fail the transfer.")
+	cpCmd.PersistentFlags().BoolVar(&raw.CheckLength, "check-length", true, "Check the length of a file transferred S2S after the transfer. If there is a mismatch, fail the transfer.")
```
JohnRusk: Please check entire text here. Some of it still refers, incorrectly now, to S2S.

ste/sender-blockBlobFromLocal.go
```diff
@@ -134,3 +134,13 @@ func (u *blockBlobUploader) Epilogue() {
 
 	u.blockBlobSenderBase.Epilogue()
 }
+
+func (u *blockBlobUploader) GetDestinationLength() (int64, error) {
```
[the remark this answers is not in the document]
JohnRusk: To be clear, I think i'm leaning towards we should stay with the latest code as it is in this PR (i.e. do check by default), but want to know if we've discussed the perf implications and what we think about those implications)

ste/xfer-anyToRemote.go
```diff
@@ -296,17 +296,27 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s ISenderBase, si
 	}
 
 	s.Epilogue() // Perform service-specific cleanup before jptm cleanup. Some services may actually require setup to make the file actually appear.
-	
-	if info.S2SDestLengthValidation {
-		if s2sc, isS2SCopier := s.(s2sCopier); isS2SCopier { // TODO: Implement this for upload and download?
+
+	if info.DestLengthValidation {
+		if s2sc, isS2SCopier := s.(s2sCopier); isS2SCopier {
```
JohnRusk: This is coded as if the method is still defined only in the S2S copier interface.  But 
(a) The method is on all sender/uploaders now, so
(b) it should be defined on the base interface, ISenderBase, and not defined anymore on the S2S (derived) interface
(c) There should be no type assertion here, just use `s` since it will have the method to get the length

ste/xfer-anyToRemote.go
```diff
@@ -296,17 +296,27 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s ISenderBase, si
...
 
 			if err != nil {
-        jptm.FailActiveSend("S2S Length check: Get destination length", err)
+				jptm.FailActiveSend("S2S length check: Get destination length", err)
+			}
+
+			if destLength != jptm.Info().SourceSize {
+				jptm.FailActiveSend("S2S length check", errors.New("destination length does not match source length"))
+			}
+		} else if ulc, isUploader := s.(uploader); isUploader { // TODO: Implement this for download?
```
JohnRusk: This branch is unnecessary if action the comment above

ste/sender-blockBlobFromLocal.go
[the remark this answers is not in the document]
adreed-msft: I don't believe we have discussed this but I agree that we should probably just leave this on by default and recommend customers turn it off if it affects throughput.

cmd/copy.go
zezha-msft: perhaps consider making a note that there's an IO cost

---
In reply to: [313170980](https://github.com/Azure/azure-storage-azcopy/pull/548#discussion_r313170980) [](ancestors = 313170980)

adreed-msft:
Squashing and merging.

PR #549, opened 2019-08-12
Unit & smoke test fixes

adreed-msft:
Attempting to fix unit tests up for Azure DevOps


adreed-msft:
@JohnRusk @zezha-msft Unit tests have been repaired. This _**should**_ be good to go

azure-pipelines.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           # run unit test and build executable
           # the set -e line is needed so that the unit tests failure would cause the job to fail properly
           set -e
-          go test -race -short -cover ./cmd ./common ./ste ./azbfs
+          go test -timeout 30m -race -short -cover ./cmd ./common ./ste ./azbfs
```
zezha-msft: >-timeout 30m [](start = 18, length = 12)

please leave a comment on why the 30 min is required.

cmd/zt_copy_s2smigration_test.go
```diff
@@ -50,7 +50,9 @@ var defaultS2SInvalideMetadataHandleOption = common.DefaultInvalidMetadataHandle
 
 func (s *cmdIntegrationSuite) SetUpSuite(c *chk.C) {
 	s3Client, err := createS3ClientWithMinio(createS3ResOptions{})
-	c.Assert(err, chk.IsNil)
+	if err != nil {
+		return
+	}
```
zezha-msft: please add a comment to explain why? It's hard to know/remember the reason otherwise

go.mod
```diff
@@ -8,21 +8,23 @@ require (
 	github.com/JeffreyRichter/enum v0.0.0-20180725232043-2567042f9cda
 	github.com/cpuguy83/go-md2man v1.0.10 // indirect
 	github.com/danieljoos/wincred v1.0.1
+	github.com/davecgh/go-spew v1.1.1 // indirect
```
zezha-msft: >github.com/davecgh/go-spew v1.1.1 // [](start = 1, length = 36)

the third party notice file needs to be updated accordingly

azure-pipelines.yml
adreed-msft: 30 isn't _super_ necessary. It was more or less a sanity check. I'll set it to 15 and leave a note.

go.mod
adreed-msft: Removed go-spew as it wasn't necessary. It's a leftover dep of minio

cmd/zt_copy_s2smigration_test.go
adreed-msft: Done

cmd/copyS2SMigrationS3Enumerator.go
```diff
@@ -229,7 +208,14 @@ func (e *copyS2SMigrationS3Enumerator) addTransferFromService(ctx context.Contex
 	// defines action need be fulfilled to enumerate bucket further
 	bucketAction := func(bucketInfo minio.BucketInfo) error {
 		// Note: Name resolving is only for destination, source's bucket name should be kept for include/exclude/wildcard.
-		resolvedBucketName, _ := r.ResolveName(bucketInfo.Name) // No error here, as already validated.
+		resolvedBucketName, err := r.ResolveName(bucketInfo.Name) // No error here, as already validated.
```
JohnRusk: This change isn't preserving all of the old behavior, and it probably should.   In particular, we sent messages to gclm in two cases in the old code: if the resolution failed, and if we change the name to resolve it.  THat glcm logging should be made to run here. (unless we have a good reason why we no longer want that logging, but I can't think of such a reason...)

cmd/zt_copy_s2smigration_test.go
```diff
@@ -196,7 +205,7 @@ func (s *cmdIntegrationSuite) TestS2SCopyFromS3ToBlobWithWildcardInSrcAndBucketN
 	// construct the raw input to simulate user input
 	rawSrcS3BucketURL := scenarioHelper{}.getRawS3BucketURL(c, "", bucketName) // Use default region
 	rawDstBlobServiceURLWithSAS := scenarioHelper{}.getRawBlobServiceURLWithSAS(c)
-	rawSrcS3BucketStrWithWirdcard := strings.Replace(rawSrcS3BucketURL.String(), invalidPrefix, "invalid*", 1)
+	rawSrcS3BucketStrWithWirdcard := strings.Replace(rawSrcS3BucketURL.String(), invalidPrefix, "invalid----*", 1)
```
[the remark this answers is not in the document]
adreed-msft: Because we were catching other buckets with similar names that would resolve to something over 63 characters and cause the test to fail.

cmd/zt_test.go
```diff
@@ -430,41 +430,58 @@ func createNewObject(c *chk.C, client *minio.Client, bucketName string, prefix s
 	return
 }
 
-func deleteBucket(c *chk.C, client *minio.Client, bucketName string) {
+func deleteBucket(c *chk.C, client *minio.Client, bucketName string, waitQuarterMinute bool) {
```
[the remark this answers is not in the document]
adreed-msft: Leaving a comment.

go.mod
```diff
@@ -13,16 +13,17 @@ require (
...
 	github.com/mattn/go-ieproxy v0.0.0-20190702010315-6dee0af9227d
-	github.com/minio/minio-go v6.0.12+incompatible
-	github.com/mitchellh/go-homedir v1.0.0 // indirect
+	github.com/minio/minio-go v6.0.14+incompatible
+	github.com/mitchellh/go-homedir v1.1.0 // indirect
+	github.com/pkg/errors v0.8.1 // indirect
 	github.com/smartystreets/goconvey v0.0.0-20190330032615-68dc04aab96a // indirect
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.2 // indirect
+	github.com/stretchr/objx v0.1.1 // indirect
```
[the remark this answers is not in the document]
adreed-msft: Adding to thirdparty.txt

cmd/copyS2SMigrationS3Enumerator.go
adreed-msft: done

PR #550, opened 2019-08-12
CI: make build fail if unit tests fail

zezha-msft:


zezha-msft:
For some reason the go test failure does not fail the job, so we must make it do so explicitly.

PR #553, opened 2019-08-16
Added commands to clean up job files

zezha-msft:
Summary of changes:

0. Allow user to specify where plan files are stored.
1. Allow user to see the job status when listing jobs
2. Added `jobs rm [jobID]` command to delete single job
3. Added `jobs clean --with-status=[Cancelled, Completed, Failed, InProgress, All]` to mass-clean job-related files.

Example:
```
./azcopy jobs clean --with-status=inprogress
INFO: Removing files for job 9f636ac6-783a-354a-4019-8e18ad8aee6c
INFO: Removing files for job be372649-87a8-9b43-7703-9c99712fb4fd
INFO: Removing files for job 1d39f399-8027-da45-464f-02e9a6040877
Successfully removed jobs with status: InProgress.
```

Note: the job status under `jobs` command are all non-enhanced (meaning simple Completed rather than CompletedWithErrors, CompletedWithSkipped, CompletedWithErrorsAndSkipped), this is because we do not read all the parts to calculate how many files got failed/skipped.

### DIFF
--- a/cmd/helpMessages.go
+++ b/cmd/helpMessages.go
@@ -150,12 +150,16 @@ Remove all files associated with the given job ID.
 
 Note that the location of log files and job plan files can be customized. Please refer to the env command.`
 
+const removeJobsCmdExample = "  azcopy jobs rm e52247de-0323-b14d-4cc8-76e0be2e2d44"
+
 const cleanJobsCmdShortDescription = "Clean up log files and job plan files (used for progress tracking and resuming) for jobs"
 
 const cleanJobsCmdLongDescription = `
 Clean up log files and job plan files (used for progress tracking and resuming) for jobs.
 
 Note that the location of log files and job plan files can be customized. Please refer to the env command.`
+
+const cleanJobsCmdExample = "  azcopy jobs clean --with-status=completed"
 
 // ===================================== LIST COMMAND ===================================== //
 const listCmdShortDescription = "List the entities in a given resource"
--- a/cmd/helpMessages.go
+++ b/cmd/helpMessages.go
@@ -143,6 +143,20 @@ const resumeJobsCmdShortDescription = "Resume the existing job with the given jo
 const resumeJobsCmdLongDescription = `
 Resume the existing job with the given job ID.`
 
+const removeJobsCmdShortDescription = "Remove all files associated with the given job ID"
+
+const removeJobsCmdLongDescription = `
+Remove all files associated with the given job ID.
+
+Note that the location of log files and job plan files can be customized. Please refer to the env command.`
+
+const cleanJobsCmdShortDescription = "Clean up log files and job plan files (used for progress tracking and resuming) for jobs"
+
+const cleanJobsCmdLongDescription = `
+Clean up log files and job plan files (used for progress tracking and resuming) for jobs.
+
+Note that the location of log files and job plan files can be customized. Please refer to the env command.`
+
 // ===================================== LIST COMMAND ===================================== //
 const listCmdShortDescription = "List the entities in a given resource"
 
--- a/cmd/jobsClean.go
+++ b/cmd/jobsClean.go
@@ -1,0 +1,132 @@
+// Copyright © Microsoft <wastore@microsoft.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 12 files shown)


cmd/jobsClean.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// remove a single job's log and plan file
+	jobsCleanCmd := &cobra.Command{
+		Use:     "clean",
+		Aliases: []string{"cl"},
+		Short:   cleanJobsCmdShortDescription,
+		Long:    cleanJobsCmdLongDescription,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 0 {
+				return errors.New("clean command does not accept arguments")
```
adreed-msft: Don't really know where to put this, but we should probably have a flag like `--keep-last-x-days` or something of that variety.

My thought is, what if somebody runs azcopy sync daily or whatever and only wants to keep the last 7 days of logs and plan files, in case of a failure?

cmd/jobsClean.go
```diff
@@ -0,0 +1,132 @@
...
+				glcm.Error(fmt.Sprintf("Failed to remove log/plan files due to error: %s.", err))
+			}
+		},
+	}
+
+	jobsCmd.AddCommand(jobsCleanCmd)
+
+	// NOTE: we have way more job status than we normally need, only show the most common ones
+	jobsCleanCmd.PersistentFlags().StringVar(&commandLineInput.withStatus, "with-status", "All",
+		"only remove the jobs with this status, available values: Cancelled, Completed, Failed, InProgress, All")
```
adreed-msft: Just a thought: change this to something like include and/or exclude status, and then have the user be able to supply multiple values?

cmd/jobsRemove.go
```diff
@@ -0,0 +1,127 @@
...
+
+func init() {
+	type JobsRemoveReq struct {
+		JobID common.JobID
+	}
+
+	commandLineInput := JobsRemoveReq{}
+
+	// remove a single job's log and plan file
+	jobsRemoveCmd := &cobra.Command{
```
adreed-msft: Another thought: Combine jobs clean and jobs rm?

cmd/helpMessages.go
```diff
@@ -143,6 +143,20 @@ const resumeJobsCmdShortDescription = "Resume the existing job with the given jo
 const resumeJobsCmdLongDescription = `
 Resume the existing job with the given job ID.`
 
+const removeJobsCmdShortDescription = "Remove all files associated with the given job ID"
```
adreed-msft: @JohnRusk @zezha-msft how do you feel about removing all of the individual `const` tokens from this file and doing a `const` block instead?

ex.
```const(
    a = "test"
    b = "test2"
)```

File: cmd/jobsClean.go (hunk 1 above)
Comment (JohnRusk) [reply cmd/jobsClean.go]: So they could do this:

`azcopy clean --keep-days 7`

I like that idea.  Not sure Ze will have time to include it in the release.  I'd be comfortable with us shipping without it if he's under time pressure.

File: cmd/jobsClean.go (hunk 2 above)
Comment (JohnRusk) [reply cmd/jobsClean.go]: Probably more sophistication than we have time for I think.  And, given the limited set of values, users can just run it, say, twice, if they want to delete those with 2 values.

File: cmd/jobsRemove.go (hunk 3 above)
Comment (JohnRusk) [reply cmd/jobsRemove.go]: I initially thought that too, but given that clean is so much wider in effect than rm, I came around to the idea of keeping them separate.

File: cmd/helpMessages.go (hunk 4 above)
Comment (JohnRusk) [reply cmd/helpMessages.go]: Maybe one day. Not a priority for this release IMHO

File: cmd/helpMessages.go [hunk 5]
```diff
@@ -143,6 +143,20 @@ const resumeJobsCmdShortDescription = "Resume the existing job with the given jo
...
+
+const removeJobsCmdLongDescription = `
+Remove all files associated with the given job ID.
+
```
JohnRusk: Would be real nice to have examples in here too, as we do for most other commands.

cmd/jobsRemove.go
```diff
@@ -0,0 +1,127 @@
...
+			return true
+		}
+		return false
+	})
+	if err != nil {
+		return err
+	}
+
+	// get rid of the logs
+	// even though we only have 1 file right now, still scan the directory since we may change the
```
JohnRusk: In debug mode we'll also have <jobid>-chunks.log (the chunk status log).  This code will, correctly, pick that up too.

main.go
```diff
@@ -38,20 +39,32 @@ var glcm = common.GetLifecycleMgr()
...
 	if azcopyLogPathFolder == "" {
 		azcopyLogPathFolder = azcopyAppPathFolder
 	}
 
+	// the user can optionally put the plan files somewhere else
+	azcopyJobPlanFolder := common.GetLifecycleMgr().GetEnvironmentVariable(common.EEnvironmentVariable.JobPlanLocation())
+	if azcopyJobPlanFolder == "" {
+		azcopyJobPlanFolder = path.Join(azcopyAppPathFolder, "plans")
+	}
+	if err := os.Mkdir(azcopyJobPlanFolder, os.ModeDir|os.ModePerm); err != nil && !os.IsExist(err) {
```
JohnRusk: We force creation of the plan dir, but I'm not sure whether we force creation of the log dir. (In the old code, forcing the creation of the plan dir automatically forced creation of log dir, because plan was child of log. But that's not the case now).

Do we need to add code to create the log dir too, now that it might not be parent of the plan dir?

cmd/jobsClean.go
zezha-msft: I thought about it too, but wasn't sure whether we should use the last modified time of the job files, or the start time of the job.

I'll leave it for now, and wait to receive some user feedback.

cmd/jobsClean.go
zezha-msft: Yah same as the keep-days, let's wait for some user feedback before adding more sophistication.

main.go
zezha-msft: Good point, we should be consistent and force create both locations

PR #557, opened 2019-08-19
Log CPK errors

adreed-msft:
Lives somewhat atop the DL refactor. Will change the base branch once DL is merged in.

### DIFF
--- a/cmd/zc_traverser_blob.go
+++ b/cmd/zc_traverser_blob.go
@@ -57,7 +57,7 @@ func (t *blobTraverser) isDirectory(isSource bool) bool {
 
 	if stgErr, ok := err.(azblob.StorageError); ok {
 		// We know for sure this is a single blob still, let it walk on through to the traverser.
-		if stgErr.ServiceCode() == "BlobUsesCustomerSpecifiedEncryption" {
+		if stgErr.ServiceCode() == common.CPK_ERROR_SERVICE_CODE {
 			return true
 		}
 	}
@@ -86,8 +86,8 @@ func (t *blobTraverser) traverse(processor objectProcessor, filters []objectFilt
 
 	if stgErr, ok := propErr.(azblob.StorageError); ok {
 		// We know for sure this is a single blob still, let it walk on through to the traverser.
-		if stgErr.ServiceCode() == "BlobUsesCustomerSpecifiedEncryption" {
-			return errors.New("this blob uses customer provided encryption keys (CPK). at the moment, AzCopy does not support CPK-encrypted blobs. " +
+		if stgErr.ServiceCode() == common.CPK_ERROR_SERVICE_CODE {
+			return errors.New("this blob uses customer provided encryption keys (CPK). At the moment, AzCopy does not support CPK-encrypted blobs. " +
 				"If you wish to make use of this blob, we recommend using one of the Azure Storage SDKs")
 		}
 	}
--- a/cmd/zc_traverser_blob.go
+++ b/cmd/zc_traverser_blob.go
@@ -58,7 +58,7 @@ func (t *blobTraverser) isDirectory(isSource bool) bool {
 	if stgErr, ok := err.(azblob.StorageError); ok {
 		// We know for sure this is a single blob still, let it walk on through to the traverser.
 		if stgErr.ServiceCode() == common.CPK_ERROR_SERVICE_CODE {
-			return true
+			return false
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/zc_traverser_blob.go
```diff
@@ -52,29 +53,45 @@ func (t *blobTraverser) isDirectory(isSource bool) bool {
...
 	util := copyHandlerUtil{}
 
 	// check if the url points to a single blob
-	blobProperties, isBlob := t.getPropertiesIfSingleBlob()
+	blobProperties, isBlob, propErr := t.getPropertiesIfSingleBlob()
+
+	if stgErr, ok := propErr.(azblob.StorageError); ok {
+		// We know for sure this is a single blob still, let it walk on through to the traverser.
+		if stgErr.ServiceCode() == "BlobUsesCustomerSpecifiedEncryption" {
+			return errors.New("this blob uses customer provided encryption keys (CPK). at the moment, AzCopy does not support CPK-encrypted blobs. " +
```
zezha-msft: >a [](start = 78, length = 1)

Capitalize.

ste/mgr-JobPartTransferMgr.go
```diff
@@ -483,7 +483,15 @@ func (jptm *jobPartTransferMgr) failActiveTransfer(typ transferErrorCode, descri
 	//  consider redesign the lifecycle management in ste
 	if !jptm.WasCanceled() {
 		jptm.Cancel()
-		status, msg := ErrorEx{err}.ErrorCodeAndString()
+		serviceCode, status, msg := ErrorEx{err}.ErrorCodeAndString()
+
+		if serviceCode == "BlobUsesCustomerSpecifiedEncryption" {
```
zezha-msft: >"BlobUsesCustomerSpecifiedEncryption" [](start = 20, length = 37)

consider making it a constant in the common package

ste/mgr-JobPartTransferMgr.go
```diff
@@ -483,7 +483,15 @@ func (jptm *jobPartTransferMgr) failActiveTransfer(typ transferErrorCode, descri
...
+			cpkAccessFailureLogGLCM.Do(func() {
+				common.GetLifecycleMgr().Info("One or more transfers have failed because AzCopy currently does not support blobs encrypted with customer provided keys (CPK). " +
+					"If you wish to access CPK-encrypted blobs, we recommend using one of the Azure Storage SDKs to do so.")
```
zezha-msft: consider making into a constant in the common package, easier to find/replace afterwards

adreed-msft:
Rebasing on dev

cmd/zc_traverser_blob.go
```diff
@@ -52,29 +53,45 @@ func (t *blobTraverser) isDirectory(isSource bool) bool {
...
 func (t *blobTraverser) traverse(processor objectProcessor, filters []objectFilter) (err error) {
 	blobUrlParts := azblob.NewBlobURLParts(*t.rawURL)
 	util := copyHandlerUtil{}
 
 	// check if the url points to a single blob
-	blobProperties, isBlob := t.getPropertiesIfSingleBlob()
+	blobProperties, isBlob, propErr := t.getPropertiesIfSingleBlob()
+
+	if stgErr, ok := propErr.(azblob.StorageError); ok {
+		// We know for sure this is a single blob still, let it walk on through to the traverser.
```
JohnRusk: Is this comment what you really meant?  Looks like a copy/paste error to me...

cmd/zc_traverser_blob.go
```diff
@@ -52,29 +53,45 @@ func (t *blobTraverser) isDirectory(isSource bool) bool {
 		return isDirDirect
 	}
 
-	_, isSingleBlob := t.getPropertiesIfSingleBlob()
+	_, isSingleBlob, err := t.getPropertiesIfSingleBlob()
+
+	if stgErr, ok := err.(azblob.StorageError); ok {
+		// We know for sure this is a single blob still, let it walk on through to the traverser.
+		if stgErr.ServiceCode() == common.CPK_ERROR_SERVICE_CODE {
+			return true
```
JohnRusk: Returning true here seems strange to me.  E.g. if the blob is the source of a copy, and user has not used --recursive, won't this result in the display of the message "cannot use directory as source without --recursive or trailing wildcard (/*)"  That's confusing to the user. 

Intuitively it feels more correct to return false here, but I'm guessing from the comment there's some particular reason why true was chosen in this case.  What was that reason, and what about the issue with it displaying the "cannot use directory as a source" message?

adreed-msft:
Squashing and merging

PR #560, opened 2019-08-21
Benchmark mode

JohnRusk:
Key elements are:
* benchmark command itself: e.g. `AzCopy bench "<containerSASUrl>"`
* benchmark enumerator to create test files of random(ish) data
* auto-tuning of concurrency (necessary to know whether we've done our best to use available network bandwidth)
* additional stats gathered in pipeline (most importantly detailed 503 stats, by type)
* performance advice generator to produce the benchmark output
* cleanup of benchmark files
* CPU usage detection (a necessary input to the performance advisor).  I've used some old experimental code that I had here.  May need to swap this out and use gopsutil instead.  Advantage of the current code is that its easier to test, because you can simulate a lack of CPU just by launching the process with constrained CPU affinity (e.g. taskset on Linux).  Can't do that if we switch to gopsutil since it will give us "whole-of-machine" usage percent, without regard to how many of the machine's CPUs our particular process is allowed to use.  Biggest disadvantage of the current CPU detection code is that it's dependent on some empirically-derived constants, and so may not hold up in all different runtime environments. (Which is why there's also an environment variable to allow opt-out of this feature when benchmarking).


cmd/benchmark.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+	rootCmd.AddCommand(benchCmd)
+
+	benchCmd.PersistentFlags().StringVar(&raw.sizePerFile, common.SizePerFileParam, "250M", "size of each auto-generated data file. Must be "+sizeStringDescription)
+	benchCmd.PersistentFlags().UintVar(&raw.fileCount, common.FileCountParam, 100, "number of auto-generated data files to use")
+
+	benchCmd.PersistentFlags().Float64Var(&raw.blockSizeMB, "block-size-mb", 0, "use this block size (specified in MiB). Default is automatically calculated based on file size. Decimal fractions are allowed - e.g. 0.25. Identical to the same-named parameter in the copy command")
+	benchCmd.PersistentFlags().StringVar(&raw.blobType, "blob-type", "None", "defines the type of blob at the destination. Used to allow benchmarking different blob types. Identical to the same-named parameter in the copy command")
+	benchCmd.PersistentFlags().BoolVar(&raw.putMd5, "put-md5", false, "create an MD5 hash of each file, and save the hash as the Content-MD5 property of the destination blob/file. (By default the hash is NOT created.) Identical to the same-named parameter in the copy command")
+	// TODO use constant for default value or, better, move loglevel param to root cmd?
```
adreed-msft: Moving to root cmd would make sense, imo. logging is a pretty global thing in AzCopy.

cmd/copy.go
```diff
@@ -536,18 +542,26 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	return cooked, nil
 }
 
+// When other commands use the copy command arguments to cook cook, set the blobType to None and validation option
+// else parsing the arguments will fail.
+func (raw *rawCopyCmdArgs) setMandatoryDefaults() {
```
adreed-msft: 👍

common/output.go
```diff
@@ -16,9 +16,14 @@ type outputMessageType uint8
 func (outputMessageType) Init() outputMessageType     { return outputMessageType(0) } // simple print, allowed to float up
 func (outputMessageType) Info() outputMessageType     { return outputMessageType(1) } // simple print, allowed to float up
 func (outputMessageType) Progress() outputMessageType { return outputMessageType(2) } // should be printed on the same line over and over again, not allowed to float up
-func (outputMessageType) Exit() outputMessageType     { return outputMessageType(3) } // exit after printing
-func (outputMessageType) Error() outputMessageType    { return outputMessageType(4) } // indicate fatal error, exit right after
-func (outputMessageType) Prompt() outputMessageType   { return outputMessageType(5) } // ask the user a question after erasing the progress
+
+// TODO: REVIEWERS: this next code was called Exit. But with followup jobs its not actually Exit.  Should we still call it exit
```
adreed-msft: I'd almost say that "exit" is something totally different now. Perhaps we should have EndOfJob and Exit, to indicate both states, as EndOfJob is now an ambiguous way to end something.

ste/JobsAdmin.go
```diff
@@ -170,8 +181,13 @@ func initJobsAdmin(appCtx context.Context, concurrency ConcurrencySettings, targ
 			lowTransferCh:    lowTransferCh,
 			normalChunckCh:   normalChunkCh,
 			lowChunkCh:       lowChunkCh,
-			suicideCh:        suicideCh,
 		},
+		poolSizingChannels: poolSizingChannels{ // all deliberately unbuffered
```
adreed-msft: Add a reason here?

ste/concurrencyTuner.go
```diff
@@ -0,0 +1,275 @@
...
+
+func (n *nullConcurrencyTuner) RequestCallbackWhenStable(callback func()) (callbackAccepted bool) {
+	return false
+}
+
+func (n *nullConcurrencyTuner) GetFinalState() (finalReason string, finalRecommendedConcurrency int, tm time.Time) {
+	return concurrencyReasonTunerDisabled, 0, time.Time{}
+}
+
+type autoConcurrencyTuner struct {
```
adreed-msft: Not super sure where to put this, but perhaps we could cache the final value of this somewhere so a user wouldn't have to re-tune every time?

ste/mgr-JobMgr.go
```diff
@@ -231,6 +256,39 @@ func (jm *jobMgr) logPerfInfo(displayStrings []string, constraint common.PerfCon
 	jm.Log(pipeline.LogInfo, msg)
 }
 
+// TODO: find a better way for JobsAdmin to log (it doesn't have direct access to the job log, because it was originally designed to support multilpe jobs
```
adreed-msft: Perhaps if/when we get around to refactoring this general area of code we can unify jobsAdmin and jobMgr

cmd/helpMessages.go
```diff
@@ -292,9 +292,49 @@ Sync an entire directory but exclude certain files from the scope (ex: every fil
...
 By default, the files are stored in a folder named 'doc' inside the current directory.
 `
+
+// ===================================== BENCH COMMAND ===================================== //
+const benchCmdShortDescription = "Performs a performance benchmark"
+
+// TODO: document whether we delete the uploaded data
+
+const benchCmdLongDescription = `
+Performs a performance benchmark by generating test data and uploading it to a specified destination.
```
zezha-msft: >Performs [](start = 0, length = 8)

Minor: change wording to be less redundant.

JohnRusk:
TODO: test after rebased onto S2S refactoring, since it has some case statements that don't know about the new Benchmark location type.  E.g. in pathUtils.go

cmd/benchmark.go
JohnRusk: Yeah, will leave as TODO for now, since its currently at child level everywhere, and I don't have time to fix (and test) everywhere.  Will leave the TODO there to remind us in the future.

ste/JobsAdmin.go
JohnRusk: Done

ste/concurrencyTuner.go
JohnRusk: Yeah, that's tempting... Tricky part is not just "where do we cache it", but also the fact that the correct value depends on:
* file size
* target account type
* target region
* other network traffic (and therefore, potentially, time-of-day)
* S2S vs upload/download

So there's no one-size fits all answer.  I think, long term, it could be nice to make the auto-tuning the default, but we should see how auto-tuning performs in benchmark mode first before making it the default.  If we make it the default, we may also want to do a variation of your suggestion - rather than cache an exact value, cache the starting value for the auto-tuning. So that time is not wasted on low values, like 8 or 16, in cases where we know that it will end up tuning to something much higher.

cmd/helpMessages.go
JohnRusk: Thanks. Done.

ste/mgr-JobMgr.go
JohnRusk: That's an interesting thought.... Let's keep it in mind.

common/output.go
JohnRusk: Yeah, I'm not entirely happy with what I've done here.  

@Zeza-msft you know a lot about how the JSON output is used, which is where my change is most likely to be a breaking one.  What do you think?

cmd/benchmark.go
```diff
@@ -0,0 +1,329 @@
...
+	blobType     string
+	output       string
+	logVerbosity string
+}
+
+const (
+	maxBytesPerFile = 4.75 * 1024 * 1024 * 1024 * 1024
+
+	// TODO would it be better to have a trailing B, eg. 12KB or 200GB? (that might make it case sensitive, or at least
+	//    necessitate making the B case sensitive, because lowercase b means bits (and we don't want to bother supporting bits)
```
zezha-msft: current approach seems fine. No ambiguity

cmd/benchmark.go
```diff
@@ -0,0 +1,329 @@
...
+	sizeStringDescription = "a number immediately followed by K, M or G. E.g. 12k or 200G"
+)
+
+// TODO move to copy handler util
```
zezha-msft: we should get rid of the util file or at least not add to it. It became a dumping ground for unrelated code.

cmd/zt_parseSize_test.go
```diff
@@ -0,0 +1,58 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package cmd
+
+import (
+	chk "gopkg.in/check.v1"
+)
+
+type parseSizeSuite struct{}
```
zezha-msft: I love unit tests :heart:

cmd/zc_traverser_benchmark.go
```diff
@@ -0,0 +1,77 @@
+// Copyright © 2017 Microsoft <wastore@microsoft.com>
```
zezha-msft: >2017 [](start = 15, length = 4)

remove 2017

cmd/zc_traverser_benchmark.go
```diff
@@ -0,0 +1,77 @@
...
+func (t *benchmarkTraverser) isDirectory(bool) bool {
+	return true
+}
+
+func (t *benchmarkTraverser) traverse(processor objectProcessor, filters []objectFilter) (err error) {
+	if len(filters) > 0 {
+		panic("filters not expected or supported in benchmark traverser") // but we still call processIfPassedFilters below, for consistency with other traversers
+	}
+
+	for i := uint(1); i <= t.fileCount; i++ {
```
zezha-msft: >for i := uint(1); i <= t.fileCount; i++ { [](start = 1, length = 41)

nice and elegant

cmd/helpMessages.go
```diff
@@ -292,9 +292,49 @@ Sync an entire directory but exclude certain files from the scope (ex: every fil
...
+
+// TODO: document whether we delete the uploaded data
+
+const benchCmdLongDescription = `
+Runs a performance benchmark, by uploading uploading test data to a specified destination.
+The test data is automatically generated.
+
+The benchmark command runs the same upload process as 'copy', except that: 
+  - there's no source parameter.  The command requires only a destination (which is typically a blob container, 
+    but may be any of the upload destinations supported by 'copy')
```
zezha-msft: update to say that only container is supported at the moment

cmd/benchmark.go
```diff
@@ -0,0 +1,329 @@
...
+	cooked.cleanupJobMessage = "Running cleanup job to delete files created during benchmarking"
+	return &cooked, err
+}
+
+type benchmarkSourceHelper struct{}
+
+// our code requires sources to be strings. So we may as well do the benchmark sources as URLs
+// so we can identify then as such using a specific domain. ".invalid" is reserved globally for cases where
+// you want a URL that can't possibly be a real one, so we'll use that
+const benchmarkSourceHost = "benchmark.invalid"
```
zezha-msft: >benchmarkSourceHost [](start = 6, length = 19)

pretty cool

common/output.go
zezha-msft: @JohnRusk EndOfJob is fine, as Stg Exp will rework the JSON parsing anyway due to Prompt messages.

JohnRusk: Adele, I realized that anything listening to our stdout will have to be monitoring for unexpected process exit anyway, so it must be already on the lookout for closure (EOF) of our stdout.  So for now, we don't need a separate exit signal. But I've left a TODO in the code to consider whether we might need one in the future.

adreed-msft: Good point. Yeah, this should probably just be left as a future work item to make output more clear in the case that we do ever start adding cleanup behavior to azcopy.

JohnRusk:
Rebased to dev

cmd/copy.go
```diff
@@ -1105,7 +1109,9 @@ func (cca *cookedCopyCmdArgs) launchFollowup(priorJobExitCode common.ExitCode) {
 		glcm.AllowReinitiateProgressReporting()
 		cca.followupJobArgs.priorJobExitCode = &priorJobExitCode
 		err := cca.followupJobArgs.process()
-		if err != nil {
+		if err == NothingToRemoveError {
+			glcm.Error("Cleanup completed (nothing needed to be deleted)")
```
zezha-msft: Did you mean to call glcm.Error? It'll have a non-zero exit code. If it's intended to be handled gracefully, Exit is fine right?

JohnRusk: Good point, thanks. I'll fix it now.

PR #564, opened 2019-08-26
Added support for Blob->Blob sync

zezha-msft:
Note: the call to get job status for sync job was removed, since we do deletes in the front end, and that code was duplicated from the copy code.


cmd/syncEnumerator.go
```diff
@@ -21,139 +21,146 @@
...
 	if err != nil {
 		return nil, err
 	}
 
 	// verify that the traversers are targeting the same type of resources
-	_, isSingleBlob := destinationTraverser.getPropertiesIfSingleBlob()
-	_, isSingleFile, _ := sourceTraverser.getInfoIfSingleFile()
-	if isSingleBlob != isSingleFile {
-		return nil, errors.New("sync must happen between source and destination of the same type: either blob <-> file, or container/virtual directory <-> local directory")
+	if sourceTraverser.isDirectory(true) != destinationTraverser.isDirectory(true) {
```
adreed-msft: Glad these additions are helping out here

adreed-msft:
LGTM. Uses all the architecture that's already been laid out and doesn't overcomplicate it.

cmd/sync.go
```diff
@@ -318,25 +322,24 @@ func (cca *cookedSyncCmdArgs) reportScanningProgress(lcm common.LifecycleMgr, th
...
 
-func (cca *cookedSyncCmdArgs) getJsonOfSyncJobSummary(summary common.ListSyncJobSummaryResponse) string {
-	// TODO figure out if deletions should be done by the enumeration engine or not
-	// TODO if not, remove this so that we get the proper number from the ste
-	summary.DeleteTotalTransfers = cca.getDeletionCount()
-	summary.DeleteTransfersCompleted = cca.getDeletionCount()
-	jsonOutput, err := json.Marshal(summary)
+func (cca *cookedSyncCmdArgs) getJsonOfSyncJobSummary(summary common.ListJobSummaryResponse) string {
+	wrapped := common.ListSyncJobSummaryResponse{ListJobSummaryResponse: summary}
+	wrapped.DeleteTotalTransfers = cca.getDeletionCount()
```
JohnRusk: Will this be a breaking change for any callers, in terms of how things are named in the JSON?  Not saying that would stop us doing it, just that if so we should be sure to list it in changelog.md (even tho there are not likely to be many users of this yet, I'm guessing)

cmd/zt_sync_blob_blob_test.go
```diff
@@ -0,0 +1,628 @@
...
+	blobList := scenarioHelper{}.generateCommonRemoteScenarioForBlob(c, srcContainerURL,
+		vdirName+common.AZCOPY_PATH_SEPARATOR_STRING)
+	c.Assert(len(blobList), chk.Not(chk.Equals), 0)
+
+	// set up the destination with the exact same files
+	scenarioHelper{}.generateBlobsFromList(c, dstContainerURL, blobList, blockBlobDefaultData)
+
+	// create an ADLS Gen2 directory at the source with the exact same name as the vdir
+	_, err := srcContainerURL.NewBlockBlobURL(vdirName).Upload(context.Background(), bytes.NewReader(nil),
+		azblob.BlobHTTPHeaders{}, azblob.Metadata{"hdi_isfolder": "true"}, azblob.BlobAccessConditions{})
```
JohnRusk: Is this something we also need to be aware of in "copy"?

PR #565, opened 2019-08-27
S2S refactor (S3, File, Blob -> Blob)

adreed-msft:
This one is considerably larger scale a portion of the refactor than previous refactor portions.

I had to make adjustments to determine the "level" of a path we're supplied, so I could handle scenarios like service -> service and container migration.

This also leads into the path generator as well.


cmd/copyEnumeratorInit.go
```diff
@@ -59,16 +76,87 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
-		src := cca.makeEscapedRelativePath(true, isDestDir, object)
-		dst := cca.makeEscapedRelativePath(false, isDestDir, object)
+		// Start by resolving the name and creating the container
+		if object.containerName != "" && cca.fromTo.From() != common.ELocation.Local() && cca.fromTo.To() != common.ELocation.Local() {
+			cName, err := s3BucketResolver.ResolveName(object.containerName)
+
+			if err != nil {
+				glcm.Error(err.Error())
+				return errors.New("failed to add transfers from service, some of the buckets have invalid names for the destination. " +
+					"Please exclude the invalid buckets in service to service copy, and copy them using bucket to container/share/filesystem copy " +
```
zezha-msft: >exclude the invalid buckets [](start = 13, length = 27)

how do they exclude buckets?

cmd/credentialUtil.go
```diff
@@ -113,10 +109,17 @@ func getBlobCredentialType(ctx context.Context, blobResourceURL string, canBePub
...
 		if isContainer {
 			containerURL := azblob.NewContainerURL(*resourceURL, p)
+			bURLparts := azblob.NewBlobURLParts(*resourceURL)
+
+			if bURLparts.ContainerName == "" || strings.Contains(bURLparts.ContainerName, "*") {
+				// Service level searches can't possibly be public.
+				return false
+			}
+
 			if _, err := containerURL.GetProperties(ctx, azblob.LeaseAccessConditions{}); err == nil {
```
zezha-msft: >containerURL.GetProperties [](start = 16, length = 26)

the blobs may be public but the container may not be. perhaps test with such a scenario to confirm behavior

cmd/zc_enumerator.go
```diff
@@ -180,7 +206,19 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 			return nil, errors.New("a valid credential and context must be supplied to create a blob traverser")
 		}
 
-		output = newBlobTraverser(sourceURL, *p, *ctx, recursive, incrementEnumerationCounter)
+		burl := azblob.NewBlobURLParts(*sourceURL)
+
+		if burl.ContainerName == "" || strings.Contains(burl.ContainerName, "*") {
+			// TODO service traverser
```
zezha-msft: >// TODO service traverser [](start = 3, length = 25)

the TODOs are done right?

cmd/zc_enumerator.go
```diff
@@ -68,17 +82,27 @@ func (storedObject *storedObject) isMoreRecentThan(storedObject2 storedObject) b
...
 }
 
+// used to get properties in a safe, but not so verbose manner
+func defaultStringIfPointerIsNil(wanted *string, instead string) string {
+	if wanted == nil {
+		return instead
+	}
+
+	return *wanted
+}
```
zezha-msft: find better place since this is not related to enumerator definitions

cmd/zc_traverser_blobfs.go
```diff
@@ -127,8 +136,23 @@ func (t *blobFSTraverser) traverse(processor objectProcessor, filters []objectFi
 					*v.ContentLength,
 					v.ContentMD5(),
 					blobTypeNA,
+					bfsURLParts.FileSystemName,
 				)
 
+				/* TODO: Enable this code segment in the case we ever do BlobFS->Blob transfers.
+
```
zezha-msft: it'll be just blob -> blob at that point, with the interop support

zezha-msft:
> // Enumerates an entire blobFS account, looking into each matching filesystem as it goes

please make sure that all the new scenarios are covered by automated tests (integration or smoke)

---
Refers to: cmd/zc_traverser_blobfs_account.go:35 in ebe53ef. [](commit_id = ebe53efe35fca1a29e5fbcffab7873c8c4fa88ad, deletion_comment = False)

cmd/copyEnumeratorInit.go
adreed-msft: You'd have to use a wildcard to do so. Perhaps we should just automatically exclude them and log that we excluded them.

zezha-msft: Yes, that sounds like a better experience, since it's hard to use wildcard for that purpose, as they have to figure out a pattern that works for all the other buckets except the one to exclude.

cmd/credentialUtil.go
adreed-msft: That scenario already is tested. You can't list blobs if a container isn't public, thus it would have to be a direct link to a blob, which we'd attempt to get properties to. I'll test this regardless though.

cmd/zc_enumerator.go
adreed-msft: Yes, removing the comment.

cmd/zc_traverser_blobfs.go
adreed-msft: Yup

adreed-msft: I more or less just left it for sanity's sake.

cmd/copyEnumeratorInit.go
adreed-msft: Done

cmd/zc_enumerator.go
adreed-msft: Done

adreed-msft: Changed to IffStringNotNil and placed in iff.go

cmd/zc_enumerator.go
```diff
@@ -191,7 +219,19 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 			return nil, errors.New("a valid credential and context must be supplied to create a file traverser")
 		}
 
-		output = newFileTraverser(sourceURL, *p, *ctx, recursive, incrementEnumerationCounter)
+		furl := azfile.NewFileURLParts(*sourceURL)
+
+		if furl.ShareName == "" || strings.Contains(furl.ShareName, "*") {
+			// TODO service traverser
```
JohnRusk: What's this TODO?

cmd/zc_enumerator.go
```diff
@@ -202,11 +242,62 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 			return nil, errors.New("a valid credential and context must be supplied to create a blobFS traverser")
 		}
 
-		output = newBlobFSTraverser(sourceURL, *p, *ctx, recursive, incrementEnumerationCounter)
+		bfsURL := azbfs.NewBfsURLParts(*sourceURL)
+
+		if bfsURL.FileSystemName == "" || strings.Contains(bfsURL.FileSystemName, "*") {
+			// TODO service traverser
```
JohnRusk: Ditto

cmd/zc_enumerator.go
```diff
@@ -202,11 +242,62 @@ func initResourceTraverser(source string, location common.Location, ctx *context
...
+		if err != nil {
+			return nil, err
+		}
+
+		if ctx == nil {
+			return nil, errors.New("a valid context must be supplied to create a S3 traverser")
+		}
+
+		if s3URLParts.BucketName == "" || strings.Contains(s3URLParts.BucketName, "*") {
+			// TODO convert to path style URL
```
JohnRusk: Ditto

cmd/zc_traverser_blob_account.go
```diff
@@ -68,7 +73,8 @@ func (t *blobAccountTraverser) traverse(processor objectProcessor, filters []obj
 			err = containerTraverser.traverse(middlemanProcessor, filters)
 
 			if err != nil {
-				return err
+				LogStdoutAndJobLog(fmt.Sprintf("failed to list blobs in container %s: %s", v.Name, err))
```
JohnRusk: FYI Adele, I logged a followup item in the backlog re this, with ID 5314752.  I don't suggest it should affect this PR.

cmd/zc_traverser_s3_service.go
```diff
@@ -72,7 +78,18 @@ func (t *s3ServiceTraverser) traverse(processor objectProcessor, filters []objec
...
 
 			if err != nil {
-				return err
+				if strings.Contains(err.Error(), "301 response missing Location header") {
+					LogStdoutAndJobLog(fmt.Sprintf("skip enumerating the bucket %q , as it's not in the region specified by source URL", v.Name))
+					continue
+				}
+
+				if strings.Contains(err.Error(), "cannot list objects, The specified bucket does not exist") {
+					LogStdoutAndJobLog(fmt.Sprintf("skip enumerating the bucket %q, as it does not exist (perhaps you have a ghost bucket?)", v.Name))
```
JohnRusk: Is "ghost bucket" a recognised AWS term?  If not, I don't think we can use that term.

cmd/zc_traverser_s3_service.go
```diff
@@ -72,7 +78,18 @@ func (t *s3ServiceTraverser) traverse(processor objectProcessor, filters []objec
 			err = bucketTraverser.traverse(middlemanProcessor, filters)
 
 			if err != nil {
-				return err
+				if strings.Contains(err.Error(), "301 response missing Location header") {
+					LogStdoutAndJobLog(fmt.Sprintf("skip enumerating the bucket %q , as it's not in the region specified by source URL", v.Name))
+					continue
```
JohnRusk: Nice

cmd/zt_copy_s2smigration_test.go
```diff
@@ -264,10 +264,20 @@ func (s *cmdIntegrationSuite) TestS2SCopyFromS3ToBlobWithBucketNameNeedBeResolve
 	rawDstBlobServiceURLWithSAS := scenarioHelper{}.getRawBlobServiceURLWithSAS(c)
 	raw := getDefaultRawCopyInput(rawSrcS3BucketURL.String(), rawDstBlobServiceURLWithSAS.String())
 
-	// bucket should be resolved, and objects should be scheduled for transfer
+	// bucket should not be resolved, and objects should not be scheduled for transfer
 	runCopyAndVerify(c, raw, func(err error) {
 		c.Assert(err, chk.NotNil)
```
JohnRusk: I'm puzzled why we get objects not scheduled for transfer here.  I understand why we get the glcm warning, and its nice that the test checks for it.  But if the bucket name can be resolved (by changing  the -- to -2- as noted in comment above) shouldn't all the transfers actually be scheduled?  I have a hunch I'm missing something... but can't think what!

cmd/copyEnumeratorInit.go
```diff
@@ -42,33 +46,119 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
+	var (
+		err      error
+		isPublic bool
+	)
+	srcCredInfo := common.CredentialInfo{}
+
+	if srcCredInfo, isPublic, err = getCredentialInfoForLocation(ctx, cca.fromTo.From(), cca.source, cca.sourceSAS, true); err != nil {
+		return nil, err
+		// If S2S and source takes OAuthToken as its cred type (OR) source takes anonymous as its cred type, but it's not public and there's no SAS
+	} else if cca.fromTo.From() != common.ELocation.Local() && cca.fromTo.To() != common.ELocation.Local() &&
```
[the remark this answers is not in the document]
adreed-msft: Good point. I was unaware of these options.

cmd/copyEnumeratorInit.go
```diff
@@ -140,6 +241,65 @@ func (cca *cookedCopyCmdArgs) initModularFilters() []objectFilter {
 	return filters
 }
 
+var existingContainers = map[string]bool{}
```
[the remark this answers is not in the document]
adreed-msft: I'm inclined to agree. It's only used by a function linked to the cca, anyway.

cmd/copyEnumeratorInit.go
```diff
@@ -140,6 +241,65 @@ func (cca *cookedCopyCmdArgs) initModularFilters() []objectFilter {
...
+		return err
+	}
+
+	dstPipeline, err := initPipeline(ctx, cca.fromTo.To(), dstCredInfo)
+	if err != nil {
+		return
+	}
+
+	// Because the only use-cases for createDstContainer will be on service-level S2S and service-level download
+	// We only need to create "containers" on local and blob.
```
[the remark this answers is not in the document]
adreed-msft: Yes, Azure Files S2S is a "later" scenario AFAIK

cmd/copyEnumeratorInit.go
```diff
@@ -203,11 +363,24 @@ func (cca *cookedCopyCmdArgs) makeEscapedRelativePath(source bool, dstIsDir bool
...
 		// Save to a directory
-		relativePath = "/" + filepath.Base(cca.source) + relativePath
+		rootDir := filepath.Base(cca.source)
+
+		if cca.fromTo.From() != common.ELocation.Local() {
+			ueRootDir, err := url.PathUnescape(rootDir)
+
+			// Realistically, err should never not be nil here.
+			if err == nil {
+				rootDir = ueRootDir
```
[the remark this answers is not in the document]
adreed-msft: Yes. The pathEncodeRules created at the start of makeEscapedRelativePath do so.

cmd/credentialUtil.go
```diff
@@ -223,8 +258,49 @@ type rawFromToInfo struct {
 	sourceSAS, destinationSAS string // Standalone SAS which might be provided
 }
 
+func getCredentialInfoForLocation(ctx context.Context, location common.Location, source, sourceSAS string, isSource bool) (credInfo common.CredentialInfo, isPublic bool, err error) {
```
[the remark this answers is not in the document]
adreed-msft: Probably a good idea. I felt a conflict of thought while using those parameter names, but couldn't think of better ones-- Good idea on the resource naming though.

cmd/copyEnumeratorInit.go
```diff
@@ -42,33 +46,119 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
+			Destination:        dstRelPath,
+			LastModifiedTime:   object.lastModifiedTime,
+			SourceSize:         object.size,
+			BlobType:           object.blobType,
+			ContentDisposition: object.contentDisposition,
+			ContentLanguage:    object.contentLanguage,
+			ContentEncoding:    object.contentEncoding,
+			CacheControl:       object.cacheControl,
+			ContentType:        object.contentType,
+		}
```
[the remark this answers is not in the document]
adreed-msft: Good catch! I'll write a function like that after I address your other comments.

cmd/s3NameResolver.go
```diff
@@ -85,9 +85,14 @@ func NewS3BucketNameToAzureResourcesResolver(s3BucketNames []string) *S3BucketNa
 // ResolveName returns resolved name for given bucket name.
 func (s3Resolver *S3BucketNameToAzureResourcesResolver) ResolveName(bucketName string) (string, error) {
 	if resolvedName, ok := s3Resolver.bucketNameResolvingMap[bucketName]; !ok {
-		return "", fmt.Errorf("%s: invalid state, cannot find resolved bucket name for %q", s3BucketNameResolveError, bucketName) // This should not happen logically
+		// Resolve the new bucket name, recurse.
```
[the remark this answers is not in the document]
adreed-msft: Good idea again-- this'll help consolidate the code into one place.

cmd/zc_enumerator.go
adreed-msft: Already done.

cmd/zc_enumerator.go
adreed-msft: AWS allows two types of URLs depending on how recently the region was made available.

You have the classic hostname style:
bucket.s3.amazonaws.com

And you have the path style:
region.s3.amazonaws.com/bucket

The tricky bit is: The hostname style doesn't require a region, but the path style does.

adreed-msft: Regions launched since a certain date (I forget exactly when, I think it was sometime in 2018?) can ONLY use the path style.

cmd/zc_traverser_blob_account.go
adreed-msft: Responded on the PBI.

cmd/zc_traverser_s3_service.go
adreed-msft: Trimmed off the portion of the string in parentheses.

cmd/zt_copy_s2smigration_test.go
adreed-msft: Only one bucket is targeted: One with an invalid name. The destination is a service, which means the container will be created before the transfer.

If container creation fails, no transfers have been scheduled.

ste/mgr-JobMgr.go
```diff
@@ -91,6 +91,7 @@ func newJobMgr(concurrency ConcurrencySettings, appLogger common.ILogger, jobID
 		overwritePrompter: newOverwritePrompter(),
 		/*Other fields remain zero-value until this job is scheduled */}
 	jm.reset(appCtx, commandString)
+	jm.logJobsAdminMessages()
```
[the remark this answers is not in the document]
adreed-msft: Fair enough. I'd assumed that we'd only wanted to do the logging BEFORE the job was initialized, but this is a good idea. I'll plonk it in getPerfInfo.

cmd/copyEnumeratorInit.go
```diff
@@ -42,33 +46,119 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
+				})
+				ste.JobsAdmin.LogToJobLog(fmt.Sprintf("failed to initialize destination container %s; the transfer will continue (but be wary it may fail): %s", cName, err))
+			}
+		}
+
+		// If above the service level, we already know the container name and don't need to supply it to makeEscapedRelativePath
+		if srcLevel != ELocationLevel.Service() {
+			object.containerName = ""
+
+			// When copying directly TO a container from a container, don't drop under a sub directory
```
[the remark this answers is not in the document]
adreed-msft: So, containerName/dstContainerName gets appended to the front of the path if present. In all scenarios, this effectively acts as a extra subdirectory, and is intended for account downloads.

cmd/s3NameResolver.go
```diff
@@ -61,49 +61,43 @@ const failToResolveMapValue = "<resolving_failed>"
...
-// S3BucketNameToAzureResourcesResolver works in such pattern:
-// 1. User provided all the bucket names returned in a certain time. (S3 has service limitation, that one S3 account can only have 100 buckets, except opening service ticket to increase the number.)
-// 2. S3BucketNameToAzureResourcesResolver resolves the names with one logic pass during creating the resolver instance.
-// 3. User can get resolved name later with ResolveName.
-// As S3BucketNameToAzureResourcesResolver need to detect naming collision, the resolver doesn't accept adding new bucket name except the initial s3BucketNames,
-// considering there could be future valid name that is not predictable during previous naming resolving.
+// Users can provide bucket names upfront and on-demand via ResolveName.
+// Previously resolved names will be returned outright by ResolveName.
 func NewS3BucketNameToAzureResourcesResolver(s3BucketNames []string) *S3BucketNameToAzureResourcesResolver {
 	s3Resolver := S3BucketNameToAzureResourcesResolver{
```
JohnRusk: Reading this prompted some more questions in my mind, about the safety of appending integers to ensure uniqueness.  Have sent you a long email with the details.  Pls leave this comment here open until the email thread is resolved.

zezha-msft:
@adreed-msft S2S copy with list-of-files flag needs to be updated to use the new format, instead of JSON:
```
failed to parse user input due to error: error invalid character 'e' in literal true (expecting 'r') unmarshalling the contents of /Users/zed/Downloads/drop 5/list.txt file passed with the list-of-file flag
```

adreed-msft:
Ah, good catch

cmd/s3NameResolver.go
adreed-msft: Thread resolved: We decided not to change the functionality at this moment.

cmd/copyEnumeratorInit.go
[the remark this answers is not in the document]
adreed-msft: Function is made, all variables are appropriately carried over.

cmd/copyEnumeratorInit.go
```diff
@@ -199,19 +199,15 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
-		}
+		transfer := common.NewCopyTransfer(
+			srcRelPath, dstRelPath,
+			object.lastModifiedTime,
+			object.size,
+			object.contentType, object.contentEncoding, object.contentDisposition, object.contentLanguage, object.cacheControl,
+			object.md5,
+			object.Metadata,
+			object.blobType,
+			object.blobAccessTier)
```
JohnRusk: Not consistent with the if stantement just below. Do you want some kind of none/null value here?

cmd/copyEnumeratorInit.go
```diff
@@ -102,35 +97,110 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
+			}
+		} else if cca.fromTo.From().IsRemote() { // if the destination has implicit container names
+			if acctTraverser, ok := traverser.(accountTraverser); ok && dstLevel == ELocationLevel.Service() {
+				containers, err := acctTraverser.listContainers()
+
+				if err != nil {
+					return nil, fmt.Errorf("failed to list containers: %s", err)
+				}
+
+				containerResolver = NewS3BucketNameToAzureResourcesResolver(containers)
```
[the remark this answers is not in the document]
adreed-msft: Yes.

cmd/copyEnumeratorInit.go
```diff
@@ -102,35 +97,110 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
+			}
+		}
+	}
+
+	filters := cca.initModularFilters()
+	processor := func(object storedObject) error {
+		// Start by resolving the name and creating the container
+		if object.containerName != "" {
+			// set up the destination container name.
+			cName := dstContainerName
```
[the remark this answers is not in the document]
adreed-msft: If a destination container name is explicitly specified, do not resolve it, do not add it to the path creation.

If we _did_  use it here, then we'd basically get
https://account.blob.core.windows.net/container/container/dstfile
instead of
https://account.blob.core.windows.net/container/dstfile

cmd/copyEnumeratorInit.go
adreed-msft: I meant to put none there, Yes. I'm correcting that now.

adreed-msft:
Hallelujah! The refactor is finally done.

PR #566, opened 2019-08-27
Create a log file even when there's nothing to do

adreed-msft:


zezha-msft:
> 		}

why is this needed? it looks like we are already handing the error gracefully a few lines down. Additional I/Os should be avoided if possible.

---
Refers to: cmd/copyEnumeratorInit.go:276 in e31e17f. [](commit_id = e31e17f1cdbe1ea59c6dab7f3357eee05162bc91, deletion_comment = False)

cmd/zc_processor.go
```diff
@@ -105,22 +109,15 @@ func (s *copyTransferProcessor) dispatchFinalPart() (copyJobInitiated bool, err
 	return true, nil
 }
 
-func (s *copyTransferProcessor) sendPartToSte() error {
+// only test the response on the final dispatch as previous ones may for some reason not have transfers.
```
JohnRusk: Do we know what that reason is?

cc @zezha-msft, I thought we avoided having empty Parts?

cmd/copyEnumeratorHelper.go
```diff
@@ -65,6 +65,13 @@ func dispatchFinalPart(e *common.CopyJobPartOrderRequest, cca *cookedCopyCmdArgs
 	Rpc(common.ERpcCmd.CopyJobPartOrder(), (*common.CopyJobPartOrderRequest)(e), &resp)
 
 	if !resp.JobStarted {
+		// Output the log location and such
+		glcm.Init(common.GetStandardInitOutputBuilder(cca.jobID.String(), fmt.Sprintf("%s%s%s.log", azcopyLogPathFolder, common.OS_PATH_SEPARATOR, cca.jobID)))
+
+		if resp.ErrorMsg == common.ECopyJobPartOrderErrorType.NoTransfersScheduledErr() {
+			return fmt.Errorf("no transfers were scheduled")
```
JohnRusk: Might be nice to add something to the output to explain it a bit more.  E.g. "because no files matched the criteria specified on the command line"

cmd/removeEnumerator.go
```diff
@@ -73,6 +73,12 @@ func newRemoveEnumerator(cca *cookedCopyCmdArgs) (enumerator *copyEnumerator, er
 	finalize := func() error {
 		jobInitiated, err := transferScheduler.dispatchFinalPart()
 		if err != nil {
+			if strings.Contains(err.Error(), "scheduled") {
```
JohnRusk: This doesn't seem super-robust - i.e. when checking error message for string content, it will break if someone one day changes the error message text. Since the error will be coming from our own code, how about we define something like
```
var NothingScheduledError = errors.New("blah blah scheduled blah") // put the real text in here
```

Then the code that raises the error can return NothingScheduledError, and this code here can become

```
if err == NothingScheduledError
```

I believe that's the normal Golang pattern for this kind of thing. I'm not quite sure whether or how it should interact with the `ECopyJobPartOrderErrorType` stuff defined elsewhere - it seems like they are both talking about the same thing, so if that's the case, maybe this check here should check against the string `ECopyJobPartOrderErrorType.NoTransfersScheduledErr` instead of the NothingScheduledError I proposed above.

cmd/sync.go
```diff
@@ -253,7 +253,7 @@ func (cca *cookedSyncCmdArgs) scanningComplete() bool {
 // if blocking is specified to false, then another goroutine spawns and wait out the job
 func (cca *cookedSyncCmdArgs) waitUntilJobCompletion(blocking bool) {
 	// print initial message to indicate that the job is starting
-	glcm.Init(common.GetStandardInitOutputBuilder(cca.jobID.String(), fmt.Sprintf("%s/%s.log", azcopyLogPathFolder, cca.jobID)))
+	glcm.Init(common.GetStandardInitOutputBuilder(cca.jobID.String(), fmt.Sprintf("%s%s%s.log", azcopyLogPathFolder, common.OS_PATH_SEPARATOR, cca.jobID)))
```
JohnRusk: Nice that this long-standing path formatting error is finally fixed!

cmd/zc_processor.go
adreed-msft: This is just me trying to work around test failures right now.

I can't seem to re-create the test failure locally at all, so I don't even have remotely a clue of what could be going on. I'm basically diagnosing a black box.

JohnRusk: Maybe just add to the comment something saying "for diagnosing root cause of test failures".  Otherwise the comment will mystify future programmers ;-)

adreed-msft: Good point.

adreed-msft:
OK so some explaining with that last commit:

For some reason, Git didn't mind merging this branch onto a branch based on master. But if I'd just rebased, every file in azcopy would have merge conflicts.

adreed-msft:
Will merge once CI passes

adreed-msft:
Squash and merge

PR #570, opened 2019-08-30
Optimized S2S copy for sparse page blob

zezha-msft:


ste/sender_pageBlobFromURL_test.go
```diff
@@ -0,0 +1,61 @@
+// Copyright © 2017 Microsoft <wastore@microsoft.com>
```
zezha-msft: remove 2017

ste/sender-pageBlobFromURL.go
```diff
@@ -109,3 +126,34 @@ func (c *urlToPageBlobCopier) GetDestinationLength() (int64, error) {
...
+	// note that the page list is ordered in increasing order (in terms of position)
+	for _, srcRange := range c.srcPageList.PageRange {
+		if givenRange.End < srcRange.Start {
+			// case 1: due to the nature of the list (it's sorted), if we've reached such a srcRange
+			// we've checked all the appropriate srcRange already and haven't found any overlapping srcRange
+			// given range:		|   |
+			// source range:			|   |
+			return false
+		} else if srcRange.End < givenRange.Start {
+			// case 2: the givenRange comes after srcRange, continue checking
```
JohnRusk: Wow, this is elegant!

ste/sender-pageBlobFromURL.go
```diff
@@ -33,16 +33,32 @@ import (
...
 	if blobSrcInfoProvider, ok := srcInfoProvider.(IBlobSourceInfoProvider); ok {
 		if blobSrcInfoProvider.BlobType() == azblob.BlobPageBlob {
+			// if the source is page blob, preserve source's blob tier.
 			destBlobTier = blobSrcInfoProvider.BlobTier()
+
+			// also get the page ranges so that we can skip the empty parts at the expense of one HTTP request
+			srcPageBlobURL := azblob.NewPageBlobURL(*srcURL, p)
+			ctx := context.WithValue(jptm.Context(), ServiceAPIVersionOverride, azblob.ServiceVersion)
+
+			// ignore the error, since according to the REST API documentation:
```
JohnRusk: I really want this comment to be true (because its convenient) but I'm not sure that it actually is.  Why? Because of what I read here:
https://blogs.msdn.microsoft.com/windowsazurestorage/2012/03/26/getting-the-page-ranges-of-a-large-page-blob-in-segments/

I'm imaging, say, an 8 TB page blob that's been used as a VM disk.  It will have had many writes to it, but what if the disk was only 1/4 full.  So there were heaps of writes to the first 2 TB, but the remaining 6 TB is still all zeroes.  Maybe, those writes to the first 2 TB would be enough to cause us a timeout, leading us to incorrectly assume that the _whole blob_ is not sparse.

Maybe we could use the approach in the URL above, using a fairly coarse-grained approach (e.g. 1 TB at a time). That way, even if we error out on one TB, we won't make multi-TBs worth of wrong assumptions about the rest of the blob.

On one hand, given time constraints and deadlines, I'm not sure we should do this fix now. But on the other hand, if we release it, say its fixed, and folks are still getting TB's more used size than they expect, they won't be very pleased and it won't really be fixed for them.  Another option is, perhaps, to defer the whole PR until we have time to do the chunking, and another is to release it as-is, with the changelog saying that its a partial fix, with a fuller one to follow.

ste/sender-pageBlobFromURL.go
```diff
@@ -33,16 +33,32 @@ import (
...
 	destBlobTier := azblob.AccessTierNone
-	// If the source is page blob, preserve source's blob tier.
+	var srcPageList *azblob.PageList
 	if blobSrcInfoProvider, ok := srcInfoProvider.(IBlobSourceInfoProvider); ok {
 		if blobSrcInfoProvider.BlobType() == azblob.BlobPageBlob {
+			// if the source is page blob, preserve source's blob tier.
 			destBlobTier = blobSrcInfoProvider.BlobTier()
+
+			// also get the page ranges so that we can skip the empty parts at the expense of one HTTP request
+			srcPageBlobURL := azblob.NewPageBlobURL(*srcURL, p)
```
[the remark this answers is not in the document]
zezha-msft: Oh shoot! That's a great catch. Thanks.

ste/sender-pageBlobFromURL.go
zezha-msft: That's some great insights! The article is from 7 years ago though, so I'll reach out to Michael Roberson to confirm whether the contents are still relevant and whether we should be mindful of the issue. 

We could also do tests to measure the current behavior, but that may be affected by how good of a day the service is having. Even if we get good results in our test, I would feel hesitant to rely on it as fact, so it's better to confirm with the service folks anyways.

I think in general it's good to ship incremental improvements. As long as we write a reasonable change log entry, I'm ok with releasing the current solution, and follow up with improvements if necessary. I see it as achieving a balance between saving cost vs performance vs complexity.

JohnRusk: Agreed.

go.mod
```diff
@@ -29,3 +29,5 @@ require (
 	gopkg.in/ini.v1 v1.42.0 // indirect
 	gopkg.in/yaml.v2 v2.2.2 // indirect
 )
+
+go 1.13
```
JohnRusk: I thought we were staying on 1.12.  Are we not? (I don't have any firm view either way. Just asking in case this commit was in error)

ste/downloader-blob.go
```diff
@@ -110,3 +132,9 @@ func (bd *blobDownloader) GenerateDownloadFunc(jptm IJobPartTransferMgr, srcPipe
 		}
 	})
 }
+
+type dummyReader struct{}
+
+func (dummyReader) Read(p []byte) (n int, err error) {
+	return len(p), nil
+}
```
JohnRusk: Ooooh.  I think we can't do this quite this simply. I think you have to zero out all the bytes in p.  Why?  Because we read into pooled byte arrays, so there might be content in there from a previous usage of the array! 

(another possible solution would be to tweak multiSliceBytePooler so that it always zeros out stuff before returning it. Maybe that's safer, at the cost of an little extra CPU activity.  As I type this, I think that's probably the best idea. If we only fix it here, then one day we may get the same mistake creeping in somewhere else in the code.  And given that the cost of fixing it "at the soucre", in multiSliceBytePooler is just some CPU cycles, I think that's probably OK, because we have not had any reports from users, at all, that indicate that CPU is a constraint on perf . What do you think?)

ste/sender-pageBlobFromURL.go
```diff
@@ -33,16 +33,26 @@ import (
...
 
 	destBlobTier := azblob.AccessTierNone
-	// If the source is page blob, preserve source's blob tier.
+	var pageRangeOptimizer *pageRangeOptimizer
 	if blobSrcInfoProvider, ok := srcInfoProvider.(IBlobSourceInfoProvider); ok {
 		if blobSrcInfoProvider.BlobType() == azblob.BlobPageBlob {
+			// if the source is page blob, preserve source's blob tier.
 			destBlobTier = blobSrcInfoProvider.BlobTier()
+
+			// also get the page ranges so that we can skip the empty parts at the expense of one HTTP request
```
JohnRusk: Minor: Can you remove this comment, because they are not fetched here.

testSuite/scripts/test_upload_page_blob.py
```diff
@@ -39,44 +39,76 @@ def test_page_blob_upload_1mb_with_oauth(self):
     # test_page_range_for_complete_sparse_file verifies the number of Page ranges for
     # complete empty file i.e each character is Null character.
     def test_page_range_for_complete_sparse_file(self):
+        # step 1: uploading a sparse file should be optimized
```
JohnRusk: Do we have a similar test for downloading?  Can we add a backlog item to create such a test next week?

ste/downloader-blob.go
```diff
@@ -59,6 +68,19 @@ func (bd *blobDownloader) Epilogue() {
 func (bd *blobDownloader) GenerateDownloadFunc(jptm IJobPartTransferMgr, srcPipeline pipeline.Pipeline, destWriter common.ChunkedFileWriter, id common.ChunkID, length int64, pacer pacer) chunkFunc {
 	return createDownloadChunkFunc(jptm, id, func() {
 
+		// If the range does not contain any data, write out empty data to disk without performing download
+		if bd.pageRangeOptimizer != nil && !bd.pageRangeOptimizer.doesRangeContainData(
+			azblob.PageRange{Start: id.OffsetInFile(), End: id.OffsetInFile() + length - 1}) {
+
+			// TODO confirm this is the right wait reason
```
JohnRusk: Yes, Body is fine. Or you can probably just not set it. Because chunked file writer will immediately change it anyway.

go.mod
zezha-msft: Sorry, go mod keeps adding it. I've read through the 1.13 release notes, and they look mostly fine. I still think we should stick to 1.12 for now, and upgrade to 1.13 in the next release.

The build seems to work fine, with Go version 1.12 and such a go.mod file with "go.13".

ste/downloader-blob.go
zezha-msft: Doh!! My bad, this is such a good catch, it would have been a data-corrupting bug.

We've actually heard about high CPU usage concerns in the past. I think we should stick to fixing it just here and not in multiSliceBytePooler, as it's not clear what'll be the perf impact. Thoughts? @JohnRusk

ste/downloader-blob.go
zezha-msft: ok

ste/downloader-blob.go
JohnRusk: Your question prompted me to do a bit of a search. It turns out that a zero-ing-out loop, written in the right form in Go, will be automatically turned into a call to memclr, which is an optimized Go runtime routine written in assembler. I.e. write a simple loop, complier turns it into a really fast memory clear operation.

Here's a description of the loop format (which is just an obvious range loop): https://github.com/golang/go/commit/f03c9202c43e0abb130669852082117ca50aa9b1

It's pretty quick - e.g. for a slice of 256 elements, it shaves 96% off the execution time of an un-optimized loop. So I think its fast enough for us to just do it.

One note/suggestion: I think what when we do this in multi-slice byte pooler, we should get the slice we are going to return, re-slice it so that its len = its capacity, use the loop to zero it all out, and only then re-slice it to the length that the user asked for. (That last bit is already in the code, but I'm suggesting we zero out to the full cap, not just to the requested len)

JohnRusk: Sorry about the slow reply @zezha-msft.  

Oh, and PS, my reason for being in favour of doing this in mulitSliceBytePooler is that, if the bug can nearly slip by us both here, it can probably slip by someone else in the future.

testSuite/scripts/test_upload_page_blob.py
zezha-msft: I'll add it now

PR #571, opened 2019-08-30
Infer --no-guess-mime-type

adreed-msft:


adreed-msft:
Changing this to dev branch since it doesn't actually rely upon the s2s refactor changes, then squashing and merging.

PR #574, opened 2019-09-03
Enable long file paths by consolidating paths to determined separators

adreed-msft:
Rather than cutting local file paths OUT of the consolidation, we include them but instead, consolidate the separators depending on what the path is actually using, not just what we use.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -65,14 +65,11 @@ type rawCopyCmdArgs struct {
 	fromTo string
 	//blobUrlForRedirection string
 
-	// TODO remove after refactoring
-	legacyInclude string
-	legacyExclude string
 	// new include/exclude only apply to file names
 	// implemented for remove (and sync) only
 	include               string
 	exclude               string
-	includePath           string
+	includePath           string // NOTE: This gets handled like list-of-files! It may LOOK like a bug, but it is not.
 	excludePath           string
 	includeFileAttributes string
 	excludeFileAttributes string
@@ -82,7 +79,6 @@ type rawCopyCmdArgs struct {
 	recursive         bool
 	stripTopDir       bool
 	followSymlinks    bool
-	withSnapshots     bool
 	autoDecompress    bool
 	// forceWrite flag is used to define the User behavior
 	// to overwrite the existing blobs or not.
@@ -105,9 +101,7 @@ type rawCopyCmdArgs struct {
 	blobType      string
 	blockBlobTier string
 	pageBlobTier  string
-	background    bool
 	output        string // TODO: Is this unused now? replaced with param at root level?
-	acl           string
 	logVerbosity  string
 	// list of blobTypes to exclude while enumerating the transfer
 	excludeBlobType string
@@ -191,7 +185,6 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 	cooked.stripTopDir = raw.stripTopDir
 	cooked.recursive = raw.recursive
 	cooked.followSymlinks = raw.followSymlinks
-	cooked.withSnapshots = raw.withSnapshots
 	err = cooked.forceWrite.Parse(raw.forceWrite)
 	if err != nil {
 		return cooked, err
@@ -260,13 +253,25 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 		}
 	}
 
+	// Prepare UTF-8 byte order marker
+	utf8BOM := string([]byte{0xEF, 0xBB, 0xBF})
+
 	go func() {
 		defer close(listChan)
 
 		if f != nil {
 			scanner := bufio.NewScanner(f)
+			checkBOM := false
 			for scanner.Scan() {
 				v := scanner.Text()
+
+				// Yes, the UTF-8 BOM has valid characters (butwhytho*10)
+				// Check it on the first line and remove it if necessary.
+				if !checkBOM {
+					v = strings.TrimPrefix(v, utf8BOM)
+					checkBOM = true
+				}
+
 				// empty strings should be ignored, otherwise the source root itself is selected
 				if len(v) > 0 {
 					listChan <- v
@@ -295,51 +300,6 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 		cooked.listOfFilesChannel = listChan
 	}
 
-	// TODO: When further in the refactor, just check this against a map
-	if cooked.fromTo.From() != common.ELocation.Local() && cooked.fromTo.To() != common.ELocation.Local() {
-		// initialize the include map which contains the list of files to be included
-		// parse the string passed in include flag
-		// more than one file are expected to be separated by ';'
-		cooked.legacyInclude = make(map[string]int)
-		if len(raw.legacyInclude) > 0 {
-			files := strings.Split(raw.legacyInclude, ";")
-			for index := range files {
-				// If split of the include string leads to an empty string
-				// not include that string
-				if len(files[index]) == 0 {
-					continue
-				}
-				// replace the OS path separator in includePath string with AZCOPY_PATH_SEPARATOR
-				// this replacement is done to handle the windows file paths where path separator "\\"
-				includePath := strings.Replace(files[index], common.OS_PATH_SEPARATOR, common.AZCOPY_PATH_SEPARATOR_STRING, -1)
-				cooked.legacyInclude[includePath] = index
-			}
-		}
-
-		// initialize the exclude map which contains the list of files to be excluded
-		// parse the string passed in exclude flag
-		// more than one file are expected to be separated by ';'
-		cooked.legacyExclude = make(map[string]int)
-		if len(raw.legacyExclude) > 0 {
-			files := strings.Split(raw.legacyExclude, ";")
-			for index := range files {
-				// If split of the include string leads to an empty string
-				// not include that string
-				if len(files[index]) == 0 {
-					continue
-				}
-				// replace the OS path separator in excludePath string with AZCOPY_PATH_SEPARATOR
-				// this replacement is done to handle the windows file paths where path separator "\\"
-				excludePath := strings.Replace(files[index], common.OS_PATH_SEPARATOR, common.AZCOPY_PATH_SEPARATOR_STRING, -1)
-				cooked.legacyExclude[excludePath] = index
-			}
-		}
-	} else {
-		// I do not like this implementation whatsoever, but just for now, we need to circumvent the legacy include and excludes.
-		raw.include = raw.legacyInclude
-		raw.exclude = raw.legacyExclude
-	}
-
 	cooked.metadata = raw.metadata
 	cooked.contentType = raw.contentType
 	cooked.contentEncoding = raw.contentEncoding
@@ -361,9 +321,6 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 
 	cooked.CheckLength = raw.CheckLength
 
-	cooked.background = raw.background
-	cooked.acl = raw.acl
-
 	// if redirection is triggered, avoid printing any output
 	if cooked.isRedirection() {
 		glcm.SetOutputFormat(common.EOutputFormat.None())
@@ -376,6 +333,25 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 	case common.EFromTo.LocalBlobFS():
 		if cooked.blobType != common.EBlobType.Detect() {
 			return cooked, fmt.Errorf("blob-type is not supported on ADLS Gen 2")
+		}
+		if cooked.preserveLastModifiedTime {
+			return cooked, fmt.Errorf("preserve-last-modified-time is not supported while uploading")
+		}
+		if cooked.blockBlobTier != common.EBlockBlobTier.None() ||
+			cooked.pageBlobTier != common.EPageBlobTier.None() {
+			return cooked, fmt.Errorf("blob-tier is not supported while uploading to ADLS Gen 2")
+		}
+		if cooked.s2sPreserveProperties {
+			return cooked, fmt.Errorf("s2s-preserve-properties is not supported while uploading")
+		}
+		if cooked.s2sPreserveAccessTier {
+			return cooked, fmt.Errorf("s2s-preserve-access-tier is not supported while uploading")
+		}
+		if cooked.s2sInvalidMetadataHandleOption != common.DefaultInvalidMetadataHandleOption {
+			return cooked, fmt.Errorf("s2s-handle-invalid-metadata is not supported while uploading")
+		}
+		if cooked.s2sSourceChangeValidation {
+			return cooked, fmt.Errorf("s2s-detect-source-changed is not supported while uploading")
 		}
 	case common.EFromTo.LocalBlob():
 		if cooked.preserveLastModifiedTime {
@@ -417,7 +393,8 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 			return cooked, fmt.Errorf("blob-type is not supported on Azure File")
 		}
 	case common.EFromTo.BlobLocal(),
-		common.EFromTo.FileLocal():
+		common.EFromTo.FileLocal(),
+		common.EFromTo.BlobFSLocal():
 		if cooked.followSymlinks {
 			return cooked, fmt.Errorf("follow-symlinks flag is not supported while downloading")
 		}
@@ -551,25 +528,20 @@ type cookedCopyCmdArgs struct {
 	destinationSAS string
 	fromTo         common.FromTo
 
-	// TODO remove after refactoring
-	legacyInclude map[string]int
-	legacyExclude map[string]int
 	// new include/exclude only apply to file names
 	// implemented for remove (and sync) only
+	// includePathPatterns are handled like a list-of-files. Do not panic. This is not a bug that it is not present here.
 	includePatterns       []string
-	includePathPatterns   []string
 	excludePatterns       []string
 	excludePathPatterns   []string
 	includeFileAttributes []string
 	excludeFileAttributes []string
 
 	// filters from flags
-	listOfFilesToCopy  []string
-	listOfFilesChannel chan string // We make it a pointer so we can check if it exists w/o reading from it & tack things onto it if necessary.
+	listOfFilesChannel chan string // Channels are nullable.
 	recursive          bool
 	stripTopDir        bool
 	followSymlinks     bool
-	withSnapshots      bool
 	forceWrite         common.OverwriteOption
 	autoDecompress     bool
 
@@ -591,8 +563,6 @@ type cookedCopyCmdArgs struct {
 	putMd5                   bool
 	md5ValidationOption      common.HashValidationOption
 	CheckLength              bool
-	background               bool
-	acl                      string
 	logVerbosity             common.LogLevel
 	// commandString hold the user given command which is logged to the Job log file
 	commandString string
@@ -791,8 +761,6 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 		AutoDecompress:  cca.autoDecompress,
 		Priority:        common.EJobPriority.Normal(),
 		LogLevel:        cca.logVerbosity,
-		Include:         cca.legacyInclude,
-		Exclude:         cca.legacyExclude,
 		ExcludeBlobType: cca.excludeBlobType,
 		BlobAttributes: common.BlobTransferAttributes{
 			BlobType:                 cca.blobType,
@@ -1246,8 +1214,7 @@ func init() {
 	// filters change which files get transferred
 	cpCmd.PersistentFlags().BoolVar(&raw.stripTopDir, "strip-top-dir", false, "strip the source's root folder from the destination path, akin to \"cp dir/*\". E.g. sourcedir/subdir1/file1 copies to subdir1/file1 on destination; whereas without --strip-top-dir, it copies to sourcedir/subdir1/file1. May be used with and without --recursive. Without --recursive, just copies files under the folder but does not recurse into sub-directories")
 	cpCmd.PersistentFlags().BoolVar(&raw.followSymlinks, "follow-symlinks", false, "follow symbolic links when uploading from local file system.")
-	cpCmd.PersistentFlags().BoolVar(&raw.withSnapshots, "with-snapshots", false, "include the snapshots. Only valid when the source is blobs.")
-	cpCmd.PersistentFlags().StringVar(&raw.legacyInclude, "include-pattern", "", "only include these files when copying. "+
+	cpCmd.PersistentFlags().StringVar(&raw.include, "include-pattern", "", "only include these files when copying. "+
 		"Support use of *. Files should be separated with ';'.")
 	cpCmd.PersistentFlags().StringVar(&raw.includePath, "include-path", "", "only include these paths when copying. "+
 		"Does not support using wildcards. Checks relative path prefix. ex. myFolder;myFolder/subDirName/file.pdf")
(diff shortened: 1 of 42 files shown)


common/LongPathHandler.go
```diff
@@ -0,0 +1,46 @@
...
+//because it fails to support long UNC paths. As a result, we opt to wrap things such as filepath.Glob()
+//to safely use them with long UNC paths.
+
+//ToExtendedPath converts short paths to an extended path.
+func ToExtendedPath(short string) string {
+	short, err := filepath.Abs(short)
+	PanicIfErr(err) //TODO: Handle errors better?
+
+	if os.PathSeparator == '\\' {
+		if strings.HasPrefix(short, EXTENDED_PATH_PREFIX) {
```
zezha-msft: >{ [](start = 52, length = 1)

please add comments (examples would be great)

common/LongPathHandler.go
```diff
@@ -0,0 +1,46 @@
...
+//Furthermore, we don't use the built-in long path support in the Go SDK commit 231aa9d6d7
+//because it fails to support long UNC paths. As a result, we opt to wrap things such as filepath.Glob()
+//to safely use them with long UNC paths.
+
+//ToExtendedPath converts short paths to an extended path.
+func ToExtendedPath(short string) string {
+	short, err := filepath.Abs(short)
+	PanicIfErr(err) //TODO: Handle errors better?
+
+	if os.PathSeparator == '\\' {
```
zezha-msft: > os.PathSeparator [](start = 3, length = 17)

perhaps test on WSL, to see how it behaves

common/LongPathHandler.go
```diff
@@ -0,0 +1,46 @@
+package common
```
zezha-msft: >pa [](start = 0, length = 2)

add copyright header

common/LongPathHandler.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	return short
+}
+
+//ToShortPath converts an extended path to a short path.
+func ToShortPath(long string) string {
+	if os.PathSeparator == '\\' {
+		if strings.HasPrefix(long, EXTENDED_UNC_PATH_PREFIX) {
+			return `\` + long[7:] //Return what we stole from it
```
zezha-msft: >7 [](start = 21, length = 1)

avoid magic numbers, as they are error-prone

cmd/zc_traverser_local.go
```diff
@@ -316,18 +315,14 @@ func newLocalTraverser(fullPath string, recursive bool, followSymlinks bool, inc
...
+	// detect if we are targeting a drive (ex: either C:\ or C:)
+	// note that on windows there must be a slash in order to target the root drive properly
+	// otherwise we'd point to the path from where AzCopy is running (if AzCopy is running from the same drive)
+	if len(localPath) == 3 && (strings.HasSuffix(localPath, `:\`) || strings.HasSuffix(localPath, ":/")) ||
 		len(localPath) == 2 && strings.HasSuffix(localPath, ":") {
-		// detect if we are targeting a drive (ex: either C:\ or C:)
-		// note that on windows there must be a slash in order to target the root drive properly
-		// otherwise we'd point to the path from where AzCopy is running (if AzCopy is running from the same drive)
-		normalizedPath += common.AZCOPY_PATH_SEPARATOR_STRING
+		normalizedPath += common.OS_PATH_SEPARATOR
```
zezha-msft: >normalizedPath += common.OS_PATH_SEPARATOR [](start = 2, length = 42)

so for C:\, we'd append another '\'?

cmd/sync.go
```diff
@@ -108,6 +109,25 @@ func (raw *rawSyncCmdArgs) cook() (cookedSyncCmdArgs, error) {
 		return cooked, fmt.Errorf("source '%s' / destination '%s' combination '%s' not supported for sync command ", raw.src, raw.dst, cooked.fromTo)
 	}
 
+	// Do this check seperately so we don't end up with a bunch of code duplication when new src/dsts are added
+	if cooked.fromTo.From() == common.ELocation.Local() {
+		result, err := filepath.Abs(cooked.source)
+
+		if err != nil {
+			return cooked, fmt.Errorf("failed to resolve absolute path from %s: %s", result, err)
+		}
```
zezha-msft: ToExtendedPath already does the filepath.Abs call

common/extensions.go
```diff
@@ -122,11 +123,23 @@ func (bs ByteSliceExtension) RemoveBOM() []byte {
 
 /////////////////////////////////////////////////////////////////////////////////////////////////
 
+func DeterminePathSeparator(path string) string {
+	// Just use forward-slash everywhere that isn't windows.
+	if runtime.GOOS == "windows" && strings.Contains(path, `\`) {
```
zezha-msft: >runtime.GOOS == "windows" [](start = 4, length = 25)

How does this differ from `os.PathSeparator == '\\'`? I see both checks being used

common/LongPathHandler.go
adreed-msft: Done

common/LongPathHandler.go
adreed-msft: I'm on WSL2 but my home machine cooperated on GA WSL.

adreed-msft: I'm changing these two to runtime.GOOS == "windows" just to be stricter.

common/LongPathHandler.go
adreed-msft: Done

common/LongPathHandler.go
adreed-msft: Will do

cmd/zc_traverser_local.go
adreed-msft: (note: edited your comment as \\ was formatted incorrectly, made readability hard.)

But yes, we would.

cmd/sync.go
adreed-msft: True. I don't know entirely why I was worrying about that here.

common/extensions.go
adreed-msft: os.PathSeparator == '\\' is actually somewhat less strict than runtime.GOOS == "windows".

I don't _THINK_ there are any OSes other than windows that use a \ separator, and definitely none that we support.

For our use case, either would work. I've already changed the long path handler to use runtime.GOOS instead since I would rather be safer than sorry.

cmd/sync.go
```diff
@@ -111,6 +110,13 @@ func (raw *rawSyncCmdArgs) cook() (cookedSyncCmdArgs, error) {
 		return cooked, fmt.Errorf("source '%s' / destination '%s' combination '%s' not supported for sync command ", raw.src, raw.dst, cooked.fromTo)
 	}
 
+	// Do this check seperately so we don't end up with a bunch of code duplication when new src/dsts are added
+	if cooked.fromTo.From() == common.ELocation.Local() {
```
JohnRusk: I suspect that the calls to cleanLocalPath, which are in the if/else ladder just above this point, should be taken out of there, and moved to here, so that these lines look like

```
cooked.source = cleanLocalPath(common.ToExtendedPath(cooked.source))
```
Why?  Because 
(a) That makes is consistent with copy, where the conversion to extended path is done befoere the clean and 
(b) That allows the panic-based check inside DeterminePathSeparator, which I've suggested in another comment. (That panic-based check is not usable if cleanLocalPath comes first.  But when ToExtendedPath comes first, it will do an Abs, wihch normalizes the delimiters to the OS default).

Does that sound reasonable?  Any downside to the above?

JohnRusk:
@Adele After submitting my review I also noticed that this branch breaks `--include-path` and `--exclude-path`

common/extensions.go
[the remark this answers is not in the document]
adreed-msft: I thought this'd cause errors with a mixed path but it seems to consolidate fine. Good thought to have.

cmd/zc_traverser_local.go
```diff
@@ -298,12 +298,11 @@ func (t *localTraverser) traverse(processor objectProcessor, filters []objectFil
 	return
 }
 
-func replacePathSeparators(path string) string {
-	if os.PathSeparator != common.AZCOPY_PATH_SEPARATOR_CHAR {
-		return strings.Replace(path, string(os.PathSeparator), common.AZCOPY_PATH_SEPARATOR_STRING, -1)
-	} else {
-		return path
-	}
+// Replace azcopy path separators (/) with the OS path separator
```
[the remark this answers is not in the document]
adreed-msft: Uhhhhh.... No?

strings.ReplaceAll(s, old, new)

You may've mis-remembered the function signature.

adreed-msft:
Good catch. I'll add tests for those in this branch and fix them up. This may well break list-of-files as well, so I should probably

- Consolidate the path in the list traverser
- Consolidate the path in the exclude path filter

as this may well break list-of-files as well.

adreed-msft:
> * Consolidate the path in the list traverser for a local source

This is already done, as it turns out. The actual problem was in initResourceTraverser trying to cut the query off of a URL. Perhaps we should write a battery of tests for this in the future just to make sure all of the string manipulation that happens in there is OK.

cmd/sync.go
adreed-msft: Well, I agree that it makes the code more consistent, and that it drops the code into one place. I think it'd actually be best to write a function called 'cleanLocation' or something of that variety and have it return the source path and SAS token. Then we'd just rail-road it down through that function and only have two calls to it, rather than 20 lines to clean up the source and destination.

@zezha-msft since you've done most of the work on the Sync function, what do you think of that solution?

cmd/zc_enumerator.go
```diff
@@ -157,14 +157,20 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 
 	// Feed list of files channel into new list traverser, separate SAS.
 	if listofFilesChannel != nil {
-		splitsrc := strings.Split(source, "?")
 		sas := ""
+		if location.IsRemote() {
+			srcURL, err := url.Parse(source)
```
JohnRusk: I noticed that this is different from how we remove the SAS's in copy.go (about lines 800 to 950).  There, we call azblob.NewBlobURLParts (or the corresponding Files or BlobFS functions).  Internally, they do something subtly different from what's here: the removed only those query parameters that actually are part of a SAS, leaving all others in place.

I'm not sure how often folks will give us additional query params.  Probably not very often.  I can imagine someone trying to use the include= parameter when listing blobs, to get deleted blobs, snapshots or whatever, and I can imagine in a single-blob scenario someone might use snaphshot=...  Although I'm not sure any of those would actually work, because I doubt we test them.

@zezha-msft What do you think?  Is it important that we always remove SAS's with the strictly correct method that does not touch any other unrelated parts of the query string that may exist?  Or, should we perhaps leave this code here as is, and in fact return errors back in copy.go if there are any additional query parameters left after we remove the SAS?  That would at least prevent users from trying stuff where the results are undefined because we don't test them.  I suspect the latter option would be quickest and safe to implement.

cmd/zt_copy_blob_upload_test.go
```diff
@@ -29,6 +29,84 @@ import (
...
+		"filea",
+		"fileb",
+		"filec",
+		"sub/filea",
+		"sub/fileb",
+		"sub/filec",
+		"sub/subsub/filea",
+		"sub/subsub/fileb",
+		"sub/subsub/filec",
+	}
```
JohnRusk: should probably also have a case here `sub/somethingelse/subsub/filex`  (which I presume would not be excluded)

cmd/zt_copy_blob_upload_test.go
```diff
@@ -29,6 +29,84 @@ import (
...
```
JohnRusk: should probably also have a case here `sub/somethingelse/subsub/filex`  (which I presume would not be included)

cmd/zc_traverser_list.go
```diff
@@ -78,6 +78,9 @@ func (l *listTraverser) traverse(processor objectProcessor, filters []objectFilt
 			return processor(object)
 		}
 
+		// The child traversers handle filters.
+		// Thus, --include-path xyz;abc --exclude-path def will actually exclude directory xyz/def and abc/def.
```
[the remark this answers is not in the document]
adreed-msft: In order to do this, we'd have to make slave traversers aware of meta-traversers handling filters.

cmd/zt_copy_blob_upload_test.go
adreed-msft: Good idea.

cmd/zc_enumerator.go
adreed-msft: I think we should have a single function to do this and unify the SAS splitting.

adreed-msft: We do it in a lot of places, and it tends to cause some code-duplication-creep.

JohnRusk:
I just popped an extra test case in, with a little commit.

cmd/zc_traverser_list.go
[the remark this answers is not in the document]
JohnRusk: Can you double check that Adele?  I just added a test (to this PR) to try to reproduce what this code comment says, I  couldn't reproduce it.  (FWIW, I think the behaviour describe in the code comment is not desireable, so I think its _good_ that my test does not manifest the behaviour in the comment. - but need you to check my test, and think through what's going on, and amend this code comment if necessary).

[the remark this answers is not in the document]
adreed-msft: Really? Your test executes differently on my machine (and CI) in that case.

I understand the behavior being non-preferable. Generally speaking, I may've been thinking of the solution all wrong. As-is, we just pass all of our filters through to the sub-traversers. Hand them a nil list and then check in the decorators.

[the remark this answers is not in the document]
adreed-msft: Basically, include-path was being handled as a list traverser, which establishes traversers under new roots. The filters were being ran against the new roots. This comment is now incorrect due to me changing it so that no filters are given to the child traverser, and all filters occur in the decorators.

adreed-msft:
FYI: Force-pushing to rebase and grab benchmark changes

cmd/zc_enumerator.go
adreed-msft: Resolved.

adreed-msft:
a TL;DR of what happened:
- Filters are now handled by the meta traverser
    - a nil list of filters is passed into the sub-traverser
    - The meta-traverser takes care of filters inside of the decorator function
- SAS splitting is handled by a single function now (SplitAuthTokenFromResource), living in pathUtils.go
- A merge conflict is to be expected with the HTTP warning PR, due to renaming the source variable in initResourceTraverser.

adreed-msft:
A merge conflict is also expected with files sync as well, now. This is caused by the replacement of sync's duplicated way to split auth tokens from resources.

cmd/copy.go
```diff
@@ -816,176 +813,27 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
-		}
-
-		jobPartOrder.SourceRoot = cleanLocalPath(getPathBeforeFirstWildcard(tmpSrc))
-		cca.source = cleanLocalPath(tmpSrc)
-
-	case common.ELocation.Benchmark():
-		// noop
-
-	case common.ELocation.Blob():
-		fromUrl, err := url.Parse(cca.source)
```
zezha-msft: nice

cmd/zc_traverser_list.go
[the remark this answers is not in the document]
JohnRusk: I'm puzzled that the test passed for me (and for CI when I re-ran it just now here: https://github.com/Azure/azure-storage-azcopy/pull/610/checks?check_run_id=235162070 ). But I'll review the latest changes anyway (i.e. won't hold off reviewing just because of test mystery)

cmd/sync.go
```diff
@@ -74,36 +71,22 @@ func (raw *rawSyncCmdArgs) parsePatterns(pattern string) (cookedPatterns []strin
...
 	cooked := cookedSyncCmdArgs{}
 
+	// this if statement ladder remains instead of being separated to help determine valid combinations for sync
+	// consider making a map of valid source/dest combos and consolidating this to generic source/dest setups, akin to the lower if statement
 	cooked.fromTo = inferFromTo(raw.src, raw.dst)
 	if cooked.fromTo == common.EFromTo.Unknown() {
 		return cooked, fmt.Errorf("Unable to infer the source '%s' / destination '%s'. ", raw.src, raw.dst)
 	} else if cooked.fromTo == common.EFromTo.LocalBlob() {
-		cooked.destination, cooked.destinationSAS = raw.separateSasFromURL(raw.dst)
+		cooked.destination, cooked.destinationSAS, _ = SplitAuthTokenFromResource(raw.dst, cooked.fromTo.To())
```
JohnRusk: The usual Go pattern, for cases like this where you intend to ignore the error, would be to write a method like this, and then call it from these places:

```
func MustSplitAuthTokenFromResource(resource string, location Location) (resource, sas, string) {
   resource, sas, err := SplitAuthTokenFromResource(resource, location)
    common.PanicIfErr(err)
}
```

For an example, see regex.Compile vs regex.MustCompile

I think we should do something similar here, given that its easily done, and ignored errors can lead to strange mysteries at support time.

cmd/copy.go
```diff
@@ -816,180 +813,27 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
-		fromURL, err := url.Parse(cca.source)
-		if err != nil {
-			return fmt.Errorf("error parsing the source url %s. Failed with error %s", fromURL.String(), err.Error())
-		}
-
-		// S3 management console encode ' '(space) as '+', which is not supported by Azure resources.
-		// To support URL from S3 managment console, azcopy decode '+' as ' '(space).
-		*fromURL = common.URLExtension{URL: *fromURL}.URLWithPlusDecodedInPath()
-		cca.source = fromURL.String()
+	cca.source, cca.sourceSAS, err = SplitAuthTokenFromResource(cca.source, from)
```
JohnRusk: What happened to the ToExtendedPath and cleanLocalPath calls that used to be here for local locations?  Have they gone?  Is that safe? (E.g. are they happening somewhere else anyway?)

cmd/copy.go
JohnRusk: Definitely!

cmd/copy.go
```diff
@@ -816,180 +813,27 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
-		bUrl := blobParts.URL()
-		cca.source = bUrl.String()
-
-		// set the clean source root
-		if strings.Contains(blobParts.ContainerName, "*") {
-			if blobParts.BlobName != "" {
-				return errors.New("cannot combine a wildcarded container name and blob path")
-			}
-
-			blobParts.ContainerName = ""
```
JohnRusk: What happened to this check, re can't combine wildcard container name and blob path?  Is something else already checking that elsewhere?

cmd/copy.go
```diff
@@ -816,180 +813,27 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
-		cca.sourceSAS = fileParts.SAS.Encode()
-		if cca.sourceSAS == "" {
-			return fmt.Errorf("azure files only supports SAS token authentication")
-		}
-		jobPartOrder.SourceSAS = cca.sourceSAS
-		fileParts.SAS = azfile.SASQueryParameters{}
-		fUrl := fileParts.URL()
-		cca.source = fUrl.String()
-
-		// set the clean source root
```
JohnRusk: Same question as above, re what is doing this check now?, and the stripping of share names with *

cmd/copy.go
```diff
@@ -816,180 +813,27 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
-		}
-		bfsParts := azbfs.NewBfsURLParts(*fromUrl)
-		cca.sourceSAS = bfsParts.SAS.Encode()
-		jobPartOrder.SourceSAS = cca.sourceSAS
-		bfsParts.SAS = azbfs.SASQueryParameters{}
-		bfsUrl := bfsParts.URL()
-		cca.source = bfsUrl.String() // this escapes spaces in the source
-
-		// set the clean source root
-		if strings.Contains(bfsParts.FileSystemName, "*") {
```
JohnRusk: Ditto and ditto.

cmd/copy.go
```diff
@@ -816,180 +813,27 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
...
+	jobPartOrder.SourceSAS = cca.sourceSAS
+	jobPartOrder.SourceRoot = cca.source
 
-		// set the clean source root
-		s3URLParts, err := common.NewS3URLParts(*fromURL)
-		if err != nil {
-			return err
-		}
-
-		if strings.Contains(s3URLParts.BucketName, "*") {
```
JohnRusk: All the S3 processing has disappeared!

cmd/pathUtils.go
```diff
@@ -119,8 +109,93 @@ func determineLocationLevel(location string, locationType common.Location, sourc
...
+	case common.ELocation.Local(),
+		common.ELocation.S3(),
+		common.ELocation.Benchmark(), // cover for benchmark as we generate data for that
+		common.ELocation.Unknown():   // cover for unknown as we treat that as garbage
+		// Local and S3 don't feature URL-embedded tokens
+		return resource, "", nil
+
+	// Use resource-specific APIs that all mostly do the same thing, just on the off-chance they end up doing something slightly different in the future.
+	// TODO: make GetAccountRoot and GetContainerName use their own specific APIs as well. It's _unlikely_ at best that the URL format will change drastically.
+	//       but just on the off-chance that it does, I'd prefer if AzCopy could adapt adequately as soon as the SDK catches the change
```
JohnRusk: Good thinking with these TODOs. And I agree that they are TODOs (i.e. shouldn't be done now).  For the first one, maybe add a backlog item, since it sounds like something that may catch us out again if we don't fix it. (Where's the second one, about find a clever way to reduce code duplication) only merits an in-code TODO, and not any backlog item).

cmd/copy.go
JohnRusk: And its not just the validation check that's gone AWOL ;-).  It's also the stripping of container names that contain * out of the source root, which is important too (else source root functionality will be broken and, even in the best case, won't therefore optimize the file size in plan files)

cmd/copy.go
adreed-msft: This change may well have been a massive mistake.

I for some reason, didn't think about the local case and went "ah, it's fine, all of this does the same thing, anyway."

cmd/zc_attr_filter_windows.go
```diff
@@ -39,9 +40,9 @@ func (f *attrFilter) doesSupportThisOS() (msg string, supported bool) {
 	return
 }
 
-// TODO: review handling of filename for long path support
 func (f *attrFilter) doesPass(storedObject storedObject) bool {
-	fileName := filepath.Join(f.filePath, storedObject.relativePath)
+	// generate the separator-consolidated file path
+	fileName := common.GenerateFullPath(f.filePath, storedObject.relativePath)
```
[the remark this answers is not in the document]
adreed-msft: I probably should change them. Filepath.Join retains the path separators whereas generatefullpath consolidates them.

cmd/copy.go
adreed-msft: I'd like to maintain what I've done so I'll re-add the missing effects in one way or another.

adreed-msft:
Force-pushed to remove the filter changes

cmd/zc_traverser_list.go
```diff
@@ -78,6 +78,9 @@ func (l *listTraverser) traverse(processor objectProcessor, filters []objectFilt
 			return processor(object)
 		}
 
+		// The child traversers handle filters.
```
JohnRusk: BTW - commenting here becuase of restrictions on where GitHub will let me comment: There was a change at the start of zc_attrfilter_windows.doesPass, which got lost when the filter changes were removed.  See todo that's there now, and first line of function.

cmd/zc_traverser_list.go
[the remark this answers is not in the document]
JohnRusk: Oh, I see. I had manipulated the test to make it pass with the current code!  It does indeed fail if I undo that manipulation of the test, and actually make it assert the behavior I think we should have. (I must have tweaked it to pass while we discussed the issue).

PR #576, opened 2019-09-04
Alter block blob cleanup behavior as discussed

adreed-msft:
If we fail the transfer, outright delete the blob.

If we cancel the transfer, only delete the blob if there's not a committed blob present.

Make a full-file single chunk upload stage a block instead of using upload-- Doing this means that if the transfer gets cancelled in a single-chunk transfer, the deletion code will still trigger correctly.

This also has the side effect of making the code paths more consistent.

Because all codepaths follow the same behavior now, putListNeed becomes functionally useless, and is more or less cluttering up otherwise functionally useful code, thus it is removed.


ste/sender-blockBlobFromLocal.go
```diff
@@ -86,26 +83,18 @@ func (u *blockBlobUploader) generatePutWholeBlob(id common.ChunkID, blockIndex i
...
-				jptm.FailActiveUpload("Getting hash", errNoHash)
-				return
-			}
-			u.headersToApply.ContentMD5 = md5Hash
-
 			// Upload the file
 			body := newPacedRequestBody(jptm.Context(), reader, u.pacer)
-			_, err = u.destBlockBlobURL.Upload(jptm.Context(), body, u.headersToApply, u.metadataToApply, azblob.BlobAccessConditions{})
+			_, err = u.destBlockBlobURL.StageBlock(jptm.Context(), encodedBlockID, body, azblob.LeaseAccessConditions{}, nil)
 		}
```
zezha-msft: I don't understand this change. Using Stage Block -> Commit block list will destroy performance with small file uploads.

ste/sender-blockBlobFromLocal.go
```diff
@@ -86,26 +83,18 @@ func (u *blockBlobUploader) generatePutWholeBlob(id common.ChunkID, blockIndex i
...
 		} else {
-			// File with content
-
-			// Get the MD5 that was computed as we read the file
-			md5Hash, ok := <-u.md5Channel
-			if !ok {
-				jptm.FailActiveUpload("Getting hash", errNoHash)
-				return
-			}
-			u.headersToApply.ContentMD5 = md5Hash
```
zezha-msft: Why removing?

ste/sender-blockBlobFromLocal.go
adreed-msft: I'm... actually not entirely sure what was running through my head when I did that.

Going to undo that change and just have the block blob cleanup behavior altered.

ste/sender-blockBlobFromLocal.go
adreed-msft: In the old changes, this would've caused two pulls from the MD5 channel.

ste/sender-blockBlob.go
```diff
@@ -176,7 +176,16 @@ func (s *blockBlobSenderBase) Cleanup() {
 		//    situation. Deletion has very different semantics then, compared to not deleting.
 		deletionContext, cancelFn := context.WithTimeout(context.Background(), 30*time.Second)
 		defer cancelFn()
-		_, _ = s.destBlockBlobURL.Delete(deletionContext, azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
+		if jptm.WasCanceled() {
```
JohnRusk: The heart of the original problem was that this section of code isn't even reached when there's a cancellation.

Also, the same issue exists (of this section not being even reached) for Files and ADLS Gen 2.

ste/sender-blockBlob.go
```diff
@@ -176,7 +176,16 @@ func (s *blockBlobSenderBase) Cleanup() {
...
+			_, err := s.destBlockBlobURL.GetProperties(deletionContext, azblob.BlobAccessConditions{})
+
+			if stgErr, ok := err.(azblob.StorageError); ok && stgErr.ServiceCode() == azblob.ServiceCodeBlobNotFound {
+				_, _ = s.destBlockBlobURL.Delete(deletionContext, azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
```
JohnRusk: Maybe pop a comment here to say that Delete can delete uncommited blobs (and that is in fact why we are doing it here).  Otherwise, its very counter intuitve for a failed existence test to be followed by an action on the "non-existent" object.

ste/sender-blockBlob.go
adreed-msft: Oh! I thought the heart of the issue was the cleanup strategy, and was unaware that epilogue code doesn't get called on a cancel.

adreed-msft: Fixed up. I used a sync.WaitGroup in order to track currently running transfers and awaited their completion in lifecycle manager. As such, this is a fairly global solution and doesn't require much per-subcommand modification.

ste/sender-blockBlob.go
adreed-msft: Done

zezha-msft:
@adreed-msft please tag me when you want me to take a second look

common/lifecyleMgr.go
```diff
@@ -371,6 +387,11 @@ func (lcm *lifecycleMgr) InitiateProgressReporting(jc WorkController) {
 		// cancelChannel will be notified when os receives os.Interrupt and os.Kill signals
 		signal.Notify(lcm.cancelChannel, os.Interrupt, os.Kill)
 
+		go func() {
```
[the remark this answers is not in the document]
adreed-msft: Not at all.

common/lifecyleMgr.go
```diff
@@ -233,7 +241,15 @@ func (lcm *lifecycleMgr) processOutputMessage() {
 	// this function constantly pulls out message to output
 	// and pass them onto the right handler based on the output format
 	for {
-		switch msgToPrint := <-lcm.msgQueue; lcm.outputFormat {
+		msgToPrint := <-lcm.msgQueue
+
+		// These are exit conditions.
+		// TODO: Await all jobs finished signal.
```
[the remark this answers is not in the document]
adreed-msft: This was from the old implementation with the waitgroup. It's no longer a necessary comment. Removing it.

common/lifecyleMgr.go
```diff
@@ -233,7 +241,15 @@ func (lcm *lifecycleMgr) processOutputMessage() {
...
+		if msgToPrint.shouldExitProcess() {
+			lcm.taskWaitGroup.Wait()
```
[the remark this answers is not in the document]
adreed-msft: This is overall no longer used

ste/sender-blockBlob.go
```diff
@@ -168,22 +169,28 @@ func (s *blockBlobSenderBase) Cleanup() {
...
+			// But if there are committed blocks, leave them there (since they still safely represent the state before our job even started)
+			blockList, err := s.destBlockBlobURL.GetBlockList(deletionContext, azblob.BlockListAll, azblob.LeaseAccessConditions{})
+			hasUncommittedOnly := err == nil && len(blockList.CommittedBlocks) == 0 && len(blockList.UncommittedBlocks) > 0
+			if hasUncommittedOnly {
+				jptm.LogAtLevelForCurrentTransfer(pipeline.LogDebug, "Deleting uncommitted destination blob due to cancellation")
+				// Delete can delete uncommitted blobs.
+				_, _ = s.destBlockBlobURL.Delete(deletionContext, azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
+			}
+		} else {
+			// TODO: review (one last time) should we really do this?  Or should we just give better error messages on "too many uncommitted blocks" errors
```
adreed-msft: >"too many uncommitted blocks" [](start = 110, length = 29)

I think if we want to log this we should check the error that causes the failure (or check the uncommitted block list length)

ste/sender.go
```diff
@@ -153,6 +154,12 @@ func createChunkFunc(setDoneStatusOnExit bool, jptm IJobPartTransferMgr, id comm
 				defer jptm.LogChunkStatus(id, common.EWaitReason.ChunkDone())
 			}
 		}
+
+		// tell the jptm that the destination should be assumed to have been modified
+		// (this is necessary for those cases where the prologue does not modify the dest, so the flag will not have been set at prologue time)
+		// It's idempotent, so we call it every time rather than, say, test OffsetInFile and assume that id.OffsetInFile == 0 will always run first.
+		jptm.SetDestinationIsModified()
```
adreed-msft: Alright, here's a thought: S2S senders, if the src length is 0 still generate a chunk func. Is this supposed to be set on these as well?

adreed-msft:
I can't approve my PR but your changes look alright, just got one or two thoughts to consider before :shipit:

ste/sender-blockBlob.go
JohnRusk: Yes, that makes sense to check the error code if we're going to log it something.  There are pros and cons of deleting here always.  Pros are that its consistent, both with itself (always does some thing regardless of uncommitted block count) and with Files and ADLS Gen2); and that it automatically handles the too many uncommitted blocks issue, without customer intervention.  The con is that its (usually) unnecessary to delete.  

I think I'm still leaning towards the following: It's easier and clearer to explain that "AzCopy always deletes the destination object if there's failure" (without having to get into the subtleties of it doing something different for Blob compared to everything else.  I think that's the same logic that brought us to the deletion approach in the first place back before GA.

Also, that's what it has been doing since GA. (While cancellation behavior is fixed by this PR, behavior on failure of a single transfer has always been to delete, even for blob).  The last thing we'd want to do is change it in this release, to something different from what it's historically been.... and then change our minds and change it back.   It makes more sense to leave it as it always has been, and only change it if/when we are certain.

So think I'm still leaning, slightly, towards we should indeed delete here.

Let's see what @zezha-msft says...

ste/sender.go
JohnRusk: Good question. I can see logical reasons for both possible answers (yes and no).  I'm leaning towards "yes, it is correct to set this" because the flag is "destination _modified_" and if we create a zero-length object at the destination then we have indeed "modified" it. (If the flag was called "destination requires cleanup", the argument would be stronger than we should _not_ set it).

BTW, if there's a cancellation in the interval of time between one of those 0-length chunk funcs and the epilogue for that blob, we'll get the block list, and realise we don't need to delete it.  So the result to the user is the same no matter what we set here (no deletion).

So I'm going to leave this code as is, and mark this resolved.

JohnRusk:
> @adreed-msft please tag me when you want me to take a second look

@zezha-msft Consider yourself tagged :-)

ste/mgr-JobPartTransferMgr.go
```diff
@@ -416,6 +425,49 @@ func (jptm *jobPartTransferMgr) SetErrorCode(errorCode int32) {
...
+// the raw status values do not reflect possible cancellation.
+// Do not call directly. Use IsDeadBeforeStart or IsDeadInflight
+// instead because they usually require different handling
+func (jptm *jobPartTransferMgr) isDead() bool {
+	return jptm.TransferStatusIgnoringCancellation() < 0 || jptm.WasCanceled()
+}
+
+// IsDeadBeforeStart is true for transfers that fail or are cancelled before any action is taken
+// that may affect the destination.
+func (jptm *jobPartTransferMgr) IsDeadBeforeStart() bool {
```
zezha-msft: >IsDeadBeforeStart [](start = 32, length = 17)

nice

PR #577, opened 2019-09-04
Warn the user when they're using HTTP

adreed-msft:
Also fixed inferring on HTTP

### DIFF
--- a/cmd/zc_enumerator.go
+++ b/cmd/zc_enumerator.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Azure/azure-pipeline-go/pipeline"
@@ -125,6 +126,9 @@ func initContainerDecorator(containerName string, processor objectProcessor) obj
 }
 
 const accountTraversalInherentlyRecursiveError = "account copies are an inherently recursive operation, and thus --recursive is required"
+const httpsRecommendedNotice = "NOTE: HTTP is in use for one or more location(s). The use of HTTP is not recommended." // TODO: I had some trouble figuring out this wording. Any ideas?
+
+var httpsRecommendationOnce sync.Once
 
 // source, location, recursive, and incrementEnumerationCounter are always required.
 // ctx, pipeline are only required for remote resources.
@@ -207,6 +211,12 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 			return nil, err
 		}
 
+		if sourceURL.Scheme == "http" {
+			httpsRecommendationOnce.Do(func() {
+				glcm.Info(httpsRecommendedNotice)
+			})
+		}
+
 		if ctx == nil || p == nil {
 			return nil, errors.New("a valid credential and context must be supplied to create a blob traverser")
 		}
@@ -227,6 +237,12 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 		sourceURL, err := url.Parse(source)
 		if err != nil {
 			return nil, err
+		}
+
+		if sourceURL.Scheme == "http" {
+			httpsRecommendationOnce.Do(func() {
+				glcm.Info(httpsRecommendedNotice)
+			})
 		}
 
 		if ctx == nil || p == nil {
@@ -250,8 +266,16 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 			return nil, err
 		}
 
-		if ctx == nil || p == nil {
+		// check if credential is also nil here (would never trigger) to tame syntax highlighting.
+		// As a precondition to pipeline p, credential must not be nil anyway.
+		if ctx == nil || p == nil || credential == nil {
 			return nil, errors.New("a valid credential and context must be supplied to create a blobFS traverser")
+		}
+
+		if sourceURL.Scheme == "http" && credential.CredentialType != common.ECredentialType.SharedKey() {
+			httpsRecommendationOnce.Do(func() {
+				glcm.Info(httpsRecommendedNotice)
+			})
 		}
 
 		bfsURL := azbfs.NewBfsURLParts(*sourceURL)
@@ -271,6 +295,12 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 		sourceURL, err := url.Parse(source)
 		if err != nil {
 			return nil, err
+		}
+
+		if sourceURL.Scheme == "http" {
+			httpsRecommendationOnce.Do(func() {
+				glcm.Info(httpsRecommendedNotice)
+			})
 		}
 
 		s3URLParts, err := common.NewS3URLParts(*sourceURL)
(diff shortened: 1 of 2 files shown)


cmd/zc_enumerator.go
```diff
@@ -111,6 +112,9 @@ func initContainerDecorator(containerName string, processor objectProcessor) obj
 }
 
 const accountTraversalInherentlyRecursiveError = "account copies are an inherently recursive operation, and thus --recursive is required"
+const httpsRecommendedNotice = "NOTE: HTTP is in use for one or more location(s). The use of HTTP is not recommended." // TODO: I had some trouble figuring out this wording. Any ideas?
```
zezha-msft: >. [](start = 116, length = 1)

maybe mention "because of security concerns"

cmd/zc_enumerator.go
```diff
@@ -255,6 +279,12 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 			return nil, err
 		}
 
+		if sourceURL.Scheme == "http" {
+			httpsRecommendationOnce.Do(func() {
+				glcm.Info(httpsRecommendedNotice)
+			})
+		}
```
zezha-msft: can we perhaps extract out the part of the code where the URL is parsed and "http" is checked?

zezha-msft: so that it's not repeated

---
In reply to: [321544752](https://github.com/Azure/azure-storage-azcopy/pull/577#discussion_r321544752) [](ancestors = 321544752)

cmd/zc_enumerator.go
```diff
@@ -200,6 +204,12 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 			return nil, err
 		}
 
+		if sourceURL.Scheme == "http" {
+			httpsRecommendationOnce.Do(func() {
```
JohnRusk: Minor: you could define this to reduce duplication in increase brevity:

```
func recommendHttpsIfNecessary(url URL) {
        if url.Scheme.EqualFold("http") {
	    httpsRecommendationOnce.Do(func() { 
				glcm.Info(httpsRecommendedNotice) 
			})
         }
}
```

Then call that everywhere you need it

cmd/zc_enumerator.go
```diff
@@ -243,10 +259,18 @@ func initResourceTraverser(source string, location common.Location, ctx *context
...
 		}
 
-		if ctx == nil || p == nil {
+		// check if credential is also nil here (would never trigger) to tame syntax highlighting.
+		// As a precondition to pipeline p, credential must not be nil anyway.
+		if ctx == nil || p == nil || credential == nil {
 			return nil, errors.New("a valid credential and context must be supplied to create a blobFS traverser")
 		}
 
+		if sourceURL.Scheme == "http" && credential.CredentialType != common.ECredentialType.SharedKey() {
```
JohnRusk: Is in necessary to have this exception for shared key auth?  IIRC Shared Key auth does protect the credentials over HTTP (which is presumably why this exception is here)… but it does not protect the data that's in transit (request/response body) nor does it protect filenames (request headers).  So I'm feeling that we should always warn, even if using SharedKey.

cmd/zc_enumerator.go
```diff
@@ -222,6 +232,12 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 			return nil, err
 		}
 
+		if sourceURL.Scheme == "http" {
+			httpsRecommendationOnce.Do(func() {
+				glcm.Info(httpsRecommendedNotice)
```
JohnRusk: A general comment: I'm noticing that all the changes in this PR reference variables named "source", never dest.  However:
(a) for uploads we should encourage HTTPS on the destination.
(b) for S2S we should (and I think already do) encourage HTTPS on the destination (because the source SAS tokens are passed to the destination - and so should not be passed over a plaintext channel).

I think (b) is already addressed somewhere in the code. With a hard error rather than a warning. That's good. Because there's just no way we want people passing SAS's over HTTP.  So (b) is done and its outside the scope of this PR. I just mention it here for completeness..  But I think (a) might be not done yet - unless I'm misreading this PR and it actually does cover destinations.

Does (a) need to be added to this PR?

cmd/zc_enumerator.go
```diff
@@ -200,6 +204,12 @@ func initResourceTraverser(source string, location common.Location, ctx *context
 			return nil, err
 		}
 
+		if sourceURL.Scheme == "http" {
```
JohnRusk: Safer perhaps to use EqualFold instead of ==?  (Not sure whether we can be sure that what we have in the URL object is always lower case.  Maybe they always are in Go. I don't know).

cmd/zc_enumerator.go
JohnRusk: Oh, I just letf a comment to the same effect :-)  Will leave my comment in too.

cmd/zc_enumerator.go
adreed-msft: I tackled this due to a PBI regarding if we should support HTTP outside of SharedKey usage... I'm not entirely sure why, but I think @zezha-msft has said that SharedKey is actually better on HTTP for some reason.

cmd/zc_enumerator.go
adreed-msft: Done

cmd/zc_enumerator.go
adreed-msft: So, initResourceTraverser is run on both the source and the destination.

I'll change the name of source to resource instead, since it does seem to be a bit confusing.

cmd/zc_enumerator.go
adreed-msft: Done

cmd/zc_enumerator.go
JohnRusk: It is better than the other forms of auth, for the reason I said in my comment.  But, also for the reasons I said in my comment, it's not better _enough_ for us to justify providing no warning.

adreed-msft: Fair point. I'll change this.

cmd/zc_enumerator.go
```diff
@@ -125,18 +126,29 @@ func initContainerDecorator(containerName string, processor objectProcessor) obj
 }
 
 const accountTraversalInherentlyRecursiveError = "account copies are an inherently recursive operation, and thus --recursive is required"
+const httpsRecommendedNotice = "NOTE: HTTP is in use for one or more location(s). The use of HTTP is not recommended due to security concerns." // TODO: I had some trouble figuring out this wording. Any ideas?
```
zezha-msft: >// TODO: I ha [](start = 144, length = 13)

looks fine to me

zezha-msft: please rm comment and merge

zezha-msft: Editing right inside Github is pretty fast.

PR #582, opened 2019-09-07
Feature/decompress by content encoding

JohnRusk:
This is the new --decompress flag

Still todo:

Rebase on latest dev
Add smoke test

But this should be enough to get initial feedback

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -1246,7 +1246,7 @@ func init() {
 	cpCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "defines the location of json which has the list of only files to be copied")
 	cpCmd.PersistentFlags().StringVar(&raw.legacyExclude, "exclude-pattern", "", "exclude these files when copying. Support use of *.")
 	cpCmd.PersistentFlags().StringVar(&raw.forceWrite, "overwrite", "true", "defines whether to overwrite the conflicting files at the destination. Could be set to true, false, or prompt.")
-	cpCmd.PersistentFlags().BoolVar(&raw.autoDecompress, "decompress", false, "automatically decompress files when downloading, if their content-encoding indicates that they are compressed.")
+	cpCmd.PersistentFlags().BoolVar(&raw.autoDecompress, "decompress", false, "automatically decompress files when downloading, if their content-encoding indicates that they are compressed. The supported content-encoding values are 'gzip' and 'deflate'.")
 	cpCmd.PersistentFlags().BoolVar(&raw.recursive, "recursive", false, "look into sub-directories recursively when uploading from local file system.")
 	cpCmd.PersistentFlags().StringVar(&raw.fromTo, "from-to", "", "optionally specifies the source destination combination. For Example: LocalBlob, BlobLocal, LocalBlobFS.")
 	cpCmd.PersistentFlags().StringVar(&raw.excludeBlobType, "exclude-blob-type", "", "optionally specifies the type of blob (BlockBlob/ PageBlob/ AppendBlob) to exclude when copying blobs from Container / Account. Use of "+
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -85,6 +85,7 @@ type rawCopyCmdArgs struct {
 	stripTopDir       bool
 	followSymlinks    bool
 	withSnapshots     bool
+	autoDecompress    bool
 	// forceWrite flag is used to define the User behavior
 	// to overwrite the existing blobs or not.
 	forceWrite string
@@ -189,6 +190,11 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	if err != nil {
 		return cooked, err
 	}
+	allowAutoDecompress := fromTo == common.EFromTo.BlobLocal() || fromTo == common.EFromTo.FileLocal()
+	if raw.autoDecompress && !allowAutoDecompress {
+		return cooked, errors.New("automatic decompression is only supported for downloads from Blob and Azure Files") // as at Sept 2019, our ADLS Gen 2 Swagger does not include content-encoding for directory (path) listings so we can't support it there
+	}
+	cooked.autoDecompress = raw.autoDecompress
 
 	// cooked.stripTopDir is effectively a workaround for the lack of wildcards in remote sources.
 	// Local, however, still supports wildcards, and thus needs its top directory stripped whenever a wildcard is used.
@@ -547,6 +553,7 @@ type cookedCopyCmdArgs struct {
 	followSymlinks     bool
 	withSnapshots      bool
 	forceWrite         common.OverwriteOption
+	autoDecompress     bool
 
 	// options from flags
 	blockSize uint32
@@ -756,6 +763,7 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 		JobID:           cca.jobID,
 		FromTo:          cca.fromTo,
 		ForceWrite:      cca.forceWrite,
+		AutoDecompress:  cca.autoDecompress,
 		Priority:        common.EJobPriority.Normal(),
 		LogLevel:        cca.logVerbosity,
 		Include:         cca.legacyInclude,
@@ -1238,6 +1246,7 @@ func init() {
 	cpCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "defines the location of json which has the list of only files to be copied")
 	cpCmd.PersistentFlags().StringVar(&raw.legacyExclude, "exclude-pattern", "", "exclude these files when copying. Support use of *.")
 	cpCmd.PersistentFlags().StringVar(&raw.forceWrite, "overwrite", "true", "defines whether to overwrite the conflicting files at the destination. Could be set to true, false, or prompt.")
+	cpCmd.PersistentFlags().BoolVar(&raw.autoDecompress, "decompress", false, "automatically decompress files when downloading, if their content-encoding indicates that they are compressed.")
 	cpCmd.PersistentFlags().BoolVar(&raw.recursive, "recursive", false, "look into sub-directories recursively when uploading from local file system.")
 	cpCmd.PersistentFlags().StringVar(&raw.fromTo, "from-to", "", "optionally specifies the source destination combination. For Example: LocalBlob, BlobLocal, LocalBlobFS.")
 	cpCmd.PersistentFlags().StringVar(&raw.excludeBlobType, "exclude-blob-type", "", "optionally specifies the type of blob (BlockBlob/ PageBlob/ AppendBlob) to exclude when copying blobs from Container / Account. Use of "+
--- a/common/decompressingWriter.go
+++ b/common/decompressingWriter.go
@@ -33,7 +33,8 @@ type decompressingWriter struct {
 	workerError chan error
 }
 
-var decompressingWriterBufferPool = NewMultiSizeSlicePool(1 * 1024 * 1024)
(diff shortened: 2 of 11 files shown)


cmd/zc_traverser_blob.go
```diff
@@ -130,13 +131,19 @@ func (t *blobTraverser) traverse(processor objectProcessor, filters []objectFilt
 				continue
 			}
 
+			encoding := ""
```
adreed-msft: Note for sometime after the s2s refactor gets merged in: use common.IffStringNotNil here instead.

cmd/zc_enumerator.go
```diff
@@ -48,6 +48,7 @@ type storedObject struct {
 	size             int64
 	md5              []byte
 	blobType         azblob.BlobType // will be "None" when unknown or not applicable
+	contentEncoding  string
```
adreed-msft: You may want to base on the s2s refactor. It already takes in parameters like this, and we may well be stepping on each other's toes. That or, whichever of our PRs gets merged in last needs to resolve this.

common/decompressingWriter.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"compress/gzip"
+	"compress/zlib"
+	"errors"
+	"io"
+	"time"
+)
+
+type decompressingWriter struct {
```
adreed-msft: Suggestion: Change name to decompressor since this doesn't implement the io.Writer interface.

common/decompressingWriter.go
```diff
@@ -0,0 +1,138 @@
...
+	pipeWriter  *io.PipeWriter
+	workerError chan error
+}
+
+const decompressingWriterCopyBufferSize = 256 * 1024 // 1/4 the size that we usually write to disk with (elsewhere in codebase). 1/4 to try to keep mem usage a bit lower, without going so small as to compromize perf
```
adreed-msft: Do you think this would cause a bottleneck? Thinking about it, we'd never actually fill up the larger buffer in theory.

common/decompressingWriter.go
```diff
@@ -0,0 +1,138 @@
...
+type CompressionType uint8
+
+func (CompressionType) None() CompressionType        { return CompressionType(0) }
+func (CompressionType) ZLib() CompressionType        { return CompressionType(1) }
+func (CompressionType) GZip() CompressionType        { return CompressionType(2) }
+func (CompressionType) Unsupported() CompressionType { return CompressionType(255) }
+
+// NewDecompressingWriter returns a WriteCloser which decompresses the data
+// that is written to it, before passing the decompressed data on to a final destination
+func NewDecompressingWriter(destination io.WriteCloser, ct CompressionType) io.WriteCloser {
```
adreed-msft: This feels wrong wrapping around another writer in a parasitic way, as it's harder to actually wrap my head around.

I understand completely what it's doing, but for some reason my gut wants it to take in a reader, not a writer, since this is effectively just a pipe through a decompressor.

adreed-msft: That being said though, this isn't me saying "don't do this"-- I just want to put my thoughts on this out there. Feel free to ping me on teams if you'd like to chat about it.

---
In reply to: [322487718](https://github.com/Azure/azure-storage-azcopy/pull/582#discussion_r322487718) [](ancestors = 322487718)

common/decompressingWriter.go
```diff
@@ -0,0 +1,138 @@
...
+	dec, err := d.decompressorFactory(tp, preader)
+	if err != nil {
+		workerError <- err
+		return
+	}
+
+	// Now read from the pipe, decompressing as we go, until
+	// reach EOF on the pipe (or encounter an error)
+	b := decompressingWriterBufferPool.RentSlice(decompressingWriterCopyBufferSize)
+	_, err = io.CopyBuffer(destination, dec, b) // returns err==nil if hits EOF, as per docs
```
adreed-msft: Consider adding a comment mentioning that this would cause the goroutine to hold until the EOF or an error is hit

ste/mgr-JobPartTransferMgr.go
```diff
@@ -158,6 +161,28 @@ func (jptm *jobPartTransferMgr) GetOverwriteOption() common.OverwriteOption {
...
+func (jptm *jobPartTransferMgr) GetSourceCompressionType() (common.CompressionType, error) {
+	encoding := jptm.Info().SrcHTTPHeaders.ContentEncoding
+	switch strings.ToLower(encoding) {
+	case "":
+		return common.ECompressionType.None(), nil
+	case "gzip":
+		return common.ECompressionType.GZip(), nil
+	case "deflate":
+		return common.ECompressionType.ZLib(), nil
+	default:
```
adreed-msft: Consider just logging that it's unsupported instead of erroring out.

cmd/zc_enumerator.go
JohnRusk: Good point - I'll hold off and rebase once yours is in.

common/decompressingWriter.go
JohnRusk: I wasn't so much worried about perf, but rather about the fact that these buffers aren't tracked within our memory budget (that we use for the main chunk buffers).  So if there are a lot of compressed files getting decompressed in parallel - e.g. a thousand (which is possible if they are small) and concurrency is high - then that starts to add up to some non-trivial mem usage.  Not sure if I understood your question properly tho...

common/decompressingWriter.go
JohnRusk: Yeah, I know the feeling.  It too me a while to get things straight in my head. Where I ended up was that it has to be an instance of the decorator pattern - i.e. have the same API as the thing that it wraps. That's so that the rest of our code doesn't care whether its dealing with a plain disk file, or this new magic thing that automatically handles the decompression.  The requirement that the rest of the code shouldn't know (or care) leads to the requirement that both the wrapper and the "wrappee" must have the same interface - which must be writeCloser because the "wrappee" is a disk file and that's a writeCloser.

So it's not really solving the problem on "what would I write if I just wanted to decompress some stuff", its solving the slightly different problem of "what would I write if I wanted to decompress some stuff _transparently_, with an interface that looks just like an ordinary writable file".

I agree that there are some intuitive hurdles here, but on reflection the decorator approach still feels like the right option for the codebase as a whole.  So I'll mark this as resolved.  Definitely agree that we noticed similar mental hurdles here :-)

ste/mgr-JobPartTransferMgr.go
JohnRusk: Once were in the STE (rather than enumerators), erroring out is actually the right thing to do - because it only errors us out of that one transfer. I.e. it will fail the transfer, but not the job. Does that make sense, and address what you were thinking of in your comment?

common/decompressingWriter.go
adreed-msft: I understand that much. Perhaps we could make it so that the buffers are split half and half when decompression is enabled.

My thought is mostly that we're trying to squeeze data from a smaller pipe into a larger pipe-- we'd never actually fill the larger pipe up, and our throughput will be lowered by virtue of this.

ste/mgr-JobPartTransferMgr.go
adreed-msft: That makes _sense_, but what I'm saying is more along the lines of "let the transfer continue, but warn the user that there's no special decompression happening"

cmd/copyEnumeratorInit.go
```diff
@@ -69,6 +69,9 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 			Destination:      dst,
 			LastModifiedTime: object.lastModifiedTime,
 			SourceSize:       object.size,
+			ContentEncoding:  object.contentEncoding,
+			// TODO: should we have just one bit of code that copies a storedObject into a common.CopyTransfer?  Currently we have at least this
+			//   plus the one in scheduleCopyTransfer that's used by sync... Seems confusing to have two?
```
zezha-msft: we should indeed consolidate them.

cmd/copy.go
```diff
@@ -1273,6 +1281,7 @@ func init() {
 	cpCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "defines the location of json which has the list of only files to be copied")
 	cpCmd.PersistentFlags().StringVar(&raw.legacyExclude, "exclude-pattern", "", "exclude these files when copying. Support use of *.")
 	cpCmd.PersistentFlags().StringVar(&raw.forceWrite, "overwrite", "true", "defines whether to overwrite the conflicting files at the destination. Could be set to true, false, or prompt.")
+	cpCmd.PersistentFlags().BoolVar(&raw.autoDecompress, "decompress", false, "automatically decompress files when downloading, if their content-encoding indicates that they are compressed.")
```
zezha-msft: >automatically [](start = 76, length = 13)

mention the supported encoding

common/decompressingWriter.go
JohnRusk: >My thought is mostly that we're trying to squeeze data from a smaller pipe into a larger pipe-- we'd never actually fill the larger pipe up, and our throughput will be lowered by virtue of this

I'm not sure I've understood that, can you elaborate?

common/decompressingWriter.go
JohnRusk: That comment already exists, above.

ste/mgr-JobPartTransferMgr.go
JohnRusk: I don't think that's the right thing to do.  In general, we should do exactly as the user asks, or else fail in a _small number_ of _easily identified_ ways(*).  Generally, textual warnings to the user don't meet those criteria - since on a large job with 100's of thousands of files, its not practical for a user to read and respond to a large number of warnings on screen.  And it's also not practical for them to read and respond to a large number of warnings in the log, since what would they search for, and how would they know that they even need to do such a search?

(*) I know we already have to break this "rule" at various places in the enumeration code.  But so far we are not routinely breaking it in the STE, and I don't want to start breaking it in the STE.

I'm gonna close this thread with no change to the code, since that's the only option that will leave it consistent with policy used in the rest of the STE.

cmd/copy.go
JohnRusk: Done

common/decompressingWriter.go
JohnRusk: BTW, it's also worth noting that they copyBuffer call that uses this buffer just keeps iterating through the data until its all done. So the only ways this could hurt perf are:
1. The CPU cost of the extra iterations (won't be an issue because it's virtually always disk or network or Service that's the bottleneck, not CPU)
2. A change in disk write performance due to the smaller buffer size.  This is the one I'm slightly concerned about. But there's probably a wide sweet-spot of suitable values, and I doubt this size pulls us out of there.  E.g. using something like 32KB might change the perf considerably, but dropping from 1 MB to 1/4 MB as we are doing here is unlikely, as far as I can tell, to make any significant change.
3. Using too much RAM, and causing paging.  This is why I made it 1/4 of what we normally use.

I haven't fully thought through the implications of trying to use the same pool as we do for the main chunk buffers. I fear there could be potential for starvation or deadlock there - e.g. all our chunk buffers are allocated, but they can't be saved because we can't allocate any decompression buffers because the chunk buffers are taking up all the RAM.  

Given all of the above, and the need to keep our workflow as simple as possible this week in terms of getting PRs completed, I propose to do no changes to this code in this PR.  Please let me know if you think we should create a new backlog item, and or PR, to test and tweak this buffer size.

cmd/zc_enumerator.go
JohnRusk: Done. Thanks for the tip.

cmd/copyEnumeratorInit.go
JohnRusk: I think I'll do this in a separate PR

JohnRusk:
Just rebased to dev

PR #584, opened 2019-09-09
Fix/misc small fixes

JohnRusk:


ste/init.go
```diff
@@ -453,6 +453,9 @@ func GetJobSummary(jobID common.JobID) common.ListJobSummaryResponse {
 		js.PercentComplete = 100
 	} else {
 		js.PercentComplete = 100 * float32(js.TotalBytesTransferred) / float32(js.TotalBytesExpected)
+		if js.PercentComplete > 100 {
+			js.PercentComplete = 100 // we had some transient conditions that temporarily gave values above 100
```
adreed-msft: Uhhh... This seems like a really hacky solution. What were the underlying conditions that would cause values above 100%? I don't feel like flashing 100% for a few seconds and then going back down (if that's even possible) once those conditions clear solves the problem.

JohnRusk: I don't know what the underlying conditions are. My repro that was working a few weeks ago didn't work when I tried it again. You're right that its hacky and ugly, for all the reasons you said.  However, I don't feel I have time to do any better in the current release cycle.

JohnRusk: Oh, I've discovered another way to reproduce it, and a more persistent (and therefore debuggable) manner.  I will investigate.

adreed-msft: That's fair. If it's not feasible to get done right now, I'm fine with a hacky solution on something this non-critical for a little while.

zezha-msft: It's indeed a bit hacky, but I'm ok with following up with a permanent solution in the next release.

JohnRusk: I've found and fixed it. There were two problems. The main one was that I was deliberately not taking into account the fact that the last chunk in each file may be smaller than the block size.  I thought it wouldn't make much difference - but as we've seen in recent days, there are cases where it can.  The second was some badly chosen integer types - surprisingly its not an error in Go to negate an unsigned int, and pass the result to another unsigned int.

adreed-msft: Good job!

I'm actually not surprised by the lack of erroring on that in Go. It doesn't error out on integer over-(or under)-flows.

ste/JobsAdmin.go
```diff
@@ -420,12 +420,16 @@ func (ja *jobsAdmin) BytesOverWire() int64 {
...
-	atomic.AddInt64(&ja.atomicSuccessfulBytesInActiveFiles, int64(n))
+func (ja *jobsAdmin) AddSuccessfulBytesInActiveFiles(n int64) {
+	atomic.AddInt64(&ja.atomicSuccessfulBytesInActiveFiles, n)
 }
 
 func (ja *jobsAdmin) SuccessfulBytesInActiveFiles() uint64 {
-	return uint64(atomic.LoadInt64(&ja.atomicSuccessfulBytesInActiveFiles))
+	n := atomic.LoadInt64(&ja.atomicSuccessfulBytesInActiveFiles)
+	if n < 0 {
+		n = 0 // should never happen, but would result in nasty over/underflow if it did
```
adreed-msft: TODO: With the supposed debug env var enabled, consider panicking instead to catch any other issues?

JohnRusk: Tempting, but given that we don't have that env var yet, and that any error would "only" relate to status reporting rather than data integrity, I propose to leave this as-is.

PR #589, opened 2019-09-10
Log response body in error message

JohnRusk:
TODO: rebase onto dev after benchmark-mode is merged


JohnRusk:
This is replaced by PR 618

PR #598, opened 2019-09-18
Update helpMessages.go

mahilleb-msft:
Azure Blob OAuth is not supported as source.


zezha-msft:
@mahilleb-msft  Thanks for the correction!

PR #603, opened 2019-09-20
Fixed empty string handling for list-of-files

zezha-msft:


cmd/copy.go
```diff
@@ -270,15 +270,21 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 			scanner := bufio.NewScanner(f)
 			for scanner.Scan() {
 				v := scanner.Text()
-				listChan <- v
+				// empty strings should be ignored, otherwise the source root itself is selected
```
adreed-msft: This just reminded me. Do we error out properly if the old format is handed in? I don't think so.

zezha-msft: The flag is internal so very few user should have visibility on it. Let's mention it in our change log.

PR #606, opened 2019-09-22
Use a lower initial starting point for auto-tuning concurrency

JohnRusk:
To better accomodate very low bandwidth networks.

We've had a few bug reports from folks with big machines (high CPU count) running of very small networks (or low-capacity proxies) and reporting problems.  This fix should help if they choose to use AUTO concurrency level, since now we'll start seeking optimum from 4 connections instead of 8 (where 8 might already be too many).

I did wonder about trying 2, but that adds one more step (and so even more time) to the tuning process for high speed cases, and 4 seems pretty low.  If you think I should make it 2, let me know.

### DIFF
--- a/ste/JobsAdmin.go
+++ b/ste/JobsAdmin.go
@@ -103,6 +103,8 @@ var JobsAdmin interface {
 	common.ILoggerCloser
 
 	CurrentMainPoolSize() int
+
+	RequestTuneSlowly()
 }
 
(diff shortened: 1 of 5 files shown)


adreed-msft:
4 seems low enough. Once we eventually get around to ranges in AUTO, we should be able to cover all cases in this setup.

PR #607, opened 2019-09-23
First draft of 10.3 changelog

JohnRusk:


ChangeLog.md
```diff
@@ -1,11 +1,105 @@
...
+   The new parameters have behaviour that is better defined in complex situations (such as
+   recursion).  The `*` wildcard is supported in the pattern parameters, but _not_ in the path ones.
+1. There have been two breaking changes to the JSON output that is produced if you request
+   JSON-formatted output. The `sync` command's output in JSON has changed for consistency reasons,
+   and the final message type, for `copy` and `sync` has changed its name from `Exit` to `EndOfJob`.
+   Tools using the JSON output format to integrate AzCopy should be aware.
+1. If downloading to "null" on Windows the target must now be named "NUL", according to standard
+   Windows conventions.  "/dev/null" remains correct on Linux. (This feature can be used to test
+   throughput or check MD5s without saving the downloaded data.) 
+1. The file format of the (still undocmented) `--list-of-files` parameter is changed.  (It remains
```
adreed-msft: TODO: Note how this has changed. I understand it's a undocumented parameter designed for ASE, but we don't know for sure if there's anybody else out there using it.

ChangeLog.md
```diff
@@ -1,11 +1,105 @@
...
+
+1. `sync` is supported from Blob Storage to Blob Storage, and from Azure Files to Azure Files.
+1. `copy` is supported from Azure Files to Azure Files, and (TBC?) from Blob Storage to Azure Files.
+1. Percent complete is displayed as each job runs.
+1. VHD files are auto-detected as page blobs.
+1. A new benchmark mode allows quick and easy performance benchmarking of your network connection to
+   Blob Storage. Run AzCopy with the paramaters `bench --help` for details.  This feature is in
+   Preview status.
+1. The location for AzCopy's "plan" files can be specified with the environment variable
+   `AZCOPY_JOB_PLAN_LOCATION`.
```
adreed-msft: Perhaps note that we still don't support moving the entire app folder-- our oAuth token is still stored in ~/.azcopy

ChangeLog.md
```diff
@@ -1,11 +1,105 @@
...
+1. Log files and plan files can be cleaned up to save disk space, using AzCopy's new `jobs rm` and
+   `jobs clean` commands.
+1. When listing jobs with `jobs show`, the status of each job is included in the output.   
+1. The `--overwrite` parameter now supports the value of "prompt" to prompt the user on a
+   file-by-file basis. (The old values of true and false are also supported.)   
+1. The environment variable `AZCOPY_CONCURRENCY_VALUE` can now be set to "AUTO". This is expected to be
```
adreed-msft: Have we documented this in our environment variable description?

ChangeLog.md
```diff
@@ -1,11 +1,105 @@
...
+    useful for customers with small networks, or those running AzCopy on
+   moderately-powered machines and transfer blobs between accounts.  This feature is in preview status.
+1. When uploading from Windows, files can be filtered by Windows-specific file attributes (such as
```
adreed-msft: Note the flag here

ChangeLog.md
```diff
@@ -1,11 +1,105 @@
...
+   "Archive", "Hidden" etc)
+1. Memory usange can be controlled by setting the new environment variable `AZCOPY_BUFFER_GB`.
```
adreed-msft: usange -> usage

ChangeLog.md
```diff
@@ -1,11 +1,105 @@
...
+   completed desination file is checked against that of the source.
+1. When downloading, AzCopy can automatically decompress blobs (or Azure Files) that have a
+   `Content-Encoding` of `gzip` or `deflate`. To enable this behaviour, supply the `--decompress`
+   parameter.
+1. The number of disk files accessed concurrently can be controlled with the new
+   `AZCOPY_CONCURRENT_FILES` environment variable. This is an advanced setting, which generally
+   should not be modified.  It does not affect the number of HTTP connections, which is still
+   controlled by `AZCOPY_CONCURRENCY_VALUE`.
+1. The values of key environment variables are listed at the start of the log file.
+1. An official 32-bit build is now released, in addition to the usual 64-bit builds. (TBC: For which OS's?)
```
adreed-msft: >(TBC: For which OS's?) [](start = 85, length = 22)

AFAIK Windows and Linux will get 32-bit builds. @zezha-msft is this accurate?

ChangeLog.md
```diff
@@ -1,11 +1,105 @@
...
+1. Logging is more concise at the default log level.
+1. Error message text, returned by Blob and File services, is now included in the log.
+1. A log file is created for copy jobs even when there was nothing to copy.
+1. In the log, UPLOAD SUCCESSFUL messages now include the name of the successful file.
+1. Clear error messages are given to show that AzCopy does not currently support Customer-Provided
+   Encryption Keys.
+1. On Windows, downloading a filename with characters not supported by the operating system will
+   result in those characters being URL-encoded to construct a Windows-compatible filename. The
+   encoding process is reversed if the file is uploaded. 
+1.  Uploading a single file to ADLS Gen 2 works now.
```
adreed-msft: Add to the end:
1. Jobs downloading from ADLS Gen 2 that result in no scheduled transfers will no longer hang.

ChangeLog.md
```diff
@@ -1,11 +1,115 @@
...
+   `Content-Encoding` of `gzip` or `deflate`. To enable this behaviour, supply the `--decompress`
+   parameter.
+1. The number of disk files accessed concurrently can be controlled with the new
+   `AZCOPY_CONCURRENT_FILES` environment variable. This is an advanced setting, which generally
+   should not be modified.  It does not affect the number of HTTP connections, which is still
+   controlled by `AZCOPY_CONCURRENCY_VALUE`.
+1. The values of key environment variables are listed at the start of the log file.
+1. An official 32-bit build is now released, in addition to the usual 64-bit builds. (TBC: For which OS's?)
+
+### Bug fixes
```
adreed-msft: This feels like a fitting place to put this, but as a part of the "log some more info" PR:

- S2S access tier preservation failures are logged
- Container name resolution is only checked on scheduled containers when using a wildcard, rather than ALL containers in an account
    - I'm starting to think this one was intentional, regarding our earlier discussion of why the name resolution was done ahead of time, actually. I don't think it's entirely something we need to worry about, however. Consider the following: A user schedules a transfer copying bucket names bucket*name, which targets `bucket-name` and `bucket.name`. `bucket.name` becomes `bucket-name-2`. Then, consider the user (somehow) scheduled a transfer which included *just* `bucket.name`. `bucket.name` resolves to `bucket-name`, and the files go to the wrong place.
    - I don't think this is so severe, though, because the same scenario would happen if `bucket-name` was deleted. There's no _possible_ way we could expect that unless we stored the map of bucket names (which I already feel is a poor idea in its own sense)

ChangeLog.md
```diff
@@ -1,11 +1,121 @@
...
+      "virtualDir".'
+   1. The second exception is when you are transferring multiple _whole_ containers (or S3 buckets). You can
+      use * as a wildcard in the container or bucket name. 
+1. The `--include` and `--exclude` parameters have been replaced by `--include-pattern` and
+   `--exclude-pattern` (for filenames) and `--include-path` and `--exclude-path` (for paths,
+   including directory and filenames).
+   The new parameters have behaviour that is better defined in complex situations (such as
+   recursion).  The `*` wildcard is supported in the pattern parameters, but _not_ in the path ones.
+1. The `remove` command now removes blobs that have snapshots (the blob and all the snapshots are
+   removed). Previously there was a bug where it would hang if snaphots existed, but now it will
```
zezha-msft: minor typo: snapshots

ChangeLog.md
```diff
@@ -1,11 +1,121 @@
...
+      _contents_ of `virtual` dir directly into the target without creating a folder named
+      "virtualDir".'
+   1. The second exception is when you are transferring multiple _whole_ containers (or S3 buckets). You can
+      use * as a wildcard in the container or bucket name. 
+1. The `--include` and `--exclude` parameters have been replaced by `--include-pattern` and
+   `--exclude-pattern` (for filenames) and `--include-path` and `--exclude-path` (for paths,
+   including directory and filenames).
+   The new parameters have behaviour that is better defined in complex situations (such as
+   recursion).  The `*` wildcard is supported in the pattern parameters, but _not_ in the path ones.
+1. The `remove` command now removes blobs that have snapshots (the blob and all the snapshots are
```
zezha-msft: change behavior to fail blobs with snapshots

ChangeLog.md
```diff
@@ -1,11 +1,121 @@
...
+   completed desination file is checked against that of the source.
+1. When downloading, AzCopy can automatically decompress blobs (or Azure Files) that have a
+   `Content-Encoding` of `gzip` or `deflate`. To enable this behaviour, supply the `--decompress`
+   parameter.
+1. The number of disk files accessed concurrently can be controlled with the new
+   `AZCOPY_CONCURRENT_FILES` environment variable. This is an advanced setting, which generally
+   should not be modified.  It does not affect the number of HTTP connections, which is still
+   controlled by `AZCOPY_CONCURRENCY_VALUE`.
+1. The values of key environment variables are listed at the start of the log file.
+1. An official 32-bit build is now released, in addition to the usual 64-bit builds. (TBC: For which
```
zezha-msft: just Windows

ChangeLog.md
```diff
@@ -1,11 +1,122 @@
...
+   removed). Previously there was a bug where it would hang if snaphots existed, but now it will
+   delete all of them.
+1. There have been two breaking changes to the JSON output that is produced if you request
+   JSON-formatted output. The `sync` command's output in JSON has changed for consistency reasons,
+   and the final message type, for `copy` and `sync` has changed its name from `Exit` to `EndOfJob`.
+   Tools using the JSON output format to integrate AzCopy should be aware.
+1. If downloading to "null" on Windows the target must now be named "NUL", according to standard
+   Windows conventions.  "/dev/null" remains correct on Linux. (This feature can be used to test
+   throughput or check MD5s without saving the downloaded data.) 
+1. The file format of the (still undocmented) `--list-of-files` parameter is changed.  (It remains
```
zezha-msft: typo: undocumented

ChangeLog.md
```diff
@@ -1,11 +1,122 @@
...
+1. The `--overwrite` parameter now supports the value of "prompt" to prompt the user on a
+   file-by-file basis. (The old values of true and false are also supported.)   
+1. The environment variable `AZCOPY_CONCURRENCY_VALUE` can now be set to "AUTO". This is expected to be
+    useful for customers with small networks, or those running AzCopy on
+   moderately-powered machines and transfer blobs between accounts.  This feature is in preview status.
+1. When uploading from Windows, files can be filtered by Windows-specific file attributes (such as
+   "Archive", "Hidden" etc)
+1. Memory usange can be controlled by setting the new environment variable `AZCOPY_BUFFER_GB`.
+   Decimal values are supported. Actual usage will be the value specified, plus some overhead. 
+1. An extra integrity check has been added for Service to Service transfers: the length of the
```
zezha-msft: update to clarify it works for upload/download too

ChangeLog.md
JohnRusk: Good point. Generally, its better if the app can warn folks about changes, rather than relying on them to read the docs. (Because folks are notoriously casual about reading technical docs).   I've just added a PR to implement such a warning.

Also, I've already put a small update into the wiki page: https://github.com/Azure/azure-storage-azcopy/wiki/Listing-specific-files-to-transfer

ChangeLog.md
JohnRusk: Done. THanks. BTW, that's only true on Windows. We use OS keyring on other OS's, right?  I hope that's right, because that's how I just worded it ;-)

ChangeLog.md
JohnRusk: Probably not. Since when I created it I wasn't sure it was the right long term approach (eg. maybe a range is better) and I didn't realise there would be so many support incidents coming through that it could address.

JohnRusk: Am still not sure I want to got out of my way to encourage it - at least not until its had a bit more bake time.

ChangeLog.md
JohnRusk: Just Windows I believe.

ChangeLog.md
JohnRusk: I think the container issue is covered by a backlog item. Search for "Make S2S bucket name resolution more robust from S3"

ChangeLog.md
adreed-msft: Good stuff.

---
In reply to: [333358972](https://github.com/Azure/azure-storage-azcopy/pull/607#discussion_r333358972) [](ancestors = 333358972)

PR #608, opened 2019-09-23
Fixed remove bug when blob has snapshots

zezha-msft:
The errors resulted from calling delete were not being handled correctly, causing the tool to hang indefinitely. 

My follow-up with StgExp: should a flag be added to specify what to do with snapshots

### DIFF
--- a/ste/xfer-deleteBlob.go
+++ b/ste/xfer-deleteBlob.go
@@ -43,12 +43,14 @@ func DeleteBlobPrologue(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pac
 		jptm.ReportTransferDone()
 	}
 
-	_, err := srcBlobURL.Delete(jptm.Context(), azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
+	// TODO confirm whether we should add a new flag to allow user to specify whether snapshots should be removed
+	_, err := srcBlobURL.Delete(jptm.Context(), azblob.DeleteSnapshotsOptionInclude, azblob.BlobAccessConditions{})
 	if err != nil {
 		// If the delete failed with err 404, i.e resource not found, then mark the transfer as success.
 		if strErr, ok := err.(azblob.StorageError); ok {
 			if strErr.Response().StatusCode == http.StatusNotFound {
 				transferDone(common.ETransferStatus.Success(), nil)
+				return
 			}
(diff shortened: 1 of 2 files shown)


ste/xfer-deleteBlob.go
```diff
@@ -43,12 +43,14 @@ func DeleteBlobPrologue(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pac
 		jptm.ReportTransferDone()
 	}
 
-	_, err := srcBlobURL.Delete(jptm.Context(), azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
+	// TODO confirm whether we should add a new flag to allow user to specify whether snapshots should be removed
```
adreed-msft: Hey wait, didn't we have a hidden flag (--with-snapshots) that never worked?

I removed that in another (un-merged) PR, but I could undo the removal of it.

ste/xfer-deleteBlob.go
```diff
@@ -43,12 +43,14 @@ func DeleteBlobPrologue(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pac
...
+	_, err := srcBlobURL.Delete(jptm.Context(), azblob.DeleteSnapshotsOptionInclude, azblob.BlobAccessConditions{})
 	if err != nil {
 		// If the delete failed with err 404, i.e resource not found, then mark the transfer as success.
 		if strErr, ok := err.(azblob.StorageError); ok {
 			if strErr.Response().StatusCode == http.StatusNotFound {
 				transferDone(common.ETransferStatus.Success(), nil)
+				return
```
adreed-msft: Consider logging a warning that sometimes 404s can be caused by failed authentication.

ste/xfer-deleteBlob.go
zezha-msft: That's fine for now, I'm trying to clarify with them whether they want the snapshot options.

ste/xfer-deleteBlob.go
zezha-msft: 404 is only given if no SAS/Oauth token is present. Either one of them had to be present for us to scan the resource.

ste/xfer-deleteBlob.go
JohnRusk: I'll pop a small note into the changelog about this too. You'll see it when you review that PR. (Which I've just added you to as a reviewer).

PR #615, opened 2019-09-27
Refactor/include exclude multi level traversal

JohnRusk:


cmd/zc_traverser_file.go
```diff
@@ -118,12 +119,17 @@ func (t *fileTraverser) traverse(processor objectProcessor, filters []objectFilt
 				relativePath = strings.TrimPrefix(relativePath, common.AZCOPY_PATH_SEPARATOR_STRING)
 
 				// We need to omit some properties if we don't get properties
+				// TODO: make it so we can (and must) call newStoredOBject here.
+				//     Reviewers: don't pass this if I have not fixed this when you review!!!
```
JohnRusk: On second thoughts. Let's not worry about calling newStoredObject in THIS PR.  If/when we fix that, let's do it in a different PR

cmd/copyUtil.go
```diff
@@ -171,187 +162,11 @@ func (copyHandlerUtil) generateObjectPath(destinationPath, fileName string) stri
 	return fmt.Sprintf("%s/%s", destinationPath, fileName)
 }
 
-// resourceShouldBeIncluded decides whether the file at given path should be included or not
```
adreed-msft: Happy to see the fat cut from here. Let's try to find a better place for these other guys soon

cmd/zc_enumerator.go
```diff
@@ -118,10 +125,10 @@ func containerNameMatchesPattern(containerName, pattern string) (bool, error) {
 	return filepath.Match(pattern, containerName)
 }
 
-func initContainerDecorator(containerName string, processor objectProcessor) objectProcessor {
-	return func(object storedObject) error {
+// newContainerDecorator constructs an objectMorpher that adds the given container name to storedObjects
+func newContainerDecorator(containerName string) objectMorpher {
```
zezha-msft: >newContainerDecorator [](start = 5, length = 21)

if we have more preprocessors to add in the future, we just chain these?

cmd/zc_enumerator.go
```diff
@@ -85,8 +85,8 @@ func (storedObject *storedObject) isMoreRecentThan(storedObject2 storedObject) b
 }
 
 // a constructor is used so that in case the storedObject has to change, the callers would get a compilation error
-func newStoredObject(name string, relativePath string, lmt time.Time, size int64, md5 []byte, blobType azblob.BlobType, containerName string) storedObject {
-	return storedObject{
+func newStoredObject(morpher objectMorpher, name string, relativePath string, lmt time.Time, size int64, md5 []byte, blobType azblob.BlobType, containerName string) storedObject {
```
adreed-msft: Suggestion: Call it objectPreProcessor instead of objectMorpher

I feel as though either name is adequate, but preProcessor is the technical term you've used to refer to it in the past

cmd/zc_traverser_list.go
```diff
@@ -73,15 +73,12 @@ func (l *listTraverser) traverse(processor objectProcessor, filters []objectFilt
 		// case 2: child2 is a directory, and it has items under it such as child2/grandchild1
 		//         the relative path returned by the child traverser would be "grandchild1"
 		//         it should be "child2/grandchild1" instead
-		preProcessor := func(object storedObject) error {
+		childPreProcessor := func(object *storedObject) {
 			object.relativePath = common.GenerateFullPath(childPath, object.relativePath)
-			return processor(object)
 		}
+		preProcessorForThisChild := preprocessor.FollowedBy(childPreProcessor) // always chain pre-processors, never replace them (this is so we avoid making any assumptions about whether an old processor actually exists)
```
adreed-msft: I'm not entirely sure how I feel about wrapping function after function in generated functions. This may well actually end up making debugging harder. In my experience, the debugger generally won't jump into them unless I add an explicit breakpoint within the generated functions. The problem is, I'm also not entirely sure what the best plan of action is to remedy this.

adreed-msft: >// always chain pre-processors, never replace them (this is so we avoid making any assumptions about whether an old processor actually exists) [](start = 73, length = 142)

Consider moving this comment to .FollowedBy (or adding it wherever .FollowedBy is used)

cmd/zc_enumerator.go
JohnRusk: True. It is inconsistent.  BTW I almost made it consistent in the other direction: "morpher" everywhere, to try to convey the idea that it's _only_ job is to modify the storedObject; not to do anything outside that object.  In the interests of time, I'm going to leave the name as-is for now.

cmd/zc_traverser_list.go
JohnRusk: I think the other way to do it would be to define objectMorpher as an interface, not a func, and to create objects that implemented that interface. E.g. a "container setter" and a "path prepender". And have newStoredObject take an array of objectMorphers, and FollowedBy just append to the array.

I think that would all work, and the code would mostly look about the same. And as you say, debuggability would be improved.

I'll put a TODO in the code to say that we should consider that.

JohnRusk: Good idea. Have moved it.

PR #616, opened 2019-09-27
Infer strip-top-dir by trailing wildcard

adreed-msft:
Introduces a generic URLParts for all resources. It's intended to give us just what we need to reduce code duplication for now, and includes a general SDK suggestion for the go devex.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -168,6 +168,55 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	return raw.cookWithId(common.NewJobID())
 }
 
+// returns result of stripping and if striptopdir is enabled
+// if nothing happens, the original source is returned
+func (raw rawCopyCmdArgs) stripTrailingWildcardOnRemoteSource(location common.Location) (result string, stripTopDir bool, err error) {
+	result = raw.src
+	// Because local already handles wildcards via a list traverser, we should only handle the trailing wildcard --strip-top-dir inference remotely.
+	// To avoid getting trapped by parsing a URL and losing a sense of which *s are real, strip the SAS token in a """unsafe""" way.
+	splitURL := strings.Split(result, "?")
+
+	// replace %2A with %00 (NULL).
+	// Azure storage doesn't support NULL, so nobody has any reason to ever do this
+	// Thus, %00 is our magic number. Understandably, this is an exception to how we handle wildcards, but this isn't a user-facing exception
+	splitURL[0] = strings.ReplaceAll(splitURL[0], "%2A", "%00")
+
+	sourceURL, err := url.Parse(splitURL[0])
+
+	if err != nil {
+		err = fmt.Errorf("failed to encode %s as URL; %s", strings.ReplaceAll(splitURL[0], "%00", "%2A"), err)
+		return
+	}
+
+	// Catch trailing wildcard in object name
+	// Ignore wildcard in container name, as that is handled by initResourceTraverser -> AccountTraverser
+	genericResourceURLParts := common.NewGenericResourceURLParts(*sourceURL, location)
+
+	if cName := genericResourceURLParts.GetContainerName(); (strings.Contains(cName, "*") || cName == "") && genericResourceURLParts.GetObjectName() != "" {
+		err = errors.New("cannot combine a specific object name with an account-level search")
+		return
+	}
+
+	// Infer stripTopDir, trim suffix so we can traverse properly
+	if strings.HasSuffix(genericResourceURLParts.GetObjectName(), "/*") || genericResourceURLParts.GetObjectName() == "*" {
+		genericResourceURLParts.SetObjectName(strings.TrimSuffix(genericResourceURLParts.GetObjectName(), "*"))
+		stripTopDir = true
+	}
+
+	// Check for other *s, error out and explain the usage
+	if strings.Contains(genericResourceURLParts.GetObjectName(), "*") {
+		err = errors.New("cannot use wildcards in the path section of the URL except in trailing \"/*\". If you wish to use * in your URL, manually encode it to %2A")
+		return
+	}
+
+	splitURL[0] = strings.ReplaceAll(genericResourceURLParts.String(), "%00", "%2A")
+	// drop URL back to string and replace our magic number
+	// re-combine underlying string
+	result = strings.Join(splitURL, "?")
+
+	return
+}
+
 func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, error) {
 
 	cooked := cookedCopyCmdArgs{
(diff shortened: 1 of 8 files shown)


cmd/copy.go
```diff
@@ -184,10 +184,50 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 
 	cooked.fromTo = fromTo
 
-	// copy&transform flags to type-safety
 	cooked.stripTopDir = raw.stripTopDir
+
+	// Check if source has a trailing wildcard on a URL
+	if fromTo.From().IsRemote() {
+		// Because local already handles wildcards via a list traverser, we should only handle the trailing wildcard --strip-top-dir inference remotely.
```
JohnRusk: Minor: I'd suggest this section should be moved into a function, for two reasons: a readability of cookWithId method and, perhaps more importantly, that will make it unit testable in isolation.

adreed-msft: Probably a good idea to separate, yeah.

I think I'll drop it to a function of cca that returns the resulting source and if strip top dir is enabled

cmd/copy.go
```diff
@@ -184,10 +183,48 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
...
+
+		// Infer stripTopDir, trim suffix so we can traverse properly
+		if strings.HasSuffix(genericResourceURLParts.GetObjectName(), "/*") || genericResourceURLParts.GetObjectName() == "*" {
+			genericResourceURLParts.SetObjectName(strings.TrimSuffix(genericResourceURLParts.GetObjectName(), "*"))
+			cooked.stripTopDir = true
+		}
+
+		// Check for other *s, error out and explain the usage
+		if strings.Contains(genericResourceURLParts.GetObjectName(), "*") {
+			return cooked, errors.New("cannot use wildcards in URL except in trailing \"/*\". If you wish to use * in your URL, manually encode it to %2A")
```
JohnRusk: Maybe change to "cannot use wildcards in _path section of_ URL except in trailing /*".  Just to avoid confusing any pedantic folks who see and, and who know we do support it in the container section.

cmd/copy.go
```diff
@@ -184,10 +183,48 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
...
+		}
+
+		// drop URL back to string and replace our magic number
```
JohnRusk: if someone gives us a wildcard that has a * in the container AND a path section (we don't allow that combinaton) does something somewhere else in the code detect that and give an error? If not, we should give one here perhaps?

cmd/copy.go
```diff
@@ -184,10 +183,48 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
...
+		genericResourceURLParts := common.NewGenericResourceURLParts(*sourceURL, fromTo.From())
+
+		// Infer stripTopDir, trim suffix so we can traverse properly
+		if strings.HasSuffix(genericResourceURLParts.GetObjectName(), "/*") || genericResourceURLParts.GetObjectName() == "*" {
+			genericResourceURLParts.SetObjectName(strings.TrimSuffix(genericResourceURLParts.GetObjectName(), "*"))
+			cooked.stripTopDir = true
+		}
+
+		// Check for other *s, error out and explain the usage
+		if strings.Contains(genericResourceURLParts.GetObjectName(), "*") {
```
JohnRusk: You mentioned that url.Parse does something with escaping, but I can't recall which way around it work. Does parse change %2A to *, which is why you had to get rid of the %2A temporarily?

cmd/helpMessages.go
```diff
@@ -76,11 +76,16 @@ Download a single file with a SAS using piping (block blobs only):
 Download an entire directory with a SAS:
   - azcopy cp "https://[account].blob.core.windows.net/[container]/[path/to/directory]?[SAS]" "/path/to/dir" --recursive=true
 
+A note about wildcards in URLs:
+
+The only usage of a wildcard in a URL that is supported is the final, trailing /*, and in the container name when not specifying a blob name.
+If you happen to use the character * in the name of a blob, please manually encode it to %2A to avoid it being treated as a wildcard character.
+
 Download a set of files with a SAS using wildcards:
```
JohnRusk: I think we should remove these two "with wildcards" examples, since they are not accomplishing the original purpose (of showing how to use wildcards in URLs) because that original purprse is no longe supported.

JohnRusk: Maybe we could change the previous example instead. I can't describe it in a comment (too difficult), so I'll push a small change  to the PR ;-)

cmd/copy.go
adreed-msft: This does get detected further down the line in job source root setup.

adreed-msft: But it probably wouldn't hurt to check it here too

cmd/copy.go
adreed-msft: Parse changes %2A to *, meaning that we don't know the difference between %2A raw and * raw.

Therefore, I replace %2A with %00 (NULL) to walk around this, since NULL is invalid in Azure Storage.

cmd/helpMessages.go
adreed-msft: I think the best way we could phrase it is
"To put the contents of a folder directly into the destination rather than under a sub-directory, append a trailing /* to the URL."

JohnRusk: Oh wow, that's perfect!  Way better than what I had in mind, so I won't push anything.

cmd/copy.go
```diff
@@ -184,10 +183,48 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
...
 
-	// copy&transform flags to type-safety
-	cooked.stripTopDir = raw.stripTopDir
+	// Check if source has a trailing wildcard on a URL
+	if fromTo.From().IsRemote() {
+		// Because local already handles wildcards via a list traverser, we should only handle the trailing wildcard --strip-top-dir inference remotely.
+		// To avoid getting trapped by parsing a URL and losing a sense of which *s are real, strip the SAS token in a """unsafe""" way.
+		splitURL := strings.Split(cooked.source, "?")
+
+		// replace %2A with %00 (NULL).
```
zezha-msft: >//  [](start = 2, length = 3)

this part came off strange at first read, please consider adding a comment to clarify that the %2As are being replaced so that they don't get turned into * when we look at the parsed url

cmd/zt_copy_blob_download_test.go
```diff
@@ -31,6 +31,132 @@ import (
...
 )
 
+func (s *cmdIntegrationSuite) TestInferredStripTopDirDownload(c *chk.C) {
+	bsu := getBSU()
+	cURL, cName := createNewContainer(c, bsu)
+
+	blobNames := []string{
+		"*", // File name that we want to retain compatibility with
+		"testFile",
+		"DoYouPronounceItDataOrData",
```
zezha-msft: >DoYouPronounceItDataOrData [](start = 3, length = 26)

omg LOL

cmd/zt_copy_blob_download_test.go
```diff
@@ -31,6 +31,132 @@ import (
...
+	cooked, err = raw.cook()
+	c.Assert(err, chk.IsNil)
+	c.Assert(cooked.stripTopDir, chk.Equals, true)
+
+	// Test and ensure only one file got scheduled
+	runCopyAndVerify(c, raw, func(err error) {
+		c.Assert(err, chk.IsNil)
+
+		c.Assert(len(mockedRPC.transfers), chk.Equals, 1)
+	})
```
zezha-msft: nice test! I like

common/genericResourceURLParts.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/Azure/azure-storage-file-go/azfile"
+
+	"github.com/Azure/azure-storage-azcopy/azbfs"
+)
+
+// GenericResourceURLParts is intended to be a generic solution to code duplication when using *URLParts
+// TODO: Consider returning to a "common" package structure for our SDKs
+//       (or, if we don't want to rely on a common package, create one that unifies URLParts and such into interfaces)
+//       The experience of trying to avoid code dupe is frankly quite poor and tends to result in more code dupe.
```
zezha-msft: common package is very problematic and will be avoided in the future. the files starting with 'zc' are the common files that are shared

cmd/copy.go
adreed-msft: Good point-- it says what it does but not why it does it.

PR #618, opened 2019-09-30
Log response body in error message

JohnRusk:
Somehow the previous PR for this got lost. (I closed it before it was approved and merged. I don't know why!)

Ze, you already approved it, so not sure if you need/want to again.
Adele, I don't think you saw the old one.


JohnRusk:
Old PR was https://github.com/Azure/azure-storage-azcopy/pull/589

PR #619, opened 2019-09-30
Prevent remove from (sometimes) crashing when there's nothing to remove

JohnRusk:


cmd/removeProcessor.go
```diff
@@ -40,7 +40,11 @@ func newRemoveTransferProcessor(cca *cookedCopyCmdArgs, numOfTransfersPerPart in
 		LogLevel: cca.logVerbosity,
 	}
 
-	reportFirstPart := func() { cca.waitUntilJobCompletion(false) }
+	reportFirstPart := func(jobStarted bool) {
+		if jobStarted {
+			cca.waitUntilJobCompletion(false)
```
JohnRusk: It's a bit ugly that we have three different ways of starting progress reporting.

* "Remove" does it here (and now does it _only_ if there's actually something to remove)
* "Sync" does it all the time (and, unlike Copy and Remove, avoids job lookup errors if there was nothing to remove, by testing a flag on the cooked sync args in its progress reporting routine.  Copy (and therefore remove) have no such flag)
* "Copy" uses code in copyEnumeratorInit which only starts progress reporting if it knows that there is work to do.

I've decided that we probably just have to leave all three of the above in place for now. There's already a TODO on Remove to refactor it to use the copy approach at some time in the future.

zezha-msft: Sounds good.

adreed-msft: Sounds good to me.

adreed-msft: Sounds good to me.

PR #624, opened 2019-09-30
Fixed unused import

JohnRusk:


zezha-msft:
@JohnRusk Thanks for fixing it! I was about to submit a commit.

PR #625, opened 2019-10-01
Added support for File -> File copy

zezha-msft:


cmd/zt_copy_file_file_test.go
```diff
@@ -0,0 +1,251 @@
...
+	defer deleteShare(c, dstShareURL)
+
+	// set up the source share with numerous files
+	fileList := scenarioHelper{}.generateCommonRemoteScenarioForAzureFile(c, srcShareURL, "")
+	c.Assert(len(fileList), chk.Not(chk.Equals), 0)
+
+	// add special files that we wish to exclude
+	filesToExclude := []string{"notGood.pdf", "excludeSub/lame.jpeg", "exactName"}
+	scenarioHelper{}.generateAzureFilesFromList(c, srcShareURL, filesToExclude)
+	excludeString := "*.pdf;*.jpeg;exactName"
```
adreed-msft: Ensure that generateCommonRemoteScenarioForAzureFile() doesn't end up generating *.pdf or *.jpeg as we don't do any manual filtering of the file list in this test

cmd/zt_copy_file_file_test.go
```diff
@@ -0,0 +1,251 @@
...
+	defer deleteShare(c, dstShareURL)
+
+	// set up the source share with numerous files
+	fileList := scenarioHelper{}.generateCommonRemoteScenarioForAzureFile(c, srcShareURL, "")
+	c.Assert(len(fileList), chk.Not(chk.Equals), 0)
+
+	// add special files that we wish to include
+	filesToInclude := []string{"important.pdf", "includeSub/amazing.jpeg"}
+	scenarioHelper{}.generateAzureFilesFromList(c, srcShareURL, filesToInclude)
+	includeString := "*.pdf;*.jpeg;exactName"
```
adreed-msft: ditto

cmd/zt_copy_file_file_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+	// add special files that we wish to exclude
+	// note that the excluded files also match the include string
+	filesToExclude := []string{"sorry.pdf", "exclude/notGood.jpeg", "exactName", "sub/exactName"}
+	scenarioHelper{}.generateAzureFilesFromList(c, srcShareURL, filesToExclude)
+	excludeString := "so*;not*;exactName"
```
adreed-msft: ditto

testSuite/scripts/test_file_copy.py
```diff
@@ -0,0 +1,65 @@
+import os
+import unittest
+
+import utility as util
+
+
+class File_Copy_User_Scenario(unittest.TestCase):
```
adreed-msft: Minor: Perhaps name this File_Service_2_Service_Copy_User_Scenario

testSuite/scripts/test_file_copy.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # verifying the copy worked, both remote source and destination should be identical to the local source
+        result = util.Command("testFile").add_arguments(content_file_path_src).add_arguments(
+            remote_src_file_path).execute_azcopy_verify()
+        self.assertTrue(result)
+        result = util.Command("testFile").add_arguments(content_file_path_src).add_arguments(
+            remote_dst_file_path).execute_azcopy_verify()
+        self.assertTrue(result)
+
+    def test_copy_entire_dir_to_dir(self):
```
adreed-msft: Consider adding a test using the trailing wildcard atop directory to directory

testSuite/scripts/test_file_copy.py
adreed-msft: (OR: Move these tests into service_2_service_copy_user_scenario)

cmd/copy.go
```diff
@@ -490,21 +490,21 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 		}
 	case common.EFromTo.BlobBlob(),
 		common.EFromTo.FileBlob(),
+		common.EFromTo.FileFile(),
```
adreed-msft: Glad that the refactor is making adding new locations and directions an absolute cinch!

cmd/copy.go
```diff
@@ -490,21 +490,21 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
...
 		if cooked.preserveLastModifiedTime {
 			return cooked, fmt.Errorf("preserve-last-modified-time is not supported while copying from service to service")
 		}
 		if cooked.followSymlinks {
 			return cooked, fmt.Errorf("follow-symlinks flag is not supported while copying from service to service")
 		}
+		// blob type is not supported if we are not doing blob -> blob
+		if cooked.blobType != common.EBlobType.Detect() && cooked.fromTo != common.EFromTo.BlobBlob() {
+			return cooked, fmt.Errorf("blob-type is not supported for the scenario (%s)", cooked.fromTo.String())
+		}
```
JohnRusk: This seems odd. I thought we supported the following:
* Blob - > Blob, all blobType parameter values are valid, and Detect means "Detect the type of the source blob"
* anything else -> Blob, all blobType parameter values are valid, and Detect means "Look for VHD/VHDX extension"
* anything -> NOT Blob, _only_ Detect is valid, and it means "No blob type specified"

I think the code here is incorrectly treating the last two cases, above, as being the same.

JohnRusk: Oh, I see that you've based this code on some code removed from a few lines below.  I believe that code was incorrect. We actually had a bug logged about it recently (via internal email).

cmd/zt_copy_file_file_test.go
zezha-msft: Yes, thanks for the reminder

testSuite/scripts/test_file_copy.py
zezha-msft: Updated the entire_dir_to_dir to use /* instead of --strip-top-dir which was being used.

cmd/copy.go
zezha-msft: I see, I don't recall seeing the bug report, but what you are saying makes sense to me.

The existing code gave me the impression that S3/File -> Blob can only be Block Blobs.

PR #630, opened 2019-10-02
Fix/ci fixes oct

JohnRusk:
Fixes busted CI

### DIFF
--- a/cmd/copyEnumeratorInit.go
+++ b/cmd/copyEnumeratorInit.go
@@ -270,7 +270,7 @@ func (cca *cookedCopyCmdArgs) initModularFilters() []objectFilter {
 	filters := make([]objectFilter, 0) // same as []objectFilter{} under the hood
 
 	if len(cca.includePatterns) != 0 {
-		filters = append(filters, &includeFilter{patterns: cca.includePatterns})
+		filters = append(filters, &includeFilter{patterns: cca.includePatterns}) // TODO should this call buildIncludeFilters?
 	}
 
 	if len(cca.excludePatterns) != 0 {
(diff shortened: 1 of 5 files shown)


cmd/copyEnumeratorInit.go
```diff
@@ -270,7 +270,7 @@ func (cca *cookedCopyCmdArgs) initModularFilters() []objectFilter {
 	filters := make([]objectFilter, 0) // same as []objectFilter{} under the hood
 
 	if len(cca.includePatterns) != 0 {
-		filters = append(filters, &includeFilter{patterns: cca.includePatterns})
+		filters = append(filters, &includeFilter{patterns: cca.includePatterns}) // TODO should this call buildIncludeFilters?
```
adreed-msft: The function? It has exclude filters as well.

cmd/zt_copy_blob_download_test.go
```diff
@@ -422,8 +422,10 @@ func (s *cmdIntegrationSuite) TestDownloadBlobContainerWithPattern(c *chk.C) {
 
 	// construct the raw input to simulate user input
 	rawContainerURLWithSAS := scenarioHelper{}.getRawContainerURLWithSAS(c, containerName)
-	rawContainerURLWithSAS.Path += "/*" // imply strip-top-dir
-	raw := getDefaultCopyRawInput(rawContainerURLWithSAS.String(), dstDirName)
+	urlString := rawContainerURLWithSAS.String()
+	// imply strip-top-dir (without the /* getting escaped by the line above)
+	urlString = strings.Replace(urlString, "?", "/*?", -1) // TODO: find a better way to do this in our tests
```
adreed-msft: >"/*?" [](start = 45, length = 5)

I'm sure this works, but our rawContainerURLWithSAS will never end with a /, right? If it does, we'd accidentally end up traversing account.blob.core.windows.net/container//*?SAS

JohnRusk: For now, given the way rawContainerURLWithSAS works, we're fine. I'll let the existing TODO remind us to find something better.

JohnRusk: Will leave it covered by the existing TODO for now.  Agree we should make it more robust against change in the generantion of the URL.

PR #635, opened 2019-10-02
Add blob -> file testing

adreed-msft:


cmd/copyEnumeratorInit.go
```diff
@@ -88,6 +89,16 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
 
+	// Disallow list-of-files and include-path on service-level traversal due to a major bug
+	// TODO: Fix the bug.
+	//       Two primary issues exist with the list-of-files implementation:
+	//       1) Account name doesn't get trimmed from the path
+	//       2) List-of-files is not considered an account traverser; therefore containers don't get made.
+	//       Resolve these two issues and service-level list-of-files/include-path will work
+	if cca.listOfFilesChannel != nil && srcLevel == ELocationLevel.Service() {
+		return nil, errors.New("cannot combine list-of-files or include-path with account traversal")
+	}
```
JohnRusk: 👍

cmd/copyEnumeratorInit.go
```diff
@@ -361,6 +374,41 @@ func (cca *cookedCopyCmdArgs) createDstContainer(containerName, dstWithSAS strin
...
+
+		fsu := azfile.NewServiceURL(*dstURL, dstPipeline)
+		shareURL := fsu.NewShareURL(containerName)
+		_, err = shareURL.GetProperties(ctx)
+
+		if err == nil {
+			return err
+		}
+
+		// Create a destination share with the default service quota
```
JohnRusk: Do we know what the default is?  I couldn't it documented online when I had a quick look just now.

adreed-msft: Actually completely unsure.

JohnRusk: Do you have one lying around from when you tested it?  Could jump into Portal and see the size there.

adreed-msft: It seems as though it just defaults to 5TiB

JohnRusk: Perfect, that's the current maximum.  Defaulting to the max is the safest in terms of a job running successfully, so that's all good.  I'll mark this conversation resolved now.

cmd/zt_copy_blob_file_test.go
```diff
@@ -0,0 +1,439 @@
...
+		c.Assert(err, chk.IsNil)
+
+		c.Assert(len(mockedRPC.transfers), chk.Equals, len(expectedTransfers))
+
+		validateS2STransfersAreScheduled(c, "/", "/", expectedTransfers, mockedRPC)
+	})
+}
+
+// TestBlobCopyToFileS2SImplicitDstShare uses a service-level URL on the destination to implicitly create the destination share.
+func (s *cmdIntegrationSuite) TestBlobCopyToFileS2SImplicitDstShare(c *chk.C) {
```
zezha-msft: >cmdIntegrationSuite) T [](start = 9, length = 22)

nice tests!

PR #638, opened 2019-10-03
Prevent collisions of actively-open destination filenames

JohnRusk:


JohnRusk:
NOT tested yet.  Is here to illustrate what we'd do if we wanted to fix this.

JohnRusk:
This turned out to need some changes to xfer-RemoteToLocal so that it could cover zero-length files (e.g. a zero-length one conflicts with a non-zero one). So that resulted in more change than I would have liked, but I can't think of any less impactful way (and the new code is actually simpler than the old, IMHO)

common/exclusiveStringMap.go
```diff
@@ -0,0 +1,82 @@
...
+	if e.caseSensitive {
+		return s
+	} else {
+		// technically, toLower is a bad way to make something case insensitive,
+		// but, in our case, the collisions we are concerned about are
+		// ones that are caused due to US (our code) mutating characters that fall within the ASCII
+		// section of the code space (specifically escaping or unescaping special characters not
+		// supported by Windows or Azure Files); rather than general case sensitivity issues in the broader sense.
+		// So we think this is OK.
+		// TODO: double check, is it really? Should we replace by going through rune-by-rune converting to unicode "fold case"?
```
zezha-msft: > TODO: double check, is it really? [](start = 4, length = 34)

hmmm, I also feel like it's ok 

[https://golang.org/pkg/strings/#ToLower](https://golang.org/pkg/strings/#ToLower) says each unicode letter is mapped to their lower case, so I guess we are fine

ste/mgr-JobPartTransferMgr.go
```diff
@@ -270,6 +275,67 @@ func (jptm *jobPartTransferMgr) FileCountLimiter() common.CacheLimiter {
 	return jptm.jobPartMgr.FileCountLimiter()
 }
 
+// WaitUntilLockDestination does two things. It respects any limit that may be in place on the number of
+// active destination files (by blocking until we are under the max count), and it
+// registers the destination as "locked" in our internal map. The reason we
+// lock internally like this is:
+// (a) it is desirable to have some kind of locking because there are (very rare) edge cases where
+// we may map two source files to one destination.  This can happen any time we mutate the destination name
```
zezha-msft: >we may map two source files to one destination. [](start = 3, length = 47)

Isn't it more common than we think? doesn't windows filesystem ignore casing, so it's possible that many blobs map to the same local file?

ste/xfer-anyToRemote.go
```diff
@@ -141,6 +141,18 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
 		}
 	}
 
+	// step 5a: lock the destination
+	// (is safe to do it relatively early here, before we run the prologue, because its just a internal lock, within the app)
+	// But must be after all of the early returns that are above here (since
+	// if we succeed here, we need to know the epilogue will definitely run to unlock later)
+	err = jptm.WaitUntilLockDestination(jptm.Context())
```
zezha-msft: >err = jptm.WaitUntilLockDestination(jptm.Context()) [](start = 1, length = 51)

I'm a bit confused here, why are we locking the destination when the destination is the remote?

The two scenarios (url encoding on windows and stripping .zip) we originally wanted to address are both for download, right?

JohnRusk: But we also do the exact same encoding stuff for uploads to Azure Files, since it has the same naming restrictions as Windows.

ste/mgr-JobPartTransferMgr.go
JohnRusk: Oh, yes. That too!!!  Technically, the

common/exclusiveStringMap.go
JohnRusk: It's not quite that simple, unfortunately, in some cases.   E.g. where there are muliple ways to represent one thing.  Maybe we should change it to work through rune-by-rune, but I didn't want to over complicate things....

ste/xfer-anyToRemote.go
JohnRusk: BTW, your comment made be realise that I'm not setting the case-sensitivity right.  Will need to think about that.

JohnRusk: OK. Have just pushed changed to correct the bug, it now takes both fromTo and OS into account when setting case sensitivity

common/exclusiveStringMap.go
JohnRusk: Technically there are cases where the number of characters in the lower and uppercase form of a "thing" are not the same.  E.g. German ß and SS are considered equal in some case insensitive comparisons…. But, not by Windows. Even with German set as the OS language, it treats ß.txt and SS.txt as separate files in a test I ran just now.

It also let me have files named İ.txt and i.txt even though, in Turkish, those differ only by case. (I didn't actually try setting Turkish as the Windows language, for fear of getting lost)

So, on reflection, I think I basically agree with you.  I feel that what we are doing with lowercasing is probably as close as we can realistically get to what the OS will do. (And that's what we need to match, not what native speakers of a language would do. This is all about how we interact with the OS - so if OS and humans disagree we should align, in this case, to the OS).

common/exclusiveStringMap.go
```diff
@@ -0,0 +1,88 @@
...
+	"sync"
+)
+
+type ExclusiveStringMap struct {
+	lock          *sync.Mutex
+	m             map[string]struct{}
+	caseSensitive bool
+}
+
+func NewExclusiveStringMap(fromTo FromTo, goos string) *ExclusiveStringMap {
```
adreed-msft: >goos string [](start = 42, length = 11)

On one hand, this feels like a really odd choice. On the other hand, I understand that it's for testing, and we need to work around the fact that runtime.GOOS is a const not a var.

WRT that, consider doing strings.EqualFold for goos == "windows"/"darwin", as we could oddly capitalize it in tests by accident.

common/exclusiveStringMap.go
adreed-msft: Does Go convert ß to SS in this scenario? If so, that's kind of worrisome. We have two places this will cover-- Unsafe characters on Windows/Files, and decompression. There's obviously no way we're going to cover EVERY edge case in this context, if Go doesn't follow the OS's interpretation of upper/lowercase characters.

---
In reply to: [331329480](https://github.com/Azure/azure-storage-azcopy/pull/638#discussion_r331329480) [](ancestors = 331329480)

ste/mgr-JobMgr.go
```diff
@@ -167,6 +168,8 @@ type jobMgr struct {
 	cancel               context.CancelFunc
 	pipelineNetworkStats *pipelineNetworkStats
 
+	exclusiveDestinationMapHolder *atomic.Value
```
adreed-msft: >exclusiveDestinationMapHolder *atomic.Value [](start = 1, length = 43)

I don't entirely understand the usage of sync/atomic here. We already have a mutex in place inside the exclusive string map to prevent multiple routines messing with the map at the same time, which should in theory prevent the data race condition. What special use does an *atomic.Value hold here?

adreed-msft: Note that I (generally) understand the usage of this package, and that I'm only really questioning this specific use of it.

---
In reply to: [331610387](https://github.com/Azure/azure-storage-azcopy/pull/638#discussion_r331610387) [](ancestors = 331610387)

common/exclusiveStringMap.go
adreed-msft: Does Go follow the OS's interpretation of upper/lower case, or does it use its' own specific implementation?

If it uses its' own specific implementation, I can forsee a scenario in which the implementations collide.

There's an OK situation: The OS doesn't consider two characters to be upper/lowercase variants of eachother, but Go does. In this case, a file just errors out. No biggie.

Then there's a worrisome situation: Go doesn't consider two characters to be upper/lowercase variants of eachother, but the OS does. Then, we get our old corruption issue again.

ste/mgr-JobMgr.go
JohnRusk: I thought there were multiple goroutines that accessed this field - i.e. that needed to get the pointer to the exclusive string map.  So I wanted the getting of the pointer to be threadsafe. (As you say, once you have the pointer, using it to access the map is fine).

I've realized just now that all of the calls to AddJobPart probably happen on the same goroutine (the one doing the enumeration) and so what I did here was unnecessary (if we are willing to rely on that assumption), since is only at AddJobPart time that we mess with it.  So I probably didn't need this.  But I don't have time to remove it at the moment.

adreed-msft: Ah! I didn't think there'd be a case where getting the pointer to it _wouldn't_ be thread-safe, aside from the initial setting, I guess.

common/exclusiveStringMap.go
JohnRusk: I agree with your logic.  A possible solution is to write our own routine to go through each string rune by rune and turn it into a folded, comparable case.  (I think to be strictly correct you must do unicode normalization before that). 

But, I can't get that done for the release date.  I'll make a backlog item to evaluate in more detail once we have time.

common/exclusiveStringMap.go
JohnRusk: Yeah, it felt odd to me too, and I did it for exactly the reason you guessed.  I've followed your suggestion.  Can you take quick look to sanity check my commit please?

JohnRusk:
@zezha-msft Just waiting on your approval for final comments on this.

ste/mgr-JobPartTransferMgr.go
JohnRusk: I've update the comment accordingly.

ste/xfer-anyToRemote.go
zezha-msft: >WaitUntilLockDestination [](start = 12, length = 24)

do we want to avoid the overhead of locking if we are targeting blob/ADLS Gen2?

JohnRusk: Good question, and yes, those would be the only two times we can skip it. (All downloads need it, for the --decompress renamings, and all sending to Azure Files needs it).  How about we consider avoiding it in those cases in the next release? (Need to think about pros/cons of maintainabilyt and complexity of turning it on and off).  I've made a backlog item.

PR #639, opened 2019-10-03
Tighten up block list access

JohnRusk:
Old code was only locking the reference to the slice, not the contents. In theory it doesn't matter, because the code is in the epilogue so nothing else will access it at that time anyway. But this change makes that assumption explicit by making sure we'll be notified if the assumption is ever violated. (with an error dereferenceing the now-nil slice field)

### DIFF
--- a/ste/sender-blockBlob.go
+++ b/ste/sender-blockBlob.go
@@ -125,6 +125,7 @@ func (s *blockBlobSenderBase) Epilogue() {
 
 	s.muBlockIDs.Lock()
 	blockIDs := s.blockIDs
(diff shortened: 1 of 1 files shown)


JohnRusk:
And we know it could not have been incorrect before (i.e. no missing blocks), since service will reject the put block list call if any is still equal to "".  The error returned is 400 XML specified is not syntactically valid

JohnRusk:
> I'm just curious, what prompted you to make this change?

When we were looking at the cancellation improvements, I spotted this and thought, "hey, that's not technically correct".  (We're getting away with it so far, but we should fix it)

PR #641, opened 2019-10-03
Fault induction tests

adreed-msft:
Added flags (, a environment variable, and a build flag to enable them) to force a failure for things like md5 checks, LMT checks, and length checks. In order to induce the failure, we modify the property internally before handing it off to the actual check.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,6 +19,7 @@ jobs:
         inputs:
           version: '1.13'
       - script: |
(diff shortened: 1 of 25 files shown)


adreed-msft:
Rebased to ensure freshness

JohnRusk:
Not sure where this sits on our backlog....?  Marking it as 10.5 for now. Let me know if we have its scheduled earlier, and I just couldn't find it.

zezha-msft:
Need to be reworked due to all the changes that occurred.

PR #643, opened 2019-10-03
Final set of edits

normesta:
It apppears as though someone already removed the AzCopy delineators from each command example. I updated only help strings. I also compiled AzCopy to ensure that I did not introduce any errors. I pull from the latest "dev" branch, so this PR shouldn't introduce any conflicts.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -1320,8 +1320,8 @@ func init() {
 	cpCmd.PersistentFlags().BoolVar(&raw.s2sPreserveAccessTier, "s2s-preserve-access-tier", true, "Preserve access tier during service to service copy. "+
 		"Please refer to [Azure Blob storage: hot, cool, and archive access tiers](https://docs.microsoft.com/azure/storage/blobs/storage-blob-storage-tiers) to ensure destination storage account supports setting access tier. "+
 		"In the cases that setting access tier is not supported, please use s2sPreserveAccessTier=false to bypass copying access tier. (default true). ")
-	cpCmd.PersistentFlags().BoolVar(&raw.s2sSourceChangeValidation, "s2s-detect-source-changed", false, "Check if source has changed after enumerating. "+
-    cpCmd.PersistentFlags().StringVar(&raw.s2sInvalidMetadataHandleOption, "s2s-handle-invalid-metadata", common.DefaultInvalidMetadataHandleOption.String(), "Specifies how invalid metadata keys are handled. Available options: ExcludeIfInvalid, FailIfInvalid, RenameIfInvalid. (default 'ExcludeIfInvalid').")
+	cpCmd.PersistentFlags().BoolVar(&raw.s2sSourceChangeValidation, "s2s-detect-source-changed", false, "Check if source has changed after enumerating. ")
+	cpCmd.PersistentFlags().StringVar(&raw.s2sInvalidMetadataHandleOption, "s2s-handle-invalid-metadata", common.DefaultInvalidMetadataHandleOption.String(), "Specifies how invalid metadata keys are handled. Available options: ExcludeIfInvalid, FailIfInvalid, RenameIfInvalid. (default 'ExcludeIfInvalid').")
 
 	// s2sGetPropertiesInBackend is an optional flag for controlling whether S3 object's or Azure file's full properties are get during enumerating in frontend or
 	// right before transferring in ste(backend).
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -1319,7 +1319,8 @@ func init() {
 	// options change how the transfers are performed
 	cpCmd.PersistentFlags().Float64Var(&raw.blockSizeMB, "block-size-mb", 0, "Use this block size (specified in MiB) when uploading to Azure Storage, and downloading from Azure Storage. The default value is automatically calculated based on file size. Decimal fractions are allowed (For example: 0.25).")
 	cpCmd.PersistentFlags().StringVar(&raw.logVerbosity, "log-level", "INFO", "Define the log verbosity for the log file, available levels: INFO(all requests/responses), WARNING(slow responses), ERROR(only failed requests), and NONE(no output logs). (default 'INFO').")
-	cpCmd.PersistentFlags().StringVar(&raw.blobType, "blob-type", "Detect", "Defines the type of blob at the destination. This is used for uploading blobs and when copying between accounts (default 'None').")
+	cpCmd.PersistentFlags().StringVar(&raw.blobType, "blob-type", "Detect", "Defines the type of blob at the destination. This is used for uploading blobs and when copying between accounts (default 'Detect'). Valid values include 'Detect', 'BlockBlob', 'PageBlob', and 'AppendBlob'. "+
+		"When copying between accounts, a value of 'Detect' causes AzCopy to use the type of source blob to determine the type of the destination blob. When uploading a file, 'Detect' determines if the file is a VHD or a VHDX file based on the file extension. If the file is ether a VHD or VHDX file, AzCopy treats the file as a page blob.")
 	cpCmd.PersistentFlags().StringVar(&raw.blockBlobTier, "block-blob-tier", "None", "upload block blob to Azure Storage using this blob tier.")
 	cpCmd.PersistentFlags().StringVar(&raw.pageBlobTier, "page-blob-tier", "None", "Upload page blob to Azure Storage using this blob tier. (default 'None').")
 	cpCmd.PersistentFlags().StringVar(&raw.metadata, "metadata", "", "Upload to Azure Storage with these key-value pairs as metadata.")
@@ -1334,9 +1335,9 @@ func init() {
 	cpCmd.PersistentFlags().StringVar(&raw.md5ValidationOption, "check-md5", common.DefaultHashValidationOption.String(), "Specifies how strictly MD5 hashes should be validated when downloading. Only available when downloading. Available options: NoCheck, LogOnly, FailIfDifferent, FailIfDifferentOrMissing. (default 'FailIfDifferent')")
 	cpCmd.PersistentFlags().StringVar(&raw.includeFileAttributes, "include-attributes", "", "(Windows only) Include files whose attributes match the attribute list. For example: A;S;R")
 	cpCmd.PersistentFlags().StringVar(&raw.excludeFileAttributes, "exclude-attributes", "", "(Windows only) Exclude files whose attributes match the attribute list. For example: A;S;R")
-	cpCmd.PersistentFlags().BoolVar(&raw.CheckLength, "check-length", true, "Check the length of a file on the destination after the transfer. If there is a mismatch between source and destination, the transfer is cancelled.")
+	cpCmd.PersistentFlags().BoolVar(&raw.CheckLength, "check-length", true, "Check the length of a file on the destination after the transfer. If there is a mismatch between source and destination, the transfer is marked as failed.")
 	cpCmd.PersistentFlags().BoolVar(&raw.s2sPreserveProperties, "s2s-preserve-properties", true, "Preserve full properties during service to service copy. "+
-		"For AWS S3 and Azure File non-single file source, the list operation doesn't return full properties of objects amd files. To preserve full properties, AzCopy needs to send one additional request per object or file.")
+		"For AWS S3 and Azure File non-single file source, the list operation doesn't return full properties of objects and files. To preserve full properties, AzCopy needs to send one additional request per object or file.")
 	cpCmd.PersistentFlags().BoolVar(&raw.s2sPreserveAccessTier, "s2s-preserve-access-tier", true, "Preserve access tier during service to service copy. "+
 		"Please refer to [Azure Blob storage: hot, cool, and archive access tiers](https://docs.microsoft.com/azure/storage/blobs/storage-blob-storage-tiers) to ensure destination storage account supports setting access tier. "+
 		"In the cases that setting access tier is not supported, please use s2sPreserveAccessTier=false to bypass copying access tier. (default true). ")
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -1301,7 +1301,6 @@ func init() {
 	cpCmd.PersistentFlags().StringVar(&raw.logVerbosity, "log-level", "INFO", "Define the log verbosity for the log file, available levels: INFO(all requests/responses), WARNING(slow responses), ERROR(only failed requests), and NONE(no output logs). (default 'INFO').")
 	cpCmd.PersistentFlags().StringVar(&raw.blobType, "blob-type", "Detect", "Defines the type of blob at the destination. This is used for uploading blobs and when copying between accounts (default 'None').")
 	cpCmd.PersistentFlags().StringVar(&raw.blockBlobTier, "block-blob-tier", "None", "upload block blob to Azure Storage using this blob tier.")
-<<<<<<< HEAD
 	cpCmd.PersistentFlags().StringVar(&raw.pageBlobTier, "page-blob-tier", "None", "Upload page blob to Azure Storage using this blob tier. (default 'None').")
 	cpCmd.PersistentFlags().StringVar(&raw.metadata, "metadata", "", "Upload to Azure Storage with these key-value pairs as metadata.")
 	cpCmd.PersistentFlags().StringVar(&raw.contentType, "content-type", "", "Specifies the content type of the file. Implies no-guess-mime-type. Returned on download.")
@@ -1313,7 +1312,8 @@ func init() {
 	cpCmd.PersistentFlags().BoolVar(&raw.preserveLastModifiedTime, "preserve-last-modified-time", false, "Only available when destination is file system.")
 	cpCmd.PersistentFlags().BoolVar(&raw.putMd5, "put-md5", false, "Create an MD5 hash of each file, and save the hash as the Content-MD5 property of the destination blob or file. (By default the hash is NOT created.) Only available when uploading.")
 	cpCmd.PersistentFlags().StringVar(&raw.md5ValidationOption, "check-md5", common.DefaultHashValidationOption.String(), "Specifies how strictly MD5 hashes should be validated when downloading. Only available when downloading. Available options: NoCheck, LogOnly, FailIfDifferent, FailIfDifferentOrMissing. (default 'FailIfDifferent')")
-
+	cpCmd.PersistentFlags().StringVar(&raw.includeFileAttributes, "include-attributes", "", "(Windows only) Include files whose attributes match the attribute list. For example: A;S;R")
+	cpCmd.PersistentFlags().StringVar(&raw.excludeFileAttributes, "exclude-attributes", "", "(Windows only) Exclude files whose attributes match the attribute list. For example: A;S;R")
 	cpCmd.PersistentFlags().BoolVar(&raw.CheckLength, "check-length", true, "Check the length of a file on the destination after the transfer. If there is a mismatch between source and destination, the transfer is cancelled.")
 	cpCmd.PersistentFlags().BoolVar(&raw.s2sPreserveProperties, "s2s-preserve-properties", true, "Preserve full properties during service to service copy. "+
 		"For AWS S3 and Azure File non-single file source, the list operation doesn't return full properties of objects amd files. To preserve full properties, AzCopy needs to send one additional request per object or file.")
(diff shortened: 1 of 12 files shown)


adreed-msft:
Is this not just an extension of your other PR?

cmd/copy.go
```diff
@@ -1280,27 +1280,48 @@ func init() {
...
+	cpCmd.PersistentFlags().Float64Var(&raw.blockSizeMB, "block-size-mb", 0, "Use this block size (specified in MiB) when uploading to Azure Storage, and downloading from Azure Storage. The default value is automatically calculated based on file size. Decimal fractions are allowed (For example: 0.25).")
+	cpCmd.PersistentFlags().StringVar(&raw.logVerbosity, "log-level", "INFO", "Define the log verbosity for the log file, available levels: INFO(all requests/responses), WARNING(slow responses), ERROR(only failed requests), and NONE(no output logs). (default 'INFO').")
+	cpCmd.PersistentFlags().StringVar(&raw.blobType, "blob-type", "Detect", "Defines the type of blob at the destination. This is used for uploading blobs and when copying between accounts (default 'None').")
 	cpCmd.PersistentFlags().StringVar(&raw.blockBlobTier, "block-blob-tier", "None", "upload block blob to Azure Storage using this blob tier.")
+<<<<<<< HEAD
```
adreed-msft: It seems as though you've accidentally left merge conflict markers in here.

cmd/copy.go
```diff
@@ -1280,27 +1280,48 @@ func init() {
...
+	cpCmd.PersistentFlags().BoolVar(&raw.s2sPreserveProperties, "s2s-preserve-properties", true, "Preserve full properties during service to service copy. "+
+		"For AWS S3 and Azure File non-single file source, the list operation doesn't return full properties of objects amd files. To preserve full properties, AzCopy needs to send one additional request per object or file.")
+	cpCmd.PersistentFlags().BoolVar(&raw.s2sPreserveAccessTier, "s2s-preserve-access-tier", true, "Preserve access tier during service to service copy. "+
+		"Please refer to [Azure Blob storage: hot, cool, and archive access tiers](https://docs.microsoft.com/azure/storage/blobs/storage-blob-storage-tiers) to ensure destination storage account supports setting access tier. "+
+		"In the cases that setting access tier is not supported, please use s2sPreserveAccessTier=false to bypass copying access tier. (default true). ")
+	cpCmd.PersistentFlags().BoolVar(&raw.s2sSourceChangeValidation, "s2s-detect-source-changed", false, "Check if source has changed after enumerating. "+
+=======
```
adreed-msft: It also appears a good number of these flags are duplicates. Please remove the duplicate flags and ensure the new help text is in place.

cmd/helpMessages.go
```diff
@@ -30,100 +30,122 @@ Copies source data to a destination location. The supported directions are:
...
+- /etc/apache2/mime.types
+- /etc/apache/mime.types
 
 On Windows, MIME types are extracted from the registry. This feature can be turned off with the help of a flag. Please refer to the flag section.
 
 ` + environmentVariableNotice
 
-const copyCmdExample = `Upload a single file using OAuth authentication. Please use 'azcopy login' command first if you aren't logged in yet:
-- azcopy cp "/path/to/file.txt" "https://[account].blob.core.windows.net/[container]/[path/to/blob]"
+const copyCmdExample = `Upload a single file by using OAuth authentication. If you have not yet logged into AzCopy, please use azcopy login command before you run the following command.
```
adreed-msft: >please use azcopy login command [](start = 116, length = 31)

"please use azcopy login command" -> "please run the azcopy login command"

cmd/helpMessages.go
```diff
@@ -30,100 +30,122 @@ Copies source data to a destination location. The supported directions are:
...
+Download an entire directory by using a SAS token:
+  
   - azcopy cp "https://[account].blob.core.windows.net/[container]/[path/to/directory]?[SAS]" "/path/to/dir" --recursive=true
 
-A note about wildcards in URLs:
+A note about using a wildcard character (*) in URLs:
 
-The only usage of a wildcard in a URL that is supported is the final, trailing /*, and in the container name when not specifying a blob name.
-If you happen to use the character * in the name of a blob, please manually encode it to %2A to avoid it being treated as a wildcard character.
+There's only two supported ways to use a wildcard character in a URL.
```
adreed-msft: >There's [](start = 0, length = 7)

"There's" -> "There are" perhaps?

cmd/helpMessages.go
```diff
@@ -30,100 +30,122 @@ Copies source data to a destination location. The supported directions are:
...
+  
   - azcopy cp "https://[account].blob.core.windows.net/[container]/[path/to/directory]?[SAS]" "/path/to/dir" --recursive=true
 
-A note about wildcards in URLs:
+A note about using a wildcard character (*) in URLs:
 
-The only usage of a wildcard in a URL that is supported is the final, trailing /*, and in the container name when not specifying a blob name.
-If you happen to use the character * in the name of a blob, please manually encode it to %2A to avoid it being treated as a wildcard character.
+There's only two supported ways to use a wildcard character in a URL. 
+You can use one just after the final forward slash (/) of a URL to obtain all of the files in a directory.
```
adreed-msft: Reword this to something along the lines of
"A trailing wildcard (in the form of /*) places all files under the directory immediately at the destination instead of under a sub-directory"

Currently, the sentence seems ambiguous, as though --recursive does the same thing (when really it does not)

cmd/helpMessages.go
```diff
@@ -30,100 +30,122 @@ Copies source data to a destination location. The supported directions are:
...
   - azcopy cp "https://[account].blob.core.windows.net/[container]/[path/to/directory]?[SAS]" "/path/to/dir" --recursive=true
 
-A note about wildcards in URLs:
+A note about using a wildcard character (*) in URLs:
 
-The only usage of a wildcard in a URL that is supported is the final, trailing /*, and in the container name when not specifying a blob name.
-If you happen to use the character * in the name of a blob, please manually encode it to %2A to avoid it being treated as a wildcard character.
+There's only two supported ways to use a wildcard character in a URL. 
+You can use one just after the final forward slash (/) of a URL to obtain all of the files in a directory. 
+You can also use one in the name of a container as long as the URL refers only to a container and not to a blob.
```
adreed-msft: This sentence says what you _can_ do, but not _why_ you'd do it. You'd use a wildcarded container name to grab files from a subset of containers on the account.

cmd/helpMessages.go
```diff
@@ -30,100 +30,122 @@ Copies source data to a destination location. The supported directions are:
...
-Download an entire storage account at once
- - azcopy cp "https://[srcaccount].blob.core.windows.net/" "/path/to/dir" --recursive
+Download an entire storage account.
 
-Download an entire storage account at once with a wildcarded container name
- - azcopy cp "https://[srcaccount].blob.core.windows.net/[container*name]" "/path/to/dir" --recursive
+  - azcopy cp "https://[srcaccount].blob.core.windows.net/" "/path/to/dir" --recursive
+
+Download an entire storage account by using a wildcard symbol (*) in the container name.
+
```
adreed-msft: Partly our fault on the wording here, but perhaps change "an entire storage account" to " a subset of containers within the storage account"

cmd/helpMessages.go
```diff
@@ -30,100 +30,122 @@ Copies source data to a destination location. The supported directions are:
...
+  - azcopy cp "https://[srcaccount].blob.core.windows.net/[container*name]" "/path/to/dir" --recursive
+
+Copy a single blob to another blob by using a SAS token.
```
adreed-msft: "by using" -> "using"

Also consider noting slightly above that Azure S2S transfers require a SAS token on the source for now.

cmd/helpMessages.go
```diff
@@ -30,100 +30,122 @@ Copies source data to a destination location. The supported directions are:
...
+ 
   - azcopy cp "https://s3.amazonaws.com/[bucket]/[folder]" "https://[destaccount].blob.core.windows.net/[container]/[path/to/directory]?[SAS]" --recursive=true
-  - Please refer to https://docs.aws.amazon.com/AmazonS3/latest/user-guide/using-folders.html for what [folder] means for S3.
+    
+    Please refer to https://docs.aws.amazon.com/AmazonS3/latest/user-guide/using-folders.html to better understand the [folder] placeholder.
 
-Copy all buckets in S3 service with access key to blob account with SAS:
-  - Set environment variable AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY for S3 source.
+Copy all buckets to Blob Storage from Amazon Web Services (AWS) by using an access key and a SAS token. First, set the environment variable AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY for AWS S3 source.
+
```
adreed-msft: Consider adding an example of wildcarded bucket names, since that has feature parity with Azure Blob and Azure Files sources

cmd/helpMessages.go
```diff
@@ -134,162 +156,177 @@ const jobsCmdLongDescription = "Sub-commands related to managing jobs."
...
 
-Log in using a VM's user-assigned identity with a Resource ID of the service identity:
-- azcopy login --identity --identity-resource-id "/subscriptions/<subscriptionId>/resourcegroups/myRG/providers/Microsoft.ManagedIdentity/userAssignedIdentities/myID"
+   - azcopy login --service-principal
 
-Log in as a service principal using a client secret:
-- Set the environment variable AZCOPY_SPA_CLIENT_SECRET to the client secret for secret based service principal auth
-- azcopy login --service-principal
+Log in as a service principal by using a certificate and a password:
+Set the environment variable AZCOPY_SPA_CERT_PASSWORD to the certificate's password for cert based service principal auth
```
adreed-msft: "and a" -> "and it's"

cmd/helpMessages.go
```diff
@@ -134,162 +156,177 @@ const jobsCmdLongDescription = "Sub-commands related to managing jobs."
...
-Please note that the sync command differs from the copy command in several ways:
-  0. The recursive flag is on by default.
-  1. The source and destination should not contain patterns(such as * or ?).
-  2. The include/exclude flags can be a list of patterns matching to the file names. Please refer to the example section for illustration.
-  3. If there are files/blobs at the destination that are not present at the source, the user will be prompted to delete them. This prompt can be silenced by using the corresponding flags to automatically answer the deletion question.
-  4. When syncing between virtual directories, add a trailing slash to the path (refer to examples) if there's a blob with the same name as one of the virtual directories 
+The sync command differs from the copy command in several ways:
+
+  1. The recursive flag is on by default.
+  2. The source and destination should not contain patterns (For example: * or ?).
```
adreed-msft: @zezha-msft this is incorrect information.
@normesta consider rewording this "sync can copy directories without the recursive flag, but will not sync sub-directories"

cmd/helpMessages.go
```diff
@@ -134,162 +156,177 @@ const jobsCmdLongDescription = "Sub-commands related to managing jobs."
...
+  3. The include and exclude flags can be a list of patterns matching to the file names. Please refer to the example section for illustration.
+  4. If there are files and blobs at the destination that are not present at the source, the user is prompted to delete them. You can use the corresponding flags to silence the prompt to automatically answer the deletion question.
```
adreed-msft: The include and exclude flags are now --include-pattern and --exclude-pattern

cmd/helpMessages.go
```diff
@@ -134,162 +156,177 @@ const jobsCmdLongDescription = "Sub-commands related to managing jobs."
...
+  5. When syncing between virtual directories, add a trailing slash to the path (refer to examples) if there's a blob with the same name as one of the virtual directories.
```
adreed-msft: # 4 is not true by default, AFAIK

cmd/helpMessages.go
normesta: Ideally, we're using more contractions (not fewer contractions). Our guidelines encourage contractions. It's part of our attempts to give content a friendlier more relaxed tone.

cmd/helpMessages.go
normesta: "by using" is actually better. Sometimes to keep titles succinct we'll omit the "by", but that's only to keep longer titles from truncating in search results or from using two lines in a TOC entry. Using this specific sentence as an example, omitting "by" leads the reader to infer that the "blob uses a SAS token". It doesn't really. The person who is typing the command applies a SAS token to the URL. Therefore, the task "copying a single blob to another blob" is performed "by using" a SAS token. This is why I've been replacing instances of "using" with "by using".

cmd/helpMessages.go
normesta: The term "subdirectory" has no hyphen. Other than that, I've updated the text with your suggestion.

cmd/copy.go
```diff
@@ -1280,50 +1280,48 @@ func init() {
...
+	cpCmd.PersistentFlags().BoolVar(&raw.s2sPreserveAccessTier, "s2s-preserve-access-tier", true, "Preserve access tier during service to service copy. "+
+		"Please refer to [Azure Blob storage: hot, cool, and archive access tiers](https://docs.microsoft.com/azure/storage/blobs/storage-blob-storage-tiers) to ensure destination storage account supports setting access tier. "+
+		"In the cases that setting access tier is not supported, please use s2sPreserveAccessTier=false to bypass copying access tier. (default true). ")
+	cpCmd.PersistentFlags().BoolVar(&raw.s2sSourceChangeValidation, "s2s-detect-source-changed", false, "Check if source has changed after enumerating. "+
+    cpCmd.PersistentFlags().StringVar(&raw.s2sInvalidMetadataHandleOption, "s2s-handle-invalid-metadata", common.DefaultInvalidMetadataHandleOption.String(), "Specifies how invalid metadata keys are handled. Available options: ExcludeIfInvalid, FailIfInvalid, RenameIfInvalid. (default 'ExcludeIfInvalid').")
```
adreed-msft: Line has no ending, causes a error when attempting to compile

adreed-msft:
Fix up the syntax error and you're good 👍

cmd/copy.go
normesta: Ok. Fixed. Thx!

cmd/copy.go
```diff
@@ -1280,50 +1280,48 @@ func init() {
...
-	cpCmd.PersistentFlags().BoolVar(&raw.recursive, "recursive", false, "look into sub-directories recursively when uploading from local file system.")
-	cpCmd.PersistentFlags().StringVar(&raw.fromTo, "from-to", "", "optionally specifies the source destination combination. For Example: LocalBlob, BlobLocal, LocalBlobFS.")
-	cpCmd.PersistentFlags().StringVar(&raw.excludeBlobType, "exclude-blob-type", "", "optionally specifies the type of blob (BlockBlob/ PageBlob/ AppendBlob) to exclude when copying blobs from Container / Account. Use of "+
-		"this flag is not applicable for copying data from non azure-service to service. More than one blob should be separated by ';' ")
+	cpCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "Defines the location of json which has the list of only files to be copied.")
+	cpCmd.PersistentFlags().StringVar(&raw.exclude, "exclude-pattern", "", "Exclude these files when copying. This option supports wildcard characters (*)")
+	cpCmd.PersistentFlags().StringVar(&raw.forceWrite, "overwrite", "true", "Overwrite the conflicting files/blobs at the destination if this flag is set to true. (default true).")
```
JohnRusk: Why are we no longer listing the possible values here?  It's not obvious, if we don't list them, that the possible values are true, false or prompt.

cmd/copy.go
```diff
@@ -1280,50 +1280,48 @@ func init() {
...
-	cpCmd.PersistentFlags().BoolVar(&raw.autoDecompress, "decompress", false, "automatically decompress files when downloading, if their content-encoding indicates that they are compressed. The supported content-encoding values are 'gzip' and 'deflate'. File extensions of '.gz'/'.gzip' or '.zz' aren't necessary, but will be removed if present.")
-	cpCmd.PersistentFlags().BoolVar(&raw.recursive, "recursive", false, "look into sub-directories recursively when uploading from local file system.")
-	cpCmd.PersistentFlags().StringVar(&raw.fromTo, "from-to", "", "optionally specifies the source destination combination. For Example: LocalBlob, BlobLocal, LocalBlobFS.")
-	cpCmd.PersistentFlags().StringVar(&raw.excludeBlobType, "exclude-blob-type", "", "optionally specifies the type of blob (BlockBlob/ PageBlob/ AppendBlob) to exclude when copying blobs from Container / Account. Use of "+
-		"this flag is not applicable for copying data from non azure-service to service. More than one blob should be separated by ';' ")
+	cpCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "Defines the location of json which has the list of only files to be copied.")
```
JohnRusk: Oh, I just noticed an other error error (not strictly related to this PR, but maybe we might fix it here).  The file is no longer json. Should say "Defines the location of text file which lists the files to be copied".  Not high priority tho, because this parameter is hidden.

cmd/copy.go
```diff
@@ -1280,50 +1280,48 @@ func init() {
...
+	cpCmd.PersistentFlags().Float64Var(&raw.blockSizeMB, "block-size-mb", 0, "Use this block size (specified in MiB) when uploading to Azure Storage, and downloading from Azure Storage. The default value is automatically calculated based on file size. Decimal fractions are allowed (For example: 0.25).")
+	cpCmd.PersistentFlags().StringVar(&raw.logVerbosity, "log-level", "INFO", "Define the log verbosity for the log file, available levels: INFO(all requests/responses), WARNING(slow responses), ERROR(only failed requests), and NONE(no output logs). (default 'INFO').")
+	cpCmd.PersistentFlags().StringVar(&raw.blobType, "blob-type", "Detect", "Defines the type of blob at the destination. This is used for uploading blobs and when copying between accounts (default 'None').")
```
JohnRusk: I don't understand why the text about detection has been removed here. That does not seem right. Also, "None" is not a valid setting any more.  The default is now "Detect".

My understanding of Detect is as follows:

- Blob - > Blob, all blobType parameter values are valid, and Detect means "Detect the type of the source blob"
- anything else -> Blob, all blobType parameter values are valid, and Detect means "Look for VHD/VHDX extension and treat that as page blob"
- anything -> NOT Blob, only Detect is valid, and it means "No blob type specified"

In addition to Detect, the valid values are BlockBlob, PageBlob, AppendBlob. Those should probably be mentioned here, beause they are not otherwise discoverable to the user.

cmd/copy.go
```diff
@@ -1280,50 +1280,48 @@ func init() {
...
+	cpCmd.PersistentFlags().BoolVar(&raw.putMd5, "put-md5", false, "Create an MD5 hash of each file, and save the hash as the Content-MD5 property of the destination blob or file. (By default the hash is NOT created.) Only available when uploading.")
+	cpCmd.PersistentFlags().StringVar(&raw.md5ValidationOption, "check-md5", common.DefaultHashValidationOption.String(), "Specifies how strictly MD5 hashes should be validated when downloading. Only available when downloading. Available options: NoCheck, LogOnly, FailIfDifferent, FailIfDifferentOrMissing. (default 'FailIfDifferent')")
+	cpCmd.PersistentFlags().StringVar(&raw.includeFileAttributes, "include-attributes", "", "(Windows only) Include files whose attributes match the attribute list. For example: A;S;R")
+	cpCmd.PersistentFlags().StringVar(&raw.excludeFileAttributes, "exclude-attributes", "", "(Windows only) Exclude files whose attributes match the attribute list. For example: A;S;R")
+	cpCmd.PersistentFlags().BoolVar(&raw.CheckLength, "check-length", true, "Check the length of a file on the destination after the transfer. If there is a mismatch between source and destination, the transfer is cancelled.")
```
JohnRusk: >cancelled. [](start = 211, length = 10)

We can't change "fail the transfer" to "the transfer is cancelled" because "failed" and "cancelled" have clear and _different_ meanings in AzCopy.  To improve the original grammar, I'd suggest "the transfer is marked as failed".

cmd/copy.go
```diff
@@ -1280,50 +1280,48 @@ func init() {
...
+	cpCmd.PersistentFlags().BoolVar(&raw.s2sPreserveProperties, "s2s-preserve-properties", true, "Preserve full properties during service to service copy. "+
+		"For AWS S3 and Azure File non-single file source, the list operation doesn't return full properties of objects amd files. To preserve full properties, AzCopy needs to send one additional request per object or file.")
```
JohnRusk: >amd [](start = 114, length = 3)

Typo: amd -> and

cmd/helpMessages.go
```diff
@@ -30,100 +30,126 @@ Copies source data to a destination location. The supported directions are:
...
+Copy an entire directory to Blob Storage from AWS S3 by using an access key and a SAS token. First, set the environment variable AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY for AWS S3 source.
+ 
   - azcopy cp "https://s3.amazonaws.com/[bucket]/[folder]" "https://[destaccount].blob.core.windows.net/[container]/[path/to/directory]?[SAS]" --recursive=true
-  - Please refer to https://docs.aws.amazon.com/AmazonS3/latest/user-guide/using-folders.html for what [folder] means for S3.
+    
+    Please refer to https://docs.aws.amazon.com/AmazonS3/latest/user-guide/using-folders.html to better understand the [folder] placeholder.
 
-Copy all buckets in S3 service with access key to blob account with SAS:
-  - Set environment variable AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY for S3 source.
+Copy all buckets to Blob Storage from Amazon Web Services (AWS) by using an access key and a SAS token. First, set the environment variable AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY for AWS S3 source.
```
JohnRusk: In a late-breaking change, due to a defect, we've just disabled the use of "[https://s3.amazonaws.com](https://s3.amazonaws.com)" to copy all buckets. Instead, users need to use the region-specific approach below.

cmd/helpMessages.go
```diff
@@ -30,100 +30,126 @@ Copies source data to a destination location. The supported directions are:
...
 
-Copy all buckets in a S3 region with access key to blob account with SAS:
-  - Set environment variable AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY for S3 source.
+Copy all buckets to Blob Storage from an Amazon Web Services (AWS) region by using an access key and a SAS token. First, set the environment variable AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY for AWS S3 source.
+ 
   - azcopy cp "https://s3-[region].amazonaws.com/" "https://[destaccount].blob.core.windows.net?[SAS]" --recursive=true
+
+Copy a subset of buckets by using a wildcard symbol (*) in the bucket name. Like the previous examples, you'll need an access key and a SAS token. Make sure to set the environment variable AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY for AWS S3 source.
+
+  - azcopy cp "https://s3.amazonaws.com/[bucket*name]/" "https://[destaccount].blob.core.windows.net?[SAS]" --recursive=true
```
JohnRusk: >https://s3.amazonaws.com/[bucket*name]/ [](start = 15, length = 39)

As per above comment, probably best to change this URL to

https://s3-[region].amazonaws.com/[bucket*name]

cmd/helpMessages.go
```diff
@@ -134,162 +160,176 @@ const jobsCmdLongDescription = "Sub-commands related to managing jobs."
...
 
 const showJobsCmdShortDescription = "Show detailed information for the given job ID"
 
 const showJobsCmdLongDescription = `
-Show detailed information for the given job ID: if only the job ID is supplied without a flag, then the progress summary of the job is returned.
-(In the progress information shown by this command, the byte counts and percent complete do not include files currently in progress.)
-If the with-status flag is set, then the list of transfers in the job with the given value will be shown.`
+If you provide only a job ID, and not a flag, then this command returns the progress summary only.
+The byte counts and percent complete that appears when you run this command don't reflect files related to jobs that are currently in progress.
+If you set the with-status flag, then only the list of transfers associated with the given status appear.`
```
JohnRusk: The edit seems to have changed the meaning here.  The byte counts and percent complete are provided for jobs in progress, but they are little innacurate because they only include the files that are _completed_ in the job. They don't not include the _partially completed_ files that are in progress right now.  (Any big live job will usually have some files that are totally completed, some that are inprogress right now, and some that are not started yet.)

cmd/helpMessages.go
```diff
@@ -134,162 +160,176 @@ const jobsCmdLongDescription = "Sub-commands related to managing jobs."
...
-  0. The recursive flag is on by default.
-  1. The source and destination should not contain patterns(such as * or ?).
-  2. The include/exclude flags can be a list of patterns matching to the file names. Please refer to the example section for illustration.
-  3. If there are files/blobs at the destination that are not present at the source, the user will be prompted to delete them. This prompt can be silenced by using the corresponding flags to automatically answer the deletion question.
-  4. When syncing between virtual directories, add a trailing slash to the path (refer to examples) if there's a blob with the same name as one of the virtual directories 
+The sync command differs from the copy command in several ways:
+
+  1. Sync can copy directories without the recursive flag, but it will not sync subdirectories.
+  2. The source and destination should not contain patterns (For example: * or ?).
+  3. The include-pattern and exclude-pattern flags can be a list of patterns matching to the file names. Please refer to the example section for illustration.
```
JohnRusk: I'm not sure what's happend with point #3 here.  The new #3 is not a difference compared to copy, because copy now has the exact same parameters.  And the old #3, about deletion, is important and still (almost) true.  I think the correct wording would be:

"If the deleteDestination flag is set to true or prompt, then sync will delete files/blobs at the destination that are not present at the source."

cmd/helpMessages.go
JohnRusk: Not sure what's happening here: the old # 1 (the one about source and destination should not contain patterns) is indeed obsolete and should have been removed.  Because now both copy and sync are the same in that regard. (I.e. it should not be the new # 2, it should be gone). 

But I think its important that the new # 1, should start with "The recursive flag is on by default", since this is indeed the opposite of what copy does.

cmd/copy.go
normesta: Thx John. I don't recall ever removing default values from these strings, so I'm not sure how to address your concern. If default values are somehow missing, can you please tell me which ones you believe need to list those and what those default values are?

cmd/helpMessages.go
normesta: I'm not sure what you mean about "old number 3". I do remember there being a statement about delete and being prompted. The statement was this "If there are files or blobs at the destination that aren't present at the source, the user will be prompted to delete them.". Adele stated that this was inaccurate so I removed it. I *think* you're asking me to add a new bullet with the string "If the deleteDestination flag is set to true or prompt, then sync will delete files/blobs at the destination that are not present at the source.". I'll proceed with that to unblock the PR.  Thx

cmd/copy.go
```diff
@@ -1319,7 +1319,8 @@ func init() {
...
 	cpCmd.PersistentFlags().StringVar(&raw.logVerbosity, "log-level", "INFO", "Define the log verbosity for the log file, available levels: INFO(all requests/responses), WARNING(slow responses), ERROR(only failed requests), and NONE(no output logs). (default 'INFO').")
-	cpCmd.PersistentFlags().StringVar(&raw.blobType, "blob-type", "Detect", "Defines the type of blob at the destination. This is used for uploading blobs and when copying between accounts (default 'None').")
+	cpCmd.PersistentFlags().StringVar(&raw.blobType, "blob-type", "Detect", "Defines the type of blob at the destination. This is used for uploading blobs and when copying between accounts (default 'Detect'). Valid values include 'Detect', 'BlockBlob', 'PageBlob', and 'AppendBlob'. "+
+		"When copying between accounts, a value of 'Detect' causes AzCopy to use the type of source blob to determine the type of the destination blob. When uploading a file, 'Detect' determines if the file is a VHD or a VHDX file based on the file extension. If the file is ether a VHD or VHDX file, AzCopy treats the file as a page blob.")
```
JohnRusk: Nice wording. Way better than anything I've had in emails and discussions on this feature!

cmd/helpMessages.go
JohnRusk: Correct.

PR #644, opened 2019-10-03
Check the destination length on upload/rename s2s related checklength variables to generic ones

adreed-msft:
This will have a merge conflict with fault induction in xfer-anyToRemote.

### DIFF
--- a/ste/xfer-anyToRemote.go
+++ b/ste/xfer-anyToRemote.go
@@ -318,11 +318,11 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s ISenderBase, si
 		destLength, err := s.GetDestinationLength()
 
 		if err != nil {
-			jptm.FailActiveSend(common.IffString(isS2SCopier, "S2S ", "Download ")+"Length check: Get destination length", err)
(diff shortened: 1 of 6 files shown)


ste/xfer-anyToRemote.go
```diff
@@ -314,16 +314,15 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s ISenderBase, si
...
-		if s2sc, isS2SCopier := s.(s2sCopier); isS2SCopier { // TODO: Implement this for upload and download?
-			destLength, err := s2sc.GetDestinationLength()
+		_, isS2SCopier := s.(s2sCopier)
+		destLength, err := s.GetDestinationLength()
 
-			if err != nil {
-				jptm.FailActiveSend("S2S Length check: Get destination length", err)
-			}
+		if err != nil {
+			jptm.FailActiveSend(common.IffString(isS2SCopier, "S2S ", "Download ")+"Length check: Get destination length", err)
```
zezha-msft: >Download [](start = 62, length = 8)

shouldn't it be Upload?

zezha-msft: approved with comment

JohnRusk: Yes, it should be.  Or just omit it, I think, and let FailActiveSend provide the contxt (upload/sys).

PR #649, opened 2019-10-04
Skip blobs that have snapshots in rm command

zezha-msft:


ste/xfer-deleteBlob.go
```diff
@@ -30,28 +29,35 @@ func DeleteBlobPrologue(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pac
...
+			jptm.Log(pipeline.LogInfo, fmt.Sprintf("DELETE SUCCESSFUL: %s", strings.Split(info.Destination, "?")[0]))
 		}
+
 		jptm.SetStatus(status)
 		jptm.ReportTransferDone()
 	}
 
 	// TODO confirm whether we should add a new flag to allow user to specify whether snapshots should be removed
-	_, err := srcBlobURL.Delete(jptm.Context(), azblob.DeleteSnapshotsOptionInclude, azblob.BlobAccessConditions{})
+	_, err := srcBlobURL.Delete(jptm.Context(), azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
```
JohnRusk: Just a thought: ther one place in the code where DeleteSnapshotsOptionInclude is still present is in the delete in Sync.  I'm wondering if it should be made consistent with this here. Not with the fancy handling of status - which we wont' have time for there right now, but simply in letting it fail if there are snapshots.  I'm in two minds about it. On one had its consistent with here. On the other hand, maybe a sync delete really means delete????  What do you think?

zezha-msft: Great question! For sync, the user actually gave consent to remove the destination files by setting `--delete-destination=true` (or saying yes to the prompt when `--delete-destination=prompt`). So I think it's ok to have this apparent discrepancy.

zezha-msft: Thoughts?

JohnRusk: I think almost I agree.  But my only hesitation is there here, in the rm-command, they also "gave their consent" by running an RM command in the first place.

JohnRusk: So basically I'm not sure...

adreed-msft: I think John's right in the sense that there's a contradiction of logic here. It would be nice if there were a flag to do away with snapshots as well, but in that same sense, I feel like it'd only make sense to have a command to create snapshots.

JohnRusk: I think there are plans for a flag in the future. THe question is what to do now, before we get around to implementing the flag.  Manybe the answer depends on whether we would expect that flag to also apply to sync?....

zezha-msft: Hmm ok, I'll leave it as is for now, and adjust if needed in a future release.

My perspective is that if customers invoke the `rm` command on an entire container, they probably don't know which blobs have snapshots and which ones don't. From this POV, their consent is probably not as strong as in the `sync` scenario. SO it's nice of us to avoid deleting the ones with snapshots, in case they want to be more cautious.

zezha-msft: The consent in `sync` is more like "these files are extra, get rid of them".

JohnRusk: That sounds logical.

:shipit:

zezha-msft:
CI failed due to unrelated flaky bfs tests.

PR #651, opened 2019-10-05
Added safety check to prevent users from performing service level sync

zezha-msft:
Forgot to add a check to prevent users from doing [Account URL] <-> [Account URL] or [Account URL] <-> [Container URL].

### DIFF
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -71,6 +71,20 @@ func (raw *rawSyncCmdArgs) parsePatterns(pattern string) (cookedPatterns []strin
 	return
 }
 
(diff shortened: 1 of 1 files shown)


cmd/sync.go
```diff
@@ -71,6 +71,20 @@ func (raw *rawSyncCmdArgs) parsePatterns(pattern string) (cookedPatterns []strin
 	return
 }
 
+// it is assume that the given url has the SAS stripped, and safe to print
```
JohnRusk: At first I wondered how confident we are in this assumption, but then I remembered we redact all our GLCM and log outputs anway.

PR #652, opened 2019-10-05
Don't lock destination when its dev/null

JohnRusk:
Since all files have same destination in that case, and the locking gets annoyed (and is totally unnecessary)

### DIFF
--- a/ste/mgr-JobPartTransferMgr.go
+++ b/ste/mgr-JobPartTransferMgr.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ste/mgr-JobPartTransferMgr.go
```diff
@@ -300,6 +301,10 @@ func (jptm *jobPartTransferMgr) FileCountLimiter() common.CacheLimiter {
 // (iii) For completeness, there's also bucket->container name resolution when copying from S3, but that is not expected to ever
 //      create collisions, since it already takes steps to prevent them.
 func (jptm *jobPartTransferMgr) WaitUntilLockDestination(ctx context.Context) error {
+	if strings.EqualFold(jptm.Info().Destination, common.Dev_Null) {
```
zezha-msft: I couldn't remember clearly, but is this string different on Windows vs Linux/Mac? Maybe double check to make sure.

JohnRusk: Yes, it is, but our constant is (now) defined as being equal to os.DevNull, so that takes care of it.

PR #653, opened 2019-10-05
Reinstate temporary extra code to get MD5s when enumerating BlobFS

JohnRusk:
because its still necessary


adreed-msft:
If we have to keep this fix up, we should pass the validation option through the enumerator initialization properly later. There are a few similarly ugly tricks I've done, but they all go to STE, were meant for CI, and were designed to just be ripped out later with relative ease, at least in the case of the blobFS flush threshold.

PR #657, opened 2019-10-07
Fix downloading to the root of a drive on Windows

adreed-msft:


common/writeThoughFile.go
```diff
@@ -25,18 +25,11 @@ import (
...
-		err = os.MkdirAll(parentDirectory, os.ModePerm)
-		if err != nil {
-			return err
-		}
-	} else if err != nil {
-		return err
-	}
-	return nil
+
+	// No need to error check if the mkdir failed.
```
JohnRusk: @Adele My initial gut instinct is not super-comfortable with this.  In particular, if there really is an issue with making a directory, and we don't report it as such, then it becomes harder for customers to diagnose at support time. E.g. instead of getting one directory creating failure error, they get one failure per file in the log.  

Can think you think of any alternatives that don't have that draw-back?

adreed-msft: Could just log to the job log that the directory couldn't be created, and then log the error.

cmd/zc_traverser_local.go
```diff
@@ -325,11 +325,14 @@ func cleanLocalPath(localPath string) string {
...
+	// path.Clean steals the first / from the // or \\ prefix.
+	if strings.HasPrefix(localPath, `\\`) || strings.HasPrefix(localPath, `//`) {
+		// return the \ we stole from the UNC/extended path.
 		normalizedPath = localPathSeparator + normalizedPath
-	} else if len(localPath) == 3 && (strings.HasSuffix(localPath, `:\`) || strings.HasSuffix(localPath, ":/")) ||
-		len(localPath) == 2 && strings.HasSuffix(localPath, ":") { // path.Clean steals the last / from C:\, C:/, and does not add one for C:
+	}
+
+	// path.Clean steals the last / from C:\, C:/, and does not add one for C:
+	if common.RootDriveRegex.MatchString(strings.ReplaceAll(common.ToShortPath(normalizedPath), common.OS_PATH_SEPARATOR, common.AZCOPY_PATH_SEPARATOR_STRING)) {
```
JohnRusk: This looks like a bug.  Because its testing normalizedPath, which will have already have the / stripped, against a regex that expects the / to be present.  Do you need a different regex?

JohnRusk: Or maybe test localPath against the regex you already have, instead of testing normalized path.

common/writeThoughFile_linux.go
```diff
@@ -32,7 +32,7 @@ func CreateFileOfSize(destinationPath string, fileSize int64) (*os.File, error)
 func CreateFileOfSizeWithWriteThroughOption(destinationPath string, fileSize int64, writeThrough bool) (*os.File, error) {
 	err := CreateParentDirectoryIfNotExist(destinationPath)
 	if err != nil {
-		return nil, err
+		return err
```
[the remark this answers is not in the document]
adreed-msft: Why did these lines revert? I could've sworn I cycled through and made sure they all were right.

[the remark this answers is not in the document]
adreed-msft: They did not revert-- You're looking at an old commit.

cmd/zc_traverser_local.go
adreed-msft: The regex does not expect the / to be present.

Please check the regexr example provided in the comment above the regex.
https://regexr.com/4mf6l

JohnRusk: Oops I missed the ? in the regex.

common/writeThoughFile_linux.go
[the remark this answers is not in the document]
JohnRusk: Weird too, I left this comment on my previous review, which was not even today. Not sure why it popped up today!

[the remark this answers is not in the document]
JohnRusk: In any case, there's no  issue.

PR #659, opened 2019-10-07
Err for wildcard + list; have list+/* play nice

adreed-msft:


cmd/zc_enumerator.go
```diff
@@ -186,6 +186,13 @@ func initResourceTraverser(resource string, location common.Location, ctx *conte
 			if err != nil {
 				return nil, err
 			}
+		} else {
+			// First, ignore all escaped stars. Stars can be valid characters on many platforms.
+			tmpResource := strings.ReplaceAll(resource, `\*`, ``)
```
JohnRusk: I'm not quite sure how this works, given that we want to detect the likes of "c:\foo\*\bar" on windows, as something that's NOT compatible with List of files.  Maybe needs to be OS specific - only do that replacement if os.PathSeparator is NOT \?

adreed-msft: Yeah, limiting this per-OS may be a good idea.

adreed-msft: Done.

PR #661, opened 2019-10-08
Fix overflow in computing flush threshold

JohnRusk:


ste/uploader-blobFS.go
```diff
@@ -141,7 +141,7 @@ func (u *blobFSUploader) RemoteFileExists() (bool, error) {
 func (u *blobFSUploader) Prologue(state common.PrologueState) (destinationModified bool) {
 	jptm := u.jptm
 
-	u.flushThreshold = int64(u.chunkSize * ADLSFlushThreshold)
+	u.flushThreshold = int64(u.chunkSize) * int64(ADLSFlushThreshold)
```
adreed-msft: Good find!

PR #665, opened 2019-10-09
Fix tier checks in smoke tests

JohnRusk:
Fixes issue in two of the smoke tests.


adreed-msft:
Seems alright to me, but don't page blobs have a tier on premium containers?

JohnRusk:
>Seems alright to me, but don't page blobs have a tier on premium containers?

They sort of do (in terms their standard sizes) but its a different set of possible values from the usual "Hot/Cold" etc, and its Hot that the tests check for by default. (So all I've done is tell the test not to check at all)

PR #669, opened 2019-10-09
Disable (for now) the use of region-less S3 URL

JohnRusk:


cmd/zc_traverser_s3.go
```diff
@@ -207,3 +212,16 @@ func newS3Traverser(rawURL *url.URL, ctx context.Context, recursive, getProperti
...
+
+// Disallow the non-region aware URL type.
+// Reason is that we had intermittent bucket location lookup issues when using that technique
+// with the apparent cause being the lookup of bucket locations, which we need to change to use minio's BucketExists.
+// For info see: https://github.com/aws/aws-sdk-go/issues/720#issuecomment-243891223
+// Once we change to bucketExists, assuming its reliable, we will be able to re allow this URL type.
+func checkS3UrlType(s3URLParts common.S3URLParts) error {
+	if strings.EqualFold(s3URLParts.Host, "s3.amazonaws.com") {
+		return errors.New("this version of AzCopy does not support transferring from the 's3.amazonaws.com' URL. " +
+			"Please used a region-specific endpoint to transfer from one specific region. E.g. s3.us-east-1.amazonaws.com or a virtual-hosted reference to a single bucket")
```
adreed-msft: used->use
one -> a
consider rewording "the 's3.amazonaws.com' URL" to "the region-unaware 's3.amazonaws.com' URL"

JohnRusk: Thanks. I've fixed the used/use.
One/a was deliberate. One of the times when this message will appear is when a user is trying to transfer their whole AWS account.  We used to tell people to do that by using the now-banned URL.  If someone tries that, and we say, "use a region-specific endpoint to transfer from a specific region" that can be read as "use a region-specific endpoint to transfer IF you are trying to transfer from a specific region". They will of course think, "But I'm not trying to do that". So the "one" wording was attempting to remove the implication of "if" and indicate that, sorry, you have to transfer from 1 region at a time.

adreed-msft: Ah.

PR #673, opened 2019-10-10
Fix URL generation for tests

JohnRusk:


JohnRusk:
This approach didn't work. Too hard to change the tests in the time we have.

PR #674, opened 2019-10-10
Soften S3 URL change to warning only

JohnRusk:
Our previously-agreed approach broke all our automated tests. This should provide the same information to users, without breaking stuff!

### DIFF
--- a/cmd/zc_traverser_s3.go
+++ b/cmd/zc_traverser_s3.go
@@ -22,10 +22,10 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/url"
 	"strings"
+	"sync"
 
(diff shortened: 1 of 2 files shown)


cmd/zc_traverser_s3.go
```diff
@@ -213,15 +211,18 @@ func newS3Traverser(rawURL *url.URL, ctx context.Context, recursive, getProperti
 	return
 }
 
-// Disallow the non-region aware URL type.
+// Discourage the non-region aware URL type. (but don't ban it, because that breaks almost all our S3 automated tests)
```
adreed-msft: _phwoof_

Hopefully we can figure out a better workaround for regionless S3 URLs in the future.

PR #682, opened 2019-10-11
Enable the use of List on Blob, Files, and BlobFS by upgrading list to the modular traverser system

adreed-msft:
This grafts the existing list structure over to the modular traverser system.

Other potential resources made available by our traversal system are blocked but can be enabled any time. Account level listing is also possible as well, due to this addition.

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -24,14 +24,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/url"
 	"strconv"
-	"strings"
+
+	"github.com/spf13/cobra"
 
 	"github.com/Azure/azure-storage-azcopy/common"
 	"github.com/Azure/azure-storage-azcopy/ste"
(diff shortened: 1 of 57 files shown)


cmd/list.go
```diff
@@ -59,11 +57,12 @@ func init() {
 			// the expected argument in input is the container sas / or path of virtual directory in the container.
 			// verifying the location type
 			location := inferArgumentLocation(sourcePath)
-			if location != location.Blob() {
+			// Only support listing for Azure locations
+			if location != location.Blob() && location != location.File() && location != location.BlobFS() {
```
zezha-msft: >location [](start = 6, length = 8)

Just to double check, are there smoke tests covering the list command?

adreed-msft: No. We probably should have them.

adreed-msft:
It's now correctly targeting master.

PR #686, opened 2019-10-14
Added back cancel-from-stdin option

zezha-msft:
More tests needed. 

Early feedbacks are welcome! Please let me know what you think of this approach.

### DIFF
--- a/common/lifecyleMgr.go
+++ b/common/lifecyleMgr.go
@@ -3,7 +3,6 @@ package common
 import (
 	"bufio"
 	"fmt"
-	"io"
 	"os"
 	"os/signal"
 	"runtime"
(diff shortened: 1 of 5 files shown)


common/lifecyleMgr.go
```diff
@@ -57,12 +65,75 @@ func GetLifecycleMgr() LifecycleMgr {
...
+		if lcm.allowCancelFromStdIn && strings.EqualFold(msg, "cancel") {
+			lcm.cancelChannel <- os.Interrupt
+		} else {
+			lcm.inputQueue <- userInput{timeReceived: timeReceived, content: msg}
+		}
+	}
+}
+
+// get the answer to a question that was asked at a certain time
+// only user input after the specified time is returned to make sure that we are getting the right answer to our question
```
JohnRusk: Nice

common/lifecyleMgr.go
```diff
@@ -57,12 +64,71 @@ func GetLifecycleMgr() LifecycleMgr {
...
+func (lcm *lifecycleMgr) getInputAfterTime(time time.Time) string {
```
adreed-msft: Now here's a thought. What if multiple questions get asked at once? If we're not careful, we could accidentally read another question's answer.

Objects in channels can only be read from a channel once. Consider adding a lock to lcm.Prompt so that only one question ever gets asked at a time.

JohnRusk: We only ask questions (currently) from with the enumeration code, right?  And that code is single-threaded.

zezha-msft: We ask questions from `ste` as well, in case of overwrite.

The lifecycleMgr already guarantees that only 1 question get asked at a time because the processors (such `processTextOutput`) outputs the question, and waits until the user answers before doing anything else.

PR #687, opened 2019-10-14
Added support for clfsload extension [DO NOT MERGE]

zezha-msft:
For CR only. No need to merge.


cmd/load.go
```diff
@@ -0,0 +1,286 @@
...
+		return "", errors.New("cannot open CLFSLoad extension")
+	}
+	defer file.Close()
+
+	//make sure we will be invoking the right CLFSLoad
+	hash := md5.New()
+	if _, err := io.Copy(hash, file); err != nil {
+		return clfsToolPath, errors.New("cannot compute hash of CLFSLoad extension")
+	}
+
```
adreed-msft: Consider adding a hidden debug flag to substitute the clfsload executable with an alternate one, in case there's say, an update and they want to test it.

cmd/loadOutputTranslation.go
```diff
@@ -0,0 +1,236 @@
...
+		if strings.Contains(line, clfsExitTag) {
+			c.processEndOfJob(line)
+		} else if strings.Contains(line, clfsInitializationTag) {
+			c.processInitialization(line)
+		} else if strings.Contains(line, clfsInfoTag) {
+			c.processInfo(line)
+		} else if strings.Contains(line, clfsProgressTag) {
+			c.processProgress(line)
+		} else if strings.Contains(line, clfsErrTag) {
+			c.processError(line)
```
adreed-msft: Consider doing strings.HasPrefix instead of strings.Contains. It's unlikely that we'll see any of these other strings inside of an output line, but it would prevent us from re-running something like processEndOfJob.

cmd/loadOutputTranslation.go
```diff
@@ -0,0 +1,236 @@
...
+	<-c.finishSignal
+}
+
+func (c *clfsExtensionOutputParser) processInitialization(line string) {
+	cleanMessage := strings.TrimPrefix(line, clfsInitializationTag)
+
+	initInfo := clfsInitialization{}
+	err := json.Unmarshal([]byte(cleanMessage), &initInfo)
+	if err != nil {
+		// don't crash if these info are not critical and are only nice-to-have
```
adreed-msft: >if [](start = 17, length = 2)

if->as

cmd/loadOutputTranslation.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+}
+
+func (c *clfsExtensionOutputParser) processInfo(line string) {
+	cleanMessage := strings.TrimPrefix(line, clfsInfoTag)
+
+	info := clfsInfo{}
+	err := json.Unmarshal([]byte(cleanMessage), &info)
+	if err != nil {
+		// don't crash if these info are not critical and are only nice-to-have
```
adreed-msft: >if [](start = 17, length = 2)

ditto

cmd/loadOutputTranslation_test.go
```diff
@@ -0,0 +1,288 @@
+// Copyright © Microsoft <wastore@microsoft.com>
```
adreed-msft: consider renaming file to zt_load_output_traanslation_test.go

cmd/load_test.go
```diff
@@ -0,0 +1,63 @@
+// Copyright © Microsoft <wastore@microsoft.com>
+//
```
adreed-msft: Consider renaming file to zt_load_test.go

cmd/loadOutputTranslation.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func (c *clfsExtensionOutputParser) processError(line string) {
+	parts := strings.Split(line, clfsErrTag)
+	if len(parts) < 2 {
+		return
+	}
+
+	c.lcm.Info(strings.TrimSpace(parts[1]))
+}
```
adreed-msft: Perhaps we should consider a higher status than Info for errors.

cmd/zt_test.go
```diff
@@ -49,7 +49,8 @@ func Test(t *testing.T) { chk.TestingT(t) }
 
 type cmdIntegrationSuite struct{}
 
-var _ = chk.Suite(&cmdIntegrationSuite{})
+// NOTE: disabled on this branch only, to speed up CI
+//var _ = chk.Suite(&cmdIntegrationSuite{})
```
adreed-msft: When this is good to go, be sure to un-comment these chk.Suite lines.

common/version.go
```diff
@@ -1,6 +1,6 @@
 package common
 
-const AzcopyVersion = "10.3.0"
+const AzcopyVersion = "10.3.0-clfsload-preview"
```
adreed-msft: Consider making this an extension user-agent like the others, since the added clfsload functionality CURRENTLY doesn't goof with any azcopy functionality.

zezha-msft:
Will update and re-open a new PR.

Issue #211, opened 2019-02-14
blob-to-blob OAuth support for both


Mmdixon:
### Which version of the AzCopy was used? 
AzCopy 10.0.7-Preview

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

### What command did you run?
```
azcopy login
azcopy copy 'https://[account1].blob.core.windows.net/[container1]/log.txt' 'https://[account2].blob.core.windows.net/[container2]/log.txt'
```
This gives the error:
```
Scanning...
Using OAuth token for authentication.

failed to perform copy command due to error: cannot start job due to error: cannot copy the entire container or directory without recursive flag. Please use --recursive flag.
```
`--recursive` is not relevant in this case, and if I add the flag I get a `404 The specified resource does not exist.` which is a typical error if you are not authenticated too. Adding SAS exclusively to the source works:
```
azcopy copy 'https://[account1].blob.core.windows.net/[container1]/log.txt?[SAS]' 'https://[account2].blob.core.windows.net/[container2]/log.txt'
```
But adding SAS to the destination does not:
```
azcopy copy 'https://[account1].blob.core.windows.net/[container1]/log.txt' 'https://[account2].blob.core.windows.net/[container2]/log.txt?[SAS]'
```

### What problem was encountered?
It would be nice if the Azure Block Blob (OAuth ) <-> Azure Block Blob (OAuth) pair could be supported.

### How can we reproduce the problem in the simplest way?
Setup 2 blobs that both require OAuth and transfer between them.

zezha-msft:
Hi @Mmdixon, thanks for reaching out!

The actual copy is done by the service, which means that the service (destination account) must be able to read your source blob. It wouldn't be safe to give the service your OAuth token, so we require the source to have a SAS. However, there will be a new feature soon that would enable your suggestion. Please stay tuned. 😄 

@jiacfan We should definitely improve the error message.

mattiasmoser8451:
We are running into this issue, too. 

@zezha-msft wrote on 2/14/2019: "However, there will be a new feature soon that would enable your suggestion." Is there any progress on this issue?

zezha-msft:
@mattiasmoser8451 thanks for the feedbacks! A [PR](https://github.com/Azure/azure-storage-azcopy/pull/689) was raised but it needs to be reworked a bit. We'll try to fit it in a future release very soon.

nakulkar-msft:
#1304

PR #689, opened 2019-10-14
Create SAS tokens for blob->* transfers on the fly in the STE

adreed-msft:
Some background:

Previous iterations of this PR have done this creation in the frontend. This would cause problems if a user attempted to resume a transfer of this style, because the SAS has to come in externally in that scenario. Furthermore, this would have caused issues with S2S transfers, as S2S transfers may potentially contain multiple source containers, and UDKs can only create SAS tokens for containers or blobs, **not** the entire service.

Therefore, I moved it all to the STE. It lives under a single-instance struct named UserDelegationAuthenticationManager, which does two things:

1) If a UDK was created, it maps container names to SAS tokens.
2) If a UDK was not created, it creates empty SAS tokens.

Furthermore, rather than just grabbing the URL once from SIP, we grab it every time we make a call to enable refreshing the underlying UDK and user delegated SAS token


adreed-msft:
Also todo: Make the tests not panic because something about OAuthTokenManager is nil

adreed-msft:
> Also todo: Make the tests not panic because something about OAuthTokenManager is nil

This didn't get caught in manual testing-- But intuition says that OAuthTokenManager didn't get set up in tests for some reason. Maybe some better nil checking or a mimic of some sort will fix it.

ste/mgr-JobPartTransferMgr.go
```diff
@@ -194,6 +208,19 @@ func (jptm *jobPartTransferMgr) Info() TransferInfo {
...
+
+	// make use of UDAM ONLY in the case of blob -> elsewhere, AND the source SAS is not present to start with.
+	fromTo := jptm.FromTo()
+	if fromTo.IsS2S() && fromTo.From() == common.ELocation.Blob() {
+		// Check srcSAS because an S2S transfer requires that the source is either public OR has a SAS token
+		if len(srcSAS) == 0 {
+			// We're going to safely ignore the error here.
+			// Realistically, there will NEVER be an error.
+			// But, if there is, we get "" back. So there's no need to worry, because srcSAS was already empty.
+			srcSAS, _ = jptm.getSrcUserDelegationSAS(src)
```
adreed-msft: TODO: Consider logging info about our attempt to create a UDK and our attempt to create a User Delegation SAS token

ste/userDelegationAuthenticationManager.go
```diff
@@ -0,0 +1,109 @@
...
+func newUserDelegationAuthenticationManager(serviceURL azblob.ServiceURL) (userDelegationAuthenticationManager, error) {
+	authManager := userDelegationAuthenticationManager{}
+	// Create a user delegation
+	authManager.startTime = time.Now()
+	authManager.expiryTime = authManager.startTime.Add(time.Hour * 24 * 7)
+
+	keyInfo := azblob.NewKeyInfo(authManager.startTime, authManager.expiryTime)
+
+	var err error
+	authManager.credential, err = serviceURL.GetUserDelegationCredential(steCtx, keyInfo, nil, nil)
```
adreed-msft: TODO: Consider dropping errors in their current use here, since errors out of UDAM are ignored anyway, and this is a holdover from the original design

ste/mgr-JobPartTransferMgr.go
```diff
@@ -186,6 +185,21 @@ func (jptm *jobPartTransferMgr) GetSourceCompressionType() (common.CompressionTy
 	return common.GetCompressionType(encoding)
 }
 
+func (jptm *jobPartTransferMgr) getSrcUserDelegationSAS(src string) (string, error) {
+	// panics if udam isn't found
+	udam := jptm.jobPartMgr.GetUserDelegationAuthenticationManagerInstance()
+
+	objectURL, err := url.Parse(src)
+
+	if err != nil {
```
adreed-msft: TODO: add note that if this fails to parse, something has gone terribly, terribly wrong.

ste/mgr-JobMgr.go
```diff
@@ -313,6 +399,21 @@ func (jm *jobMgr) AddJobPart(partNum PartNumber, planFile JobPartPlanFileName, s
...
+
+	// This is actually a really fantastic place to set up our user delegation authentication manager.
+	// This is because we have no awareness of our source root until now,
+	// and two job parts will never have different sourceroots.
+	// Note that setupUDAM checks if UDAM is nil before attempting to create it.
+	if jpm.Plan().FromTo.IsS2S() && jpm.Plan().FromTo.From() == common.ELocation.Blob() && len(jpm.sourceSAS) == 0 {
+		// So now you may be thinking "Can't you copy safely from a public resource?"
+		// The answer is yes.
+		// If a user is copying from a public resource in this case, one of two things will happen:
+		// 1) setupUDAM will fail cleanly, leaving an empty UDAM instance that does nothing.
```
adreed-msft: TODO: Consider the implications of a dummy UDAM instance

ste/mgr-JobMgr.go
```diff
@@ -188,10 +192,92 @@ type jobMgr struct {
...
+			LogError: func(str string) { jm.Log(pipeline.LogError, str) },
+			Panic:    jm.Panic,
+			CallerID: "userDelegationAuthenticationManager",
+			Cancel:   jm.Cancel,
+		}
+
+		// Why craft our own credential info here?
+		// the frontend's may be impure, in the case that we're doing OAuth -> SAS
+		// This means that we'll get the OAuthTokenInfo but the credentialType will be anonymous.
+		credInfo := common.CredentialInfo{CredentialType: common.ECredentialType.OAuthToken()}
```
adreed-msft: TODO: consider a better way to get an OAuth token in the STE instead of just carrying one from STE

ste/mgr-JobMgr.go
```diff
@@ -188,10 +192,92 @@ type jobMgr struct {
...
+		// no reason for this to fail, ever.
+		common.PanicIfErr(err)
+
+		// get a service-level URL
+		srcBlobURLParts := azblob.NewBlobURLParts(*srcURL)
+		srcBlobURLParts.BlobName = ""
+		srcBlobURLParts.ContainerName = ""
+
+		// create credentials, xfer options
+		credOption := common.CredentialOpOptions{
```
adreed-msft: TODO: consider this logging strategy

It only ends up outputting the actual attempted request itself. This works to show success or failure at the creation of a UDK, but it's something that is only understood by the developer at this point.

ste/mgr-JobMgr.go
```diff
@@ -188,10 +192,92 @@ type jobMgr struct {
 
 	// only a single instance of the prompter is needed for all transfers
 	overwritePrompter *overwritePrompter
+	// only a single instance of the user delegation authentication manager is needed for all transfers
+	udam             *userDelegationAuthenticationManager
+	udamCreationLock sync.Mutex // This is a sanity check because it *may* be possible to have AddJobPart called multiple times at once if transfers get scheduled fast enough.
```
adreed-msft: TODO: Consider if this is even necessary

common/environment.go
```diff
@@ -190,7 +190,7 @@ func (EnvironmentVariable) CredentialType() EnvironmentVariable {
 func (EnvironmentVariable) DefaultServiceApiVersion() EnvironmentVariable {
 	return EnvironmentVariable{
 		Name:         "AZCOPY_DEFAULT_SERVICE_API_VERSION",
-		DefaultValue: "2018-03-28",
+		DefaultValue: "2018-11-09",
```
adreed-msft: TODO: Consider the implications of adjusting the default service API version

adreed-msft: It's necessary in this case, but I don't know if it'll cause any other problems that need to be fixed before this can ship.

cmd/sync.go
```diff
@@ -479,13 +480,24 @@ func (cca *cookedSyncCmdArgs) process() (err error) {
...
-		if tokenInfo, err := uotm.GetTokenInfo(ctx); err != nil {
-			return err
+	if cca.fromTo.IsS2S() && cca.sourceSAS == "" {
+		if cca.fromTo.From() == common.ELocation.Blob() && cca.credentialInfo.OAuthTokenInfo != (common.OAuthTokenInfo{}) {
+			glcm.Info("A SAS token is required as a part of the source in S2S sync, unless the source is a public resource. " +
+				"Because no SAS token is present on your blob source, we'll attempt to generate a SAS token at transfer time. Please note that " +
+				"transfers may fail if generating the user delegation SAS token fails.")
 		} else {
-			cca.credentialInfo.OAuthTokenInfo = *tokenInfo
+			return errors.New("a SAS token is required as a part of the source in S2S sync, unless the source is a public resource (or a blob source and you have storage blob data owner permissions)")
```
adreed-msft: TODO: Note that SBDO is not actually required here (https://docs.microsoft.com/en-us/rest/api/storageservices/get-user-delegation-key)

ste/userDelegationAuthenticationManager.go
adreed-msft: Also TODO: Consider refreshing the UDK, in case customers have long-running scenarios

adreed-msft: The latter is actually a humongous pain to do--  STE may have already taken expired SAS tokens if we don't refresh

A workaround is to just keep resuming the job whenever the token expires.

adreed-msft: WRT to this-- Perhaps passing around pointers to atomic.Values (and ensuring validity before we use a SAS token) would be the better solution in terms of handling SAS tokens in the STE if we need to refresh this.

adreed-msft: Although, thinking about it, this isn't a huge concern-- If somebody somehow runs into this, the suggestion is literally just "create a SAS token that lasts longer"

adreed-msft:
Speaking of which, this'll resolve #211

ste/userDelegationAuthenticationManager.go
adreed-msft: Deciding NOT to handle things this way. As-is, despite the lack of use of errors, there _may_ be a use-case in the future.

ste/mgr-JobMgr.go
adreed-msft: The answer is: not really, but it's still good to have this.

common/environment.go
adreed-msft: It seems fine to me, there are no changes to existing APIs, just additions.

ste/mgr-JobMgr.go
adreed-msft: Added a note here with this TODO and created a PBI. This isn't exactly what this PR is about, anyway.

ste/mgr-JobMgr.go
adreed-msft: Dummy UDAM instances are _ok_, but we should add some extra logging and stdout warnings for it.

adreed-msft:
Force-pushing to ensure PR is not stale

cmd/credentialUtil.go
```diff
@@ -286,14 +286,15 @@ func getCredentialInfoForLocation(ctx context.Context, location common.Location,
...
 
-	if credInfo.CredentialType == common.ECredentialType.OAuthToken() {
-		uotm := GetUserOAuthTokenManagerInstance()
-
-		if tokenInfo, err := uotm.GetTokenInfo(ctx); err != nil {
-			return credInfo, false, err
-		} else {
-			credInfo.OAuthTokenInfo = *tokenInfo
-		}
+	// Get the OAuth token info anyway. This may be used later in STE.
```
JohnRusk: This change doesn't smell right. Is it necessary?

If user has specifically told us on the command line NOT to use OAuth, why would we try to look up a token?

common/environment.go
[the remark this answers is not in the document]
adreed-msft: Nope. The transfers will fail as the user delegation SAS tokens can't be used on the earlier revision.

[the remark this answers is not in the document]
adreed-msft: We could however, just bump it for this type of transfer.

ste/mgr-JobMgr.go
```diff
@@ -188,10 +192,98 @@ type jobMgr struct {
...
+				RetryDelay:    UploadRetryDelay,
+				MaxRetryDelay: UploadMaxRetryDelay},
+			JobsAdmin.(*jobsAdmin).pacer,
+			jm.HttpClient(),
+			nil)
+
+		// finally, we have a service URL.
+		bsu := azblob.NewServiceURL(srcBlobURLParts.URL(), p)
+
+		// Ignoring the error. If we can't create UDAM, we don't have adequate permissions in the first place.
```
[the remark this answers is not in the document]
adreed-msft: This was a comment made prior to the err check just below. The error isn't really handled, the failure is just logged.

ste/mgr-JobPartTransferMgr.go
```diff
@@ -185,6 +186,25 @@ func (jptm *jobPartTransferMgr) GetSourceCompressionType() (common.CompressionTy
 	return common.GetCompressionType(encoding)
 }
 
+func (jptm *jobPartTransferMgr) getSrcUserDelegationSAS(src string) (string, error) {
```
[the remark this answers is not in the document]
adreed-msft: Yup, that came from the old implementation, prior to the usage of SIP

ste/userDelegationAuthenticationManager.go
```diff
@@ -0,0 +1,132 @@
...
+		sas, err := u.createUserDelegationSASForURL(blobURLParts.ContainerName)
+		return sas, err
+	}
+}
+
+func (u *userDelegationAuthenticationManager) createUserDelegationSASForURL(containerName string) (string, error) {
+	// If it's not present, we need to generate a SAS query and store it, then return.
+	sasQuery, err := azblob.BlobSASSignatureValues{
+		Version:       DefaultServiceApiVersion,
+		Protocol:      azblob.SASProtocolHTTPSandHTTP, // A user may be inclined to use HTTP for one reason or another. We already warn them about this.
```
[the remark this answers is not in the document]
JohnRusk: I unresolved it because the latest version has the constant changed (great) but still the old comment. Please fix/remove old comment at the end of the line. It only confuses readers if we have code that does one thing and comment that does another.

ste/mgr-JobMgr.go
```diff
@@ -313,6 +430,30 @@ func (jm *jobMgr) AddJobPart(partNum PartNumber, planFile JobPartPlanFileName, s
...
+		jm.initState = &jobMgrInitState{}
+
+		// For UDAM to be initialized, there are some conditions:
+		// 1) The source must not be public (It's unsafe to send the OAuth token to a public source)
+		// 2) It should be a S2S transfer from blob (There's no need for a SAS on a download, and user delegation doesn't work on files)
+		// 3) The source should lack a SAS token (There's no need for SAS generation when we already have)
+		if !jpm.Plan().SourcePublic && jpm.Plan().FromTo.IsS2S() && jpm.Plan().FromTo.From() == common.ELocation.Blob() && len(jpm.sourceSAS) == 0 {
+			// Upgrade the default version for this transfer.
+			// Note that S2S transfers with a User delegation SAS will fail on a lower revision.
+			DefaultServiceApiVersion = "2018-11-09"
```
JohnRusk: >DefaultServiceApiVersion = "2018-11-09" [](start = 3, length = 39)

Oh, we need the more recent service version not to create the user delegation SAS, but to USE it?  Have I understood that correctly.  If so, maybe my earlier suggestion not to change the defualt globally was a bad idea.  I think I'd perfer a global change as you origienally did it to changing it here like this.

common/LFUCache.go
```diff
@@ -0,0 +1,147 @@
+package common
+
+import (
+	"sync"
+)
+
+// LFUCache implements a thread-safe least frequently used map cache.
```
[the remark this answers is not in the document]
JohnRusk: This file and its related tests, are on longer used, right? If that's correct, please remove them from the PR.

ste/mgr-JobMgr.go
```diff
@@ -188,10 +198,117 @@ type jobMgr struct {
...
+		}
+
+		srcURL, err := url.Parse(src)
+
+		// no reason for this to fail, ever.
+		common.PanicIfErr(err)
+
+		// Check if the source is pointing at Azure. We shouldn't send the OAuth token anywhere else.
+		// format is account.[blob.core.windows.net], [] is the selection
+		if srcURL.Host[strings.Index(srcURL.Host, ".")+1:] != "blob.core.windows.net" {
```
JohnRusk: Not sure we can use this.  Since it doesn't work for sovereign clouds. E.g. For Azure China the accounts end in .blob.core.chinacloudapi.cn

antonipp:
Hi, has this PR been abandoned? This is a very useful feature for us that we'd definitely like to see in `azcopy`.

JohnRusk:
@zezha-msft for comment

zezha-msft:
Thanks @ antonipp for the feedbacks!

This PR has unfortunately fallen out of sync with the underlying dev branch now, so it needs to be rebased and reworked. @adreed-msft will take care of it when her bandwidth frees up.

adreed-msft:
Closing this because the PR has been refreshed: #1375

PR #710, opened 2019-10-22
Updated output for rm of ADLS Gen2

zezha-msft:
Goal: output proper JSON to allow StgExp to parse it more easily.

### DIFF
--- a/cmd/removeEnumerator.go
+++ b/cmd/removeEnumerator.go
@@ -172,6 +172,9 @@ func removeBfsResources(cca *cookedCopyCmdArgs) (err error) {
 
 			return successMsg
 		}, common.EExitCode.Success())
+
+		// explicitly exit, since in our tests Exit might be mocked away
+		return nil
 	}
 
 	// list of files is given, record the parent path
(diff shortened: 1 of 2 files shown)


cmd/removeEnumerator.go
```diff
@@ -150,29 +151,79 @@ func removeBfsResources(cca *cookedCopyCmdArgs) (successMessage string, err erro
...
+		successMsg, err := removeSingleBfsResource(urlParts, p, ctx, cca.recursive)
+		if err != nil {
+			return err
+		}
+
+		glcm.Exit(func(format common.OutputFormat) string {
+			if format == common.EOutputFormat.Json() {
+				summary := common.ListJobSummaryResponse{
+					JobStatus:          common.EJobStatus.Completed(),
+					TotalTransfers:     1,
```
JohnRusk: Is it just the format that needs to be consistent with other types of removal?  NOT the count?  E.g. remove a folder with this and the count will be 1. Remove a folder with anything else, and the count will be recursive count of contents.

zezha-msft: Yes, we couldn't get the info on how many child items were deleted. They jus need the data to be easier to process.

PR #723, opened 2019-11-04
Allow cancellation during enumeration

JohnRusk:
Works by allowing the transition to cancelled, in this case.  That will then cause application exit. So there is no change to the actual enumeration code, since it gets ended just by the application exit.

### DIFF
--- a/ste/mgr-JobMgr.go
+++ b/ste/mgr-JobMgr.go
@@ -157,8 +157,9 @@ type jobMgr struct {
 	// TODO: added for debugging purpose. remove later
 	atomicCurrentConcurrentConnections int64
 	// atomicAllTransfersScheduled defines whether all job parts have been iterated and resumed or not
-	atomicAllTransfersScheduled int32
-	atomicTransferDirection     common.TransferDirection
+	atomicAllTransfersScheduled     int32
+	atomicFinalPartOrderedIndicator int32
+	atomicTransferDirection         common.TransferDirection
 
 	concurrency          ConcurrencySettings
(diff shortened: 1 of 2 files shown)


ste/mgr-JobMgr.go
```diff
@@ -324,6 +324,26 @@ func (jm *jobMgr) AddJobPart(partNum PartNumber, planFile JobPartPlanFileName, s
...
+	oldVal := atomic.SwapInt32(&jm.atomicFinalPartOrderedIndicator, newVal)
+	if newVal == 0 && oldVal == 1 {
+		// we just cleared the flag. Sanity check that.
+		if partNum == 0 {
+			// We can't complain because, when resuming a job, there are actually TWO calls made the ResurrectJob.
+			// The effect is that all the parts are ordered... then all the parts are ordered _again_ (with new JobPartManagers replacing those from the first time)
+			// (The first resurrect is from GetJobFromTo and the second is from ResumeJobOrder)
+			// So we don't object if the _first_ part clears the flag. The assumption we make, by allowing this special case here, is that
+			// the first part will be scheduled before any higher-numbered part.  As long as that assumption is true, this is safe.
+			// TODO: do we really need to to Resurrect the job twice?
```
zezha-msft: > TODO: do we really need to to Resurrect the job twice? [](start = 5, length = 55)

I find that pretty confusing.. to clarify, is the first resurrection useless then?

JohnRusk: I think its just used to get the FromTo, and nothing else.

PR #727, opened 2019-11-05
After auth error, cancel cleanly instead of immediately exiting

JohnRusk:
So that our usual cancel-time cleanup code will run (e.g. to delete partially-completed destination files)

### DIFF
--- a/ste/mgr-JobPartTransferMgr.go
+++ b/ste/mgr-JobPartTransferMgr.go
@@ -642,7 +642,17 @@ func (jptm *jobPartTransferMgr) failActiveTransfer(typ transferErrorCode, descri
 		if status == http.StatusForbidden {
 			// quit right away, since without proper authentication no work can be done
 			// display a clear message
-			common.GetLifecycleMgr().Error(fmt.Sprintf("Authentication failed, it is either not correct, or expired, or does not have the correct permission %s", err.Error()))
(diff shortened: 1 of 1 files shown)


ste/mgr-JobPartTransferMgr.go
```diff
@@ -642,7 +642,17 @@ func (jptm *jobPartTransferMgr) failActiveTransfer(typ transferErrorCode, descri
 		if status == http.StatusForbidden {
 			// quit right away, since without proper authentication no work can be done
 			// display a clear message
-			common.GetLifecycleMgr().Error(fmt.Sprintf("Authentication failed, it is either not correct, or expired, or does not have the correct permission %s", err.Error()))
+			common.GetLifecycleMgr().Info(fmt.Sprintf("Authentication failed, it is either not correct, or expired, or does not have the correct permission %s", err.Error()))
+			// and use the normal cancelling mechanism so that we can exit in a clean and controlled way
+			jobId := jptm.jobPartMgr.Plan().JobID
+			CancelPauseJobOrder(jobId, common.EJobStatus.Cancelling())
+			// TODO: this results in the final job output line being: Final Job Status: Cancelled
+			//     That's not ideal, because it would be better if it said Final Job Status: Failed
```
zezha-msft: Reporting it as Cancelled seems fine, since that's exactly what we did. Cancellation doesn't have to be initiated by user

PR #728, opened 2019-11-05
Include SDDL library written on side

adreed-msft:
Includes tests for stringifying and parsing just about every single type, and uses a tiered structure so that errors get carried down cleanly.

There's definitely still some messiness regarding errors, and one TODO regarding stringifying nested conditionals.

### THE IMPORTANT STUFF:
- `sddl.go`
- `sid.go`
- `acls.go`
- `ace.go`
- `ace_resource_attribute_conditional.go`
- `ace_resource_attribute_values.go`

These files are the meat of the actual parsing.
`ace_resource_attribute_conditional.go` is a bit of a beast on its own. Basically, it splits a conditional up as much as possible, and then starts the real work, dividing it across as many go routines as we have CPU cores. QUESTION: Should we make this something tune-able within this library?

### THE TEST STRUCTURE:
All of the tests basically rip down an array containing input and output cases. Most stringifying tests rely upon their parsing test to succeed. If they don't, they're considered unreliable.

This is because the stringifying can be somewhat random-ordered but will ultimately still have the same effects, structure wise.

This allows us to write tests based upon what they're trying to do, rather than worrying about the structure of the test.

### THE ENUM STRUCTURE:
There are two types of enums in this library-- combinable and uncombinable.

There are .Add, .Remove, etc. functions on the combinable enums, and none of that on the uncombinables.

Enums hide their root e* type, and reveal it to the outside world with a capital E* constant. Parts of an enum are exposed via the constant's sub-functions. This prevents any messiness with the actual enum type.


adreed-msft:
Closed because we decided to curve away from this path.

PR #734, opened 2019-11-10
Allow autotuning to proceed when there are many job parts

JohnRusk:


JohnRusk:
Fixing in 10.3.2 because the bug undermines AUTO  - which is the preferred fix we currently offer for the "small network, big machine" case, where earlier versions could make too many connections.

PR #737, opened 2019-11-11
Add diagnostics around the completion of downloads

JohnRusk:
To help with debugging the issue where files are not getting deleted


zezha-msft:
re-triggering CI before merging

PR #739, opened 2019-11-13
Auto set MD5 strict checking in smoke tests

JohnRusk:
@zezha-msft can you see if this looks OK, in spite of my my sketchy Python skills

### DIFF
--- a/testSuite/scripts/utility.py
+++ b/testSuite/scripts/utility.py
@@ -25,6 +25,16 @@ class Command(object):
         if argument == None:
             return
         self.args.append(argument)
+
+        # auto-set MD5 checking flags, so that we always check when testing
+        ct = str.lower(self.command_type)
+        is_copy_or_sync = ct == "copy" or ct == "cp" or ct == "sync"
+        if is_copy_or_sync and (not str.startswith(argument, "h")):  # this is a local location
+            if len(self.args) == 1:
+                self.add_flags("put-md5", "true")  # this is an upload
+            else:
+                self.add_flags("check-md5", "FailIfDifferentOrMissing")
+
         return self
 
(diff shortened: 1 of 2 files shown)


testSuite/scripts/utility.py
```diff
@@ -25,6 +25,16 @@ def add_arguments(self, argument):
         if argument == None:
             return
         self.args.append(argument)
+
+        # auto-set MD5 checking flags, so that we always check when testing
+        ct = str.lower(self.command_type)
+        is_copy_or_sync = ct == "copy" or ct == "cp" or ct == "sync"
+        if is_copy_or_sync and (not str.startswith(argument, "h")):  # this is a local location
```
adreed-msft: >(not str.startswith(argument, "h")) [](start = 31, length = 35)

I would be a little more clear about this... Perhaps check if it doesn't start with "https://"

testSuite/scripts/utility.py
```diff
@@ -25,6 +25,16 @@ def add_arguments(self, argument):
         if argument == None:
             return
         self.args.append(argument)
+
+        # auto-set MD5 checking flags, so that we always check when testing
+        ct = str.lower(self.command_type)
+        is_copy_or_sync = ct == "copy" or ct == "cp" or ct == "sync"
```
adreed-msft: >is_copy_or_sync = ct == "copy" or ct == "cp" or ct == "sync" [](start = 8, length = 60)

I don't believe that md5 checks occur on S2S transfers, at least from S3 to Azure.

testSuite/scripts/utility.py
JohnRusk: will change to http.  (Not https, in case we sometimes only use http)

PR #740, opened 2019-11-13
Make RootDriveRegex case insensitive at is definition, instead of at sites of usage

JohnRusk:
So we know for sure its always doing what we want, and we don't have to rely on code order to give us expected capitalization beforehand


adreed-msft:
Can't approve from mobile for one reason or another, so approved

PR #743, opened 2019-11-14
Fix targeting IPv4 addresses

adreed-msft:


cmd/validators.go
```diff
@@ -130,6 +133,9 @@ func inferArgumentLocation(arg string) common.Location {
 				return common.ELocation.BlobFS()
 			case strings.Contains(host, benchmarkSourceHost):
 				return common.ELocation.Benchmark()
+				// enable targeting an emulator/stack
+			case IPv4Regex.MatchString(host):
+				return common.ELocation.Blob()
```
zezha-msft: There's no guarantee that it's Blob. 

Btw we should cover custom domains too: a specified URL (foobar.com) could point to blob/file/ADLS endpoints. I believe the right solution in this case is to take what the user has given as `--from-to` and discard the inference.

cmd/validators.go
```diff
@@ -24,14 +24,17 @@ import (
...
 	"net/url"
+	"regexp"
 	"strings"
 
 	"github.com/Azure/azure-storage-azcopy/common"
 )
 
 func validateFromTo(src, dst string, userSpecifiedFromTo string) (common.FromTo, error) {
-	inferredFromTo := inferFromTo(src, dst)
+	var inferredFromTo common.FromTo
```
JohnRusk: The goal here is to always respect what the user says, right?  This looks like it would work, but it seems like an odd way to do it, because there are still checks against inferredFromTo below, in the case where userFromTo is not "".  Those checks seem pointless now, since inferredFromTo will always be Unknown when userFromTo has a value.  I'd suggest that all usages of inferredFromTo below line 43 should probably be removed, and the declaration of inferredFromTo should be moved to line 36.

This doesn't affect correctness, but it does affect maintainability, since right now we have one part of the routine working the new way (your changes in this PR) but a whole lot of code and comments further below that appear to work the other way (but will never be triggered).. That's confusing to maintain..

cmd/validators.go
```diff
@@ -130,6 +127,9 @@ func inferArgumentLocation(arg string) common.Location {
 				return common.ELocation.BlobFS()
 			case strings.Contains(host, benchmarkSourceHost):
 				return common.ELocation.Benchmark()
+				// enable targeting an emulator/stack
+			case IPv4Regex.MatchString(host):
+				return common.ELocation.Unknown()
 			}
```
zezha-msft: To confirm my understanding, this just makes the error reporting clearer, and the user can specify their own fromtTo, correct?

adreed-msft: Yup!

PR #744, opened 2019-11-14
Catch missed properties during a S2S sync

### LINKED ISSUE #729: AZCopy Sync not setting correct content type on Blob Container to Blob Container data move.
AzCopy 10.3.1

Windows Powershell

### What command did you run?
$sourceStore = "<Source Store>"
$sourceSAS = "<Source SAS>"
$destStore = "<Destination Store>"
$destSAS = "<Destination SAS>"
$blob = "<container Name>"
$sourceURI = "https://"+"$sourceStore"+".blob.core.windows.net/"+"$blob"+"$sourceSAS"
$destURI = "https://"+"$destStore"+".blob.core.windows.net/"+"$blob"+"$destSAS"

azcopy sync "$sourceURI" "$destURI" --recursive

The Source Files are of varying file types (.jpg , .pdf ,.docx, .ppt, etc..)  and so have content types relevant to their file types e.g. 'image/jpeg' 
However  when they are sync'd to the destination store all the files are given the content type 'application/octet-stream'.

We are migrating this data from a classic storage account and Blob into an ARM based storage account and Blob

You shold be able to run the powershell provided (Adding your own stores/blob and SAS details)

We are migrating from a classic Blob stor to an  ARM Store so this may be related

-Not thus far.- 
You are able to manually change the file type through Storage explorer.. however we have 18k+ files in just one Store, and so we need an automated update route

adreed-msft:
Fixes #729

### DIFF
--- a/cmd/zc_processor.go
+++ b/cmd/zc_processor.go
@@ -75,15 +75,23 @@ func (s *copyTransferProcessor) scheduleCopyTransfer(storedObject storedObject) 
 
 	// only append the transfer after we've checked and dispatched a part
 	// so that there is at least one transfer for the final part
-	s.copyJobTemplate.Transfers = append(s.copyJobTemplate.Transfers, common.CopyTransfer{
-		Source:           s.escapeIfNecessary(storedObject.relativePath, s.shouldEscapeSourceObjectName),
-		Destination:      s.escapeIfNecessary(storedObject.relativePath, s.shouldEscapeDestinationObjectName),
-		SourceSize:       storedObject.size,
-		LastModifiedTime: storedObject.lastModifiedTime,
-		ContentMD5:       storedObject.md5,
-		BlobType:         storedObject.blobType,
-		ContentEncoding:  storedObject.contentEncoding,
-	})
+	s.copyJobTemplate.Transfers = append(s.copyJobTemplate.Transfers, common.NewCopyTransfer(
+		false,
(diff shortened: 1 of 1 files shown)


JohnRusk:
>I think John has another item to do something about this so that we don't forget properties anymore.

I see from this PR that my other item was only addressing half the problem. It was forcing us to fully construct storedObjects, but it was not forcing us to fully _use_ the properties of the storedObjects, which is what this PR is doing.  I'll see if I can add anything to my upcoming PR to help with that side of the problem too.

cmd/zc_processor.go
```diff
@@ -75,15 +75,23 @@ func (s *copyTransferProcessor) scheduleCopyTransfer(storedObject storedObject)
...
+		storedObject.size,
+		storedObject.contentType,
+		storedObject.contentEncoding,
+		storedObject.contentDisposition,
+		storedObject.contentLanguage,
+		storedObject.cacheControl,
+		storedObject.md5,
+		storedObject.Metadata,
+		storedObject.blobType,
+		storedObject.blobAccessTier,
```
JohnRusk: BTW, I just realised this is a bug.  Compare with the other usage of NewCopyTransfer. I'll create a new PR to fix it.

JohnRusk: ... because I just merged this one!

JohnRusk: The fix is in #761

PR #751, opened 2019-11-18
Manually unfurl paths to fix long file paths on UNC locations

adreed-msft:


cmd/zc_traverser_local.go
```diff
@@ -68,6 +69,43 @@ func (t *localTraverser) getInfoIfSingleFile() (os.FileInfo, bool, error) {
...
+
+		if fi.Mode()&os.ModeSymlink != 0 {
+			result, err := os.Readlink(item)
+
+			if err != nil {
+				return result, err
+			}
+
+			sameDir := filepath.Join(filepath.Dir(item), result)
+			if _, err = os.Stat(sameDir); err == nil {
```
zezha-msft: Sorry, what does this mean exactly? A comment would be nice! 😉

cmd/zc_traverser_local.go
```diff
@@ -105,7 +146,7 @@ func WalkWithSymlinks(fullPath string, walkFunc filepath.WalkFunc) (err error) {
 			computedRelativePath = strings.TrimPrefix(computedRelativePath, common.AZCOPY_PATH_SEPARATOR_STRING)
 
 			if fileInfo.Mode()&os.ModeSymlink != 0 {
-				result, err := filepath.EvalSymlinks(filePath)
+				result, err := UnfurlSymlinks(filePath)
```
JohnRusk: Does this also need to be called in the non-recursive codepath, below?

adreed-msft:
PR revived, on 10.5 backlog

cmd/zc_traverser_local.go
adreed-msft: Yes it did, I've added the call.

JohnRusk:
What's the bug that this fixes?

cmd/zc_traverser_local.go
```diff
@@ -68,6 +69,46 @@ func (t *localTraverser) getInfoIfSingleFile() (os.FileInfo, bool, error) {
...
+
+			if err != nil {
+				return result, err
+			}
+
+			// We need to discern whether the item symlinked is under the same directory as the symlink.
+			// This is because when this happens, we only get a relative path, not a full path.
+			// So we'll check for the existence of the item under the relative path, and otherwise treat it as a full path.
+			sameDir := filepath.Join(filepath.Dir(item), result)
+			if _, err = os.Stat(sameDir); err == nil {
```
JohnRusk, line 95: Two thoughts/questions about this line:

1. I'm a bit surprised that the call is to Stat. I would have (just intuitively) expected LStat here.  In case it's a symlink to a symlink.  In that case, I'm not totally sure the current code would be wrong. But it does feel odd. What do you think?

2. Is there anything at the start of the path, that is returned by ReadLink, that would tell us if this is, syntactically, a relative path. E.g. does it start with "."?  Asking because I feel a bit uncomfortable with just appending and assuming its relative if the append points to something that exists. It's theoretically possible that appending an absolute (non-relatev) path may also point to something that exists, almost by chance. E.g. we are in //foo/bar and ReadLink returns //aaa/bbb.  If a file or directory //foo/bar/aaa/bbb happens to exist, I'm pretty sure that we will then mistakenly thing that the path was relative. (I think the double //'s before the aaa will be treated as a single / by Windows, but I could be wrong on that).

What about calling filepath.IsAbs() on the value returned by ReadLink, instead of probing for existence?  Would that work?

adreed-msft:
> What's the bug that this fixes?

Multi-symlinking into a long file path

cmd/zc_traverser_local.go
adreed-msft: I don't entirely agree with the safeness of this either.

I'm going to do a bit of testing and see if I can syntactically identify a relative path without too much pain. I recall that it felt tricky and like a bit of a complication, however.

adreed-msft: I can't verify this was ever super necessary, oddly. I must've had to do it for some reason.

cmd/zc_enumerator.go
```diff
@@ -659,9 +659,24 @@ func passedFilters(filters []objectFilter, storedObject storedObject) bool {
...
+	}
+
+	return false, err
+}
+
 func processIfPassedFilters(filters []objectFilter, storedObject storedObject, processor objectProcessor) (err error) {
 	if passedFilters(filters, storedObject) {
 		err = processor(storedObject)
+	} else {
+		err = ignoredError
```
zezha-msft, line 679: I'm a bit confused here, the comment above says that the ignoredError is only used for symlink processing, but here it appears that any storedObject that didn't pass the filters (e.g. include/exclude) would get it. 

I'd imagine that we'd

zezha-msft: We've clarified the implementation offline: the ignoredError is all about the situation where the include/exclude flag interact with a folder and a symlink that points to it; they need to be considered separately.

PR #757, opened 2019-11-25
Improve proxy lookup performance and reliability

JohnRusk:


common/ProxyLookupCache.go
```diff
@@ -0,0 +1,166 @@
...
+	if c.refreshInterval == 0 {
+		return
+	}
+
+	for {
+		time.Sleep(c.refreshInterval)
+		value := c.getProxyNoCache(representativeFullRequest)
+		c.m.Store(key, value)
+	}
+}
```
adreed-msft: This could be simplified with a time.Ticker and a list of URLs, potentially.

JohnRusk: Yes, I think if where were lots of URLs expected, the we'd have to do something like that.  Or maybe just "range" through the contents of the map.

In our case, where will usually only be 1 or 2 URLs in the map, so I'm leaning towards leaving it as-is.

common/ProxyLookupCache.go
```diff
@@ -0,0 +1,166 @@
...
+	err error
+}
+
+// proxyLookupCache caches the result of proxy lookups
+// It's here, rather than contributed to mattn.go-ieproxy because it's not general-purpose.
+// In particular, it assumes that there will not be lots and lots of different hosts in the cache,
+// and so it has no ability to clear them or reduce the size of the cache, and it runs one
+// permanent GR per cache entry. That assumption makes sense in AzCopy, but is not correct in general (e.g.
+// if used in something with usage patterns like a web browser).
+// TODO: should we one day find a better solution, so it can be contributed to mattn.go-ieproxy instead of done here?
```
zezha-msft: >TODO [](start = 3, length = 4)

Our scenario indeed seems way too specific for go-ieproxy

PR #759, opened 2019-11-27
Added configuration for user agent prefix

zezha-msft:


cmd/credentialUtil.go
```diff
@@ -420,7 +420,7 @@ func createFilePipeline(ctx context.Context, credInfo common.CredentialInfo) (pi
 				MaxRetryDelay: ste.UploadMaxRetryDelay,
 			},
 			Telemetry: azfile.TelemetryOptions{
-				Value: common.UserAgent,
+				Value: glcm.GetEnvironmentVariable(common.EEnvironmentVariable.UserAgentPrefix()) + common.UserAgent,
```
JohnRusk: Minor: might be nice to have a function 
```
func glcm.AddUserAgenPrefix(rawUserAgentString string) {
   return ... //join the env var with the passed inrawUserAgentString
}

File: ste/mgr-JobPartMgr.go [hunk 2]
```diff
@@ -436,6 +436,7 @@ func (jpm *jobPartMgr) createPipelines(ctx context.Context) {
 	} else if fromTo.From() == common.ELocation.Benchmark() || fromTo.To() == common.ELocation.Benchmark() {
 		userAgent = common.BenchmarkUserAgent
 	}
+	userAgent = common.GetLifecycleMgr().GetEnvironmentVariable(common.EEnvironmentVariable.UserAgentPrefix()) + userAgent
```
adreed-msft: >userAgent = common.GetLifecycleMgr().GetEnvironmentVariable(common.EEnvironmentVariable.UserAgentPrefix()) + userAgent [](start = 1, length = 118)

You might want to consider adding a space in between these two if one isn't present in the env var-- Otherwise it might come out like "StgExpAzCopy" rather than "StgExp AzCopy..."

common/environment.go
```diff
@@ -47,6 +47,7 @@ var VisibleEnvironmentVariables = []EnvironmentVariable{
 	EEnvironmentVariable.CertificatePassword(),
 	EEnvironmentVariable.AutoTuneToCpu(),
 	EEnvironmentVariable.CacheProxyLookup(),
+	EEnvironmentVariable.UserAgentPrefix(),
```
JohnRusk: Are you sure we want this visible?  Previously I assumed you'd left it hidden on purpose, so that we would be less likely to get customers adding random prefixes that we are not interested in.

JohnRusk: Ze replied offline: sometimes people start integrating or embedding AzCopy without even telling us, so its good that this is discoverable for them.

PR #760, opened 2019-11-29
Force all property setting for storedObject to go through the constructor

JohnRusk:
So we can't forget any necessary properties, and to make the call sites more concise

### DIFF
--- a/cmd/zc_enumerator.go
+++ b/cmd/zc_enumerator.go
@@ -47,6 +47,7 @@ import (
 
 // represent a local or remote resource object (ex: local file, blob, etc.)
 // we can add more properties if needed, as this is easily extensible
+// ** DO NOT instantiate directly, always use newStoredObject ** (to make sure its fully populated and any preprocessor method runs)
 type storedObject struct {
 	name             string
 	lastModifiedTime time.Time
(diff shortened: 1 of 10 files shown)


cmd/zc_enumerator.go
```diff
@@ -47,6 +47,7 @@ import (
 
 // represent a local or remote resource object (ex: local file, blob, etc.)
 // we can add more properties if needed, as this is easily extensible
+// ** DO NOT instantiate directly, always use newStoredObject ** (to make sure its fully populated and any preprocessor method runs)
```
zezha-msft: >its [](start = 79, length = 3)

minor: it's

cmd/zc_traverser_blob.go
```diff
@@ -105,21 +105,12 @@ func (t *blobTraverser) traverse(preprocessor objectMorpher, processor objectPro
 			"",
 			blobProperties.LastModified(),
 			blobProperties.ContentLength(),
-			blobProperties.ContentMD5(),
-			blobProperties.BlobType(),
+			blobProperties,
+			blobPropertiesResponseAdapter{blobProperties},
```
zezha-msft: >blobPropertiesResponseAdapter{blobProperties}, [](start = 3, length = 46)

this is nice!

PR #761, opened 2019-11-29
Fix/make tier processing consistent between copy and sync

JohnRusk:
Without this, there'll be a bug in that it will impossible to, for instance, sync between a standard blob storage account and a premium block blob account (because they don't have the same access tiers).

Note that there are two commits in this PR. Only the first is technically necessary to fix the bug. But the second reduces code duplication and should improve maintainability as we add more properties in future.  The second basically just takes `common.NewCopyTransfer` and makes it a method of `storedObject` so we don't have to pass heaps of parameters invididually


adreed-msft:
Should comfortably reduce some code dupe between sync and copy.

PR #763, opened 2019-12-02
Benchmark mode for Azure Files and ADLS Gen2

JohnRusk:


zezha-msft:
Re-triggering CI before merging.

PR #771, opened 2019-12-05
Improve azcopy's logging when the JPPFN sanity check is hit

adreed-msft:


ste/JobPartPlanFileName.go
```diff
@@ -215,6 +216,10 @@ func (jpfn JobPartPlanFileName) Create(order common.CopyJobPartOrderRequest) {
 
 	// Write each transfer to the Job Part Plan file (except for the src/dst strings; comes come later)
 	for t := range order.Transfers {
+		if len(order.Transfers[t].Source) > math.MaxInt16 || len(order.Transfers[t].Destination) > math.MaxInt16 {
+			panic(fmt.Sprintf("The file %s exceeds azcopy's current maximum path length on either the source or the destination.", order.Transfers[t].Source))
+		}
+
```
JohnRusk: Can you add the same check for metadata length please?

PR #780, opened 2019-12-09
Add draft code for detection of early EOFs

JohnRusk:


JohnRusk:
Closing, since will not progress this in 10.0.4

PR #782, opened 2019-12-09
Prevent race detector triggering on proxy cache test code

JohnRusk:


JohnRusk:
Merging on a single approval, to get builds running again.

PR #789, opened 2019-12-12
Converted numbers to strings in JSON output

zezha-msft:
JS is a funny language. 🤣

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,13 @@
 
 # Change Log
 
+## Version 10.x.x
+
+### Breaking changes
+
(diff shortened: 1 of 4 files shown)


adreed-msft:
```
[] == 0
0 == "0"
"0" == []
```

Try running those lines in a js console 👀

zezha-msft:
I need to update our smoke tests which depend on the JSON output.

adreed-msft:
Good catch!

adreed-msft:
Is there anything else that might rely upon this json output?

zezha-msft:
@adreed-msft yes, I think the other partner teams will have to make some changes when we release. So we have to call this out as a breaking change.

PR #792, opened 2019-12-13
Safely catch potential errors with the destination blob tier

adreed-msft:


zezha-msft:
Nice, this was a pain point for many customers.

An automated smoke test would be nice.

common/fe-ste-models.go
```diff
@@ -595,6 +595,8 @@ func (TransferStatus) SkippedFileAlreadyExists() TransferStatus { return Transfe
 
 func (TransferStatus) SkippedBlobHasSnapshots() TransferStatus { return TransferStatus(-4) }
 
+func (TransferStatus) TierAvailabilityCheckFailure() TransferStatus { return TransferStatus(-5) }
```
JohnRusk: Please add to the changelog a point that says that this new status can be returned (ssince they are visible in azcopy jobs show and, IIRC, in the json format output

go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/Azure/azure-storage-azcopy
 
 require (
 	github.com/Azure/azure-pipeline-go v0.2.1
-	github.com/Azure/azure-storage-blob-go v0.7.0
+	github.com/Azure/azure-storage-blob-go v0.8.1-0.20191213204130-762620a866ba
```
JohnRusk: What's the plan for getting back onto a released version?  If that plan requires us to remember to do something later, do we have a backlog entry to serve as that reminder?

ste/sender-pageBlob.go
```diff
@@ -220,17 +223,41 @@ func (s *pageBlobSenderBase) Prologue(ps common.PrologueState) (destinationModif
...
+
+			// First, let's check if the intended tier is a premium page blob tier.
+			// These aren't available on non-premium storage accounts, but we can still copy the data.
+
+			shouldSet := true
+			if premiumPageBlobTierRegex.MatchString(string(s.destBlobTier)) {
+				infoResp, err := s.destPageBlobURL.GetAccountInfo(ctxWithLatestServiceVersion)
+				if err != nil {
+					// If GetAccountInfo fails, this transfer should fail because we lack at least one available permission
+					// (https://docs.microsoft.com/en-us/rest/api/storageservices/get-account-information#authorization)
```
JohnRusk: I notice that the doc page only mentions shared key and SAS auth. For that reason, can you please test it with OAuth and note the result here.

ste/sender-pageBlob.go
```diff
@@ -220,17 +223,41 @@ func (s *pageBlobSenderBase) Prologue(ps common.PrologueState) (destinationModif
...
-						glcm := common.GetLifecycleMgr()
-						glcm.Error("One or more blobs have failed blob tier replication at the destination. Try transferring with the flag --s2s-preserve-access-tier=false")
-					})
+
+			// First, let's check if the intended tier is a premium page blob tier.
+			// These aren't available on non-premium storage accounts, but we can still copy the data.
+
+			shouldSet := true
+			if premiumPageBlobTierRegex.MatchString(string(s.destBlobTier)) {
+				infoResp, err := s.destPageBlobURL.GetAccountInfo(ctxWithLatestServiceVersion)
```
JohnRusk: It would be preferrable to get the account information once for the source account, and once for the destination.  Rather than once for every blob.  THis may not be an issue on page blobs, but it will be an issue on block blobs. As I'll note in another comment. 
I think ideally we'd compute some kind of boolean flag saying whether the source and dest accounts have compatible tier types (maybe wrapped with a sync.Once) and then just drive most of the logic here off that flag.

go.mod
adreed-msft: It will require us to make a release of the blob SDK before azcopy, and to fix up the go.mod for it.

I'll create the backlog item after I address other comments on this PR.

ste/sender-pageBlob.go
adreed-msft: Because not all destination blobs will have the same type or tier in case of S2S preservation, we can't just run the logic once. We can however, just grab the account details once.

adreed-msft:
You're right in that we need it to handle pageblob access tier. However, there aren't separate premium/standard block blob access tiers.

ste/sender-pageBlob.go
adreed-msft: It fails. I have added an alternate code path in case the user is potentially unable to obtain the account details. Even if set tier fails, the transfer will still succeed via OAuth.

adreed-msft:
@zezha-msft I know you already reviewed, but significant enough changes have occurred to justify a re-review.

ste/xfer-anyToRemote.go
```diff
@@ -30,12 +31,78 @@ import (
...
+func prepareDestAccountInfo(bURL azblob.BlobURL, jptm IJobPartTransferMgr, ctx context.Context, mustGet bool) {
+	getDestAccountInfo.Do(func() {
+		infoResp, err := bURL.GetAccountInfo(ctx)
+		if err != nil {
+			// If GetAccountInfo fails, this transfer should fail because we lack at least one available permission
+			// UNLESS the user is using OAuth. In which case, the account owner can still get the info.
+			// If we fail to get the info under OAuth, don't fail the transfer.
+			// (https://docs.microsoft.com/en-us/rest/api/storageservices/get-account-information#authorization)
+			if mustGet {
+				jptm.FailActiveSendWithStatus("Checking destination tier availability (Set PageBlob tier) ", err, common.ETransferStatus.TierAvailabilityCheckFailure())
```
JohnRusk: It seems odd to use FailActiveSendWithStatus inside a sync.Once.  Would it make more sense just to set up destAccountSKU, destAccountKind and *destAccountTypeError* inside the Once, and then use FailActiveSendWithStatus outside the sync.Once?

common/fe-ste-models.go
JohnRusk: I just realised, adding or removing a transfer status requires changes to the aggregation logic in `GetJobSummary`.  But do we really want to do that?  What about just using the existing failure status?  Especially given that these types of failure will be rare.

ChangeLog.md
```diff
@@ -3,6 +3,10 @@
 
 ## Version 10.x.x
 
+### New features
+
+1. Azcopy can now detect when setting a blob tier would be impossible. If azcopy cannot check the destination account type, a new job failure status will be set: `TierAvailabilityCheckFailure`
```
zezha-msft: >job [](start = 130, length = 3)

*it's a transfer status, not a job status

common/fe-ste-models.go
adreed-msft: This actually is just one line.
image

ste/xfer-anyToRemote.go
```diff
@@ -30,12 +31,83 @@ import (
...
+		}
+	})
+
+	if getDestAccountInfoError != nil {
+		jptm.FailActiveSendWithStatus("Checking destination tier availability (Set PageBlob tier) ", getDestAccountInfoError, common.ETransferStatus.TierAvailabilityCheckFailure())
+	}
+}
+
+// TODO: Infer availability based upon blob size as well, for premium page blobs.
+func BlobTierAllowed(destTier azblob.AccessTierType) bool {
```
zezha-msft: >BlobTierAllowed [](start = 5, length = 15)

Normally I don't recommend this kind of client-side logic, but in this case I think it's fine since we can only have new rules with new service versions, i.e. no cx will be blocked because the service rolled out a new rule

adreed-msft:
Reviving pull request

ste/xfer-anyToRemote-file.go
```diff
@@ -31,12 +32,83 @@ import (
...
+		if strings.Contains(destAccountKind, "Block") {
+			// No tier setting is allowed.
+			return false
+		}
+
+		// Any other storage type would have to be file storage, and we can't set tier there.
+		panic("Cannot set tier on azure files.")
+	} else {
+		// Standard storage account. If it's Hot, Cool, or Archive, we're A-OK.
+		return destTier == azblob.AccessTierArchive || destTier == azblob.AccessTierCool || destTier == azblob.AccessTierHot
```
JohnRusk, line 108: Hot etc is not valid on Standard for Page blobs.  Our code in sender-pageBlob already prevents us from sending one of these if the destination is a page blob. So on that basis is probably not necessary to change the code here  (unless you want to, for completeness). But maybe a comment here about page blobs would help future readers.

ste/sender-blockBlob.go
```diff
@@ -153,19 +153,38 @@ func (s *blockBlobSenderBase) Epilogue() {
...
-		ctxWithLatestServiceVersion := context.WithValue(jptm.Context(), ServiceAPIVersionOverride, azblob.ServiceVersion)
-		_, err := s.destBlockBlobURL.SetTier(ctxWithLatestServiceVersion, s.destBlobTier, azblob.LeaseAccessConditions{})
-		if err != nil {
-			if s.jptm.Info().S2SSrcBlobTier != azblob.AccessTierNone {
-				s.jptm.LogTransferInfo(pipeline.LogError, s.jptm.Info().Source, s.jptm.Info().Destination, "Failed to replicate blob tier at destination. Try transferring with the flag --s2s-preserve-access-tier=false")
-				s2sAccessTierFailureLogStdout.Do(func() {
-					glcm := common.GetLifecycleMgr()
-					glcm.Error("One or more blobs have failed blob tier replication at the destination. Try transferring with the flag --s2s-preserve-access-tier=false")
-				})
+		ctxWithLatestServiceVersion := context.WithValue(s.jptm.Context(), ServiceAPIVersionOverride, azblob.ServiceVersion)
```
JohnRusk, line 156: This code is now duplicated in the blockBLob and page blob senders. At about 35 lines, that's more duplication than we'd normally be comfortable with.  Can you refactor it into a routine that exists in one place, and gets called by the block and page blob senders?

JohnRusk: PS. The reason is not just that these are 35 duplicated lines.  The other part of the reason is that we have to be very vigilant about duplicated code. If we're not careful about it, we'll end up with a whole lot of duplication and that leads to maintenance problems. (As we saw, prior to the various refactorings of 2019, since before those refactorings duplication was a problem).

ste/sender-blockBlob.go
```diff
@@ -153,19 +153,38 @@ func (s *blockBlobSenderBase) Epilogue() {
...
+
+		// Let's check if we can confirm we'll be able to check the destination blob's account info.
+		// A SAS token, even with write-only permissions is enough. OR, OAuth with the account owner.
+		// We can't guess that last information, so we'll take a gamble and try to get account info anyway.
+		destParts := azblob.NewBlobURLParts(s.destBlockBlobURL.URL())
+		mustGet := destParts.SAS.Encode() != ""
+
```
JohnRusk, line 163: Nice

ste/sender-blockBlob.go
```diff
@@ -153,19 +153,38 @@ func (s *blockBlobSenderBase) Epilogue() {
...
+		prepareDestAccountInfo(s.destBlockBlobURL.BlobURL, s.jptm, ctxWithLatestServiceVersion, mustGet)
+		tierAvailable := BlobTierAllowed(s.destBlobTier)
+
+		if tierAvailable {
+			_, err := s.destBlockBlobURL.SetTier(ctxWithLatestServiceVersion, s.destBlobTier, azblob.LeaseAccessConditions{})
+			if err != nil {
+				if s.jptm.Info().S2SSrcBlobTier != azblob.AccessTierNone {
+					s.jptm.LogTransferInfo(pipeline.LogError, s.jptm.Info().Source, s.jptm.Info().Destination, "Failed to replicate blob tier at destination. Try transferring with the flag --s2s-preserve-access-tier=false")
```
JohnRusk, line 171: Does having a SrcAccess tier really proved that the thing that failed was the attempt to replicate (i.e. copy) the tier?  
 I don't think it does prove that, because what if the user had set an explicit destination tier using a command line parameter? They could have specified one that's not available at the destination.  But error messages saying "replicate" would be confusing in that case, because replicate means "copy", but this is not a case of copying the tier from the source.

If we want to give an error specific to replication, I would have expected the if statement to say something like `if s.jptm.Info().S2SSrcBlobTier == s.destBlobTier`.  That will return true if we are indeed _replicating_ the tier.

(When writing the above, I'm _assuming_ that the command line parameters affect s.destBlobTier but not S2SSRcBlobTier. You should check my assumption before relying on it ;-)

ste/sender-blockBlob.go
```diff
@@ -153,19 +153,38 @@ func (s *blockBlobSenderBase) Epilogue() {
...
+					s2sAccessTierFailureLogStdout.Do(func() {
+						glcm := common.GetLifecycleMgr()
+						glcm.Error("One or more blobs have failed blob tier replication at the destination. Try transferring with the flag --s2s-preserve-access-tier=false")
```
JohnRusk, line 174: Why is this one glcm.Error but the other copy of this code has glcm.Info?  I would have thought both should be info. (If that's correct, it may indicate why avoiding the code duplication will help with maintenance).

ste/sender-blockBlob.go
```diff
@@ -153,19 +153,38 @@ func (s *blockBlobSenderBase) Epilogue() {
...
+					})
+				}
+
+				// If we know the destination tier is possible, something's wrong and we should error out.
+				if !tierSetPossibleFail {
```
JohnRusk, line 179: Minor: if a if statement contains both an if section and an else section, it's conventional to do the test in the affirmative (i.e. in this case `if tierSetPossibleFail`) and put the "true" case first.  I.e. to remove the ! and reverse the two cases.  That conventional generally makes the code easier to read.

ste/xfer-anyToRemote-file.go
```diff
@@ -31,12 +32,83 @@ import (
...
+	// If the account is premium, Storage/StorageV2 only supports page blobs (Tiers P1-80). Block blob does not support tiering whatsoever.
+	if strings.Contains(destAccountSKU, "Premium") {
+		// storage V1/V2
+		if destAccountKind == "StorageV2" {
+			// P1-80 possible.
+			return premiumPageBlobTierRegex.MatchString(string(destTier))
+		}
+
+		if destAccountKind == "Storage" {
+			// No tier setting is allowed.
```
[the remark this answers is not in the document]
adreed-msft: Nope, I'm certain.

```
2020/07/01 22:42:49 ==> REQUEST/RESPONSE (Try=1/37.4951ms, OpTime=37.4951ms) -- RESPONSE STATUS CODE ERROR
   PUT https://adreedstgv1.blob.core.windows.net/test/test.vhd?comp=tier&se=2020-07-02t22%3A26%3A12z&sig=-REDACTED-&sp=racwdl&sr=c&st=2020-07-01t22%3A26%3A12z&sv=2019-02-02&timeout=901
   User-Agent: [AzCopy/10.4.86-preview Azure-Storage/0.7 (go1.14.3; Windows_NT)]
   X-Ms-Access-Tier: [Hot]
   X-Ms-Client-Request-Id: [d5c2e113-965d-455e-4dc5-cbcd333d8ce4]
   X-Ms-Version: [2018-11-09]
   RESPONSE Status: 400 Blob access tier is not supported on this storage account type.
   Content-Length: [272]
   Content-Type: [application/xml]
   Date: [Wed, 01 Jul 2020 22:42:47 GMT]
   Server: [Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0]
   X-Ms-Error-Code: [BlobAccessTierNotSupportedForAccountType]
   X-Ms-Request-Id: [a5088d9d-201e-001a-1cf8-4f8696000000]
   X-Ms-Version: [2018-11-09]
Response Details: ﻿<Code>BlobAccessTierNotSupportedForAccountType</Code><Message>Blob access tier is not supported on this storage account type. </Message>
```

ste/xfer-anyToRemote-file.go
```diff
@@ -31,12 +32,83 @@ import (
...
+				destAccountKind = "failget"
+			}
+		} else {
+			destAccountSKU = string(infoResp.SkuName())
+			destAccountKind = string(infoResp.AccountKind())
+		}
+	})
+
+	if getDestAccountInfoError != nil {
+		jptm.FailActiveSendWithStatus("Checking destination tier availability (Set PageBlob tier) ", getDestAccountInfoError, common.ETransferStatus.TierAvailabilityCheckFailure())
```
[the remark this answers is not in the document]
adreed-msft: Typo likely

ste/sender-blockBlob.go
adreed-msft: This really shouldn't be error anyway, since that kills azcopy.

ste/sender-blockBlob.go
adreed-msft: Your assumption was right, and I've implemented it.

ste/sender-blockBlob.go
adreed-msft: The code is now de-duped.

ste/xfer-anyToRemote-file.go
adreed-msft: I think it's fine to leave this. I'm going to add a note here.

ste/xfer-anyToRemote-file.go
```diff
@@ -21,22 +21,140 @@
...
+						glcm := common.GetLifecycleMgr()
+						glcm.Info("One or more blobs have failed blob tier replication at the destination. Try transferring with the flag --s2s-preserve-access-tier=false")
+					})
+				}
+
+				// If we know the destination tier is possible, something's wrong and we should error out.
+				if tierSetPossibleFail {
+					jptm.LogTransferInfo(pipeline.LogWarning, jptm.Info().Source, jptm.Info().Destination, "Cannot set destination block blob to the pending access tier ("+string(blobTier)+"), because either the destination account or blob type does not support it. The transfer will still succeed.")
+				} else {
+					jptm.FailActiveSendWithStatus("Setting BlockBlob tier", err, common.ETransferStatus.BlobTierFailure())
```
zezha-msft, line 150: >BlockBlob [](start = 44, length = 9)

minor: not necessarily block blob

ChangeLog.md
```diff
@@ -90,6 +90,10 @@ contained directly in a container or virtual directory include `/*` at the end o
    you should only add Microsoft Azure domains. 
 1. When transferring over a million files, AzCopy will reduces its progress reporting frequency from every 2 seconds to every 2 minutes.   
 
+### New features
+
+1. Azcopy can now detect when setting a blob tier would be impossible. If azcopy cannot check the destination account type, a new transfer failure status will be set: `TierAvailabilityCheckFailure`
+
```
zezha-msft, line 96: minor: not at the right place

cg6:
This appears to be the source of this regression: https://github.com/Azure/azure-storage-azcopy/issues/1116
When using a Container based SAS token the account check fails.

Issue #801, opened 2019-12-20
Receiving runtime error: invalid memory address or nil pointer dereference when using azcopy copy on blob container


tsuggs1:

tsuggs1:
![Screen Shot 2019-11-22 at 11 59 54 AM](https://user-images.githubusercontent.com/59095744/71278127-b4f56f80-2324-11ea-9c1c-4e87427a192c.png)
<img width="1006" alt="Screen Shot 2019-11-25 at 9 19 53 PM" src="https://user-images.githubusercontent.com/59095744/71278131-ba52ba00-2324-11ea-8eb3-1b37307d768f.png">


Screenshots for reference.

tsuggs1:
<img width="641" alt="Screen Shot 2019-12-20 at 12 27 55 PM" src="https://user-images.githubusercontent.com/59095744/71278170-cfc7e400-2324-11ea-9185-42833cb82428.png">
<img width="636" alt="Screen Shot 2019-12-20 at 12 27 32 PM" src="https://user-images.githubusercontent.com/59095744/71278174-d3f40180-2324-11ea-847a-3fede7e4ccf8.png">

tsuggs1:
Lastly, I have attempted 3 times and each time it has copied a different number of files before throwing that same error. THe first time it was about 74,000 files, then 64,000 then 43,000.

Thanks!

adreed-msft:
Hi there, could you show the full stack trace that azcopy dumped when it crashed? Thanks!

tsuggs1:
Hi,

So I am actually running it again to replicate the issue and so far so good.. I will update shortly if it ends up getting all the way through.

In the meantime, is it normal for there to be a number for 'Failed' and what it that means? I also think we have over 2 million blob items in this blob container, would there be any issue why it wouldn't be able to copy all of this?

Thanks for the help!

tsuggs1:
Unfortunately the same error came up:
<img width="643" alt="Screen Shot 2019-12-20 at 4 39 02 PM" src="https://user-images.githubusercontent.com/59095744/71294737-5ee5f380-2347-11ea-96ee-4f94982b19c6.png">


Not sure where to locate that full stack trace, I did find a file in the logs but it's ~400 MB so it won't allow me to attach it here. Let me know if that's what you're looking for.

xuetao8997:
#### Yes, i really got the same error, the detail as follows:
panic: runtime error: invalid memory address or nil pointer dereference
[...]

PR #808, opened 2019-12-30
Fix nil derefs users were reporting to support further investigation of their cases

adreed-msft:
NOTE: This PR doesn't entirely _fix_ the users' cases in #801, but it'll certainly help with us understanding what's going wrong in their cases, by fixing some cases of nil derefs that were causing dirty crashes.

### DIFF
--- a/ste/xfer-anyToRemote.go
+++ b/ste/xfer-anyToRemote.go
@@ -226,6 +226,7 @@ func scheduleSendChunks(jptm IJobPartTransferMgr, srcPath string, srcFile common
 				// As long as the prefetch error is nil, we'll attempt a prefetch.
 				// Otherwise, the chunk reader didn't need to be made.
 				// It's a waste of time to prefetch here, too, if we already know we can't upload.
+				// Furthermore, this prevents prefetchErr changing from under us.
 				if prefetchErr == nil {
 					// create reader and prefetch the data into it
 					chunkReader = createPopulatedChunkReader(jptm, sourceFileFactory, id, adjustedChunkSize, srcFile)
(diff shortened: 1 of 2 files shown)


ste/xfer-anyToRemote.go
```diff
@@ -223,16 +223,21 @@ func scheduleSendChunks(jptm IJobPartTransferMgr, srcPath string, srcFile common
 			if jptm.WasCanceled() {
 				prefetchErr = jobCancelledLocalPrefetchErr
 			} else {
-				// create reader and prefetch the data into it
-				chunkReader = createPopulatedChunkReader(jptm, sourceFileFactory, id, adjustedChunkSize, srcFile)
-
-				// Wait until we have enough RAM, and when we do, prefetch the data for this chunk.
-				prefetchErr = chunkReader.BlockingPrefetch(srcFile, false)
+				// As long as the prefetch error is nil, we'll attempt a prefetch.
+				// Otherwise, the chunk reader didn't need to be made.
```
JohnRusk: Nice

ste/xfer-anyToRemote.go
```diff
@@ -259,9 +264,15 @@ func scheduleSendChunks(jptm IJobPartTransferMgr, srcPath string, srcFile common
 				if chunkReader != nil {
 					_ = chunkReader.Close()
 				}
+
+				// Copy the error so that we can avoid the prefetch err being overwritten under us.
```
JohnRusk: Doesn't the code above stop it from ever being changed?  In which case, why introduce this func here, to protect us from changes in something that will no longer change?

And if we did have to copy the error, why not do this (simpler than making separate func)

copyOfError := prefetchErr
cf = createSendToRemoteChunkFunc(jptm, id, func() { jptm.FailActiveSend("chunk data read", copyOfError ) })

adreed-msft: It _shouldn't_ be overwritten. I'll remove this.

Issue #811, opened 2019-12-31
Precheck Missing MD5 for FailIfDifferentOrMissing


justinormont:
Currently the remote MD5 value is checked, after the file download is finished, for both _different_ and _missing_.

When the remote MD5 is missing, the download process can take hours, then fail.

**Request**:
Would it be possible to pre-check for the existence of the remote MD5 value then fail fast, before the download. This would save time.

MD5 check for `--check-md5=FailIfDifferentOrMissing`:
https://github.com/Azure/azure-storage-azcopy/blob/d4e856e572f7f63521329703d18d9f96bac21613/ste/md5Comparer.go#L64-L65

**Repro**:
`azcopy.exe cp "https://lilablobssc.blob.core.windows.net/snapshotserengeti-v-2-0/SnapshotSerengeti_S10_v2_0_part2.zip" "F:\datasets\SnapshotSerengeti\raw\" --check-md5=FailIfDifferentOrMissing --overwrite=false`

File is 166GB and public ([src](http://lila.science/datasets/snapshot-serengeti)).

**Error**: _(only in log; should be raised to user level)_
`000 : no MD5 was stored in the Blob/File service against this file. So the downloaded data cannot be MD5-validated. This application is currently configured to treat missing MD5 hashes as errors. When Checking MD5 hash. X-Ms-Request-Id: `

adreed-msft:
Whoof, yeah, hours is pretty rough for this kind of a thing. I think we could perform that early during transfer init. I'll pop open a PR for that here in a moment.

JohnRusk:
Closing since this work has now been done.

PR #812, opened 2019-12-31
Check if the MD5 is missing before starting the transfer to save time

adreed-msft:
Saves a lot of time in the case of extra-large files (#811)

### DIFF
--- a/ste/xfer-remoteToLocal.go
+++ b/ste/xfer-remoteToLocal.go
@@ -73,6 +73,17 @@ func remoteToLocal(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, d
 		}
 	}
 
+	if jptm.MD5ValidationOption() == common.EHashValidationOption.FailIfDifferentOrMissing() {
+		// We can make a check early on MD5 existence and fail the transfer if it's not present.
+		// This will save hours in the event a user has say, a several hundred gigabyte file.
+		if len(info.SrcHTTPHeaders.ContentMD5) == 0 {
+			jptm.LogDownloadError(info.Source, info.Destination, errExpectedMd5Missing.Error(), 0)
+			jptm.SetStatus(common.ETransferStatus.Failed())
+			jptm.ReportTransferDone()
+			return
+		}
+	}
+
 	// step 4a: mark destination as modified before we take our first action there (which is to create the destination file)
 	jptm.SetDestinationIsModified()
 
--- a/testSuite/scripts/test_service_to_service_copy.py
+++ b/testSuite/scripts/test_service_to_service_copy.py
@@ -1147,12 +1147,17 @@ class Service_2_Service_Copy_User_Scenario(unittest.TestCase):
         validate_dir_name = "validate_copy_file_from_%s_bucket_to_%s_bucket_propertyandmetadata_%s" % (srcType, dstType, preserveProperties)
         local_validate_dest_dir = util.create_test_dir(validate_dir_name)
         local_validate_dest = local_validate_dest_dir + fileName
+        # Because the MD5 is checked early, we need to clear the check-md5 flag.
         if srcType == "S3":
             result = util.Command("copy").add_arguments(dstFileURL).add_arguments(local_validate_dest). \
-                add_flags("log-level", "info").execute_azcopy_copy_command()
+                add_flags("log-level", "info")  # Temporarily set result to Command for the sake of modifying the md5 check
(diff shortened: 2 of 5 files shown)


ste/xfer-remoteToLocal.go
```diff
@@ -73,6 +73,17 @@ func remoteToLocal(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, d
 		}
 	}
 
+	if jptm.MD5ValidationOption() == common.EHashValidationOption.FailIfDifferentOrMissing() {
+		// We can make a check early on MD5 existence and fail the transfer if it's not present.
+		// This will save hours in the event a user has say, a several hundred gigabyte file.
+		if len(info.SrcHTTPHeaders.ContentMD5) == 0 {
+			jptm.LogDownloadError(info.Source, info.Destination, errExpectedMd5Missing.Error(), 0)
```
justinormont: Does `LogDownloadError` emit these errors to the `azcopy` stdout? The previous error for missing MD5 was only emitted to the log file instead of the more user accessible stdout.

ste/xfer-remoteToLocal.go
```diff
@@ -73,6 +73,17 @@ func remoteToLocal(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, d
...
+			jptm.SetStatus(common.ETransferStatus.Failed())
+			jptm.ReportTransferDone()
+			return
+		}
+	}
```
justinormont: Should the previous error location be removed? As it should be an unreachable condition since it's caught here first. 

Or perhaps changed to an `unexpected condition` style error?

ste/xfer-remoteToLocal.go
adreed-msft: No, it logs them to the job log.

ste/xfer-remoteToLocal.go
adreed-msft: I'll promote the old check to a panic, since it's now in a place that shouldn't even be reached to start with.

ste/xfer-remoteToLocal.go
justinormont: Should errors like these reach the stdout? 

When running I was thinking it was a transient issue before looking further into the logs, and finding the source files didn't have an MD5 value. Emitting the actual error to the stdout would let me know the actual error more quickly.

JohnRusk: I suspect that question is a bigger, and more general one.  I'd guess that for this PR, we should stay consistent with existing conventions in the app. If one day, we hypothetically LogDownloadError to use stderr, then that would take care of not just this case, but all other relevant ones where stderr could be used in this way.

testSuite/scripts/test_service_to_service_copy.py
```diff
@@ -1147,12 +1147,17 @@ def util_test_copy_file_from_x_bucket_to_x_bucket_propertyandmetadata(
...
-                add_flags("log-level", "info").execute_azcopy_copy_command()
+                add_flags("log-level", "info")  # Temporarily set result to Command for the sake of modifying the md5 check
+            result.flags["check-md5"] = ""
+            result = result.execute_azcopy_copy_command()  # Wrangle result to a bool for checking
         else:
             result = util.Command("copy").add_arguments(srcFileURL).add_arguments(local_validate_dest). \
-                add_flags("log-level", "info").execute_azcopy_copy_command()
+                add_flags("log-level", "info")  # Temporarily set result to Command for the sake of modifying the md5 check
+            result.flags["check-md5"] = ""
+            result = result.execute_azcopy_copy_command()  # Wrangle result to a bool for checking
```
JohnRusk: I haven't quite understood what's going on here.  Specifically, anything that fails now should have failed before - right?  (All this PR does is move the failure forward in time).  So:
(a) why is this change necessary? (Since anything that fails now must have also failed before).
(b) if this change is necessary, why is it only necessary in this one test (and not many tests)?

adreed-msft: a) When we choose not to persist the properties and metadata about the file, we lose the MD5. The MD5 is set to "FailIfDifferentOrMissing", so the file transfer never takes place, and these tests fail.

b) This is the only S2S test utility that opts to not copy properties and metadata occasionally. Otherwise, we're always putting and checking MD5s.

zezha-msft:
Re-trigger CI before merging.

adreed-msft:
Squashing and merging.

PR #813, opened 2020-01-02
Enable the creation of dummy transfers to deliver errors earlier in the process

adreed-msft:
This also extends the available path length on the source and destination, and metadata up to the unsigned 32-bit integer limit.

### On the backend:

If `ExpectFailure` is set in a `TransferInfo` object, the hijacking code in `xfer.go:68 (expectFailureXferDecorator)` will cancel the transfer pre-prologue state with the error of `TransferInfo.FailureReason`.

`xfer.go:130 (computeJobXfer)` also computes the transfer function and wraps it in the expected failure decorator.

### On the frontend:

storedObject now has the `expectedFailure` and `failureReason` properties. Setting these will cause them to also be set on the copy transfer that comes from `storedObject.ToNewCopyTransfer`.

To quickly create an intentional failure, call `newForcedErrorStorageObject(err string)`.


zezha-msft:
Could you please explain the changes a bit in the PR description?

cmd/zc_traverser_s3_service.go
```diff
@@ -104,17 +106,38 @@ func (t *s3ServiceTraverser) traverse(preprocessor objectMorpher, processor obje
...
+				errStr = fmt.Sprintf("skip enumerating the bucket %q, as it does not exist.", v)
+			} else {
+				errStr = err.Error()
 			}
 
-			if strings.Contains(err.Error(), "cannot list objects, The specified bucket does not exist") {
-				LogStdoutAndJobLog(fmt.Sprintf("skip enumerating the bucket %q, as it does not exist.", v))
-				continue
+			// Schedule a dummy transfer so the user knows something went wrong with enumeration.
+			dummyObj := newStoredObject(
```
JohnRusk: It seems odd to see code like this repeated several times.  Any chance of  a reusable func newForcedErrorStoredObject()

common/extensions.go
```diff
@@ -16,7 +16,7 @@ type URLStringExtension string
 
 func (s URLStringExtension) RedactSecretQueryParamForLogging() string {
 	u, err := url.Parse(string(s))
-	if err != nil {
+	if err != nil || !strings.HasPrefix(u.Scheme, "http") {
```
JohnRusk: Just out of curiosity, was there a case where we wanted to avoid the redaction, or is this just to save time?

ste/JobPartPlanFileName.go
```diff
@@ -214,42 +215,74 @@ func (jpfn JobPartPlanFileName) Create(order common.CopyJobPartOrderRequest) {
...
+	case common.ELocation.BlobFS():
+		maxDestLength = 1024
+		// Leave it as is if it's a default.
+	}
+
 	// Write each transfer to the Job Part Plan file (except for the src/dst strings; comes come later)
 	for t := range order.Transfers {
-		if len(order.Transfers[t].Source) > math.MaxInt16 || len(order.Transfers[t].Destination) > math.MaxInt16 {
-			panic(fmt.Sprintf("The file %s exceeds azcopy's current maximum path length on either the source or the destination.", order.Transfers[t].Source))
+		if uint32(len(order.Transfers[t].Source)) > uint32(math.MaxUint32) || uint32(len(order.Transfers[t].Destination)) > maxDestLength {
```
JohnRusk: Remember, if we change _anything_ in the plan files, we must always change const DataSchemaVersion common.Version

cmd/copyEnumeratorInit.go
```diff
@@ -112,9 +111,30 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
 	var containerResolver = NewS3BucketNameToAzureResourcesResolver(nil)
 	existingContainers := make(map[string]bool)
-	var logDstContainerCreateFailureOnce sync.Once
 	seenFailedContainers := make(map[string]bool) // Create map of already failed container conversions so we don't log a million items just for one container.
 
+	createContainerFailureTransfer := func(dstContainerName, failureReason string) error {
+		// Don't spam logs on failed containers
+		if _, ok := seenFailedContainers[dstContainerName]; !ok {
+			// Schedule a intentionally failing transfer at the container level with a empty dummy object
+			dummyObj := newStoredObject(nil, "", "", time.Now(), 0, nil, azblob.BlobNone, dstContainerName)
```
JohnRusk: I think you'll need to tweak all your usage of newStorageObject because commit 5241069e8918 has changed it signature.  (As noted in aother comment, I'd also suggest centrolizing them in one place).  Pls ping me for re-review once that's done, since I've only done a quick skim-read today.

cmd/zc_traverser_s3_service.go
adreed-msft: Yeah, that sounds like a better option considering how it's used.

common/extensions.go
adreed-msft: Yeah, this is a fix. If you push a non-URL through logging, it will be encoded as though it's a URL path segment. I intended to avoid that here.

ste/JobPartPlanFileName.go
adreed-msft: I'd assumed we'd already changed it this revision-- I'll increment it.

cmd/copyEnumeratorInit.go
```diff
@@ -112,9 +110,34 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 	// Giving it nothing to work with as new names will be added as we traverse.
 	var containerResolver = NewS3BucketNameToAzureResourcesResolver(nil)
 	existingContainers := make(map[string]bool)
-	var logDstContainerCreateFailureOnce sync.Once
 	seenFailedContainers := make(map[string]bool) // Create map of already failed container conversions so we don't log a million items just for one container.
 
+	createContainerFailureTransfer := func(dstContainerName, failureReason string) error {
```
JohnRusk: Minor: if you happen to be in here making any other changes, how about renaming this to `addContainerFailureTransfer` since the existing "create" name makes it sound like a factory method, and a factory method would not be expected to call `addTransfer` (whereas this does call `addTransfer`)

cmd/zc_enumerator.go
```diff
@@ -179,6 +185,25 @@ func newStoredObject(morpher objectMorpher, name string, relativePath string, lm
...
 }
 
+func newForcedErrorStoredObject(err, name, relativePath, containerName string) storedObject {
+	so := newStoredObject(
+		nil, // No need for a morpher on a intended failure
+		name,
+		relativePath,
+		time.Now(),
+		0,                        // A intended failure has no size,
+		emptyPropertiesAdapter{}, // and no content properties
```
JohnRusk: Probably best to use the existing constants defined for this purpose, namely: `cmd.noContentProps`and `cmd.noBlobProps` and `cmd.noMetadata`

cmd/copyEnumeratorInit.go
```diff
@@ -129,13 +152,10 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
-			if _, ok := seenFailedContainers[dstContainerName]; err != nil && ste.JobsAdmin != nil && !ok {
-				logDstContainerCreateFailureOnce.Do(func() {
-					glcm.Info("Failed to create one or more destination container(s). Your transfers may still succeed if the container already exists.")
-				})
-				ste.JobsAdmin.LogToJobLog(fmt.Sprintf("failed to initialize destination container %s; the transfer will continue (but be wary it may fail): %s", dstContainerName, err))
-				seenFailedContainers[dstContainerName] = true
+			if err != nil {
+				// No need to worry about an error from here.
+				_ = createContainerFailureTransfer(dstContainerName,
+					fmt.Sprintf("failed to initialize destination container %s; the transfer will continue (but be wary it may fail): %s", dstContainerName, err))
```
JohnRusk: The old on-screen wording, of "You transfers may still success if the container already exists" is nice because it gives the user a clearer picture whether they need to be alarmed.  Can you change the wording here to that please?

JohnRusk: Also, please use a constant or something instead of repeating the same message string in 3 places.

adreed-msft:
> My gut feel is that, when we were initially discussing the concept of ExpectedFailure, the intent was that it would always produce a failure status for the job, and therefore it should only be used for real _errors_ not _warnings_.

I would argue the inability to transfer a container for one reason or another is a failure. However, it may be a bit too extreme for just container creation failures, since there are scenarios where that can still have a successful transfer. IDK. I'll ponder some more on this here in a bit and get back to you.

adreed-msft:
Decided to think a bit more about this.

> "Are expected failures included in the total transfer count in the output? I think it would confuse uses if we did include them. E.g."I have 5 files, but AzCopy said I have 7 transfers". But its it really practical to exclude them?"

Yes, they are included. I don't think it'd make sense to not include them in the transfer list. In the folders as first class citizens PR, folders are treated as files (and individual transfers). I would argue that a container would count as a folder of sorts as well, and thus should be treated as a failed transfer if something goes wrong while transferring it.

However, failing to create the container definitely is a little extreme to throw a failing transfer that early. Failing to enumerate the container, however, I'd say would be a good time to throw the failing transfer.

> "What is the purpose of an expected failure? is it to produce a failure in the job status? if so, I does't seem correct to use it for these container ones - because they don't necessarily mean that the job has failed. I notice I see them a lot ( with the old code) just when I'm normally using AzCopy."

The purpose of an expected failure is supposed to be for something that we cannot, no matter what, transfer, and need to track an error for the transfer. Hence why I'd mentioned that in hindsight, it's probably not good to throw an expected failure there.

> "My gut feel is that, when we were initially discussing the concept of ExpectedFailure, the intent was that it would always produce a failure status for the job, and therefore it should only be used for real errors not warnings."

Yup, that's it. I'll reduce the failed creation back down to a job log and stdout log, and make the other requested changes.

adreed-msft:
This will now be piggybacked off to help when the ADLSG2 traverser is missing information.

JohnRusk:
@adreed-msft Does this relate to one of the 10.5 Backlog Items?  If so, can you pls mark it with Milestone 10.5. And if not, can you mark it with label "future release"?

adreed-msft:
An important few lines disappeared when merging, this should be good if CI passes... IF.

PR #824, opened 2020-01-10
Feature/folders as first class citizens

JohnRusk:
Here's a first draft of folders-as-first-class citizens, for review.

**What's included:**
Everything except:
1. Merge/Rebase to get it up to date with Dev
1. Checks of how it works with sync. The code is hooked in, but i need to work through whether any changes to the actual sync comparision logic are required, and think about sync deletion in a folder-aware context.
1. Implementation of directory awareness in `azcopy rm` (*)
1. The ability to set properties on one existing directory (as discussed) (*)
1. Making sure our ADLS Gen2 to ADLS Gen2 copy (via blob enpoints) is folder-safe.. As discussed. (*)

I propose to do this marked * in separate PRS, and the other as a future addition to this PR. But am open to suggestion on that.

**Architecture:**
* Plan files have an extra field for each transfer: entityType
* xfer-anyToRemote has been split into file and folder versions, since they didn't have enough in common to justify using just one routine.
* Enumerators of folder-aware sources should always include folders in their enumeration results.  storedObject.ToTransfer will filter the folders out, before they make it to the plan file, IF the destination is not folder-aware
* To send folders to a service, the existing "sender" implementation(s) need to add support for the new IFolderSender methods.  
* To read folder info from a service, the existing "source info provider" needs to implement the new EntityType method (to say which kind of entity its dealing with) and be able to return folder properties from its existing Properties method when EntityType == Folder.
* Folder overwrite logic is as discussed, and as reflected in FolderCreationTracker.ShouldSetProperties


ste/sender-azureFile.go
```diff
@@ -34,14 +34,27 @@ import (
 	"github.com/Azure/azure-storage-azcopy/common"
 )
 
+type URLHolder interface {
+	URL() url.URL
+	String() string
+}
+
+// TODO: review the decision to have this implement both interfaces
```
zezha-msft: makes sense

ste/mgr-JobPartTransferMgr.go
```diff
@@ -694,16 +711,18 @@ const (
...
-		" Dst: " + common.URLStringExtension(jptm.Info().Destination).RedactSecretQueryParamForLogging()
+		" Dst: " + common.URLStringExtension(info.Destination).RedactSecretQueryParamForLogging()
 
 	jptm.Log(level, fullMsg)
 }
 
 func (jptm *jobPartTransferMgr) logTransferError(errorCode transferErrorCode, source, destination, errorMsg string, status int) {
 	// order of log elements here is mirrored, in subset, in LogForCurrentTransfer
-	msg := fmt.Sprintf("%v: ", errorCode) + common.URLStringExtension(source).RedactSecretQueryParamForLogging() +
+	info := jptm.Info() // TODO we are getting a lot of Info calls and its (presumably) not well-optimized.  Profile that?
```
adreed-msft: >// TODO we are getting a lot of Info calls and its (presumably) not well-optimized.  Profile that? [](start = 21, length = 98)

Not a bad idea. I wonder if we could get away with just memoizing it without too large of a hit on memory usage..

ste/sender-azureFile.go
```diff
@@ -34,14 +34,27 @@ import (
...
+// azureFileSenderBase implements both IFolderSender and (most of) IFileSender.
+// Why implement both interfaces in the one type, even though they are largely unrelated? Because it
+// makes functions like newAzureFilesUploader easier to reason about, since they always return the same type.
+// It may also make it easier to describe what's needed when supporting an new backend - e.g. "to send to a new back end
+// you need a sender that implements IFileSender and, if the back end is folder aware, it should also implement IFolderSender"
+// (The alternative would be to have the likes of newAzureFilesUploader call sip.EntityType and return a different type
+// if the entity type is folder).
 type azureFileSenderBase struct {
```
adreed-msft: I honestly feel like this is a good choice. It puts everything in one place.

ste/sender-azureFile.go
```diff
@@ -34,14 +34,27 @@ import (
...
-	jptm      IJobPartTransferMgr
-	fileURL   azfile.FileURL
-	chunkSize uint32
-	numChunks uint32
-	pipeline  pipeline.Pipeline
-	pacer     pacer
-	ctx       context.Context
+	jptm         IJobPartTransferMgr
+	fileOrDirURL URLHolder
```
adreed-msft: >fileOrDirURL URLHolder [](start = 1, length = 22)

Some work may need to be done on the user delegation PR to fit compatibility with this.

adreed-msft:
I don't think this'll have any immediate conflicts with SDDL persistence. There will need to be code in place specifically for folders to put the SDDLs.

ste/mgr-JobPartTransferMgr.go
JohnRusk: Is on the backlog

ste/sender-azureFile.go
JohnRusk: OK.  How about we get that PR done and merged before this one?  Would that be sensible?

ste/sender-azureFile.go
JohnRusk: Just checking, @zezha-msft . Did you mean "makes sense" to leave it as it is?  Or "make sense" to review and reconsider the current approach?

zezha-msft: Sorry, makes sense to leave as it is

cmd/jobsShow.go
```diff
@@ -118,8 +118,14 @@ func PrintJobTransfers(listTransfersResponse common.ListJobTransfersResponse) {
 		var sb strings.Builder
 		sb.WriteString("----------- Transfers for JobId " + listTransfersResponse.JobID.String() + " -----------\n")
 		for index := 0; index < len(listTransfersResponse.Details); index++ {
-			sb.WriteString("transfer--> source: " + listTransfersResponse.Details[index].Src + " destination: " +
-				listTransfersResponse.Details[index].Dst + " status " + listTransfersResponse.Details[index].TransferStatus.String() + "\n")
+			folderChar := ""
+			if listTransfersResponse.Details[index].IsFolderProperties {
+				folderChar = "/"
+				// TODO: reviewers, is it OK that this is hard coded to fwd slash (even for windows local sources?)
```
adreed-msft: >// TODO: reviewers, is it OK that this is hard coded to fwd slash (even for windows local sources?) [](start = 4, length = 99)

idea: We have a function to detect the resource's path separator. Consider using it instead

cmd/jobsShow.go
```diff
@@ -118,8 +118,14 @@ func PrintJobTransfers(listTransfersResponse common.ListJobTransfersResponse) {
...
+				// TODO: reviewers, would it be better to output something like "type: folder"?  I did't because it seemed verbose.
```
adreed-msft: I wouldn't be too concerned about it.

cmd/zc_traverser_blob.go
```diff
@@ -161,6 +162,7 @@ func (t *blobTraverser) traverse(preprocessor objectMorpher, processor objectPro
 				preprocessor,
 				getObjectNameOnly(blobInfo.Name),
 				relativePath,
+				common.EEntityType.File(),
 				blobInfo.Properties.LastModified,
```
adreed-msft: Here's a thought: Blobfuse and ADLSG2 support pseudo folders on blob with a blob that has the metadata property hdi_isfolder = true.

Since blob is semi-capable of being folder aware in that way, should we consider supporting those pseudo folders?

IMO it would make sense to do so, especially if a user has blobs under these pseudo folders.

adreed-msft: at a bare minimum, we should not enumerate these blobs.

at most, we should copy their metadata as well.

---
In reply to: [376626007](https://github.com/Azure/azure-storage-azcopy/pull/824#discussion_r376626007) [](ancestors = 376626007)

adreed-msft: I however, don't believe that this would make blob something that empty folders can be copied to, considering how application specific it is.

---
In reply to: [376626284](https://github.com/Azure/azure-storage-azcopy/pull/824#discussion_r376626284) [](ancestors = 376626284,376626007)

ste/sender-azureFile.go
adreed-msft: That sounds good to me. So that'd make the order of these SDDLs -> Folders -> User Delegation.

---
In reply to: [369332636](https://github.com/Azure/azure-storage-azcopy/pull/824#discussion_r369332636) [](ancestors = 369332636)

ste/sender-azureFile.go
```diff
@@ -164,13 +200,22 @@ func (u *azureFileSenderBase) GetDestinationLength() (int64, error) {
 	return prop.ContentLength(), nil
 }
 
+func (u *azureFileSenderBase) EnsureFolderExists() error {
+	return AzureFileParentDirCreator{}.CreateDirToRoot(u.ctx, u.dirURL(), u.pipeline, u.jptm.GetFolderCreationTracker())
+}
+
+func (u *azureFileSenderBase) SetFolderProperties() error {
+	// TODO
+	return nil
```
adreed-msft: Is there no properties or metadata that need to be transferred S2S already?

Or is this left as a TODO because it might be better to implement when handling the remainder of SDDLs and SMB properties?

ste/xfer-anyToRemote-folder.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		jptm.LogSendError(info.Source, info.Destination, err.Error(), 0)
+		jptm.SetStatus(common.ETransferStatus.Failed())
+		jptm.ReportTransferDone()
+		return
+	}
+	s := sBase.(IFolderSender)
+
+	/*TODO
+	// step 3: check overwrite option
```
adreed-msft: I imagine this is to be done soon :tm: as well? (as in not this PR but another)

common/folderCreationTracker.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+type simpleFolderTracker struct {
+	mu       *sync.Mutex
+	contents map[string]struct{}
+}
+
+func (f *simpleFolderTracker) RecordCreation(folder string) {
+	f.mu.Lock()
```
JohnRusk: This needs to have a mode where it does NOT add, because when there are no folderPropertiesTransfers (e.g. due to source or dest being not folder aware) nothing will ever call StopTracking. So we'll "leak" ram here.

ste/sender-azureFile.go
JohnRusk: Yes, the latter.  We can do it in the followup PR

ste/xfer-anyToRemote-folder.go
JohnRusk: Oh, it was supposed to be in here!  Let me check. Thanks for spotting it.

cmd/jobsShow.go
JohnRusk: Turns out there's not necessarily a path separator in every result string. Specifically, those at root level don't have one.  Given that fact, and the fact that `list` will mostly be used on cloud-based sources not local dirs, I'm going to leave this code as is.

cmd/zc_traverser_blob.go
JohnRusk: Tracked already as a separate backlog item.

ste/xfer-anyToRemote-folder.go
JohnRusk: Correction. It's done. I just forgot to remove the old TODO.  

It's done in simpleFolderTracker.ShouldSetProperties

JohnRusk:
> Also need to consider detecting "properties changed but content not" on Windows (for both folders and files, although for folders what we're looking at is specifically permissions/attributes changed but no LMT update due to no file changes to children)

This will be tracked and addressed separately.

adreed-msft:
I honestly don't have much else to say about the new content. Good that we're finding some use in the job manager initialization bit.

cmd/syncProcessor.go
```diff
@@ -175,9 +176,32 @@ type localFileDeleter struct {
 	rootPath string
 }
 
+// As at version 10.4.0, we intentionally don't delete directories in sync,
+// even if our folder properties option suggests we should.
+// Why? The key difficulties are as follows, and its the third one that we don't currently have a solution for.
+// 1. Timing (solvable in theory with FolderDeletionManager)
+// 2. Identifying which should be removed when source does not have concept of folders (e.g. BLob)
+//    Probably solution is to just respect the folder properties option setting (which we already do in our delete processors)
+// 3. In Azure Files case (and to a lesser extent on local disks) users may have ACLS or other properties
```
adreed-msft: Remember that hdi_isfolder blobs exist. If we're respecting the folder properties option, we should perhaps respect these folder-esque representations.

The other problem with this is that we're trying to sync from something that's objectively not a filesystem to something that objectively is a filesystem.

cmd/syncProcessor.go
```diff
@@ -175,9 +176,32 @@ type localFileDeleter struct {
...
+//    set on the directories, and wish to retain those even tho the directories are empty. (Perhaps less of an issue
+//    when syncing from folder-aware sources that DOES NOT HAVE the directory. But still an issue when syncing from
+//    blob. E.g. we delete a folder because there's nothing in it right now, but really user wanted it there,
+//    and have set up custom ACLs on it for future use.  If we delete, they lose the custom ACL setup.
+// TODO: shall we add folder deletion support at some stage? (In cases where folderPropertiesOption says that folders should be processed)
```
adreed-msft: This sounds like it goes hand-in-hand with the overwrite option.

adreed-msft:
Looks good to me.

cmd/syncProcessor.go
JohnRusk: Yeah, let's address it as part of that. (I.e. not in this PR)

cmd/syncProcessor.go
JohnRusk: Yes. We have a separate backlog item for the hdi_isfolder case.

PR #836, opened 2020-01-21
Added new option ifSourceNewer for overwrite flag

zezha-msft:
Open question: if source and destination lmts are equal, should the file be transferred?

### DIFF
--- a/ste/remoteObjectExists.go
+++ b/ste/remoteObjectExists.go
@@ -22,6 +22,7 @@ package ste
 
 import (
 	"net/http"
+	"time"
 )
 
 // an error with an HTTP Response
@@ -29,18 +30,22 @@ type responseError interface {
 	Response() *http.Response
 }
 
+type lastModifiedTimerProvider interface {
+	LastModified() time.Time
+}
+
 // remoteObjectExists takes the error returned when trying to access a remote object, sees whether is
 // a "not found" error.  If the object exists (i.e. error is nil) it returns (true, nil).  If the
 // error is a "not found" error, it returns (false, nil). Else it returns false and the original error.
 // The initial, dummy, parameter, is to allow callers to conveniently call it with functions that return a tuple
 // - even though we only need the error.
-func remoteObjectExists(_ interface{}, errWhenAccessingRemoteObject error) (bool, error) {
+func remoteObjectExists(props lastModifiedTimerProvider, errWhenAccessingRemoteObject error) (bool, time.Time, error) {
 
 	if typedErr, ok := errWhenAccessingRemoteObject.(responseError); ok && typedErr.Response().StatusCode == http.StatusNotFound {
-		return false, nil // 404 error, so it does NOT exist
+		return false, time.Time{}, nil // 404 error, so it does NOT exist
 	} else if errWhenAccessingRemoteObject != nil {
-		return false, errWhenAccessingRemoteObject // some other error happened, so we return it
+		return false, time.Time{}, errWhenAccessingRemoteObject // some other error happened, so we return it
 	} else {
-		return true, nil // If err equals nil, the file exists
+		return true, props.LastModified(), nil // If err equals nil, the file exists
 	}
 }
--- a/ste/xfer-anyToRemote.go
+++ b/ste/xfer-anyToRemote.go
@@ -73,26 +73,6 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
 		panic("must always schedule one chunk, even if file is empty") // this keeps our code structure simpler, by using a dummy chunk for empty files
 	}
 
-	// Do LMT verfication before transfer, when:
-	// 1) Source is local, so get source file's LMT is free.
-	// 2) Source is remote, i.e. S2S copy case. And source's size is larger than one chunk. So verification can possibly save transfer's cost.
-	if copier, isS2SCopier := s.(s2sCopier); srcInfoProvider.IsLocal() ||
-		(isS2SCopier && info.S2SSourceChangeValidation && srcSize > int64(copier.ChunkSize())) {
-		lmt, err := srcInfoProvider.GetLastModifiedTime()
-		if err != nil {
-			jptm.LogSendError(info.Source, info.Destination, "Couldn't get source's last modified time-"+err.Error(), 0)
-			jptm.SetStatus(common.ETransferStatus.Failed())
-			jptm.ReportTransferDone()
-			return
-		}
-		if lmt.UTC() != jptm.LastModifiedTime().UTC() {
-			jptm.LogSendError(info.Source, info.Destination, "File modified since transfer scheduled", 0)
-			jptm.SetStatus(common.ETransferStatus.Failed())
-			jptm.ReportTransferDone()
-			return
-		}
-	}
-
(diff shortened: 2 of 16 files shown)


ste/remoteObjectExists.go
```diff
@@ -22,25 +22,30 @@ package ste
...
 	"net/http"
+	"time"
 )
 
 // an error with an HTTP Response
 type responseError interface {
 	Response() *http.Response
 }
 
+type lastModifiedTimerProvider interface {
```
JohnRusk: Nice.

ste/xfer-anyToRemote.go
```diff
@@ -94,6 +94,11 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
 				parsed, _ := url.Parse(info.Destination)
 				parsed.RawQuery = ""
 				shouldOverwrite = jptm.GetOverwritePrompter().shouldOverwrite(parsed.String())
+			} else if jptm.GetOverwriteOption() == common.EOverwriteOption.IfSourceNewer() {
+				// only overwrite if source lmt is newer (after) the destination
+				if jptm.LastModifiedTime().After(dstLmt) {
```
JohnRusk: These (the jptm.LastModifiedTimes) can be stale in the sense that if the job is long-running, they date from the time of scanning rather than "right now".  Prior to this PR, such "stale" cases result in a failed transfer, due to the LMT check below (or, in the case for small remote sources, where we don't run the check below, I think we fail on the LMT IfModified condition at time of transfer).

This change will change that behaviour: when using ifSourceModified, if the source is modified between scanning and time of transfer, it will just be processed here as a "skip" (if the scan-time LMT was <= the dest).  That smells incorrect, for two reasons:
(a) now we have something that would have be noted as a fail, noted instead as a skip. and, perhaps more importantly,
(b) Skip was probably the wrong thing to do, since the source files _real_ LMT may now be after the destination.  This is the most serious aspect, since user will believe we have fully synced source to test, when really we haven't.

I can think of two possible solutions. The low-impact, short-term one, is to take the existing code that checks source LMTs (starting on line 129) and move it into a function. Call that function from line 129 (to fully preserve all existing behavoiur) but ALSO call it here, if jptm.LastModifiedTime() is NOT after dstLMT.  (If it IS after, then there's no reason to call it).  That will cause the problem cases (where the LMT is modified between scan and transfer) to be reported as fail, which is consistent with the currrent behaviour.)

The longer term-solution (e.g. in v10.5) might be to change the way we deal with things that change between scanning and transfer - but as discussed previously that's non-trivial from a perf point of view in the general case (e.g. small remote files), and also non-trivial in that we'd need to pick up not just fresh LMTs but also fresh _sizes_ and maybe even fresh mime-types.

ste/xfer-remoteToLocal.go
```diff
@@ -53,14 +53,19 @@ func remoteToLocal(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, d
...
 			// if the error is nil, then file exists locally
 			shouldOverwrite := false
 
 			// if necessary, prompt to confirm user's intent
 			if jptm.GetOverwriteOption() == common.EOverwriteOption.Prompt() {
 				shouldOverwrite = jptm.GetOverwritePrompter().shouldOverwrite(info.Destination)
+			} else if jptm.GetOverwriteOption() == common.EOverwriteOption.IfSourceNewer() {
+				// only overwrite if source lmt is newer (after) the destination
+				if jptm.LastModifiedTime().After(dstProps.ModTime()) {
+					shouldOverwrite = true
```
JohnRusk: I suspect the same issue applies here as I mentioned for the anyToRemote case, and again the safest short-term solution appears to be to retrive the latest LMT here, using sourceInfoProvider (check that they all requery and don't rely on the scan-time result!)

ste/xfer-anyToRemote.go
zezha-msft: I totally understand your concern, I asked myself the same question too, here is how I reasoned:

1. Our assumption so far is that the user's intention is to transfer the exact files at the time when they invoked the copy command. For example, if the file changed after scanning, we fail the transfer. In addition, let's say they started a job, but paused it after scanning was done; when they resume it 2 weeks later, they expect the same files to be transferred, right? 
2. From the same POV, `overwrite=ifSourceNewer` has the effect of selecting which files to transfer, and that should be determined with respect to the state of the files at the time of scanning. 

The unfortunate part is that we don't do this comparison right away, instead we save the state of the source file at the time of scanning, and then compare it against the destination file at the time of transferring. I agree that this is inconsistent, but it also feels inconsistent to do it the other way..

zezha-msft: Hmm, furthermore, if we checked the source lmt at different times (since jobs can last a long time), there's no predictability on whether a file would be included, right? At least the scanning calls are back-to-back, and we know the lmts were obtained around the same time.

JohnRusk: I'm feeling somewhat persuaded by your two points.  Especially because doing it the other way doesn't guarantee that at the end of the run everything is in sync because, for example, consider a job that runs for 12 hours.  A file transferred in the first hour may be changed at the source later in the other 11 hours of the job, so at the end of the job, there's still no guarantee that everything matches. (And there cannot be such a guarantee).

But what do you think about the inconsistency that for changes in source LMT between scan and send, report them as "fails" for all existing OverwriteOptions, and as "skips" with IfSourceNewer?  (My first proposed solution removes this inconsistency, by making them always get reported as "fails", and the expense of a little extra complexity in the code. Which may or may not be a desirable tradeoff.  What I like about it is that its consistent, and therefore more explainable and understandable).

zezha-msft: What if we simply moved up the logics starting on line 129 (checking lmt of source) to precede the overwrite logics? Then the behavior is consistent for files that change between scan and send, thoughts?

JohnRusk: I have not checked whether it looks safe to move that block.  If it is safe, then that sounds fine in principle. But it only affects one direction (any to remote). It doesn't solve remoteTolocal.  I'd imagine we'd make the change for both for for neither.

zezha-msft: Ohh, download doesn't do such check because each block gets checked already thanks to the access conditions on the GET requests.

Let me re-consider this, maybe it makes more sense to simply get the latest source lmt.

JohnRusk: If the saved LMT in the plan file is already known to be >= the destination one, then no need to see if there's an even more recent source LMT.

BTW, I'm not sure I'm suggesting a total "reconsider" here......  Just that that we present "stale" LMTs as failures in the IfSourceNewer case, for consistency.

JohnRusk: Having said that, there may be a better option that you might be able to find....

ste/xfer-remoteToLocal.go
JohnRusk: Resolving. My logic was incorrrect (or at least, unnecessary).

ste/xfer-anyToRemote.go
JohnRusk: Resolving, as per review comment.

PR #840, opened 2020-01-22
Enable job cleanup on windows by unmapping files and closing logs during a list call.

### LINKED ISSUE #814: Command "azcopy jobs clean --with-status=completed" not cleaning log and plan files
Hi,  I am unable to clean log and plan files with command 
"azcopy jobs clean --with-status=completed", it is giving error.

10.3.3

windows

### What command did you run?
azcopy jobs clean --with-status=completed

Error come up : Failed to remove log/plan files due to error: remove C:\*/e8244365-cc14-5a4b-4584-eca6d277ebe3--00000.steV10: Access is denied..

Simply first run azcopy copy command, it will generate a log and a plan file, then run command
"azcopy jobs clean --with-status=completed", it will throw error.

No

adreed-msft:
Fixes #814

### DIFF
--- a/ste/JobsAdmin.go
+++ b/ste/JobsAdmin.go
@@ -636,9 +636,7 @@ func (ja *jobsAdmin) ResurrectJob(jobId common.JobID, sourceSAS string, destinat
 		}
 		mmf := planFile.Map()
 		jm := ja.JobMgrEnsureExists(jobID, mmf.Plan().LogLevel, "")
-		jm.AddJobPart(partNum, planFile, sourceSAS, destinationSAS, false)
-		// Note that unmapping this mmf is perfectly fine. It just gets re-mapped in the job part managers.
-		mmf.Unmap()
+		jm.AddJobPart(partNum, planFile, mmf, sourceSAS, destinationSAS, false)
 	}
 	return true
 }
@@ -667,9 +665,7 @@ func (ja *jobsAdmin) ResurrectJobParts() {
 		mmf := planFile.Map()
 		//todo : call the compute transfer function here for each job.
 		jm := ja.JobMgrEnsureExists(jobID, mmf.Plan().LogLevel, "")
-		jm.AddJobPart(partNum, planFile, EMPTY_SAS_STRING, EMPTY_SAS_STRING, false)
-		// This only unmaps this particular instance of the mmf, which has already been re-mapped for the job part.
-		mmf.Unmap()
+		jm.AddJobPart(partNum, planFile, mmf, EMPTY_SAS_STRING, EMPTY_SAS_STRING, false)
 	}
 }
 
--- a/ste/JobsAdmin.go
+++ b/ste/JobsAdmin.go
@@ -637,6 +637,8 @@ func (ja *jobsAdmin) ResurrectJob(jobId common.JobID, sourceSAS string, destinat
 		mmf := planFile.Map()
 		jm := ja.JobMgrEnsureExists(jobID, mmf.Plan().LogLevel, "")
 		jm.AddJobPart(partNum, planFile, sourceSAS, destinationSAS, false)
+		// Note that unmapping this mmf is perfectly fine. It just gets re-mapped in the job part managers.
+		mmf.Unmap()
 	}
 	return true
 }
@@ -666,6 +668,8 @@ func (ja *jobsAdmin) ResurrectJobParts() {
 		//todo : call the compute transfer function here for each job.
 		jm := ja.JobMgrEnsureExists(jobID, mmf.Plan().LogLevel, "")
 		jm.AddJobPart(partNum, planFile, EMPTY_SAS_STRING, EMPTY_SAS_STRING, false)
+		// This only unmaps this particular instance of the mmf, which has already been re-mapped for the job part.
+		mmf.Unmap()
 	}
 }
 
--- a/cmd/jobsClean.go
+++ b/cmd/jobsClean.go
@@ -25,8 +25,9 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


ste/JobsAdmin.go
```diff
@@ -637,6 +637,8 @@ func (ja *jobsAdmin) ResurrectJob(jobId common.JobID, sourceSAS string, destinat
 		mmf := planFile.Map()
 		jm := ja.JobMgrEnsureExists(jobID, mmf.Plan().LogLevel, "")
 		jm.AddJobPart(partNum, planFile, sourceSAS, destinationSAS, false)
+		// Note that unmapping this mmf is perfectly fine. It just gets re-mapped in the job part managers.
+		mmf.Unmap()
```
JohnRusk: This seems unexpected, to close here so immediately and so automatically.  I find it hard to reason about the correctness of this in general.  Is there any way to replace this Unmap and the one added to RessurectJobParts, with explict close calls like you added to the ListJobs function?  E.g. closing from something in the call chain to this routine here, rather than unmapping right here?

ste/JobsAdmin.go
```diff
@@ -666,6 +668,8 @@ func (ja *jobsAdmin) ResurrectJobParts() {
 		//todo : call the compute transfer function here for each job.
 		jm := ja.JobMgrEnsureExists(jobID, mmf.Plan().LogLevel, "")
 		jm.AddJobPart(partNum, planFile, EMPTY_SAS_STRING, EMPTY_SAS_STRING, false)
+		// This only unmaps this particular instance of the mmf, which has already been re-mapped for the job part.
+		mmf.Unmap()
```
JohnRusk: Sam comment as RessurectJob.

ste/JobsAdmin.go
JohnRusk: By "in general" I mean, in cases such as copy and sync and resume.

adreed-msft: This MMF never gets linked to a job part manager, and for all intents and purposes, never actually gets used again.

There is no explicit close call to be made, unless I'm misunderstanding your question?

adreed-msft: If anything, the file handle is otherwise leaked here.

JohnRusk: Do you mean "This MMF never gets linked to a job part manager when we run jobs clean"; or do you mean "This MMF never gets linked to a job part manager, and that's true no matter how this routine is called and used".

adreed-msft: I mean this MMF is literally never used again, no matter how it's called. I tested copy and resume before I sent in this PR because I had similar concerns.

JohnRusk: And is that also true (and tested) for the similar change that's been made to ResurrectJobParts?

adreed-msft: Yup. ResurrectJobParts is only used by ListJobs at the moment, and even then, it's _effectively_ just repeatedly calling ResurrectJob.

JohnRusk: So why can't the unmapping be moved into ListJobs?  That way, the ressurect* methods would be _safe_ to call in future from other places. Right now, they're not really safe to reuse, because (with this PR) they are making the assumption that nothing will actually use the plan file. Assuming "my caller will never actually use my data" seems like a likely way to surprise some future maintainer of the code.  
Thoughts on moving the unmapping to ListBlobs?

adreed-msft: The thing is that these mmfs literally don't get returned. They never get used again, because they can't be used again. Closing the mmfs here isn't unsafe, not closing them is just leaking file handles.

adreed-msft: The only thing this mmf is created for is to get the log level.

adreed-msft: It's _absolutely_ obtuse, and it would be great if AddJobParts just took in the MMF instead of repeating mapping in the file.

adreed-msft: Would you prefer if I set it up that way, so that the close call occurring later in the list call is the only place we need to close the mmf?

JohnRusk: By "that way", do you mean "leave the Resurrect* methods back the way they were before this PR, and instead close their MMFs in the List call"?  If so, yes, I think I'd find that easier to review and approve. 

(I'll find it hard to approve if the methods named Resurrect* actually give back objects that are partially "dead" - as is the case with the current draft of this PR.  I take your point that the existing logic and naming, prior to this PR, is not ideal.  But I'm not sure I can bring myself to get on board with Resurrect* returning a "dead"-ish object - since that's opposite of what 'resurrect" means).

adreed-msft: I think labeling these returned objects as "partially dead" is an incorrect assumption.

These MMFs aren't tracked. They aren't returned. They aren't reused anywhere. They are, for all intents and purposes, leaked memory and handles. Killing them doesn't kill anything about the underlying job managers, job part managers, or job part transfer managers. They're left _entirely_ in tact.

My suggestion is to make it so that calling AddJobPart allows you to take in an existing MMF, which would allow us to avoid that leakage.

zezha-msft:
This should target master instead (just the 1 commit).

adreed-msft:
> This latest version looks nice. Clean and simple.
> 
> BTW, what closes the mmfs that are passed in to AddJobPart? Is it the code in ListJobs, that calls v.Close in a loop, which closes them?

v.Close closes the mmfs. Interesting sidenote: That's the only place that we close job parts aside from just expecting the OS to do it for us.

adreed-msft:
> This should target master instead (just the 1 commit).

wait, why should this alone target master?

adreed-msft:
Rebased on master, retargeted to master, squashing and merging.

PR #842, opened 2020-01-23
Prototyped new end-to-end testing

zezha-msft:
The goals of the prototype is to propose a new approach to:

1. allow us to run tests in parallel (on multiple OSs at the same time, using the same account)
2. avoid the need to update credentials periodically on the CI config
3. understand why tests are failing (no "False is not True" LOL)
4. easily allow devs to run tests locally, no need to configure 20 environment variables
5. have faster tests so that it's easier to add coverage for our scenarios
6. have more concise tests (avoid repetitive setups)

😃 Don't be alarmed, there aren't that many new lines. Only 4 short files matter here:

1. config.go: defines all the inputs to the end-to-end testing. We should take in credentials that don't expire (goal 2 and 4), to lower the number of inputs.
2. factory.go: generate test resources on the fly. Each test should create its own randomly named resource, so that tests can run in parallel (goal 1 and 5). We need to think about how account -> account scenarios fit in (idea: use a SPN identity to create new accounts on the fly)
3. runner.go: takes care of invoking the AzCopy binary, and parse the outputs to tell us about the results (goal 3). More improvements are due here.
4. validator.go: helper methods that validate results. 

zt_copy_blob_upload_test.go shows a basic working test.

The rest of the files are ported from the cmd package. Their contents will be extracted out into the appropriate places ultimately.

### DIFF
--- a/e2etest/config.go
+++ b/e2etest/config.go
@@ -1,0 +1,73 @@
+// Copyright © Microsoft <wastore@microsoft.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package e2etest
+
+import (
+	"fmt"
+	"os"
+	"reflect"
+
+	"github.com/JeffreyRichter/enum/enum"
+)
+
+// clearly define all the inputs to the end-to-end tests
+// it's ok to panic if the inputs are absolutely required
+// the general guidance is to take in as few parameters as possible
+type GlobalInputManager struct{}
+
+func (GlobalInputManager) GetAccountAndKey(accountType AccountType) (string, string) {
+	var name, key string
+
+	switch accountType {
+	case EAccountType.Standard():
+		name = os.Getenv("AZCOPY_E2E_ACCOUNT_NAME")
+		key = os.Getenv("AZCOPY_E2E_ACCOUNT_KEY")
+	default:
+		panic("Only the standard account type is supported for the moment.")
+	}
+
+	if name == "" || key == "" {
+		panic(fmt.Sprintf("Name and key for %s account must be set before running tests", accountType))
+	}
+
+	return name, key
+}
+
+func (GlobalInputManager) GetExecutablePath() string {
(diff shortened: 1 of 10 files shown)


e2etest/config.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (GlobalInputManager) GetAccountAndKey(accountType AccountType) (string, string) {
+	var name, key string
+
+	switch accountType {
+	case EAccountType.Standard():
+		name = os.Getenv("AZCOPY_E2E_ACCOUNT_NAME")
+		key = os.Getenv("AZCOPY_E2E_ACCOUNT_KEY")
+	default:
+		panic("Only the standard account type is supported for the moment.")
```
JohnRusk: Would we end up with 3 key pairs in the end: standard, another pair for a HNS-enabled account, and S3?

e2etest/config.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/JeffreyRichter/enum/enum"
+)
+
+// clearly define all the inputs to the end-to-end tests
+// it's ok to panic if the inputs are absolutely required
+// the general guidance is to take in as few parameters as possible
+type GlobalInputManager struct{}
+
+func (GlobalInputManager) GetAccountAndKey(accountType AccountType) (string, string) {
```
JohnRusk: How do we test OAuth, if we start with account keys?  If that's not possible, should we abandon account keys, and go with OAuth tokens as our single type of Azure Auth. (Probably with Serivce Principal auth, I'm guesssing.  Might be able to give that Principal rights to make accounts even, which would help with account-to-accountt scenarios).  Then we'd need just one sent of SPA creds (e.g. a cert) and the account/key for S3

zezha-msft: Yes, that's what I had in mind too. SPN with access to the entire subscription should be the easiest approach to do account -> account tests.

However, it's somewhat expensive (long) to create new accounts, so for regular tests, I lean towards using an account like we do now. Let's discuss offline.

adreed-msft: I think a secret supplied from an environment variable would work just fine. As long as the service principal has the ability to create user delegation keys and such, it'll work just fine.

---
In reply to: [370330965](https://github.com/Azure/azure-storage-azcopy/pull/842#discussion_r370330965) [](ancestors = 370330965)

e2etest/config.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (GlobalInputManager) GetExecutablePath() string {
+	path := os.Getenv("AZCOPY_E2E_EXECUTABLE_PATH")
+	if path == "" {
+		panic("Cannot test AzCopy if AZCOPY_E2E_EXECUTABLE_PATH is not provided")
+	}
+
+	return path
+}
```
adreed-msft: It would be nice if the test suite would compile the executable for us.

e2etest/runner.go
```diff
@@ -0,0 +1,154 @@
...
+		panic(err)
+	}
+
+	jobSummary := common.ListJobSummaryResponse{}
+	err = json.Unmarshal([]byte(finalMsg.MessageContent), &jobSummary)
+	if err != nil {
+		panic(err)
+	}
+
+	return CopyCommandResult{jobID: jobSummary.JobID, finalStatus: jobSummary}
```
adreed-msft: >return CopyCommandResult{jobID: jobSummary.JobID, finalStatus: jobSummary} [](start = 1, length = 74)

If the job summary isn't success, we should log out the command (with SAS redacted)

adreed-msft:
> 

We should consider general testing of jobs resume, clean, and otherwise.

---
Refers to: e2etest/runner.go:155 in b10bbe8. [](commit_id = b10bbe852a398bcd99d84c5ba5011561457c2767, deletion_comment = False)

e2etest/zt_copy_blob_upload_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	result, err := runner.ExecuteCopyCommand(dirPath, containerURLWithSAS.String())
+	c.Assert(err, chk.IsNil)
+	c.Assert(int(result.finalStatus.TransfersCompleted), chk.Equals, len(filesToInclude))
+
+	transfers := result.GetTransferList(common.ETransferStatus.Success())
+	srcRoot := dirPath
+	dstRoot := fmt.Sprintf("%s/%s", containerURL.String(), filepath.Base(dirPath))
+	Validator{}.ValidateCopyTransfersAreScheduled(c, false, true, srcRoot, dstRoot,
+		filesToInclude, transfers)
```
adreed-msft: As john suggested during yesterday's meeting, we should have more than just trusting what azcopy says it did. (ex. verify the files manually)

PR #850, opened 2020-01-28
Un-hide list-of-files flag for copy command

zezha-msft:


JohnRusk:
Removed from 10.4, pending perf tuning

zezha-msft:
@JohnRusk do we need to benchmark this? Is there some kind of numbers we have to hit?

JohnRusk:
@zezha-msft There are two concerns here:
1. Is it fast enough to expose?  I think it is for lists of hundreds of files, but not for lists of thousands. (It was super slow when I tested with 3000 some time ago).
2. If we fix the perf, will that entail a breaking change to the existing --list-of-files?  (I suspect it probably will).  If it will, that becomes a strong reason to _not_ make the current implementation public.

zezha-msft:
Closing for now, not ready to expose this flag.

PR #853, opened 2020-01-29
Remove all usages of legacy getCredType code

adreed-msft:
When re-using the credential info given by the old code to enumerate the source, we'd accidentally send an OAuth token to public sources. This wasn't problematic prior to the fromto fix, but now it is because it can go to non-azure locations when enumerating.

This PR is a draft until we determine which of the three solutions to take to this still being caused when overriding the credential type.

### DIFF
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -214,12 +214,12 @@ type cookedSyncCmdArgs struct {
 	// deletion count keeps track of how many extra files from the destination were removed
 	atomicDeletionCount uint32
 
-	source         string
-	sourceSAS      string
-	destination    string
-	destinationSAS string
-	fromTo         common.FromTo
-	credentialInfo common.CredentialInfo
+	source            string
+	sourceSAS         string
+	destination       string
+	destinationSAS    string
+	fromTo            common.FromTo
+	dstCredentialInfo common.CredentialInfo
 
 	// filters
 	recursive             bool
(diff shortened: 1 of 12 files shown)


adreed-msft:
It was decided to roll with just ignoring the ability to achieve this effect by overriding the auth type.

cmd/sync.go
```diff
@@ -477,36 +477,21 @@ Final Job Status: %v%s%s
...
-	// verifies credential type and initializes credential info.
-	// For sync, only one side need credential.
-	cca.credentialInfo.CredentialType, err = getCredentialType(ctx, rawFromToInfo{
-		fromTo:         cca.fromTo,
-		source:         cca.source,
-		destination:    cca.destination,
-		sourceSAS:      cca.sourceSAS,
-		destinationSAS: cca.destinationSAS,
-	})
+	// Initializes credential info and oauth token for the destination. This should not be re-used on the source.
```
JohnRusk: I'm finding this confusing: in that it's named dst... but sometimes we put source credentials into it.

Is there anything that would stop us from defining both cca.srcCredentialInfo and cca.dstCredentialInfo, and always putting the source creds in the source one and the dest creds in the dest one?

JohnRusk:
Any chance we could do glcm.Info in all cases that says something like:

```
Credential type for source: SAS/Anon
Credential type for dest: OAuth
```

(And for local files we might say: none (local filesystem))

Might need to say EITHER SAS OR anon, by seeing if the URL has a SAS.

JohnRusk:
To be thorough, I think we should think about these lines in getBlobCredentialType:
```
	} else if !canBePublic { // oauth token found, if it can not be public resource, return token credential
		return common.ECredentialType.OAuthToken(), false, nil
```

This triggers for the destination.  If the user was to accidentally give a public destination (even when we didn't expect that) should we detect that fact and decline to provide the auth creds....???

cmd/sync.go
JohnRusk: In fact, I think we really ought to define both.  Right now we are getting issues - e.g. I ran a copy based download, and the isSource flag was wrong, becuase of the way this stuff currently works.  Getting separate fields would help.

JohnRusk:
Please discuss with me before resolving the comments I've left.  Since there may be non-trivial conflicts with another PR, and there may also be another significant change I'll need to suggest. (I just need to check something first).

PR #855, opened 2020-01-30
Set minio proxy in addition to http's proxy.

adreed-msft:


adreed-msft:
@zezha-msft stop working you're on paternal leave

PR #878, opened 2020-02-28
Make RM process empty folders

JohnRusk:


common/folderDeletionManager.go
```diff
@@ -0,0 +1,226 @@
...
+// FolderDeletionManager handles the fact that (in most locations) we can't delete folders that
+// still contain files.  So it allows us to request deletion of a folder, and have that be attempted
+// after the last file is removed.  Note that maybe the apparent last file isn't the last (e.g.
+// there are other files, still to be deleted, in future job parts), in which case any failed deletion
+// will be retried if there's a new "candidate last child" removed.
+// Takes URLs rather than strings because that ensures correct (un)escaping, and makes it clear that we
+// don't support Windows & MacOS local paths (which have cases insensitivity that we don't support here).
+type FolderDeletionManager interface {
+
+	// RecordChildExists takes a child name and records it counts it against the child's immediate parent
```
rickle-msft: "records it counts it" sounds like a typo

common/folderDeletionManager.go
```diff
@@ -0,0 +1,226 @@
...
+	folderStatePtr, alreadyKnown := s.contents[folder]
+	if !alreadyKnown {
+		// we are not tracking this child, so there is nothing that we should do in response
+		// to its deletion (may happen in the recursive calls from tryDeletion, when they recurse up to parent dirs)
+		s.mu.Unlock()
+		return
+	}
+	folderStatePtr.childCount--
+	if folderStatePtr.childCount < 0 {
+		// should never happen. If it does it means someone called RequestDeletion and Recorded a child as deleted, without every registering the child as known
```
rickle-msft: "without every registering"

common/folderDeletionManager.go
```diff
@@ -0,0 +1,226 @@
...
+		s.tryDeletion(folderStr, deletionFunc)
+	}
+}
+
+func (s *standardFolderDeletionManager) tryDeletion(folder string, deletionFunc FolderDeletionFunc) {
+	success := deletionFunc(s.ctx, s.logger) // for safety, deletionFunc should be coded to do nothing, and return false, if the directory is not empty
+
+	if success {
+		s.mu.Lock()
+		delete(s.contents, folder)
```
rickle-msft: Is this locking around a network call?

JohnRusk: Good question.  No, it's just locking around the delete from the in-memory map.

common/folderDeletionManager.go
JohnRusk: Yeah, must be typo because it sure doesn't sound like English! ;-) i'll fix it.

cmd/copyEnumeratorInit.go
```diff
@@ -521,11 +521,13 @@ func (cca *cookedCopyCmdArgs) makeEscapedRelativePath(source bool, dstIsDir bool
 	return pathEncodeRules(relativePath)
 }
 
-func newFolderPropertyOption(bothFolderAware bool, recursive bool, stripTopDir bool, filters []objectFilter) (common.FolderPropertyOption, string) {
-	if bothFolderAware {
+func newFolderPropertyOption(fromTo common.FromTo, recursive bool, stripTopDir bool, filters []objectFilter) (common.FolderPropertyOption, string) {
+	bothFolderAware := fromTo.AreBothFolderAware()
+	isRemoveFromFolderAware := fromTo == common.EFromTo.FileTrash()
+	if bothFolderAware || isRemoveFromFolderAware {
```
adreed-msft: Is it possible to consider the trash location (I know we track it as unknown) folder-aware?

It's not like we're using it as a source anywhere.

JohnRusk: What would be the benefit of that?  Simplifying this code here?  Something else?

cmd/zt_remove_file_test.go
```diff
@@ -76,14 +76,19 @@ func (s *cmdIntegrationSuite) TestRemoveFilesUnderShare(c *chk.C) {
 	raw := getDefaultRemoveRawInput(rawShareURLWithSAS.String())
 	raw.recursive = true
 
+	includeRootInTransfers := true // this is our current behaviour (schedule it, but STE does nothing for that removal transfer).
+	// TODO: review
+
```
adreed-msft: why wouldn't STE remove the root? It seems counter-intuitive to recursively delete a folder but not actually delete the folder.

Unless this is the container root

cmd/zt_scenario_helpers_for_test.go
```diff
@@ -696,7 +696,12 @@ func validateRemoveTransfersAreScheduled(c *chk.C, isSrcEncoded bool, expectedTr
 		// look up the source from the expected transfers, make sure it exists
 		_, srcExist := lookupMap[srcRelativeFilePath]
 		c.Assert(srcExist, chk.Equals, true)
+
+		delete(lookupMap, srcRelativeFilePath)
 	}
+	//if len(lookupMap) > 0 {
+	//	panic("set breakpoint here to debug")
+	//}
```
adreed-msft: should this be removed?

ste/mgr-JobMgr.go
```diff
@@ -152,6 +152,7 @@ func (jm *jobMgr) logConcurrencyParameters() {
 type jobMgrInitState struct {
 	securityInfoPersistenceManager *securityInfoPersistenceManager
 	folderCreationTracker          common.FolderCreationTracker
+	folderDeletionManager          common.FolderDeletionManager
 }
```
adreed-msft: Glad to see the use this is getting. It's a wonderful little way for us to globally track state without querying every item to do so.

ste/xfer-deleteBlob.go
```diff
@@ -14,20 +14,29 @@ import (
...
-
-	info := jptm.Info()
-	// Get the source blob url of blob to delete
-	u, _ := url.Parse(info.Source)
-
-	srcBlobURL := azblob.NewBlobURL(*u, p)
+func DeleteBlob(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer) {
 
 	// If the transfer was cancelled, then reporting transfer as done and increasing the bytestransferred by the size of the source.
 	if jptm.WasCanceled() {
```
adreed-msft: Not really sure what you meant with this comment. Clarify?

ste/xfer-deleteFile.go
```diff
@@ -11,24 +12,77 @@ import (
...
 	// Sets the transfer status and Report Transfer as Done.
 	// Internal function is created to avoid redundancy of the above steps from several places in the api.
 	transferDone := func(status common.TransferStatus, err error) {
+		if status == common.ETransferStatus.Success() {
+			jptm.FolderDeletionManager().RecordChildDeleted(srcUrl)
+			// TODO: doing this only on success raises the possibility of the
+			//   FolderDeletionManager's internal may growing rather large if there are lots of failures
+			//   on a big folder tree. Is living with that preferable to the "incorrectness" of calling
+			//   RecordChildDeleted when it wasn't actually deleted.  Yes, probably.  But think about it a bit more.
+		}
```
adreed-msft: I'm somewhere in between.

We don't record the folder deletion as a transfer failure, so it wouldn't hurt to call this and have the folder deletion fail

Alternative naming suggestion:

RecordChildDeletionAttempted

cmd/zt_scenario_helpers_for_test.go
JohnRusk: I was leaving it there as a way to remind myself, in the future, how to debug that test.

cmd/copyEnumeratorInit.go
adreed-msft: Simplifying the code here for one. This is a really minor nitpick, so feel free to dismiss.

ste/xfer-deleteBlob.go
adreed-msft: (er, this apparently is older code)

cmd/copyEnumeratorInit.go
JohnRusk: Thanks for clarifying.  Given time constraints, I'm going to leave it as-is, since any change would take a non-zero amount of time ;-)

cmd/zt_remove_file_test.go
JohnRusk: I have clarified the comment, thanks.

Enumerators always include the root.  If the root is a directory, then the STE will delete it. But if the root is the share itself (container root as per your comment) then the STE will skip it.

ste/xfer-deleteBlob.go
JohnRusk: Yeah, old code. Probably a copy and paste error in the comment (copied from a non-deletion case).  Will leave as is.

ste/xfer-deleteFile.go
JohnRusk: Let's favor correctness over memory-efficiency for now, and leave the code as it is.  If we find that memory usage is an issue in cases with lots of failures, we can revisit.

I have updated the TODO a little.

PR #880, opened 2020-03-04
Support Azure files snapshots

JohnRusk:
There's a lot of code changed here, but there's only two key parts to it.

1. Add a field to the job part plan file, to hold extra query params (that are not part of the SAS). This is where something like `sharesnapshot=2020-03-03T20%3A24%3A13.0000000Z` ends up if it is included in the source URL.
2. Refactor the handing of URLs in the front end (enumeration) code. Instead of having two fields for source - source and sourceSAS - have one struct with *three* fields: the main value, the SAS, and extra query params. Ditto for the destination.  By using this struct, not only do we have somewhere to store the extra query params, but also it is explicit when we are dealing with just the "non-query" part of the URL, vs the full URL, including all query params.

Without the clear separation between the different elements, provided by item 2 above, we ended up with some confusion.  E.g. this line

```
	} else if !source && !cca.stripTopDir {
		// We ONLY need to do this adjustment to the destination.
		// The source SAS has already been removed. No need to convert it to a URL or whatever.
		// Save to a directory
		rootDir := filepath.Base(cca.source)  // <--- HERE
```

Was including the snapshot identifier as part of "rootDir", because the extra query params were in cca.source.

### DIFF
--- a/cmd/syncTraverser.go
+++ b/cmd/syncTraverser.go
@@ -39,9 +39,9 @@ func newLocalTraverserForSync(cca *cookedSyncCmdArgs, isSource bool) (*localTrav
 	var fullPath string
 
 	if isSource {
-		fullPath = cca.source
+		fullPath = cca.source.ValueLocal()
 	} else {
-		fullPath = cca.destination
+		fullPath = cca.destination.ValueLocal()
 	}
 
 	if strings.ContainsAny(strings.TrimPrefix(fullPath, common.EXTENDED_PATH_PREFIX), "*?") {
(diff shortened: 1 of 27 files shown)


cmd/syncTraverser.go
```diff
@@ -76,15 +76,9 @@ func newBlobTraverserForSync(cca *cookedSyncCmdArgs, isSource bool) (t *blobTrav
 	// figure out the right URL
 	var rawURL *url.URL
 	if isSource {
-		rawURL, err = url.Parse(cca.source)
-		if err == nil && cca.sourceSAS != "" {
-			copyHandlerUtil{}.appendQueryParamToUrl(rawURL, cca.sourceSAS)
-		}
+		rawURL, err = cca.source.FullURL()
 	} else {
```
adreed-msft: I do like how this is reducing the amount of code duplication in some areas.

common/rpc-models.go
```diff
@@ -37,6 +39,71 @@ func (c *RpcCmd) Parse(s string) error {
 	return err
 }
 
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// ResourceString represents a source or dest string, that can have
+// three parts: the main part, a sas, and extra query parameters that are not part of the sas.
+type ResourceString struct {
+	Value      string
```
adreed-msft: Do we have tests for this?

adreed-msft:
The shift to a single container type for our paths was a thought I'd had back when I was working on extended paths for windows. Glad to see it realized.

common/rpc-models.go
JohnRusk: No (other than implicitly testing it in the existing enumeration tests - although some of those will be calling the enumerators directly and so missing this).  I don't have any clear ideas in my head on how to add extra tests.  Do you have any suggestions?

JohnRusk:
>The shift to a single container type for our paths was a thought I'd had back when I was working on extended paths for windows. Glad to see it realized.

I think I remember us having a conversation about it!

common/rpc-models.go
JohnRusk: I'll merge this now, but can add tests separately if we can come up with thoughts on how to do so.

PR #883, opened 2020-03-04
Merge 10.3.4 master changes into dev

JohnRusk:
We had some changes that only went to master, for the hot fixes for 10.3.4.  This gets them into dev.


JohnRusk:
Last smoke test run had a single failure, which was not present in the first run on this branch. Nor is it explained by any changes since that first run.  So I'm merging.

PR #884, opened 2020-03-04
Restrict implicit auth forms to known good destinations

JohnRusk:
Still to do: change from env var to command line param.

### DIFF
--- a/cmd/credentialUtil.go
+++ b/cmd/credentialUtil.go
@@ -26,6 +26,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/minio/minio-go/pkg/s3utils"
 	"net/http"
 	"net/url"
 	"strings"
@@ -258,14 +259,12 @@ type rawFromToInfo struct {
 }
 
 const trustedSuffixesNameAAD = "trusted-aad-suffixes"
-const trustedSuffixesNameS3 = "trusted-s3-suffixes"
 const trustedSuffixesAAD = "*.core.windows.net;*.core.chinacloudapi.cn;*.core.cloudapi.de;*.core.usgovcloudapi.net"
-const trustedSuffixesS3 = "*.amazonaws.com;*.amazonaws.com.cn"
 
 // checkAuthSafeForTarget checks our "implicit" auth types (those that pick up creds from the environment
 // or a prior login) to make sure they are only being used in places where we know those auth types are safe.
 // This prevents, for example, us accidentally sending OAuth creds to some place they don't belong
-func checkAuthSafeForTarget(ct common.CredentialType, resource, extraSuffixesAAD, extraSuffixesS3 string) error {
+func checkAuthSafeForTarget(ct common.CredentialType, resource, extraSuffixesAAD string) error {
 
 	getSuffixes := func(list string, extras string) []string {
 		extras = strings.Trim(extras, " ")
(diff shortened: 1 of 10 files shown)


cmd/credentialUtil.go
```diff
@@ -257,35 +257,146 @@ type rawFromToInfo struct {
...
+const trustedSuffixesAAD = "*.core.windows.net;*.core.chinacloudapi.cn;*.core.cloudapi.de;*.core.usgovcloudapi.net"
+const trustedSuffixesS3 = "*.amazonaws.com;*.amazonaws.com.cn"
+
+// checkAuthSafeForTarget checks our "implicit" auth types (those that pick up creds from the environment
+// or a prior login) to make sure they are only being used in places where we know those auth types are safe.
+// This prevents, for example, us accidentally sending OAuth creds to some place they don't belong
+func checkAuthSafeForTarget(ct common.CredentialType, resource, extraSuffixesAAD, extraSuffixesS3 string) error {
+
+	getSuffixes := func(list string, extras string) []string {
+		extras = strings.Trim(extras, " ")
```
adreed-msft: >Trim(extras, " ") [](start = 19, length = 17)

strings.TrimSpace exists

cmd/credentialUtil.go
```diff
@@ -257,35 +257,146 @@ type rawFromToInfo struct {
...
+
+	default:
+		panic("unknown credential type")
+	}
+
+	return nil
+}
+
+func logAuthType(ct common.CredentialType, location common.Location, isSource bool) {
+	if location == common.ELocation.Unknown() {
```
adreed-msft: We may want to do this once per auth location, if not already done.

cmd/credentialUtil.go
```diff
@@ -294,64 +405,40 @@ func getCredentialInfoForLocation(ctx context.Context, location common.Location,
 		} else {
 			credInfo.OAuthTokenInfo = *tokenInfo
 		}
+	} else if credInfo.CredentialType == common.ECredentialType.S3AccessKey() {
+		// TODO didn't seem to be any code for this in the original getCredentialInfoForLocation
+		//   but presumably we need some, based only on the fact that there are fields in credInfo for S3 stuff ,but maybe
+		//   those fields are never used?
 	}
```
adreed-msft: S3 auth is a weird, complicated beast. It happens when we get our S3 client, unlike the rest of our auth.

It'd be cool if we could refactor this in line with the rest of our auth, but it's not entirely critical.

cmd/credentialUtil.go
```diff
@@ -294,64 +405,40 @@ func getCredentialInfoForLocation(ctx context.Context, location common.Location,
...
-		}
-	case common.EFromTo.LocalFile(), common.EFromTo.FileLocal(), common.EFromTo.FileTrash(), common.EFromTo.FilePipe(), common.EFromTo.PipeFile(), common.EFromTo.BenchmarkFile(),
-		common.EFromTo.FileFile(), common.EFromTo.BlobFile():
-		if credentialType, err = getAzureFileCredentialType(); err != nil {
-			return common.ECredentialType.Unknown(), err
-		}
+// TODO: consider replace with calls to getCredentialInfoForLocation
+// (right now, we have tweaked this to be a wrapper for that function, but really should remove this one totally)
+func getCredentialType(ctx context.Context, raw rawFromToInfo) (credType common.CredentialType, err error) {
+
```
adreed-msft: There's another PR for this in the works, but we should really get to that after we get to this.

cmd/credentialUtil.go
JohnRusk: Oh!  I will remember that for next time. Thanks.

cmd/credentialUtil.go
JohnRusk: It does it once per message, which amounts to the same thing.

cmd/credentialUtil.go
JohnRusk: Ah, I see. Thanks. I'll change the TODO into a comment explaining that.

PR #891, opened 2020-03-09
SMB property persistence

adreed-msft:
Includes the persistence of last write time, creation time, and file attribute bits.

Relies on the completion of the files PR and a release of files-go


ste/sourceInfoProvider.go
```diff
@@ -74,10 +76,13 @@ type IBlobSourceInfoProvider interface {
 	BlobType() azblob.BlobType
 }
 
-type ISDDLBearingSourceInfoProvider interface {
+type ISMBPropertyBearingSourceInfoProvider interface {
 	ISourceInfoProvider
 
 	GetSDDL() (string, error)
+	GetFileSMBCreationTime() (time.Time, error)
```
JohnRusk: Question/suggestion: what do you think about

(a) Modifying azFiles.SMBPropertyHolder to drop the FilePermissionKey() that it currently has. Nothing seems to be using that, as far as I can tell. That leaves an interface with just creationTime, lastWriteTime and attributes.
(b) then modifiy ISMBPropertyBearingSourceInfoProvider, so that it has a GetSMBProperties() method, which returns a SMBPropertyHolder.

This would align the AzCopy code with the interface that's already in the files SDK, and I think it would also make some of the code simpler here with less duplication. E.g. in the local source info provider, we'd hit the filesystem just once to ask for the properties, instead of multiple times as we do now. (And that can matter, locally, in some cases with large numbers of small files).

ste/downloader.go
```diff
@@ -43,12 +43,14 @@ type downloader interface {
...
 // Code outside of windows-specific files shouldn't implement this ever.
-type sddlAwareDownloader interface {
+type smbPropertyAwareDownloader interface {
 	downloader
 
-	PutSDDL(sip ISDDLBearingSourceInfoProvider, txInfo TransferInfo) error
+	PutSDDL(sip ISMBPropertyBearingSourceInfoProvider, txInfo TransferInfo) error
+
+	PutFileTimes(sip ISMBPropertyBearingSourceInfoProvider, txInfo TransferInfo) error
 }
```
JohnRusk: Nice how this is just slotting into the existing structure used for the SDDL

ste/downloader-azureFiles_windows.go
```diff
@@ -4,13 +4,69 @@ package ste
...
+
+	// =========== set file times ===========
+
+	smbCreation, err := sip.GetFileSMBCreationTime()
+
+	if err != nil {
+		return err
+	}
+
+	// reviewers: We already persist last modified time (I think?) to some extent.
```
JohnRusk: We have preserve-last-modified-time.  I'd suggest you make it an error to set both that AND preserve-smb-attributes. (I.e. a normal "cook time" error).

ste/downloader-azureFiles_windows.go
```diff
@@ -4,13 +4,69 @@ package ste
...
+	"syscall"
 
 	"golang.org/x/sys/windows"
 )
 
-// This file implements the windows-triggered sddlAwareDownloader interface.
+// This file implements the windows-triggered smbPropertyAwareDownloader interface.
 
-func (bd *azureFilesDownloader) PutSDDL(sip ISDDLBearingSourceInfoProvider, txInfo TransferInfo) error {
+func (bd *azureFilesDownloader) PutFileTimes(sip ISMBPropertyBearingSourceInfoProvider, txInfo TransferInfo) error {
```
JohnRusk: What about calling this PutFileProperties (to align it with the flag name)?

ste/downloader-azureFiles_windows.go
```diff
@@ -4,13 +4,69 @@ package ste
...
+
+	defer windows.Close(fd)
+
+	// windows.NsecToFileTime does the opposite of FileTime.Nanoseconds, and adjusts away the unix epoch for windows.
+	smbCreationFileTime := windows.NsecToFiletime(smbCreation.UnixNano())
+	smbLastWriteFileTime := windows.NsecToFiletime(smbLastWrite.UnixNano())
+
+	err = windows.SetFileTime(fd, &smbCreationFileTime, nil, &smbLastWriteFileTime)
+
+	fmt.Printf("attribs: %d lwtime: %s ctime: %s\n", attribs, smbLastWrite.Local(), smbCreation.Local())
```
JohnRusk: :eyes:

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/Azure/azure-storage-azcopy
 require (
 	github.com/Azure/azure-pipeline-go v0.2.1
 	github.com/Azure/azure-storage-blob-go v0.7.0
-	github.com/Azure/azure-storage-file-go v0.6.1-0.20191219012846-66d72a9c7823
+	github.com/Azure/azure-storage-file-go v0.6.1-0.20200306202433-9355d1265351
```
JohnRusk: Will this be the final version number? Or is there an update to follow once the storage-file-go PR is merged?

cmd/sync.go
```diff
@@ -573,7 +580,8 @@ func init() {
 
 	rootCmd.AddCommand(syncCmd)
 	syncCmd.PersistentFlags().BoolVar(&raw.recursive, "recursive", true, "True by default, look into sub-directories recursively when syncing between directories. (default true).")
-	syncCmd.PersistentFlags().BoolVar(&raw.preserveNTFSACLs, "preserve-ntfs-acls", false, "False by default. Preserves NTFS ACLs between aware resources (Windows and Azure Files)")
+	syncCmd.PersistentFlags().BoolVar(&raw.preserveSMBPermissions, "preserve-smb-permissions", false, "False by default. Preserves NTFS ACLs between aware resources (Windows and Azure Files)")
```
JohnRusk: Change this to SMB too pls

cmd/copy.go
```diff
@@ -634,15 +641,15 @@ func (raw *rawCopyCmdArgs) setMandatoryDefaults() {
...
 	if toPreserve && !(fromTo == common.EFromTo.LocalFile() ||
 		fromTo == common.EFromTo.FileLocal() ||
 		fromTo == common.EFromTo.FileFile()) {
-		return fmt.Errorf("preserve-ntfs-acls is set but the job is not between NTFS ACL aware resources")
+		return fmt.Errorf("%s is set but the job is not between aware resources", flagName)
 	}
 
 	if toPreserve && (fromTo.IsUpload() || fromTo.IsDownload()) && runtime.GOOS != "windows" {
-		return fmt.Errorf("preserve-ntfs-acls is set but ACL persistence for up/downloads is a windows-only feature")
+		return fmt.Errorf("%s is set but persistence for up/downloads is a windows-only feature", flagName)
```
JohnRusk: Capital W needed in our product name ;-)

cmd/copy.go
```diff
@@ -634,15 +641,15 @@ func (raw *rawCopyCmdArgs) setMandatoryDefaults() {
 	raw.forceWrite = common.EOverwriteOption.True().String()
 }
 
-func validatePreserveNTFSACLs(toPreserve bool, fromTo common.FromTo) error {
+func validatePreserveSMBPropertyOption(toPreserve bool, fromTo common.FromTo, flagName string) error {
 	if toPreserve && !(fromTo == common.EFromTo.LocalFile() ||
 		fromTo == common.EFromTo.FileLocal() ||
 		fromTo == common.EFromTo.FileFile()) {
-		return fmt.Errorf("preserve-ntfs-acls is set but the job is not between NTFS ACL aware resources")
+		return fmt.Errorf("%s is set but the job is not between aware resources", flagName)
```
JohnRusk: Change to "between SMB-aware resources", I'd suggest.

cmd/copy.go
```diff
@@ -453,7 +460,7 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 			cooked.pageBlobTier != common.EPageBlobTier.None() {
 			return cooked, fmt.Errorf("blob-tier is not supported while uploading to ADLS Gen 2")
 		}
-		if cooked.preserveNTFSACLs {
+		if cooked.preserveSMBPermissions {
```
JohnRusk: Next line needs NTFS replaced with SMB. Can you also to a search for NTFS to see if any others have been missed?

ste/sourceInfoProvider.go
adreed-msft: I think that'd be a good idea.

ste/sourceInfoProvider-File.go
```diff
@@ -157,5 +220,7 @@ func (p *fileSourceInfoProvider) GetFileLastModifiedTime() (time.Time, error) {
 		return time.Time{}, err
 	}
 
+	// We ignore smblastwrite because otherwise the tx will fail s2s
```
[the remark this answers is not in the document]
adreed-msft: During a previous point in the PR, I'd considered the SMB last write time as a part of the last modified time. Turns out, this can vary from the REST last write time, and cause an item that's not modified during transfer to think it is.

ste/sourceInfoProvider-File.go
```diff
@@ -48,7 +51,73 @@ func newFileSourceInfoProvider(jptm IJobPartTransferMgr) (ISourceInfoProvider, e
...
+	}
+
+	return azfile.ParseFileAttributeFlagsString(props.FileAttributes()), nil
+}
+
+// for reviewers: should we worry about proactively getting the latest version of this?
+// If not, then we shouldn't worry about whether this is newer than the REST last write time
+func (p *fileSourceInfoProvider) GetFileSMBLastWriteTime() (time.Time, error) {
+	// Because this is subject to change quite often, it's worthwhile to get a not-cached version for this.
+	presigned, err := p.PreSignedSourceURL()
```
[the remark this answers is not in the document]
adreed-msft: Yeah, it was more for the afforementioned check.

This was changed with the consolidation of SMB properties you suggested in the pre-review comment.

ste/JobPartPlan.go
```diff
@@ -59,7 +59,8 @@ type JobPartPlanHeader struct {
 	DstBlobData           JobPartPlanDstBlob  // Additional data for blob destinations
 	DstLocalData          JobPartPlanDstLocal // Additional data for local destinations
 
-	PreserveNTFSACLs bool
+	PreserveSMBPermissions bool
```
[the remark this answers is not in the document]
adreed-msft: I decided to not update it for this PR since I'd already done it in the SDDL PR, which this was a companion to.

go.mod
adreed-msft: There will need to be an update once storage-file-go is merged.

cmd/copy.go
```diff
@@ -445,6 +445,11 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 		return cooked, err
 	}
 
+	// Since we already don't support preserving LMT on upload or S2S, it's fine to limit this to download.
+	if cooked.fromTo.IsDownload() && cooked.preserveSMBProperties && cooked.preserveLastModifiedTime {
+		return cooked, errors.New("SMB properties include a last modified time. Please prefer either the SMB property or the REST property")
```
JohnRusk: I'm not sure that users will know what this means.  Can we use the actual parameter names here, telling to use either param x or param y, but not both?

adreed-msft: Sure.

ste/sourceInfoProvider-File.go
```diff
@@ -48,7 +51,73 @@ func newFileSourceInfoProvider(jptm IJobPartTransferMgr) (ISourceInfoProvider, e
...
+	if err != nil {
+		return 0, err
+	}
+
+	return azfile.ParseFileAttributeFlagsString(props.FileAttributes()), nil
+}
+
+// for reviewers: should we worry about proactively getting the latest version of this?
+// If not, then we shouldn't worry about whether this is newer than the REST last write time
+func (p *fileSourceInfoProvider) GetFileSMBLastWriteTime() (time.Time, error) {
```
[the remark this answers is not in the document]
JohnRusk: Re-opening this conversation, as discussed.

cmd/copy.go
```diff
@@ -433,11 +435,21 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
...
 	}
 
+	cooked.preserveSMBProperties = raw.preserveSMBProperties
+	if err = validatePreserveSMBPropertyOption(cooked.preserveSMBProperties, cooked.fromTo, "preserve-smb-properties"); err != nil {
+		return cooked, err
+	}
+
+	// Since we already don't support preserving LMT on upload or S2S, it's fine to limit this to download.
+	if cooked.fromTo.IsDownload() && cooked.preserveSMBProperties && cooked.preserveLastModifiedTime {
+		return cooked, errors.New("SMB properties include a last modified time. Please use either --preserve-smb-properties OR --preserve-last-modified-time")
```
JohnRusk: I think I've just realised this is incorrect.  Why? Because sync also sets preserveLastModifiedTime in the job order - and its hard coded to always set it. So we can reject it being set, in the case of sync.

So, I think we probably need to change the code in the STE where we preserve the LMT, to say that we only preserve it if we are NOT preserving SMB properties.  I.e. let user set both, but only do the preservation once?  Thoughts?

ste/sourceInfoProvider-File.go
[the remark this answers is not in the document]
adreed-msft: As per discussion in teams, I made this change.

cmd/copy.go
adreed-msft: I think I agree with preferring the SMB property if explicitly specified.

PR #893, opened 2020-03-11
Re-order fields in job part transfer

JohnRusk:
Just to see if that helps with segfault on ARM in atomic operations


JohnRusk:
It did not help. So discarding this PR

PR #898, opened 2020-03-12
Preserve SMB properties and permissions for folders

JohnRusk:


ste/downloader.go
```diff
@@ -43,17 +40,26 @@ type downloader interface {
 	Epilogue()
 }
 
+type folderDownloader interface {
+	SetFolderProperties(jptm IJobPartTransferMgr) error
+}
+
+//It seems the most simple option simple to say that all downloaders must implement all methods
+//Those that don't support folders should panic in SetFolderProperties (since it will never be called on them)
+type downloaderBase interface {
```
adreed-msft: Suggestion: Rather than making all downloaders implement un-necessary folder methods that just panic and serve no purpose, I suggest we reverse this and make it so folderDownloader includes fileDownloader, making it a logical extension to fileDownloader.

Sure, this means in our xferAnyToRemote and remoteToLocal, we have to type wrangle, BUT we can now add this panic to the xfer functions instead.

ste/sender.go
```diff
@@ -82,11 +70,17 @@ type IFileSender interface {
...
 /////////////////////////////////////////////////////////////////////////////////////////////////
 type IFolderSender interface {
-	ISenderBase
 	EnsureFolderExists() error
 	SetFolderProperties() error
 }
 
+// its clearest to say that all senders must implement both interfaces, with those that don't support folders
+// just panicing in the folder-related methods (which will never be called on those types).
+type ISenderBase interface {
```
adreed-msft: ditto

ste/downloader-azureFiles.go
```diff
@@ -78,14 +78,26 @@ func (bd *azureFilesDownloader) Epilogue() {
 			err := spdl.PutFileSMBProperties(bd.sip.(ISMBPropertyBearingSourceInfoProvider), bd.txInfo)
```
adreed-msft: Perhaps we should change this to PutDestinationSMBProperties, if it's going to be re-used for folders.

ste/downloader-azureFiles_windows.go
```diff
@@ -11,7 +11,8 @@ import (
 
 // This file implements the windows-triggered smbPropertyAwareDownloader interface.
 
-func (bd *azureFilesDownloader) PutFileSMBProperties(sip ISMBPropertyBearingSourceInfoProvider, txInfo TransferInfo) error {
+// works for both folders and files
+func (_ *azureFilesDownloader) PutFileSMBProperties(sip ISMBPropertyBearingSourceInfoProvider, txInfo TransferInfo) error {
```
adreed-msft: >_ [](start = 6, length = 1)

This _ isn't actually necessary

ste/downloader.go
JohnRusk: Thanks!  Will do.

ste/downloader-azureFiles.go
JohnRusk: Is it "File" as in Azure files, or "file" as in thing-that's-not-a-folder ;-)  I don't know!

JohnRusk: I've renamed it to PutSMBProperties. That seemed consistent with the other method in the interface:  PutSDDL

adreed-msft: This sounds good.

---
In reply to: [391993472](https://github.com/Azure/azure-storage-azcopy/pull/898#discussion_r391993472) [](ancestors = 391993472)

ste/downloader-azureFiles_windows.go
```diff
@@ -98,5 +107,10 @@ func (bd *azureFilesDownloader) PutSDDL(sip ISMBPropertyBearingSourceInfoProvide
 		nil,
 	)
 
+	if err != nil && strings.HasPrefix(sddlString, "O:SYG:SYD:") {
+		// TODO: awaiting replies re where this SSDL comes from
+		return errorCantSetLocalSystemSddl
+	}
```
adreed-msft: We should get to the bottom of this before we merge this PR.

adreed-msft: Also, you should remove the trailing "D:" because it's possible for the DACL section to not exist (I think)

---
In reply to: [393339778](https://github.com/Azure/azure-storage-azcopy/pull/898#discussion_r393339778) [](ancestors = 393339778)

JohnRusk: Have removed the D.

I think I'll merge now, since it may take a few days to get to the bottom of it. We can address that in a separate PR

JohnRusk:
Ignoring the one failed smoke test. I believe its a dodgy test.

PR #911, opened 2020-03-17
Fix/read-only handling for Azure Files and Windows

JohnRusk:


ste/xfer-anyToRemote-file.go
```diff
@@ -350,6 +350,8 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s sender, sip ISo
 		}
 	}
 
+	// TODO: should we refactor to force this to accept jptm isLive as a parameter, to encourage it to be checked?
+	//  or should we redefine epilogue to be success-path only, and only call it in that case?
 	s.Epilogue() // Perform service-specific cleanup before jptm cleanup. Some services may actually require setup to make the file actually appear.
```
adreed-msft: It would be a net positive to be certain that we have a jptm instance in all parts of senders and downloaders, even if not used.

ste/xfer-deleteFile.go
```diff
@@ -99,7 +99,11 @@ func doDeleteFile(jptm IJobPartTransferMgr, p pipeline.Pipeline) {
 	}
 
 	// Delete the source file
-	_, err := srcFileUrl.Delete(jptm.Context())
+	helper := &azureFileSenderBase{}
+	err := helper.DoWithOverrideReadOnly(jptm.Context(),
+		func() (interface{}, error) { return srcFileUrl.Delete(jptm.Context()) },
+		srcFileUrl,
+		jptm.GetForceIfReadOnly())
 	if err != nil {
```
adreed-msft: Just thinking, can this ever get called multiple times on a single file?

Perhaps we should have a sync.Once for clearing the flags, if so.

ste/xfer-anyToRemote-file.go
JohnRusk: I think for senders we always do have an instance, because they are constructed with new... factory methods that take the jptm as a parameter.  But yes, I agree that the IsLive check could be improved.  And for downloaders the factory method doesn't take a jptm, so we are not guarnteed to even always have an instance, which is where some of the issues fixed in this PR came from.
I don't plan to fix the downloader factory parameters for 10.4. That refactoring can wait I think.

ste/xfer-deleteFile.go
JohnRusk: It can't get called multiple times on a single file.

PR #915, opened 2020-03-18
Feature/backup mode

JohnRusk:


adreed-msft:
> 

Since it's disallowed on not-windows anyway, perhaps we should panic since this should never happen?

---
Refers to: common/writeThoughFile_linux.go:64 in 8ddd858. [](commit_id = 8ddd858308b723f63c8a6a7cf9e39165e70b0eb4, deletion_comment = False)

JohnRusk:
> Since it's disallowed on not-windows anyway, perhaps we should panic since this should never happen?

The current pattern is that the method is always called, and just does nothing on non-supported platforms. (You can't supply the --backup parameter on non-Win, but we DO call SetBackUpMode on non-Win, since the check againts the flag value is made inside that routine.)

ste/downloader-azureFiles_windows.go
```diff
@@ -103,9 +103,9 @@ func (*azureFilesDownloader) PutSDDL(sip ISMBPropertyBearingSourceInfoProvider,
 		nil,
 	)
 
-	if err != nil && strings.HasPrefix(sddlString, "O:SYG:SY") {
-		// TODO: awaiting replies re where this SSDL comes from
-		return errorCantSetLocalSystemSddl
+	if err != nil {
+		return fmt.Errorf("permissions could not be restored. It may help to run from a elevated command prompt, and set the '%s' flag. Error message was: %w",
+			common.BackupModeFlagName, err)
 	}
```
adreed-msft: We decided to persist these anyway?

JohnRusk: Yes, subject to possible review in a future PR, the plan for now is that when Files gives us a default ACL, we will use it.  (Otherwise, if some have default ACLs and some have user-specifed ACLS, we end up setting no ACL on the defaulted ones - which is probably counter-intuitive to the user)

PR #918, opened 2020-03-20
Fix/files enumeration perf

JohnRusk:
This fixes the Azure Files enumeration perf issue.

The idea is to isolate the parallelism into just a tiny part of the traverser, so we don't have the think about multi-threading anything else (e.g. don't want multi-threaded additions to the plan file(s).)

There's actually two parts to it. One allows directory crawls to be parallelized, and the other parallelizes the extra network roundtrip where we get the LMT and MD5.  Testing indicates both are necessary to get a significant speedup.

### DIFF
--- a/cmd/zc_traverser_file.go
+++ b/cmd/zc_traverser_file.go
@@ -93,7 +93,9 @@ func (t *fileTraverser) traverse(preprocessor objectMorpher, processor objectPro
 
 	// else, its not just one file
 
-	processEntity := func(f azfileEntity) error {
+	// This func must be threadsafe/goroutine safe
+	convertToStoredObject := func(input parallel.InputObject) (parallel.OutputObject, error) {
+		f := input.(azfileEntity)
 		// compute the relative path of the file with respect to the target directory
 		fileURLParts := azfile.NewFileURLParts(f.url)
 		relativePath := strings.TrimPrefix(fileURLParts.DirectoryOrFilePath, targetURLParts.DirectoryOrFilePath)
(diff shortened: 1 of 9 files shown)


cmd/zc_traverser_file.go
```diff
@@ -142,47 +147,64 @@ func (t *fileTraverser) traverse(preprocessor objectMorpher, processor objectPro
...
+		return nil
+	}
+
+	// run the actual enumeration.
+	// First part is a parallel directory crawl
+	// Second part is parallel conversion of the directories and files to stored objects. This is necessary because the conversion to stored object may hit the network and therefore be slow in not parallelized
+	parallelism := 1
+	if enumerationParallelism > 1 {
+		parallelism = enumerationParallelism / 2 // half for crawl, half for transform
+	}
```
adreed-msft: I wonder if it'd be preferable to do 1/3 or 1/4th for transform. My guess is that network calls are more costly than the actual transform.

adreed-msft:
I think this'd work fine if we had to repurpose it for other traversers. The structure feels a little awkward (personally), but I get what it's aiming for and don't really know how I'd do it better immediately. Perhaps a blurb inside the parallel package about how it's intended to be used may make it more understandable in the future.

cmd/zc_traverser_file.go
JohnRusk: I have not yet managed to figure out how much of the network activity is in the transform (where we make the get properties call) and how much in the directory enumeration - I suspect the balance will vary based on directory structure.  So will leave at 1/2 for now.

PR #922, opened 2020-03-23
Fix managed disk corruption risk, if upload two to same URL

adreed-msft:


ste/sender-pageBlob.go
```diff
@@ -53,6 +53,12 @@ type pageBlobSenderBase struct {
 	// Using a automatic pacer here lets us find the right rate for this particular page blob, at which
 	// we won't be trying to move the faster than the Service wants us to.
 	filePacer autopacer
+
+	// destPageRangeOptimizer is necessary for managed disk imports,
+	// as it helps us identify where we actually need to write all zeroes to.
+	// Previously, if a page prefetched all zeroes, we'd ignore it.
+	// This would cause corruption on managed disks.
```
JohnRusk: This comment "would cause corruption" is pretty scary, and also too vague.  To make it more precise, and therefore also less scary, can you reword it to something like: "this could cause corruption if non-zero values had previously been uploaded to the same page in the same md-impex URL. E.g. by previously uploading a different VHD to the same URL"

ste/sender-pageBlob.go
```diff
@@ -53,6 +53,12 @@ type pageBlobSenderBase struct {
...
+	destPageRangeOptimizer *pageRangeOptimizer
```
JohnRusk: Nice name. Can you please rename the other one (the one in sender-pageBlobFromURL) to sourcePageRangeOptimizer, so that it fits into the same naming scheme.

PR #926, opened 2020-03-25
Add diagnostic code for rare putBlockList failures

JohnRusk:


ste/sender-blockBlob.go
```diff
@@ -142,7 +142,22 @@ func (s *blockBlobSenderBase) Epilogue() {
...
+			// Experimental debugging code, to be shipped in release 10.4.
+			// Wait a bit (actually a lot) and retry.
+			time.Sleep(20 * time.Second)
+			err = doCommit()
+			if err == nil {
+				// Log that it worked. Since the outcome of the second call, either positive or negative, will help us understand the rare cases where the first call has failed
+				jptm.LogAtLevelForCurrentTransfer(pipeline.LogInfo, "PUTBLOCKLISTRETRYSUCCESSFUL The putBlockList operation was retried and was successful. There is no error here. This line simply records that the retry was successful.")
+			}
+		}
+		if err != nil {
```
adreed-msft: Just a suggestion, but perhaps if our log level is on DEBUG, we could consider getting a list of uncommitted blocks and comparing it to our list of blocks, and log what's in each list but not the other.

adreed-msft: That was more my initial thought for this diagnostic code when you mentioned it.

ste/sender-blockBlob.go
```diff
@@ -142,7 +142,22 @@ func (s *blockBlobSenderBase) Epilogue() {
...
-		if _, err := s.destBlockBlobURL.CommitBlockList(jptm.Context(), blockIDs, s.headersToApply, s.metadataToApply, azblob.BlobAccessConditions{}); err != nil {
+		doCommit := func() error {
+			_, commitErr := s.destBlockBlobURL.CommitBlockList(jptm.Context(), blockIDs, s.headersToApply, s.metadataToApply, azblob.BlobAccessConditions{})
+			return commitErr
+		}
+		err := doCommit()
+		if strErr, ok := err.(azblob.StorageError); ok && strErr.ServiceCode() == azblob.ServiceCodeInvalidBlockList {
+			// Experimental debugging code, to be shipped in release 10.4.
+			// Wait a bit (actually a lot) and retry.
+			time.Sleep(20 * time.Second)
```
adreed-msft: Is this a service bug, or an azcopy bug we're trying to work around here?

JohnRusk: We don't know what's going on. But, after eliminating all the client-side causes I could think of, this code is here in case its server-side.

ste/sender-blockBlob.go
JohnRusk: Good idea, but more than I can find time to code at the moment. (And also, I suspect that we are providing the _right_ list, and something else is going wrong here).

PR #928, opened 2020-03-26
High blob count optimizations part 1 (to master)

JohnRusk:
These are the optimizations discussed.

The most important is to allow prefix-based searches.  I realized we didn't need any new parameters or syntax at all for that, can just treat it as an optimization of the existing include-pattern.

The other is to back off the progress reporting frequency at high file counts.  Because otherwise, looping through all the transfers to count those at the various statuses becomes expensive. (It still will be expensive if there are skipped or failed files, since as per known bug, they are re-listed every time. That bug is not fixed here.  I.e. this will be good for cases where there aren't too many skips or fails).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,7 +9,7 @@
 first `*` will be used as a server-side prefix, to filter the search results more efficiently. E.g. "--include-path abc*" will be implemented 
 as a prefix search for "abc". In a more complex example, "--include-path abc\*123", will be implemented as a prefix search for "abc", followed
 by client-side filtering to find exact matches to abc\*123.
-2. When processing over a million files, AzCopy will report on its progress once a minute instead of once every 2 seconds.  This reduces the CPU 
+2. When processing over a million files, AzCopy will report on its progress once ever 2 minutes instead of once every 2 seconds.  This reduces the CPU 
 load associated with progress reporting.
 
 ## Version 10.3.4
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,17 @@
 
 # Change Log
 
+## Version 10.3.61 (private drop)
+
+### Performance optimizations
+
+1. Any `--include-pattern` that contains only `*` wildcards will be performance optimized when querying blob storage.  The section before the 
+first `*` will be used as a server-side prefix, to filter the search results more efficiently. E.g. "--include-path abc*" will be implemented 
+as a prefix search for "abc". In a more complex example, "--include-path abc\*123", will be implemented as a prefix search for "abc", followed
+by client-side filtering to find exact matches to abc\*123.
+2. When processing over a million files, AzCopy will report on its progress once a minute instead of once every 2 seconds.  This reduces the CPU 
+load associated with progress reporting.
+
 ## Version 10.3.4
 
 ### New features
--- a/cmd/zc_filter.go
+++ b/cmd/zc_filter.go
@@ -128,7 +128,7 @@ func (f *includeFilter) doesPass(storedObject storedObject) bool {
 		matched := false
 
 		var err error
-		matched, err = path.Match(pattern, checkItem)
+		matched, err = path.Match(pattern, checkItem) // note: getEnumerationPreFilter below encodes assumptions about the valid wildcards used here
 
 		// if the pattern failed to match with an error, then we assume the pattern is invalid
 		// and ignore it
@@ -146,6 +146,24 @@ func (f *includeFilter) doesPass(storedObject storedObject) bool {
 	return false
 }
 
+// getEnumerationPreFilter returns a prefix, if any, which can be used service-side to pre-select
+// things that will pass the filter. E.g. if there's exactly one include pattern, and it is
+// "foo*bar", then this routine will return "foo", since only things starting with "foo" can pass the filters.
+// Service side enumeration code can be given that prefix, to optimize the enumeration.
+func (f *includeFilter) getEnumerationPreFilter() string {
+	if len(f.patterns) == 1 {
+		pat := f.patterns[0]
+		if strings.ContainsAny(pat, "?[\\") {
+			// this pattern doesn't just use a *, so it's too complex for us to optimize with a prefix
+			return ""
+		}
+		return strings.Split(pat, "*")[0]
+	} else {
+		// for simplicity, we won't even try computing a common prefix for all patterns (even though that might help in theory in some cases)
+		return ""
+	}
+}
+
 func buildIncludeFilters(patterns []string) []objectFilter {
 	validPatterns := make([]string, 0)
 	for _, pattern := range patterns {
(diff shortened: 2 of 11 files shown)


adreed-msft:
Disregard testing failure, it's just the macos build (which will always fail on master until we merge dev, as it has the change fixing the macos image version)

cmd/zc_filter.go
```diff
@@ -146,6 +146,24 @@ func (f *includeFilter) doesPass(storedObject storedObject) bool {
 	return false
 }
 
+// getEnumerationPreFilter returns a prefix, if any, which can be used service-side to pre-select
+// things that will pass the filter. E.g. if there's exactly one include pattern, and it is
+// "foo*bar", then this routine will return "foo", since only things starting with "foo" can pass the filters.
+// Service side enumeration code can be given that prefix, to optimize the enumeration.
+func (f *includeFilter) getEnumerationPreFilter() string {
```
adreed-msft: Actually, I redact my approval.

Our include filter isn't repeatedly added. It's a single filter, with multiple potential patterns.

For instance, if I supplied

a*;b*

I suspect this might introduce a bug where none of the b* entries are found.

cmd/zc_filter.go
```diff
@@ -156,3 +174,26 @@ func buildIncludeFilters(patterns []string) []objectFilter {
...
+// The result can be used to optimize enumeration, since anything without this prefix will fail the filterSet
+func (fs filterSet) GetEnumerationPreFilter() string {
+	prefix := ""
+	for _, f := range fs {
+		if participatingFilter, ok := f.(preFilterProvider); ok {
+			// this filter knows how to participate in our scheme
+			if prefix == "" {
+				prefix = participatingFilter.getEnumerationPreFilter()
+			} else {
+				// prefix already has a value, which means there must be two participating filters, and we can't handle that.
```
adreed-msft: I'm also feeling a bit nitpicky about this. If we can find a good system to select a common root prefilter in the individual prefilters, then we can do it here.

So, if the above comment gets changes, this one should too.

cmd/zc_traverser_blob.go
```diff
@@ -142,11 +142,14 @@ func (t *blobTraverser) traverse(preprocessor objectMorpher, processor objectPro
...
 
+	// as a performance optimization, get an extra prefix to do pre-filtering. It's typically the start portion of a blob name.
+	extraSearchPrefix := filterSet(filters).GetEnumerationPreFilter()
+
 	for marker := (azblob.Marker{}); marker.NotDone(); {
 		// look for all blobs that start with the prefix
 		// TODO optimize for the case where recursive is off
 		listBlob, err := containerURL.ListBlobsFlatSegment(t.ctx, marker,
-			azblob.ListBlobsSegmentOptions{Prefix: searchPrefix, Details: azblob.BlobListingDetails{Metadata: true}})
+			azblob.ListBlobsSegmentOptions{Prefix: searchPrefix + extraSearchPrefix, Details: azblob.BlobListingDetails{Metadata: true}})
```
adreed-msft: This... is actually overall a bug, and I don't think we can get past it.

Unless we're in a virtual folder with files like this, a filter of a* can cause a return of no results.

Rather than, as a user might expect, returning results of all files in sub-virtual-directories containing this pattern of a*.

adreed-msft: I think the best path forward _may_ be to just dynamically decrease the size of our estimated job part, as I'd suggested on teams, on second thought.

cmd/zc_filter.go
JohnRusk: Yes. I ignored the potential optimization on purpose in that case.  I felt I couldn't safely code and test that more complex version in the time available.  (I think I even left a comment in the code saying I'd deliberately not done it). 
We could add it to this code when we cherry pick this into 10.4, but I'm not comfortable with adding the complexity of that to this PR - given that the customer we are doing this PR for is time-sensitive, and has no need for a multi-part filter.
@adreed-msft Can you reconsider your approval in the light of the above?

cmd/zc_filter.go
JohnRusk: This is something we could do in 10.4.  I don't think we should do it in this PR, for the reasons outlined in my other answer.

cmd/zc_traverser_blob.go
JohnRusk: Oh, yes, well spotted. It is indeed a bug.

JohnRusk: Tweaking job part size solves a different problem (slow to start first transfer).  But in this PR we're trying to find a way to efficiently select only a subset of what is in the container.

cmd/zc_filter.go
zezha-msft: In this kind of scenario, I'd definitely agree to dumb-down the logic as much as possible to avoid potential bugs.

adreed-msft: Yeah, I'll reconsider approval after I look at the new push. I suspect with our discussion lastnight, I'll be happy with this

ChangeLog.md
```diff
@@ -1,6 +1,17 @@
...
 
+## Version 10.3.61 (private drop)
+
+### Performance optimizations
+
+1. Any `--include-pattern` that contains only `*` wildcards will be performance optimized when querying blob storage.  The section before the 
+first `*` will be used as a server-side prefix, to filter the search results more efficiently. E.g. "--include-path abc*" will be implemented 
+as a prefix search for "abc". In a more complex example, "--include-path abc\*123", will be implemented as a prefix search for "abc", followed
+by client-side filtering to find exact matches to abc\*123.
+2. When processing over a million files, AzCopy will report on its progress once ever 2 minutes instead of once every 2 seconds.  This reduces the CPU
```
adreed-msft: This is inaccurate information. Because the \ is present, we'd ignore attempting to prefilter, because we think it's too complex. This isn't a necessary change right now (we can do it when we get to the rest of this PR, or when we update the changelogs otherwise)

cmd/zc_filter.go
```diff
@@ -156,3 +174,38 @@ func buildIncludeFilters(patterns []string) []objectFilter {
...
+		return ""
+		// we don't/can't support recursive cases yet, with a strict prefix-based search.
+		// Because if the filter is, say "a*", then, then a prefix of "a"
+		// will find: enumerationroot/afoo and enumerationroot/abar
+		// but it will not find: enumerationroot/virtualdir/afoo
+		// even though we want --include-pattern to find that.
+		// So, in recursive cases, we just don't use this prefix-based optimization.
+		// TODO: consider whether we need to support some way to separately invoke prefix-based optimization
+		//   and filtering.  E.g. by a directory-by-directory enumeration (with prefix only within directory),
+		//   using the prefix feature in ListBlobs.
```
adreed-msft: This might be difficult since directories don't actually exist in blob, and in order to discover them, we have to do a full flat listing anyway, iirc.

JohnRusk: There's a solution, with the Prefix field, I understand.

PR #930, opened 2020-03-27
Fix storage explorer's credential location issue when downloading

adreed-msft:
This introduces a breaking change, in that it splits off the credential type overrides into source and destination credential type overrides. This is more consistent with our newer, internal usage of credential types (and with the introduction of User Delegation SAS Token generation, will be beneficial).

This furthermore, ensures we're no longer confused when a user overrides the credential type without both sides being of that credential type.

### DIFF
--- a/cmd/credentialUtil.go
+++ b/cmd/credentialUtil.go
@@ -264,7 +264,7 @@ const trustedSuffixesAAD = "*.core.windows.net;*.core.chinacloudapi.cn;*.core.cl
 // checkAuthSafeForTarget checks our "implicit" auth types (those that pick up creds from the environment
 // or a prior login) to make sure they are only being used in places where we know those auth types are safe.
 // This prevents, for example, us accidentally sending OAuth creds to some place they don't belong
-func checkAuthSafeForTarget(ct common.CredentialType, resource, extraSuffixesAAD string) error {
+func checkAuthSafeForTarget(ct common.CredentialType, resource, extraSuffixesAAD string, resourceType common.Location) error {
 
 	getSuffixes := func(list string, extras string) []string {
 		extras = strings.Trim(extras, " ")
@@ -298,6 +298,20 @@ func checkAuthSafeForTarget(ct common.CredentialType, resource, extraSuffixesAAD
 		return nil
 	case common.ECredentialType.OAuthToken(),
 		common.ECredentialType.SharedKey():
+		// Files doesn't currently support OAuth.
+		// TODO: When adding files OAuth support to AzCopy eventually, fix this if statement.
+		if resourceType != common.ELocation.Blob() && resourceType != common.ELocation.BlobFS() {
+			// There may be a reason for files->blob to specify this.
+			// Files doesn't take OAuth currently, anyway.
+			// And for storage explorer, they just always supply it.
+			// We however, disinclude S3 from this list because S3 is a remote non-azure resource.
+			if resourceType == common.ELocation.Local() || resourceType == common.ELocation.File() {
+				return nil
+			}
+
+			return fmt.Errorf("azure OAuth authentication to %s is not enabled in AzCopy", resourceType.String())
+		}
+
 		// these are Azure auth types, so make sure the resource is known to be in Azure
 		domainSuffixes := getSuffixes(trustedSuffixesAAD, extraSuffixesAAD)
 		if host, ok := isResourceInSuffixList(domainSuffixes); !ok {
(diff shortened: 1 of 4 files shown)


cmd/zt_credentialUtil_test.go
```diff
@@ -34,46 +34,50 @@ var _ = chk.Suite(&credentialUtilSuite{})
...
+		{common.ECredentialType.OAuthToken(), common.ELocation.Blob(), "http://somethingelseinazure.windows.net", "", false},
+		{common.ECredentialType.S3AccessKey(), common.ELocation.S3(), "http://somethingelseinaws.amazonaws.com", "", false},
 
 		// As should these (they are nothing to do with the expected URLs)
-		{common.ECredentialType.OAuthToken(), "http://example.com", "", false},
-		{common.ECredentialType.S3AccessKey(), "http://example.com", "", false},
+		{common.ECredentialType.OAuthToken(), common.ELocation.Blob(), "http://example.com", "", false},
+		{common.ECredentialType.S3AccessKey(), common.ELocation.S3(), "http://example.com", "", false},
+		// Test that we don't want to send an S3 access key to a blob resource type.
+		{common.ECredentialType.S3AccessKey(), common.ELocation.Blob(), "http://example.com", "", false},
```
JohnRusk: Should this url be a blob URL, if its trying to test (only) what the comment above it says?

cmd/credentialUtil.go
```diff
@@ -298,6 +298,20 @@ func checkAuthSafeForTarget(ct common.CredentialType, resource, extraSuffixesAAD
 		return nil
 	case common.ECredentialType.OAuthToken(),
 		common.ECredentialType.SharedKey():
+		// Files doesn't currently support OAuth.
+		// TODO: When adding files OAuth support to AzCopy eventually, fix this if statement.
+		if resourceType != common.ELocation.Blob() && resourceType != common.ELocation.BlobFS() {
+			// There may be a reason for files->blob to specify this.
+			// Files doesn't take OAuth currently, anyway.
```
JohnRusk: Why do we need to consider Azure Files as a special case here?  Why can't we just do if resourceType == common.ELocation.Local() at the very first if statement (the one that currently looks at !Blob && !BlobFS) and then just return nil.  (And that would be the whole thing).

I.e. I'm asking why we can't let the Files locations just run through the host name test. (All Files urls will pass those checks, since the hard-coded domain suffixes deliberately dont' restrict to blob or blobFS).

That way, there would be nothing here that we would need to remember to maintain in future.  Comments of the nature "TODO: when we do X later (in some other part of the codebase) remember to change this" tend to be dangeorous when it comes to security, since they can be forgotten.  (They're equally likely to be overlooked in non-security situations - its just that the consequences are more severe with security).

adreed-msft: Good point that it does match the pattern. No reason to have a special case here.

cmd/zt_credentialUtil_test.go
```diff
@@ -34,46 +34,50 @@ var _ = chk.Suite(&credentialUtilSuite{})
...
 	for i, t := range tests {
-		err := checkAuthSafeForTarget(t.ct, t.resource, t.extraSuffixesAAD)
+		err := checkAuthSafeForTarget(t.ct, t.resource, t.extraSuffixesAAD, t.resourceType)
 		c.Assert(err == nil, chk.Equals, t.expectedOK, chk.Commentf("Failed on test %d for resource %s", i, t.resource))
 	}
 }
 
 func (s *credentialUtilSuite) TestCheckAuthSafeForTargetIsCalledWhenGettingAuthType(c *chk.C) {
-	mockGetCredTypeFromEnvVar := func() common.CredentialType {
+	mockGetCredTypeFromEnvVar := func(isSource bool) common.CredentialType {
```
JohnRusk: This looks like a part of the earlier change, that should be reversed. (Since in the end that param wasn't added)

JohnRusk: I've fixed this in the small commit I just pushed to this PR.

cmd/zt_credentialUtil_test.go
```diff
@@ -34,46 +34,50 @@ var _ = chk.Suite(&credentialUtilSuite{})
...
-		{common.ECredentialType.OAuthToken(), "http://example.com", "", false},
-		{common.ECredentialType.S3AccessKey(), "http://example.com", "", false},
+		{common.ECredentialType.OAuthToken(), common.ELocation.Blob(), "http://example.com", "", false},
+		{common.ECredentialType.S3AccessKey(), common.ELocation.S3(), "http://example.com", "", false},
+		// Test that we don't want to send an S3 access key to a blob resource type.
+		{common.ECredentialType.S3AccessKey(), common.ELocation.Blob(), "http://example.com", "", false},
 
 		// But the same Azure one should pass if the user opts in to them (we don't support any similar override for S3)
-		{common.ECredentialType.OAuthToken(), "http://example.com", "*.foo.com;*.example.com", false},
+		{common.ECredentialType.OAuthToken(), common.ELocation.Blob(), "http://example.com", "*.foo.com;*.example.com", false},
```
JohnRusk: I just noticed something - this test has been wrong all along!  When I wrote it, I was supposed to have "true" at the end.  I've just pushed a small commit to fix it.  URL changes to abc.example.com so that the dot before example is there (which is expected by the *.example.com).

PR #936, opened 2020-03-31
Revert "Add diagnostic code for rare putBlockList failures (#926)"

JohnRusk:
This reverts commit 74b25450d659bc6c14248d89bbce72deb643b45f because the whole retry idea looks like it was a red herring. (The case that prompted us to write it was actually caused by user hitting same blob with two instances of AzCopy at the same time.  Nothing to do with anything that any retry could have helped with).


adreed-msft:
We may well have had the same customer, somebody reported a really similar issue to me (and I told them it wasn't really supported in azcopy)

JohnRusk:
>it wasn't really supported in azcopy

You mean two processes write to same blob?  Yeah, it kind of always has to be a failure for the "losing" process.  Maybe one day we'll make the message clearer.

adreed-msft:
Correct, that's what I meant.

PR #937, opened 2020-03-31
Disallow combination of perserve-smb-permissions with follow-symlinks

JohnRusk:


cmd/copy.go
```diff
@@ -699,6 +703,13 @@ func validatePreserveSMBPropertyOption(toPreserve bool, fromTo common.FromTo, ov
 	return nil
 }
 
+func crossValidateSymlinksAndPermissions(followSymlinks, preservePermissions bool) error {
+	if followSymlinks && preservePermissions {
+		return errors.New("cannot follow symlinks when preserving permissions (since the correct permission inheritance behaviour for symlink targets is hard to define)")
```
adreed-msft: >hard to define [](start = 147, length = 14)

I would prefer undefined, here.

cmd/sync.go
```diff
@@ -153,6 +153,9 @@ func (raw *rawSyncCmdArgs) cook() (cookedSyncCmdArgs, error) {
 	}
 
 	cooked.followSymlinks = raw.followSymlinks
+	if err = crossValidateSymlinksAndPermissions(cooked.followSymlinks, true /* replace with real value when available */); err != nil {
```
adreed-msft: >true [](start = 69, length = 4)

Shouldn't this be false, right now, if sync doesn't support persist smb permissions?

adreed-msft:
Consider it approved once the sync side of this is fixed

cmd/sync.go
JohnRusk: Sync also doesn't support follow symlinks right now.  

So, by hard coding true, here, it will immediately fail when someone first makes sync support symlinks, and then they'll know to hook in the smp property correctly.

I.e. anything we hard-code here is incorrect, but better to have something that will obviously fail when symlink support is first tested (by us, not by customer!) than something which will silently do the wrong thing, and therefore never get fixed.

adreed-msft: Ah, I suppose that does make sense. I thought we did have symlinks enabled in sync. Perhaps you should add a comment here describing that.

cmd/copy.go
JohnRusk: Done

PR #940, opened 2020-04-01
Fix/symlink processsing

JohnRusk:
In 10.4, we have already switched all our directory tree crawling to use the symlink-aware code, to reduce code duplication. It respects the followSymlinks flag that is passed into it.

But, this meant that any memory usage issues in that code would always have an effect. So this PR fixes two memory usage issues:
   1. previously the code ran the dedupe map even when followSymlinks was false AND
   1. previously it put files into the dedupe map as well as directories.

It also fixes a minor bug where we did not pick up directory properties for the directory that is the target of a symlink; and another minor bug where the Folders as First Class Citizens work had caused it to output a "duplicate skipped" message _every_ time.

This PR does not fix the other existing bug that was found, which is that: The existing code would not process a symlink to a file. (It would always treat it as a duplicate).  That can wait, and be tested properly, in 10.5.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -72,7 +72,6 @@ contained directly in a container or virtual directory include `/*` at the end o
 include/exclude patterns were disallowed, but exclude-path was not. That was incorrect. All should have been
 disallowed because none (other than include-path) are respected. 
    
-
 ## Version 10.3.4
 
 ### New features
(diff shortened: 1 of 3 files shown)


JohnRusk:
Abandonning. The change to not save file names, while good for RAM usage, is hard to get right with the existing code structure

JohnRusk:
One test is still failing now, and the implementation as it stands is this PR is incorrect

PR #942, opened 2020-04-02
Fix/symlink processsing (PR 2)

JohnRusk:
This is a new PR, to replace an earlier abandoned one.

This one is less ambitious. Specifically it does not reduce memory usage for the follow-symlinks case.

### DIFF
--- a/cmd/zc_traverser_local.go
+++ b/cmd/zc_traverser_local.go
@@ -94,6 +94,15 @@ func (r *realSeenPathsRecorder) HasSeen(path string) bool {
 	return ok
 }
 
+type symlinkTargetFileInfo struct {
+	os.FileInfo
+	name string
+}
+
+func (s symlinkTargetFileInfo) Name() string {
+	return s.name // override the name
+}
+
 // WalkWithSymlinks is a symlinks-aware version of filePath.Walk.
 // Separate this from the traverser for two purposes:
 // 1) Cleaner code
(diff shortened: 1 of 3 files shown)


cmd/zc_traverser_local.go
```diff
@@ -122,33 +165,69 @@ func WalkWithSymlinks(fullPath string, walkFunc filepath.WalkFunc, followSymlink
...
-					glcm.Info(fmt.Sprintf("Ignored already linked directory pointed at %s (link at %s)", result, common.GenerateFullPath(fullPath, computedRelativePath)))
+					glcm.Info(fmt.Sprintf("Symlinks to individual files are not currently supported, so will ignore file at %s (link at %s)", result, common.GenerateFullPath(fullPath, computedRelativePath)))
+					// TODO: remove the above info call and enable the below, with suitable multi-OS testing
+					//    including enable the test: TestWalkWithSymlinks_ToFile
+					/*
+							// It's a symlink to a file. Just process the file because there's no danger of cycles with links to individual files.
+							// (this does create the inconsistency that if there are two symlinks to the same file we will process it twice,
+							// but if there are two symlinks to the same directory we will process it only once. Because only directories are
+							// deduped to break cycles.  For now, we are living with the inconsistency. The alternative would be to "burn" more
+							// RAM by putting filepaths into seenDirs too, but that could be a non-trivial amount of RAM in big directories trees).
```
adreed-msft, line 199: I think it makes sense to not dedupe files here, in this case.

cmd/zc_traverser_local.go
```diff
@@ -122,33 +165,69 @@ func WalkWithSymlinks(fullPath string, walkFunc filepath.WalkFunc, followSymlink
...
+
+							// TODO: this code here won't handle the case of (file-type symlink) -> (another file-type symlink) -> file
+						    //    But do we WANT to handle that?  (since it opens us to risk of file->file cycles, and we are deliberately NOT
+						    //    putting files in our map, to reduce RAM usage).  Maybe just detect if the target of a file symlink its itself a symlink
+						    //    and skip those cases with an error message?
+							// Make file info that has name of source, and stats of dest (to mirror what os.Stat calls on source will give us later)
```
adreed-msft, line 205: Actually, maybe instead of putting files in the map, we could just temporarily hold a map for this particular operation and then discard it.

cmd/zt_generic_traverser_test.go
```diff
@@ -198,7 +206,49 @@ func (s *genericTraverserSuite) TestWalkWithSymlinks(c *chk.C) {
 
 	// 3 files live in base, 3 files live in symlink
 	c.Assert(fileCount, chk.Equals, 6)
+	c.Assert(sawLinkTargetDir, chk.Equals, true)
+}
+
+// Next test is temporarily disabled, to avoid changing functionality near 10.4 release date
+/*
```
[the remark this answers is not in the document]
adreed-msft: We do need at least _basic_ symlink following for files-- even if it's just that temporary map for handling individual file symlinks.

---
In reply to: [402580645](https://github.com/Azure/azure-storage-azcopy/pull/942#discussion_r402580645) [](ancestors = 402580645)

cmd/zc_traverser_local.go
JohnRusk: Let's review for Files in 10.4.1, as discussed.

cmd/zt_generic_traverser_test.go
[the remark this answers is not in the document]
JohnRusk: As discussed, it's not a breaking change because following file symlinks was already broken in 10.3.4 ;-)   So let's not complicate things by fixing that now, but let's fix it in, say 10.4.1 or something.

PR #946, opened 2020-04-04
Fix/parallel enumeration

JohnRusk:
Old implementation was prone to deadlocks due to blocking nature of fixed-capacity channel

### DIFF
--- a/common/parallel/TreeCrawler.go
+++ b/common/parallel/TreeCrawler.go
@@ -32,7 +32,7 @@ type crawler struct {
 	parallelism int
 	cond        *sync.Cond
 	// the following is protected by cond (and must only be accessed when cond.L is held)
-	unstartedDirs      chan Directory // protected by cond.L because we use len() on this, and need to hold lock while making len-based decisions
+	unstartedDirs      []Directory // not a channel, because channels have length limits, and those get in our way
 	dirInProgressCount int64
 }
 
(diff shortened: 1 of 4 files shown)


common/parallel/TreeCrawler.go
```diff
@@ -156,11 +163,26 @@ func (c *crawler) processOneDirectory(ctx context.Context) (bool, error) {
...
+	// (It's impossible to know exactly what to do here, because we don't know whether more workers would _clear_
+	// the queue more quickly; or _add to_ the queue more quickly.  It depends on whether the directories we process
+	// next contain mostly child directories or if they are "leaf" directories containing mostly just files.  But,
+	// if we slowly reduce parallelism the end state is roughly equivalent to a single-threaded depth-first traversal, which
+	// is generally fine in terms of memory usage on most folder structures)
+	const maxQueueDirectories = 1000 * 1000
+	shouldShutSelfDown := len(c.unstartedDirs) > maxQueueDirectories && // we are getting way too much stuff queued up
+		workerIndex > (c.parallelism/4) && // never shut down the last ones, since we need something left to clear the queue
+		time.Since(c.lastAutoShutdown) > time.Second // adjust slightly gradually
+	if shouldShutSelfDown {
```
adreed-msft, line 182: Just a thought, but would only working on the newest items help as well?

I mean, the thinking is that we are more likely to clear a single branch (and probably remove a bunch of items from the pool) if we prefer only newer items, rather than just older ones, since we're perpetually opening a whole new, likely larger can of worms with every lower-level branch we open.

adreed-msft: This is only really a thought since we now have the grace of being able to implement this as a double ended queue

PR #948, opened 2020-04-06
Fix an S3 text processing issue caused by the text processing PR

adreed-msft:


common/extensions.go
```diff
@@ -30,10 +31,18 @@ type URLExtension struct {
...
-	if u.RawPath != "" && strings.Contains(u.RawPath, "+") {
-		u.RawPath = strings.Replace(u.RawPath, "+", " ", -1)
-		u.Path = url.PathEscape(u.RawPath)
+	// url.RawPath is not always present. Which is likely, if we're _just_ using +.
+	if u.RawPath != "" {
+		var err error
+		u.RawPath = strings.ReplaceAll(u.RawPath, "+", "%20")
+		u.Path, err = url.PathUnescape(u.RawPath)
+
+		PanicIfErr(err)
```
JohnRusk, line 40: In theory, this may not be _totally_ safe, because it makes an assumption about RawPath that the core Go library tries to avoid. Namely it assumes that, on entry to this function, RawPath is either null or is a valid encoding of Path.  THat's the way things are supposed to be, be the Go SDK carefully avoids making that assumption.  Insetead, the Go SDK help says this:

>In general, code should call EscapedPath instead of reading u.RawPath directly.

EscapedPath returns RawPath if it exists, and if it is valid.  Else it returns Path.

So, technically, I _think_ we should make sure RawPath is valid, by changing the if statement here to this:
```
if u.RawPath != "" && u.RawPath == u.EscapedPath() {   // only look at RawPath if it exists AND is valid
```

I'm not sure how necessarry it really is to do this.  I feel nervous when we make an assumption that the Go SDK deliberately doesn't.  What do you think?

common/s3URLParts.go
```diff
@@ -114,6 +119,11 @@ func NewS3URLParts(u url.URL) (S3URLParts, error) {
 
 		if bucketEndIndex := strings.Index(path, "/"); bucketEndIndex != -1 {
 			up.BucketName = path[:bucketEndIndex]
+			if rawPath != "" {
+				up.RawBucketName = rawPath[:strings.Index(rawPath, "/")]
+			} else {
+				up.RawBucketName = up.BucketName
+			}
```
JohnRusk, line 126: I haven't understood yet why getting RawBucketName is necessary, and why getting it from rawPath like this is safe. (E.g. what if RawPath contains an escaped / ?  I haven't managed to think that through)

Can you explain a bit why its necessary and safe?

common/extensions.go
adreed-msft: I would say that if the rawpath were invalid and we were this far, that's something we should panic over.

Why? Because we've already lost data about what the user wanted at that point. We can't verify any individual character of RawPath means anything with the context of Path.

adreed-msft: Furthermore, it'd act as a safeguard for any more URL manipulation we end up doing prior to this point in the future

JohnRusk: OK. So this would mean inside the existing if, doing this:

```
if u.RawPath != u.EscapedPath() {
   panic(…)
}
```

It feels mildly unconventional, but I'm comfortable with you doing it.

common/s3URLParts.go
adreed-msft: If it were escaped in RawPath, it would be unescaped in path

https://play.golang.org/p/h5UGb923Qh4

Arguably using path is slightly less safe in a way?

JohnRusk: >if it were escaped in RawPath, it would be unescaped in path

How does that help, given that if RawPath exists, it becomes our only source of RawBucketName?

JohnRusk: I guess I don't really understand what we're trying to do here, in terms of why we need to obtain raw bucket name. It's just for the test in zc_enumerator right?

JohnRusk: THe test there would feel safer to me if, instead of being this:

```
if s3URLParts.RawBucketName == "" || strings.Contains(s3URLParts.RawBucketName, "*") {
```

it was this instead

```
if s3URLParts.BucketName == "" || strings.Contains(s3URLParts.BucketName, "*")  || strings.Contains(s3URLParts.RawBucketName, "*") {
```
I.e. use BucketName for the "is empty" part of the test, then check both Raw and non-raw for the "is star" part.  That feels safe to me with this current implementation of s3UrlParts.  What do you think?

adreed-msft: Raw bucket name wasn't actually necessary after the changes made above. This was more just mild internal panic.

PR #950, opened 2020-04-07
Fix permission protection at download root, in case where downloading whole share

JohnRusk:
Our robocopy-like behavior, of protecting permissions at the download root, was failing when the downlaod root was the share itself - because share itself did not return any SDDL, so there was nothing for us to protect.


JohnRusk:
Test failure appears to be same transient failure that was fixed on dev branch.  So merging.

PR #953, opened 2020-04-09
Update changelog to include missing entries and rewording

adreed-msft:


JohnRusk:
I made a few small corrections.

PR #958, opened 2020-04-13
Feature/parallel enumeration for local file systems

JohnRusk:
Experimental code to improve perf for high-file count SAN/NAS customers

(May also improve enumeration in backup mode, because it uses a backup-mode aware file open function)

### DIFF
--- a/cmd/zc_traverser_file.go
+++ b/cmd/zc_traverser_file.go
@@ -183,7 +183,7 @@ func (t *fileTraverser) traverse(preprocessor objectMorpher, processor objectPro
 
 	// run the actual enumeration.
 	// First part is a parallel directory crawl
-	// Second part is parallel conversion of the directories and files to stored objects. This is necessary because the conversion to stored object may hit the network and therefore be slow in not parallelized
+	// Second part is parallel conversion of the directories and files to stored objects. This is necessary because the conversion to stored object may hit the network and therefore be slow if not parallelized
 	parallelism := enumerationParallelism // for Azure Files we'll run two pools of this size, one for crawl and one for transform
 
 	workerContext, cancelWorkers := context.WithCancel(t.ctx)
--- a/cmd/zc_traverser_file.go
+++ b/cmd/zc_traverser_file.go
@@ -184,10 +184,8 @@ func (t *fileTraverser) traverse(preprocessor objectMorpher, processor objectPro
 	// run the actual enumeration.
 	// First part is a parallel directory crawl
 	// Second part is parallel conversion of the directories and files to stored objects. This is necessary because the conversion to stored object may hit the network and therefore be slow in not parallelized
-	parallelism := 1
-	if enumerationParallelism > 1 {
-		parallelism = enumerationParallelism / 2 // half for crawl, half for transform
-	}
+	parallelism := enumerationParallelism // for Azure Files we'll run two pools of this size, one for crawl and one for transform
+
 	workerContext, cancelWorkers := context.WithCancel(t.ctx)
 
 	cCrawled := parallel.Crawl(workerContext, directoryURL, enumerateOneDir, parallelism)
--- a/common/osOpen_windows.go
+++ b/common/osOpen_windows.go
@@ -6,6 +6,8 @@ import (
 	"os"
 )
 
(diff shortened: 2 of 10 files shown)


common/osOpen_windows.go
```diff
@@ -23,6 +25,11 @@ func OSOpenFile(name string, flag int, perm os.FileMode) (*os.File, error) {
 	return file, nil
 }
 
+// TODO: os.Stat(name string) already uses backup semantics.  Can/should we just use that?
+//  That would work around the issue where here we are using OSOpen, which technically returns a bad result for
+//  directories (but which we seem to get away with in the os.Stat(handle) call - which goes down the "I am a file" code
+//  path, but still returns a valid result for a directory.
+//  See also todo on OsOpen, and edit or remove it
 func OSStat(name string) (os.FileInfo, error) {
```
adreed-msft, line 33: It would make sense to remove this, if os.Stat() works just fine here.

common/osOpen_windows.go
```diff
@@ -6,6 +6,8 @@ import (
 	"os"
 )
 
+// TODO: this is not reliable to use for directories, except inside OSStat (since the stat gets the right result)
+//   (But if the raw result of this routine is used, checking its internal dir flag, that flag is wrong for directories)
 func OSOpenFile(name string, flag int, perm os.FileMode) (*os.File, error) {
```
adreed-msft, line 11: What does it look like, when this incorrectly happens? Is there a known solution?

cmd/zc_traverser_file.go
```diff
@@ -184,10 +184,8 @@ func (t *fileTraverser) traverse(preprocessor objectMorpher, processor objectPro
 	// run the actual enumeration.
 	// First part is a parallel directory crawl
 	// Second part is parallel conversion of the directories and files to stored objects. This is necessary because the conversion to stored object may hit the network and therefore be slow in not parallelized
```
zezha-msft, line 186: >in [](start = 187, length = 2)

minor typo: if not parallelized

common/parallel/zt_FileSystemCrawlerTest_test.go
```diff
@@ -0,0 +1,107 @@
...
+func Test(t *testing.T) { chk.TestingT(t) }
+
+type fileSystemCrawlerSuite struct{}
+
+var _ = chk.Suite(&fileSystemCrawlerSuite{})
+var ctx = context.Background()
+
+func (s *fileSystemCrawlerSuite) TestParallelEnumeration(c *chk.C) {
+	var err error
+	dir := "/lib"
```
zezha-msft, line 44: > "/lib" [](start = 7, length = 7)

smart test!

common/parallel/FileSystemCrawler.go
```diff
@@ -0,0 +1,135 @@
...
+	)
+}
+
+// Walk is similar to filepath.Walk.
+// But note the following difference is how WalkFunc is used:
+// 1. If fileError passed to walkFunc is not nil, then here the filePath passed to that function will usually be ""
+//    (whereas with filepath.Walk it will usually (always?) have a value)
+// 2. If the return value of walkFunc function is not nil, enumeration will always stop, not matter what the type of the error.
+//    (Unlike filepath.WalkFunc, where returning filePath.SkipDir is handled as a special case)
+func Walk(root string, parallelism int, walkFn filepath.WalkFunc) {
```
zezha-msft, line 55: Very nicely written

common/osOpen_windows.go
JohnRusk: Solution is to use os.Stat(string) instead.  Basically the statue of the os.File is messed up, in that it actually points to directory, but it thinks it points to a File.  (On Windows, Golang uses very different internal representations for files vs directories, inside the implementation of os.File)

Some things just don't work. E.g. the readdir call in the parallel enumerator.  But even for things that do work, I'm not sure I'd want to trust them long term. (Right now, we seem to get away with it, because of how we happen to be using the os.Files, but I don't think we can rely on usage patterns staying the same long term).

common/osOpen_windows.go
JohnRusk: Covered in a separate PR

PR #967, opened 2020-04-17
Prevent transfers overwriting self, that could cause data loss

adreed-msft:


ste/xfer-anyToRemote-file.go
```diff
@@ -41,6 +41,25 @@ var s2sAccessTierFailureLogStdout sync.Once
...
+	// Ensure that the transfer isn't the same item, and fail it if it is.
+	// This scenario can only happen with S2S. We'll parse the URLs and compare the host and path.
+	if fromTo.IsS2S() {
+		srcURL, err := url.Parse(info.Source)
+		common.PanicIfErr(err)
+		dstURL, err := url.Parse(info.Destination)
+		common.PanicIfErr(err)
+
+		if srcURL.Hostname() == dstURL.Hostname() &&
+			srcURL.EscapedPath() == dstURL.EscapedPath() {
```
JohnRusk, line 55: Am curious why its EscapedPath that is compared, rather than Path?  Aren't there multiple possible escapings of one path? I.e. isn't it possible for two URLs to differ in EscapedPath, but not in Path?  Or am I thinking of this around the wrong way?

adreed-msft: __technically__ yes. But I suspect it's OK here due to the manipulation of the path that we do in enumeration.

It would be a worthwhile test to try overwriting a file with itself using different encodings with the same result, I suppose.

PR #969, opened 2020-04-20
Added support for load command based on  CLFSLoad extension

zezha-msft:


cmd/loadOutputTranslation.go
```diff
@@ -0,0 +1,236 @@
...
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if strings.Contains(line, clfsExitTag) {
+			c.processEndOfJob(line)
+		} else if strings.Contains(line, clfsInitializationTag) {
+			c.processInitialization(line)
+		} else if strings.Contains(line, clfsInfoTag) {
+			c.processInfo(line)
```
JohnRusk, line 63: Nice

cmd/loadOutputTranslation.go
```diff
@@ -0,0 +1,236 @@
...
+	err := json.Unmarshal([]byte(cleanMessage), &info)
+	if err != nil {
+		// don't crash if these info are not critical and are only nice-to-have
+		return
+	}
+
+	c.lcm.Info(fmt.Sprintf("%s phase has started.", info.Phase))
+}
+
+func (c *clfsExtensionOutputParser) processError(line string) {
```
JohnRusk, line 113: Do errors also get logged somewhere?  E.g. by the Python code, to a log file?

cmd/loadOutputTranslation.go
```diff
@@ -0,0 +1,236 @@
...
+			return summary.parseIntoString()
+		} else if format == common.EOutputFormat.Json() {
+			// re-marshal it the Go way
+			jsonOutput, err := json.Marshal(summary)
+			common.PanicIfErr(err)
+			return string(jsonOutput)
+		}
+
+		return ""
+	}, common.EExitCode.Success())
```
JohnRusk, line 167: Is this always returning success? If so, isn't that inconsistent with what AzCopy normally does?  Don't we return failure if even one file failed?  Should we, and could we, follow the same pattern here?

cmd/loadOutputTranslation.go
```diff
@@ -0,0 +1,236 @@
...
+	FilesTotal          int     `json:"files_total"`
+	PercentComplete     float64 `json:"pct_complete"`
+	Phase               string  `json:"phase"`
+	ThroughputMbps      float64 `json:"throughput_Mbps"`
+	ThroughputDeltaSecs float64 `json:"throughput_delta_secs"`
+	TimeEta             float64 `json:"time_eta"`
+	TimeReport          float64 `json:"time_report"`
+}
+
+func (progress clfsProgress) parseIntoString() string {
```
JohnRusk, line 196: Nice. I like these parseIntoString methods

cmd/loadCLFS.go
```diff
@@ -0,0 +1,256 @@
...
+		glcm.Exit(func(format common.OutputFormat) string {
+			return ""
+		}, common.EExitCode.Success())
+	},
+}
+
+func init() {
+	loadCmd.AddCommand(loadClfsCmd)
+	loadClfsCmd.PersistentFlags().BoolVar(&loadCmdRawInput.newSession, "new-session", true, "start a new job rather than continuing an existing one whose tracking information is kept at --state-path.")
+	loadClfsCmd.PersistentFlags().StringVar(&loadCmdRawInput.statePath, "state-path", "", "required path to a local directory for job state tracking. The path must point to an existing directory in order to resume a job. It must be empty for a new job.")
```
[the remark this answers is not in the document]
zezha-msft: It's the python code, otherwise an error is thrown.

cmd/loadOutputTranslation.go
zezha-msft: Yes! it's in the state path.

cmd/loadOutputTranslation.go
zezha-msft: Sounds good I've fixed it.

zezha-msft:
All the comments have been addressed.

Issue #860, opened 2020-02-03
azcopy sync with mixed SAS and AD authentication - Authentication information is not given in the correct format


t-l-k:
##### Note: The version is visible when running AzCopy without any argument
10.3.4

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows 

##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.

I had a SAS URL specified as the source system, and was intending to rely on AD authentication and RBAC for the target system. `azcopy login` was run first.

```
azcopy sync "<storage-url-with-sas>" "<storage-url-without-sas-but-with-azcopy-login>" --recursive --s2s-preserve-access-tier=false
```

### What problem was encountered?

I discovered today that you can't mix SAS and AD authentication methods between source and target systems in an `azcopy sync`. 

### How can we reproduce the problem in the simplest way?

The error was non-descriptive considering the above command, plus the `RESPONSE ERROR` was related to the source system, so I _thought I had messed up the SAS format_:

```
Description=Authentication information is not given in the correct format. Check the value of Authorization header.
```

All other request parameters looked well formed. I used Azure CLI to generate SAS tokens in Powershell:

``` powershell
az storage container generate-sas -n '<container-name>' --account-name <account-name> --https-only --permissions dlrw --expiry  (Get-Date).AddHours(1/4).ToString("yyyy-MM-ddTHH:mm:ssZ") -o tsv
```

### Have you found a mitigation/solution?

Use only one authentication method for both source and target system. Switching to 2x SAS tokens worked, and using 2x AD Authentication with RBAC worked.

JohnRusk:
>using 2x AD Authentication with RBAC worked.

That's not supposed to work yet ;-)  Are you sure it worked for you?

t-l-k:
@JohnRusk 

> That's not supposed to work yet ;-) Are you sure it worked for you?

Erm... yeah, in 1 subscription, copying from one storage account to another, using just container urls with no SAS, and having done an `azcopy login` authenticating as me using a device code, and after having set Storage Blob Data Owner to my user account, and yeah it just "works".

I can double check today and confirm ...

JohnRusk:
Ah, I think it's because your source and destination were in the same account.  I was assuming they were in different accounts.

JohnRusk:
[...]

PR #970, opened 2020-04-20
Update sync authorization code to prevent inevitably failing transfers & fix sas->oauth transfers in sync

adreed-msft:
Targets #860

### DIFF
--- a/cmd/syncEnumerator.go
+++ b/cmd/syncEnumerator.go
@@ -41,7 +41,7 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 	}
 
 	if cca.fromTo.IsS2S() {
-		if cca.fromTo.From() == common.ELocation.Blob() || cca.fromTo.From() == common.ELocation.File() {
+		if cca.fromTo.From() != common.ELocation.S3() {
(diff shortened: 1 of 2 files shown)


cmd/syncEnumerator.go
```diff
@@ -34,10 +34,30 @@ import (
...
 func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *syncEnumerator, err error) {
 
+	srcCredInfo, srcIsPublic, err := getCredentialInfoForLocation(ctx, cca.fromTo.From(), cca.source.Value, cca.source.SAS, true)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if cca.fromTo.IsS2S() {
+		if cca.fromTo.From() == common.ELocation.Blob() || cca.fromTo.From() == common.ELocation.File() {
```
JohnRusk, line 44: What's the purpose of this test?  Is it to exclude the S3 scenario?  If so, why not write is as != s3?  That way, if/when we add ADLS Gen 2 sync one day, this code will still be correct for ADLS Gen 2 (instead of just getting skipped when we add ADLS Gen 2).

adreed-msft: That's probably a good point. I'll adjust it.

PR #973, opened 2020-04-22
Ignore AuthFailed errror while listing properties of parent directory

nakulkar-msft:
Transfer to a filestore with a read-only SAS fails because we try
listing properties for parent directory, and fail with permission mismatch.
With this change, we ignore the error and try creating parent directories.
Also, we try matching the source and destination file lengths in
epilogue. The error here will be selectively ignored as well.

### DIFF
--- a/ste/xfer-anyToRemote-file.go
+++ b/ste/xfer-anyToRemote-file.go
@@ -25,12 +25,14 @@ import (
 	"errors"
 	"fmt"
 	"hash"
+	"net/http"
 	"net/url"
 	"runtime"
 	"strings"
 	"sync"
 
 	"github.com/Azure/azure-pipeline-go/pipeline"
+	"github.com/Azure/azure-storage-file-go/azfile"
 	"github.com/Azure/azure-storage-azcopy/common"
 )
 
@@ -363,12 +365,15 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s sender, sip ISo
 		_, isS2SCopier := s.(s2sCopier)
 		destLength, err := s.GetDestinationLength()
 
-		if err != nil {
+		if stgErr, stgErrOk := err.(azfile.StorageError); stgErrOk && stgErr.Response() != nil {
 			wrapped := fmt.Errorf("could not read destination length. If destination is write-only, use --check-length=false on the AzCopy command line. %w", err)
-			jptm.FailActiveSend(common.IffString(isS2SCopier, "S2S ", "Upload ")+"Length check: Get destination length", wrapped)
+			if stgErr.Response().StatusCode != http.StatusForbidden {
+				// We ignore this error, because the destination may be write-only
+				jptm.FailActiveSend(common.IffString(isS2SCopier, "S2S ", "Upload ")+"Length check: Get destination length", wrapped)
+			}
 		}
 
-		if destLength != jptm.Info().SourceSize {
+		if err == nil && destLength != jptm.Info().SourceSize {
 			jptm.FailActiveSend(common.IffString(isS2SCopier, "S2S ", "Upload ")+"Length check", errors.New("destination length does not match source length"))
 		}
 	}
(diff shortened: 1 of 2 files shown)


JohnRusk:
BTW, @nakulkar-msft , for future PRs, please don't use numerical branch names (like b6974627).  Please name the branch with a meaningful name, and start the name with `fix/` if its a bug fix and `feature/` if its a new feature.

ste/xfer-anyToRemote-file.go
```diff
@@ -363,12 +365,15 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s sender, sip ISo
 		_, isS2SCopier := s.(s2sCopier)
 		destLength, err := s.GetDestinationLength()
 
-		if err != nil {
+		if stgErr, stgErrOk := err.(azfile.StorageError); stgErrOk && stgErr.Response() != nil {
```
JohnRusk, line 368: There are cases where err will have a value, but that value will not be a StorageError. E.g. it might be a network error.

As far as i can tell this current code deosn't really handle those errors.  It just skips the length check if they happen, but it does say anything.

I'd feel more comfortable with this if we hav some kind of Boolean, say shouldCheckLength := true.  And then immediately after s.GetDestinationLength we set it to false if the error is a StorageError with http.StatusForbidden. If shouldCheckLength is true, then we'd do the existing logic.
I.e. 
```
shouldCheckLength := true
destlength, err := s.GetDestinatinoLength
if stgErr, stgErrOk := err.(pipeline.Response); stgErrOk && stgErr.Response() != nil && stgErr.Response().StatusCode != http.StatusForbidden {
  shouldCheckLength = false
}
if shouldCheckLength { 
     if err != nil { 
        existing error code block, from before this PR..
     } else {
        do the length check 
    }
}
```

JohnRusk, line 368: It's probably cleaner, and better idomatic Go, to use pipeline.Response instead of azfile.StorageError.  Why? Because this code doesn't just run for Azure Files, it will also run for Blobs and ADLS Gen 2 (which is known as blobFS in the AzCopy codebase).

You'd probably get away with the existing code, because azfile.StorageError has the same structure as azblob.StorageError and azblobfs.StorageError, but it's still a bit messy. It's cleaner to just use a "narrower" interface, with only the method you need. As it happens, all those error types implement pipeline.Response, which contains exactly the one method that you need.  So using pipeline.Response would be cleaner and (I think) better idiomatic Go.

ste/xfer-anyToRemote-file.go
```diff
@@ -361,15 +362,22 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s sender, sip ISo
...
 
-		if err != nil {
-			wrapped := fmt.Errorf("could not read destination length. If destination is write-only, use --check-length=false on the AzCopy command line. %w", err)
-			jptm.FailActiveSend(common.IffString(isS2SCopier, "S2S ", "Upload ")+"Length check: Get destination length", wrapped)
+		if resp, respOk := err.(pipeline.Response); respOk && resp.Response() != nil &&
+			resp.Response().StatusCode != http.StatusForbidden {
+			shouldCheckLength = false
 		}
+		if shouldCheckLength {
+			if err != nil {
```
zezha-msft, line 373: To make things clearer, I would suggest to move the error handling code together, and

1. If status is permission forbidden, we should write a message to the screen (using glcm.Info) just once, telling the user that we didn't have permission to check length, and the read permission is required.
2. If the error is anything else than permission problem, still write a message to the screen, but just say the length couldn't be obtained because of xxxx reason (no need to explain how to fix)

nakulkar-msft: Made these changes. Also, the permission-error will be displayed exactly once.

adreed-msft:
> 				_, err := curDirURL.Create(ctx, azfile.Metadata{}, azfile.SMBProperties{})

This is an old TODO. Feel free to remove it.

---
Refers to: ste/sender-azureFile.go:443 in 126ceaf. [](commit_id = 126ceafce2e2c6ed4506880936c2160ad19e174f, deletion_comment = False)

PR #978, opened 2020-04-24
Only write ACLs for one file at a time when downloading

JohnRusk:


ste/downloader-azureFiles_windows.go
```diff
@@ -162,6 +165,12 @@ func (a *azureFilesDownloader) PutSDDL(sip ISMBPropertyBearingSourceInfoProvider
 	}
 
 	// Then let's set the security info.
+	// TODO: review and potentially remove the use of the global mutex here, once we finish drilling into issues
+	//   observed when setting ACLs concurrently on a test UNC share.
+	//   BTW, testing indicates no measurable perf difference, between using the mutex and not, in the cases tested.
+	//   So it's safe to leave it here for now.
+	globalSetAclMu.Lock()
+	defer globalSetAclMu.Unlock()
 	err = windows.SetNamedSecurityInfo(txInfo.Destination,
```
adreed-msft, line 174: I also put this on SMB properties, but I'm not certain how important that is. I don't recall noticing any alteration there.

PR #988, opened 2020-04-28
Reverse the digits in the generated filenames for benchmarking

JohnRusk:


cmd/zc_traverser_benchmark.go
```diff
@@ -47,14 +47,26 @@ func (t *benchmarkTraverser) isDirectory(bool) bool {
...
 func (t *benchmarkTraverser) traverse(preprocessor objectMorpher, processor objectProcessor, filters []objectFilter) (err error) {
 	if len(filters) > 0 {
 		panic("filters not expected or supported in benchmark traverser") // but we still call processIfPassedFilters below, for consistency with other traversers
 	}
 
 	for i := uint(1); i <= t.fileCount; i++ {
 
-		name := fmt.Sprintf("%d", i)
+		name := t.toReversedString(i) // this gives an even distribution through the namespace (compare the starting characters, for 0 to 199, when reversed or not). This is useful for performance when High Throughput Block Blob pathway does not apply
 		relativePath := name
```
adreed-msft, line 70: How exactly does this effect performance? I'm a little confused reading your comment.

Furthermore, is it closer to real world performance of azcopy, or further?

JohnRusk: Here's why it matters:
https://docs.microsoft.com/en-us/azure/storage/blobs/storage-performance-checklist#partitioning

The change means we are doing best-case perf.  Before the change we were doing worst case.  Real-world perf will be somewhere in between.  But not point benchmarking with worst-case.

adreed-msft: Ah. Makes sense then.

PR #989, opened 2020-04-28
Fix folder overwrite prompt

zezha-msft:


common/writeThoughFile.go
```diff
@@ -51,18 +51,22 @@ func CreateDirectoryIfNotExist(directory string, tracker FolderCreationTracker)
...
 	if _, err := OSStat(directory); err != nil {
 		// if the error is present, try to create the directory
 		// stat errors can be present in write-only scenarios, when the directory isn't present, etc.
 		// as a result, we care more about the mkdir error than the stat error, because that's the tell.
 		// It'd seem that mkdirall would be necessary to port like osstat and osopenfile for new folders in a already no-access dest,
-		//     But in testing, this isn't the case.
-		err := os.MkdirAll(directory, os.ModePerm)
-		// if MkdirAll succeeds, no error is dropped-- it is nil.
-		// therefore, returning here is perfectly acceptable as it either succeeds (or it doesn't)
+		// But in testing, this isn't the case.
```
JohnRusk, line 60: (very) minor: comment on previous line refers to mkdirall.  Could ideally be removed or amended, since we don't call mkdirall now.

common/writeThoughFile.go
```diff
@@ -71,7 +75,16 @@ func CreateDirectoryIfNotExist(directory string, tracker FolderCreationTracker)
 			// consistent.
 			tracker.RecordCreation(directory)
 		}
-		return err
+
+		// another routine might have created the directory at the same time
+		// check whether the directory now exists
+		if _, err := OSStat(directory); err != nil {
```
JohnRusk, line 81: Do you wanna do this OSStat only if mkDirErr != nil?  Three reasons for that:
* when mkDirErr == nil, the test doesn't accomplish anything useful  (since we would have returned nil without the test anyway)
* hypothetically (I haven't tested it myself, but I think someone else has) Stat may fail but mkdir may succeed. (Due to being in a write-only situation.  So replacing nil mkdir error with a non-nill Stat err would be misleading in that case)
* It will save a disk IO operation

zezha-msft: Good call, I forgot a return statement

PR #1000, opened 2020-05-06
Fix/faster linux parallel enumeration

JohnRusk:


JohnRusk:
@zezha-msft I still need to add some automated tests for this.  But if you could review what's there so far, that would be awesome thanks.

common/parallel/dirReader_default.go
```diff
@@ -0,0 +1,45 @@
+// +build !linux
+
+// TODO: this currently applies on Darwin. Should it? Or should we use the Linux implementation there?
```
zezha-msft, line 3: >TODO: this currently applies on Darwin. [](start = 3, length = 39)

I'm not sure, would you like me to check?

common/parallel/dirReader_linux.go
```diff
@@ -0,0 +1,139 @@
...
+	if len(res) == 0 {
+		return nil, fmt.Errorf("readdir all LStat calls failed with last error %w", lastErr) // TODO: doing this because it's not valid for this method to return empty result and nil error, IIRC, when n > 0. Check that?
+	}
+
+	return res, nil
+}
+
+// worker reads linuxDirEntries from channel, LStat's them, and sends the result back on the
+// entry-specific channel
+func (r linuxDirReader) worker() {
```
zezha-msft, line 121: >(r linuxDirReader) worker( [](start = 5, length = 26)

nice, quite elegant implementation

common/parallel/dirReader_default.go
JohnRusk: Yes, if you could check, that would be great thanks.

zezha-msft:
@JohnRusk the logging changes look good to me 😉

JohnRusk:
@zezha-msft  and @adreed-msft , can you please re-review the changes that have happened here since your last review, noting the following points:

Two commits have been both added and reversed (by later commits), so you can ignore them. They relate to "Count Sleeping GRs" and "Logging on slow Directories".

The key changes that are new, are as follows:

Don't do early LMT Refresh on small files: ac76fc0

Log pre-scheduling state transitions for uploads: b3e2be0

Fix logic error in timeouts which could result in all workers blocking for ever: 802e3b2

Make the use of parallel file.Stat calls configurable: 7ffc553

adreed-msft:
The new changes look good to me. I'm a little concerned about not refreshing on small files, at least for the general consumer, but for a private drop I'm less concerned.

JohnRusk:
Thanks @Adele.  Can you elaborate on your concern re not re-querying the LMT before transferring the small files?

adreed-msft:
If a customer has lots of small files, I can fully imagine a transfer waiting a very long time to hit this bit of code. For instance, minutes. In that time, it might potentially become modified.

JohnRusk:
Yes, but even with the refresh, we did not USE the refreshed date to proceed with the transfer. We just FAILED the transfer more promptly.   I.e. we ALWAYS fail the transfer if the file has changed since the time of enumeration.  This is a long-standing design decision. For uploads, its driven mainly by the desire to be consistent with S2S and downloads, for which refresh of LMT wouldn't be performant.  Also, refreshing LMT on its own wouldn't be enough. We'd also need to refresh file size (and for S2S and downlaods, MD5 as well).   It's likely that we'll do those refreshes one day, but only if were are running in a job that has been _resumed_.   For the initial run of a job (i.e. not a resume), the current behaviour is probably what we'll go with for ever.

So what's the point of looking at the LMT before uploading, if we're not going to actually USE the new value?  It's a useful check for a large file.  We check the LMT before uploading, and fail the transfer immediately if's changed.  The alternative, if we didn't check the LMT before uploading, would be that the whole file would get uploaded, then we'd fail it after the upload, because we always do an LMT check after the upload.  But that wastes a lot of time, to upload the file first, then fail it. Which is why the current check exists, to allow an early failure.

But for a small file, the cost of uploading the file is much lower, so the benefit of failing it _before_ uploading is lower.  That's why I removed the pre-upload check for small files.

Does that sound reasonable?

JohnRusk:
Based on comments above, and via email, I'm going to merge this now.

PR #1001, opened 2020-05-07
Exclude windows system files from being copied

nakulkar-msft:
This will stop copying files from Recycle bin, System Info and other such from being copied. Also fixes the Permission error obtained while copying from drive root.


cmd/copyEnumeratorInit.go
```diff
@@ -313,6 +313,11 @@ func (cca *cookedCopyCmdArgs) initModularFilters() []objectFilter {
 		filters = append(filters, buildAttrFilters(cca.includeFileAttributes, cca.source.ValueLocal(), true)...)
 	}
 
+	//Always exclude systems files on Windows
+	if runtime.GOOS == "windows" && cca.fromTo.From().IsLocal() {
+		cca.excludeFileAttributes = append(cca.excludeFileAttributes, "S")
+	}
+
```
adreed-msft, line 320: What if a user specifically adds an include for the system attribute? Do we accidentally end up ignoring it?

adreed-msft: On a second note, this doesn't apply for Sync at the moment (and it probably should). 😉

---
In reply to: [421772865](https://github.com/Azure/azure-storage-azcopy/pull/1001#discussion_r421772865) [](ancestors = 421772865)

nakulkar-msft: Yes, we'll end up ignoring the specific request for system files. On a related note, do we have a valid case where we want to copy system files. We may need do that with a flag --include-sys-fiels. 

I'll take care of Sync :(

JohnRusk: @nakulkar-msft Have you added the sync fix to this PR yet?  Just wondering whether I should review it yet.

JohnRusk: I see that's added.

BTW, @nakulkar-msft, what are your thoughts re Adele's first comment above?  How hard would it be do to "If user has explicitly INcluded S, then don't automatically EXclude it?"

nakulkar-msft: I would require a lot of special handling in the filter code. Primary concern is that it would change the meaning of the filters, compared to current usage. Today, if the user has included S, we 'transfer system files alone'. If we go with this approach, the meaning would be  'transfer system files as well.' 
That said, would it be better if we introduce a command line bool parameter.

JohnRusk: Can you elaborate on this:
>If we go with this approach, the meaning would be 'transfer system files as well.'

nakulkar-msft: How do we want the below command to work with the fix?
`azcopy copy <win local path> <dst filestore> --include-attributes "S"` 
User has explicitly included `S`. Do we transfer system files alone, as we do today, or do we transfer both system files and non-system files?

JohnRusk: I guess I'd just assumed that your new if statements would be reconfigured to look something like this:

```
	if runtime.GOOS == "windows" && cca.fromTo.From().IsLocal() && <user did not include S in the INclude attributes>{
		cca.excludeFileAttributes = append(cca.excludeFileAttributes, "S")
	}
```

So, if they INclude it, then we don't Exclude it.  Doesn't that mean we transfer system files alone, as we do today, if they use the command line in your example?

nakulkar-msft: Okay. I'll do that. I was assuming that the user intends to transfer 'system files along with regular files'. And I see that it was wrong.

JohnRusk: @nakulkar-msft Have you made that change? (Wondering if I should re-review)

JohnRusk:
@nakulkar-msft Is this PR still "live" (i.e. to be reviewed)?  I think you mentioned something about changing the way it works?

nakulkar-msft:
Breaks the way we handle folders.

PR #1007, opened 2020-05-14
Benchmark for downloads

nakulkar-msft:
The existing infra for workloads is used to test downloads as well.
The current usage is 'azcopy bench <target> --download'. The exact
usage will be decided.

### DIFF
--- a/cmd/benchmark.go
+++ b/cmd/benchmark.go
@@ -23,22 +23,22 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"net/url"
+	"os"
+	"strconv"
+	"strings"
+
 	"github.com/Azure/azure-storage-azcopy/azbfs"
 	"github.com/Azure/azure-storage-azcopy/common"
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/Azure/azure-storage-file-go/azfile"
 	"github.com/spf13/cobra"
-	"net/url"
-	"strconv"
-	"strings"
 )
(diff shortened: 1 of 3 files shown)


cmd/benchmark.go
```diff
@@ -317,6 +336,7 @@ func init() {
 	benchCmd.PersistentFlags().StringVar(&raw.blobType, "blob-type", "Detect", "defines the type of blob at the destination. Used to allow benchmarking different blob types. Identical to the same-named parameter in the copy command")
 	benchCmd.PersistentFlags().BoolVar(&raw.putMd5, "put-md5", false, "create an MD5 hash of each file, and save the hash as the Content-MD5 property of the destination blob/file. (By default the hash is NOT created.) Identical to the same-named parameter in the copy command")
 	benchCmd.PersistentFlags().BoolVar(&raw.checkLength, "check-length", true, "Check the length of a file on the destination after the transfer. If there is a mismatch between source and destination, the transfer is marked as failed.")
+	benchCmd.PersistentFlags().StringVar(&raw.mode, "mode", "upload", "Defines if Azcopy should test uploads or downloads from this target. Valid values are 'upload' and 'download'. Defaulted option is 'upload'.")
```
JohnRusk, line 339: Looks good. BTW, every time we modify the command parameters, we also need to look at the help text and examples in cmd/helpMessages.go and update them too. Can you do that please, for the section of helpMessages.go that relates to benchmark?

PR #1008, opened 2020-05-14
Allow rm to delete blob directory stubs with hdi_isfolder

zezha-msft:


cmd/copyEnumeratorInit.go
```diff
@@ -57,7 +57,7 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 	jobPartOrder.DestLengthValidation = cca.CheckLength
 	jobPartOrder.S2SInvalidMetadataHandleOption = cca.s2sInvalidMetadataHandleOption
 
-	traverser, err = initResourceTraverser(cca.source, cca.fromTo.From(), &ctx, &srcCredInfo, &cca.followSymlinks, cca.listOfFilesChannel, cca.recursive, getRemoteProperties, func(common.EntityType) {})
+	traverser, err = initResourceTraverser(cca.source, cca.fromTo.From(), &ctx, &srcCredInfo, &cca.followSymlinks, cca.listOfFilesChannel, cca.recursive, getRemoteProperties, false, func(common.EntityType) {})
```
JohnRusk, line 60: Minor: why not use cca.includeDirectoryStubs here, insteadof hard coding the "false"? (In case we move remove over to use initCopyEnumerator).

cmd/zc_traverser_blob_account.go
```diff
@@ -94,7 +94,7 @@ func (t *blobAccountTraverser) traverse(preprocessor objectMorpher, processor ob
 
 	for _, v := range cList {
 		containerURL := t.accountURL.NewContainerURL(v).URL()
```
JohnRusk, line 96: Minor: hard coding the false here assumes we'll never implement account-level remove. If we ever might do that, maybe pass the includeBlobStubs in?

cmd/zc_traverser_blob.go
```diff
@@ -65,24 +68,34 @@ func (t *blobTraverser) isDirectory(isSource bool) bool {
...
 
 func (t *blobTraverser) traverse(preprocessor objectMorpher, processor objectProcessor, filters []objectFilter) (err error) {
 	blobUrlParts := azblob.NewBlobURLParts(*t.rawURL)
 	util := copyHandlerUtil{}
 
+	// trim away the trailing slash before we check whether it's a single blob
+	// so that we can detect the directory stub in case there is one
+	hadTrailingSlash := strings.HasSuffix(blobUrlParts.BlobName, common.AZCOPY_PATH_SEPARATOR_STRING)
+	blobUrlParts.BlobName = strings.TrimSuffix(blobUrlParts.BlobName, common.AZCOPY_PATH_SEPARATOR_STRING)
+	*t.rawURL = blobUrlParts.URL()
```
JohnRusk, line 95: It feels odd to trim this _here_.  In particular, the other usage of t.rawURL is isDirectory.  I can't think through in my head whether isDirectory is called before or after this trimming.  I did note that that isDirectory in turn calls getPropertiesIfSingleBlob AND isDirectory relies on the trailing slashes.

Might be be easier to reason about if you moved the stripping to getPropertiesIfSingleBlob.... and did not ever mutate t.rawURL (i.e. leave the / on it).

I'm not absolutely sure there's a problem here - just that it's weird to see one part of the code trimming a suffix when another depends on it being there. 

*** Also *** can you take a look at that dependency on trailing slashes in urlIsContainerOrVirtualDirectory, and see how that releates to the above ****

cmd/copyEnumeratorInit.go
adreed-msft: Actually isn't this for blob? So logically it would use initCopyEnumerator already, since azbfs is handled in its own case.

---
In reply to: [433610287](https://github.com/Azure/azure-storage-azcopy/pull/1008#discussion_r433610287) [](ancestors = 433610287)

adreed-msft: (I stand corrected, forgot we had remove-specific enumerator init)

---
In reply to: [438900923](https://github.com/Azure/azure-storage-azcopy/pull/1008#discussion_r438900923) [](ancestors = 438900923,433610287)

cmd/removeEnumerator.go
```diff
@@ -39,13 +39,15 @@ var NothingToRemoveError = errors.New("nothing found to remove")
 // provide an enumerator that lists a given resource (Blob, File)
 // and schedule delete transfers to remove them
 // TODO: Make this merge into the other copy refactor code
+// TODO: initEnumerator is significantly more verbose at this point, evaluate the impact of switching over
 func newRemoveEnumerator(cca *cookedCopyCmdArgs) (enumerator *copyEnumerator, err error) {
```
adreed-msft, line 43: yeah, there is quite a bit of uh, "weightiness" to initEnumerator, especially considering the individual little hacks to retain the same functionality as the older revisions of azcopy

cmd/zc_traverser_blob.go
adreed-msft: It doesn't strictly seem like t.rawURL is re-used, unless t.traverse is called somehow a second time. Though, I agree that this modifies the state of the traverser, and should be localized to the get properties call.

---
In reply to: [433614556](https://github.com/Azure/azure-storage-azcopy/pull/1008#discussion_r433614556) [](ancestors = 433614556)

zezha-msft: Ok!

JohnRusk:
I was already reviewing, when you requested the review :-)

PR #1015, opened 2020-05-22
Add include-after parameter for Copy

JohnRusk:


cmd/zc_filter.go
```diff
@@ -225,3 +227,75 @@ func (fs filterSet) GetEnumerationPreFilter(recursive bool) string {
...
+	msg = ""
+	supported = true
+	return
+}
+
+func (f *includeAfterDateFilter) appliesOnlyToFiles() bool {
+	return true
+	// because we don't currently (May 2020) have meaningful LMTs for folders. The meaningful time for a folder is the "change time" not the "last write time", and the change time can only be obtained via NtGetFileInformation, which we don't yet call.
+	// TODO: the consequence of this is that folder properties and folder acls can't be moved when using this filter.
+	//       Can we live with that, for now?
```
zezha-msft, line 249: sure, maybe this could be communicated in the flag's description?

cmd/zc_filter.go
```diff
@@ -225,3 +227,75 @@ func (fs filterSet) GetEnumerationPreFilter(recursive bool) string {
...
+	}
+
+	return storedObject.lastModifiedTime.After(f.threshold) ||
+		storedObject.lastModifiedTime.Equal(f.threshold) // >= is easier for users to understand than >
+}
+
+// ParseISO8601 parses ISO 8601 dates. This routine is needed because GoLang's time.Parse* routines require all expected
+// elements to be present.  I.e. you can't specify just a date, and have the time default to 00:00. But ISO 8601 requires
+// that and, for usability, that's what we want.  (So that users can omit the whole time, or at least the seconds portion of it, if they wish)
+func (_ includeAfterDateFilter) ParseISO8601(s string) (time.Time, error) {
```
zezha-msft, line 265: Minor: ParseISO8601 can be done inside the constructor (which can be created for includeAfterDateFilter), so that the user of the struct does not have to worry about calling parsing.

cmd/zc_filter.go
JohnRusk: Sure. Will do.

cmd/zc_filter.go
JohnRusk: The reason I didn't do that, is that I wanted the parsing to happen a parameter "cook" time.  So that command line errors would be detected there - just like they are for all other param errors.  In fact, IIRC, we don't even return error values from filter constructors, so couldn't check it there.

For those reasons, I'll leave this as is.

PR #1016, opened 2020-05-24
Stop treating zero-item copy as failure

nakulkar-msft:
* Edit log entry to note we are transferring zero items

* Mark status of jobPartMgr to Completed() so that we wind down
  transfer and print the summary

### DIFF
--- a/ste/init.go
+++ b/ste/init.go
@@ -144,8 +144,12 @@ func ExecuteNewCopyJobPartOrder(order common.CopyJobPartOrderRequest) common.Cop
 	jpm := JobsAdmin.JobMgrEnsureExists(order.JobID, order.LogLevel, order.CommandString) // Get a this job part's job manager (create it if it doesn't exist)
 
 	if len(order.Transfers) == 0 && order.IsFinalPart {
-		jpm.Log(pipeline.LogError, "ERROR: No transfers were scheduled.")
-		return common.CopyJobPartOrderResponse{JobStarted: false, ErrorMsg: common.ECopyJobPartOrderErrorType.NoTransfersScheduledErr()}
+		/*
+		 * We set the status of this jobPart to Completed()
+		 * immediately after it is scheduled, and wind down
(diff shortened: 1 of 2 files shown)


ste/init.go
```diff
@@ -144,8 +144,12 @@ func ExecuteNewCopyJobPartOrder(order common.CopyJobPartOrderRequest) common.Cop
 	jpm := JobsAdmin.JobMgrEnsureExists(order.JobID, order.LogLevel, order.CommandString) // Get a this job part's job manager (create it if it doesn't exist)
 
 	if len(order.Transfers) == 0 && order.IsFinalPart {
-		jpm.Log(pipeline.LogError, "ERROR: No transfers were scheduled.")
-		return common.CopyJobPartOrderResponse{JobStarted: false, ErrorMsg: common.ECopyJobPartOrderErrorType.NoTransfersScheduledErr()}
+		/*
+		 * We set the status of this jobPart to Completed()
```
JohnRusk, line 148: Since this code is no longer returning NoTransfersScheduledError, can you please look for usages of that, fix them if necessary and, if there are no remaining usages, remove the definition of NoTransfersScheduledErr?

nakulkar-msft: The other usages of NoTransferScheduledErr is in remove and sync enumerators. We kind of use the error to formulate a better information on lifecycle mgr. We would need some indicator from ste to know that there are no files to act on, and error appears to be a better way. [Here](https://drive.google.com/file/d/1nZdCOtpl6dWpllpChoMa3mEdMIOAIOOf/view?usp=sharing) is an example for sync.

zezha-msft: @nakulkar-msft Sync is fine as is, but perhaps we should update it for remove, which the StgExp folks also use.

nakulkar-msft: Remove is updated. We no longer receive the error message but display that zero files were removed. That is the intended behaviour right?

PR #1017, opened 2020-05-25
Experimental/shuffle job parts

JohnRusk:


JohnRusk:
Closing, since this is still very experimental.  It's here for record keeping of what we've experimented with.

PR #1033, opened 2020-06-02
Fix daylight saving bug in include-after

JohnRusk:


cmd/copy.go
```diff
@@ -424,7 +424,9 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 	}
 
 	if raw.includeAfter != "" {
-		parsedIncludeAfter, err := includeAfterDateFilter{}.ParseISO8601(raw.includeAfter)
+		// must set chooseEarliest = true, so that if there's an ambiguous local date, the earliest will be returned
+		// (since that's safest for includeAfter.  Better to choose the earlier time and do more work, than the later one and fail to pick up a changed file
```
zezha-msft, line 428: This might mean that we could also be copying more than the user intended. Should we document this information somewhere?

JohnRusk: It could.  But it only affects cases where both the file change and the running of AzCopy happen during the 2-hour "ambiguous time" window that happens once a year in the middle of the night.  On that basis, I was planning to not document it.  Do you think we should?

zezha-msft: Hmm, I agree with you, it's too cluttering for our docs

zezha-msft:
@adreed-msft @nakulkar-msft could you please take a look and merge accordingly?

PR #1035, opened 2020-06-03
Feature/gcp connector

dheeraj-msft:
This PR will add feature to transfer Objects from Google Cloud Platform to Azure Block Blob Storage.


dheeraj-msft:
To test export GOOGLE_APPLICATION_CREDENTIALS=<path/to/serviceAccountKey.json>. 
Eg cmd to transfer file:
azcopy copy 'https://storage.cloud.google.com/<bucket-name>/<object-name>' '<destination-url>'

Issue #1032, opened 2020-06-01
Unexpected upload destination when uploading a drive


JasonYeMSFT:
### Which version of the AzCopy was used? 
10.4.2

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

### What command did you run?
azcopy cp "D:\\" "https://account.blob.core.windows.net/container"

### What problem was encountered?
The files/directories under D: drive is not put directly under the root of container. Instead, the upload destination will contains two // before the actual paths to the files/directories (e.g. D:\file.txt will end up at //file.txt).

### How can we reproduce the problem in the simplest way?
Run the command or go to storage explorer and drag your D: drive to a blob container.

### Have you found a mitigation/solution?
No.

PR #1042, opened 2020-06-09
Issue #1032: Removed "\\" from the start of filename when uploaded from root volume in windows

mohsha-msft:
https://github.com/Azure/azure-storage-azcopy/issues/1032


ste/JobPartPlan.go
```diff
@@ -144,11 +145,20 @@ func (jpph *JobPartPlanHeader) TransferSrcDstStrings(transferIndex uint32) (sour
 	sh.Cap = sh.Len
 	dstRelative := string(dstSlice)
 
+	// Incase user is trying to copy complete directory in windows i.e D:\\
```
JohnRusk, line 148: It feels a bit odd to do this processing here, inside the JobPartPlanHeader. I believe it would be more consistent, with the way paths are usually handled in AzCopy, to do this processing at the point where the root directory is first set up. Which is here: https://github.com/Azure/azure-storage-azcopy/blob/9a88118d55716be9408a2d33fd1a0cffd50f9221/cmd/copyEnumeratorInit.go#L535

What do you think about moving the processing to that point?

common/windowsPathUtil.go
```diff
@@ -0,0 +1,48 @@
...
+
+// IsPathSeparator reports whether c is a directory separator character.
+func IsPathSeparator(c uint8) bool {
+	// NOTE: Windows accept / as path separator.
+	return c == '\\' || c == '/'
+}
+
+// VolumeName function takes path string as a parameter and returns the name of the volume. 
+func VolumeName(path string) (v string) {
+	if len(path) < 2 {
```
JohnRusk, line 16: The VolumeName function already exists in Go as `filepath.VolumeName`.  Can you please investigate using that one, rather than writing our own?

JohnRusk:
Also, please reword the PR name, to reflect what it actually does.  (Not just a number). E.g. "Construct destination path correctly when uploading a whole drive."

common/windowsPathUtil.go
mohsha-msft: Thanks for the inputs. Yeah I used **```filepath.VolumeName```** method to get the required results. It's just that self defined methods provides more customization and tweaking so used it earlier.

ste/JobPartPlan.go
```diff
@@ -144,11 +145,20 @@ func (jpph *JobPartPlanHeader) TransferSrcDstStrings(transferIndex uint32) (sour
...
+	if runtime.GOOS == "windows" && startsWith(dstRelative, "/%5C") {
+		dstRelative = strings.Replace(common.VolumeName(srcRoot), common.EXTENDED_PATH_PREFIX, "", 1) + dstRelative[4:]
+	}
+
 	return common.GenerateFullPathWithQuery(srcRoot, srcRelative, srcExtraQuery),
 		common.GenerateFullPathWithQuery(dstRoot, dstRelative, dstExtraQuery),
 		isFolder
 }
 
+func startsWith(s string, t string) bool {
```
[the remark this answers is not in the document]
mohsha-msft: Done.

cmd/copyEnumeratorInit.go
```diff
@@ -233,6 +233,14 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 		srcRelPath := cca.makeEscapedRelativePath(true, isDestDir, object)
 		dstRelPath := cca.makeEscapedRelativePath(false, isDestDir, object)
 
+		if runtime.GOOS == "windows" && strings.HasPrefix(dstRelPath, "/%5C") {
```
zezha-msft, line 236: Sorry I'm a bit confused here. Could you please add a comment to explain what is coming in, and what are we doing to it? An example would be best.

mohsha-msft: Added comments for better understanding.

So basically when a user tries to copy whole volume in windows, "//" gets added to the name of all the files/directories in the destination. "D:\\file.txt" will get uploaded as "\\file.txt" and "D:" will not be created. So I've tried to fix it by getting the volume name from source and then stripping "/%5C" from dstRelPath to create valid path.

PR #1043, opened 2020-06-09
Overwrite content type for static website files

zezha-msft:
Also tested manually in addition to the current unit test.

### DIFF
--- a/ste/mgr-JobPartMgr.go
+++ b/ste/mgr-JobPartMgr.go
@@ -585,10 +585,10 @@ func (jpm *jobPartMgr) resourceDstData(fullFilePath string, dataFileToXfer []byt
 
 // TODO do we want these charset=utf-8?
 var builtinTypes = map[string]string{
-	".css":  "text/css; charset=utf-8",
+	".css":  "text/css;",
 	".gif":  "image/gif",
-	".htm":  "text/html; charset=utf-8",
-	".html": "text/html; charset=utf-8",
+	".htm":  "text/html;",
+	".html": "text/html;",
 	".jpeg": "image/jpeg",
 	".jpg":  "image/jpeg",
 	".js":   "application/javascript",
@@ -598,7 +598,7 @@ var builtinTypes = map[string]string{
 	".svg":  "image/svg+xml",
 	".wasm": "application/wasm",
 	".webp": "image/webp",
-	".xml":  "text/xml; charset=utf-8",
+	".xml":  "text/xml;",
 }
 
 func (jpm *jobPartMgr) inferContentType(fullFilePath string, dataFileToXfer []byte) string {
@@ -606,7 +606,7 @@ func (jpm *jobPartMgr) inferContentType(fullFilePath string, dataFileToXfer []by
 
 	// short-circuit for common static website files
 	// mime.TypeByExtension takes the registry into account, which is most often undesirable in practice
-	if override, ok := builtinTypes[fileExtension]; ok {
+	if override, ok := builtinTypes[strings.ToLower(fileExtension)]; ok {
 		return override
 	}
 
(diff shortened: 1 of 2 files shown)


ste/mgr-JobPartMgr.go
```diff
@@ -583,8 +583,34 @@ func (jpm *jobPartMgr) resourceDstData(fullFilePath string, dataFileToXfer []byt
 	return common.ResourceHTTPHeaders{ContentType: jpm.inferContentType(fullFilePath, dataFileToXfer), ContentLanguage: jpm.httpHeaders.ContentLanguage, ContentDisposition: jpm.httpHeaders.ContentDisposition, ContentEncoding: jpm.httpHeaders.ContentEncoding, CacheControl: jpm.httpHeaders.CacheControl}, jpm.metadata
 }
 
+// TODO do we want these charset=utf-8?
```
zezha-msft, line 586: I believe we should, since that's consistent with the current behavior. @JohnRusk I'm interested in hearing your thoughts on this since you dealt with an issue on this.

JohnRusk: I'm beginning to suspect that we should NOT specify any charset at all.  The details vary slightly from file type to file type, but the theme is generally that specifying a charset is really only useful if you know it's a non-standard one (i.e. not UTF-8).  If it is UTF-8, you don't need to specify it. (I _think_)

CSS, XML, HTML and Javascript all have their own ways of specifying and detecting charset within the file itself.  We don't want to have to write custom parsers for all those file types, to detect the encoding that is specified within the file (if any).  And what would be the value of us doing so?  It's already specified _within_ the file in those cases, so you don't need it in the content-type.

I've checked the IANA Mime Type registry, and charset is _optional_ for all four of the above Mime types.

In general, the situation where we would add value by specifying a charset  in the content type is when it's _not_ specified inside the file, and the correct encoding is something exotic (i.e. _not_ UTF 8).  But I suspect those cases are rare. More importantly, there's nothing we could realistically do to implement them correctly anyway. How would we know what the right _non_-UTF-8 encoding is?

Browsers will do their own sniffing on content bytes if there's no other indication of the type.  It's not realistic to think that a data transfer application (AzCopy) could do that sniffing substantially better than a browser could - so it's better that we say nothing and let the browser figure it out.

So I think we should :
(a) Never specify a charset in these hard-coded types and 
(b) Actually strip out any charset that is returned to us by mime.TypeByExtension. (This would fix the related issue, which I've been involved, with in all cases except for a possible edge case from our later call to mime.TypeByExtension)

What do you think? Shall we do (a) and (b)?

cc @nakulkar-msft

nakulkar-msft: Comments [here](https://github.com/golang/go/issues/27078) say utf8 on text files was a conscious decision.

ste/mgr-JobPartMgr.go
```diff
@@ -583,8 +583,34 @@ func (jpm *jobPartMgr) resourceDstData(fullFilePath string, dataFileToXfer []byt
...
+	".xml":  "text/xml; charset=utf-8",
+}
+
 func (jpm *jobPartMgr) inferContentType(fullFilePath string, dataFileToXfer []byte) string {
-	if guessedType := mime.TypeByExtension(filepath.Ext(fullFilePath)); guessedType != "" {
+	fileExtension := filepath.Ext(fullFilePath)
+
+	// short-circuit for common static website files
+	// mime.TypeByExtension takes the registry into account, which is most often undesirable in practice
+	if override, ok := builtinTypes[fileExtension]; ok {
```
JohnRusk, line 609: @zezha-msft I just realised, we should do strings.ToLower() on the extension before using it to look up in the map.  Otherwise we'll still be broken if the extension is .HTML or something.

ste/mgr-JobPartMgr.go
JohnRusk: Actually, my point (b) is irrelevant.  The only types for which mime.TypeByExtension will return a charset are ones that this PR overrides.  So my feedback simply becomes, let's remove the the charset from the list in this PR.

(As for it being a conscious decision: as far as I can tell, in our context, it wasn't the right decision)

adreed-msft, line 586: I agree we should probably drop the charset variables in there. It may be undesirable to a customer who opts to not use UTF-8 for one reason or another.

zezha-msft: Ok I'm convinced!

PR #1049, opened 2020-06-10
Port ATOMiCNebula's work from the blob SDK to azbfs WRT SAS token time precision

adreed-msft:


azbfs/zt_url_file_test.go
```diff
@@ -232,3 +234,58 @@ func (s *FileURLSuite) TestUploadDownloadRoundTrip(c *chk.C) {
 	c.Assert(download[:2048], chk.DeepEquals, contentD1[:])
 	c.Assert(download[2048:], chk.DeepEquals, contentD2[:])
 }
+
+func (s *FileURLSuite) TestBlobURLPartsSASQueryTimes(c *chk.C) {
```
zezha-msft, line 238: Minor: test name should be TestBfsURLPartsSASQueryTimes

PR #1054, opened 2020-06-16
Fix GetFileInformation()  error on Windows drive root

nakulkar-msft:
GetFileInformationByHandle() on Windows returns EINVAL if the file handle is created with an Extended path for a drive root. Change the path to a short path if we're referring to a Drive root.


JohnRusk:
BTW, I suspect isDriveRoot will return true for the root of a UNC share too.  E.g. if you pass "\\host\share" to VolumeName it will return the value you passed in.  But I think that's fine, because they'll always be short enough that the conversion to a short name will be safe.

PR #1058, opened 2020-06-17
Allow finer values for job status in jptm

nakulkar-msft:
- jpm-runtime will have a progressInfo, consisting of atomic integers transfersSkipped, transfersFailed, transfersCompleted.
- Each jptm on completion reports to jpm through jpm.ReportTransferDone().This is enhanced to include  TransferStatus(Success/Failed/Skipped)argument. Depending on TransferStatus, above flags/counters are atomically updated throuth jpm.updateJobPartProgress()
- jobMgr.ReportJobPartDone() will now run as a thread (named reportJobPartDoneHandler). This will listen for progressInfo from each part.
- This handler will update its local counters maintaining Skipped/Failed/Completed transfers.
- On receiving the progressInfo from final jobPart, the handler will use above counters to update jobStatus in jpph(0) appropriately to Completed/CompletedWithErrors/CompletedWithSkipped.
- The instances where we close a transfer in jptm without setting TransferStatus to Success/Failed/Skipped are fixed.
- Modify 'azcopy jobs list' to report improved status

### DIFF
--- a/cmd/jobsClean.go
+++ b/cmd/jobsClean.go
@@ -90,7 +90,7 @@ func handleCleanJobsCommand(givenStatus common.JobStatus) error {
 
 	// we must query the jobs and find out which one to remove
 	resp := common.ListJobsResponse{}
-	Rpc(common.ERpcCmd.ListJobs(), nil, &resp)
+	Rpc(common.ERpcCmd.ListJobs(), common.EJobStatus.All(), &resp)
 
 	if resp.ErrorMessage != "" {
 		return errors.New("failed to query the list of jobs")
--- a/cmd/jobsClean.go
+++ b/cmd/jobsClean.go
@@ -79,7 +79,7 @@ func init() {
 	// NOTE: we have way more job status than we normally need, only show the most common ones
 	jobsCleanCmd.PersistentFlags().StringVar(&commandLineInput.withStatus, "with-status", "All",
 		"only remove the jobs with this status, available values: All, Cancelled, Failed, Completed"+
-			" CompletedWithErrors, CompletedWithFailures, CompletedWithErrorsAndSkipped")
+			" CompletedWithErrors, CompletedWithSkipped, CompletedWithErrorsAndSkipped")
 }
 
 func handleCleanJobsCommand(givenStatus common.JobStatus) error {
--- a/cmd/jobsClean.go
+++ b/cmd/jobsClean.go
@@ -78,7 +78,8 @@ func init() {
 
 	// NOTE: we have way more job status than we normally need, only show the most common ones
 	jobsCleanCmd.PersistentFlags().StringVar(&commandLineInput.withStatus, "with-status", "All",
-		"only remove the jobs with this status, available values: Cancelled, Completed, Failed, InProgress, All")
+		"only remove the jobs with this status, available values: All, Cancelled, Failed, Completed"+
+			" CompletedWithErrors, CompletedWithFailures, CompletedWithErrorsAndSkipped")
 }
 
 func handleCleanJobsCommand(givenStatus common.JobStatus) error {
--- a/cmd/jobsList.go
+++ b/cmd/jobsList.go
@@ -36,6 +36,12 @@ type ListResponse struct {
 }
 
 func init() {
+	type JobsListReq struct {
+		withStatus string
+	}
+
+	commandLineInput := JobsListReq{}
+
 	// lsCmd represents the listJob command
 	lsCmd := &cobra.Command{
 		Use:     "list",
@@ -52,7 +58,13 @@ func init() {
 			return nil
 		},
 		Run: func(cmd *cobra.Command, args []string) {
-			err := HandleListJobsCommand()
+			withStatus := common.EJobStatus
+			err := withStatus.Parse(commandLineInput.withStatus)
+			if err != nil {
+				glcm.Error(fmt.Sprintf("Failed to parse --with-status due to error: %s.", err))
+			}
+
+			err = HandleListJobsCommand(withStatus)
 			if err == nil {
 				glcm.Exit(nil, common.EExitCode.Success())
 			} else {
@@ -62,18 +74,21 @@ func init() {
 	}
 
 	jobsCmd.AddCommand(lsCmd)
+
+	jobsCmd.PersistentFlags().StringVar(&commandLineInput.withStatus, "with-status", "All",
+		"only remove the jobs with this status, available values: Cancelled, Completed, Failed, InProgress, All")
 }
 
 // HandleListJobsCommand sends the ListJobs request to transfer engine
 // Print the Jobs in the history of Azcopy
-func HandleListJobsCommand() error {
+func HandleListJobsCommand(jobStatus common.JobStatus) error {
 	resp := common.ListJobsResponse{}
 	Rpc(common.ERpcCmd.ListJobs(), nil, &resp)
-	return PrintExistingJobIds(resp)
+	return PrintExistingJobIds(resp, jobStatus)
 }
 
 // PrintExistingJobIds prints the response of listOrder command when listOrder command requested the list of existing jobs
-func PrintExistingJobIds(listJobResponse common.ListJobsResponse) error {
+func PrintExistingJobIds(listJobResponse common.ListJobsResponse, jobStatus common.JobStatus) error {
 	if listJobResponse.ErrorMessage != "" {
 		return fmt.Errorf("request failed with following error message: %s", listJobResponse.ErrorMessage)
 	}
@@ -92,6 +107,9 @@ func PrintExistingJobIds(listJobResponse common.ListJobsResponse) error {
 		sb.WriteString("Existing Jobs \n")
 		for index := 0; index < len(listJobResponse.JobIDDetails); index++ {
 			jobDetail := listJobResponse.JobIDDetails[index]
+			if jobDetail.JobStatus != jobStatus {
+				continue
+			}
 			sb.WriteString(fmt.Sprintf("JobId: %s\nStart Time: %s\nStatus: %s\nCommand: %s\n\n",
 				jobDetail.JobId.String(),
 				time.Unix(0, jobDetail.StartTime).Format(time.RFC850),
--- a/cmd/jobsList.go
+++ b/cmd/jobsList.go
@@ -76,19 +76,20 @@ func init() {
 	jobsCmd.AddCommand(lsCmd)
 
 	jobsCmd.PersistentFlags().StringVar(&commandLineInput.withStatus, "with-status", "All",
-		"only remove the jobs with this status, available values: Cancelled, Completed, Failed, InProgress, All")
+		"List the jobs with given status, available values: All, Cancelled, Failed, InProgress, Completed,"+
+			" CompletedWithErrors, CompletedWithFailures, CompletedWithErrorsAndSkipped")
 }
 
 // HandleListJobsCommand sends the ListJobs request to transfer engine
 // Print the Jobs in the history of Azcopy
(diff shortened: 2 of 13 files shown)


cmd/jobsList.go
```diff
@@ -62,18 +74,21 @@ func init() {
 	}
 
 	jobsCmd.AddCommand(lsCmd)
+
+	jobsCmd.PersistentFlags().StringVar(&commandLineInput.withStatus, "with-status", "All",
+		"only remove the jobs with this status, available values: Cancelled, Completed, Failed, InProgress, All")
```
adreed-msft, line 79: >remove [](start = 8, length = 6)

Remove? Shouldn't this be "list"?

nakulkar-msft: Yes, it should've. I'll put this in the next commit.

ste/mgr-JobPartMgr.go
```diff
@@ -355,7 +364,7 @@ func (jpm *jobPartMgr) ScheduleTransfers(jobCtx context.Context) {
 			// If source doesn't exists, skip the transfer
 			_, ok := includeTransfer[src]
 			if !ok {
-				jpm.ReportTransferDone() // Don't schedule transfer which is not mentioned to be included
+				jpm.ReportTransferDone(ts) // Don't schedule transfer which is not mentioned to be included
```
JohnRusk, line 367: I'm not sure we actually _use_ includeTransfer and exclude transfer.

@zezha-msft can you confirm whether we do?

@nakulkar-msft If if turns out that we don't use those, maybe you could just panic here, saying that "includeTransfer" lists are unsupported. (And the same below for exclude below).

The reasons I suggest this are that:
(a) in these cases, I would expect `ts` to be `Started` (the default value) for these, and that will get a panic in ReportTransferDone.
(b) Having these exclude and include lists is icky in general, and I think even more icky because they don't cleanly fit into any of the statues that you're looking at.

(I guess an alternative might be to report them done as "skipped"??? Maybe see what Ze thinks.  As noted above, I'm not sure the code in question every even gets called these days.  It was an old idea, and I'm not sure we still support it.  Maybe one day in the future, not in this PR, we could remove the code completely IF its obsolete.)

ste/mgr-JobPartMgr.go
```diff
@@ -369,7 +378,7 @@ func (jpm *jobPartMgr) ScheduleTransfers(jobCtx context.Context) {
 			// skip the transfer
 			_, ok := excludeTransfer[src]
 			if ok {
-				jpm.ReportTransferDone() // Don't schedule transfer which is mentioned to be excluded
+				jpm.ReportTransferDone(ts) // Don't schedule transfer which is mentioned to be excluded
```
JohnRusk, line 381: See comment above

ste/mgr-JobMgr.go
```diff
@@ -467,30 +472,48 @@ func (jm *jobMgr) ResetAllTransfersScheduled() {
 }
 
 // ReportJobPartDone is called to report that a job part completed or failed
-func (jm *jobMgr) ReportJobPartDone() uint32 {
+func (jm *jobMgr) ReportJobPartDone(progressInfo jobPartProgressInfo) {
```
JohnRusk, line 475: I like the way you've moved the processing out to a channel-reading routine

ste/mgr-JobMgr.go
```diff
@@ -467,30 +472,48 @@ func (jm *jobMgr) ResetAllTransfersScheduled() {
...
+		if shouldComplete {
+			break
+		} //Else log and wait for next part to complete
+
 		if shouldLog {
 			jm.Log(pipeline.LogInfo, fmt.Sprintf("is part of Job which %d total number of parts done ", partsDone))
 		}
-		return partsDone
+
+		jm.chunkStatusLogger.FlushLog()
```
JohnRusk, line 511: I'd suggest removing this call to `FlushLog`, and just leave the one call at the end of reportJobPartDoneHandler.  Why? Because this Flush() function has a sticky mode: after its been called once, it always flushes _after every single write_ from then on. Which could have a perf cost. So Flush() should only be called once at the end of the job.

(Admittedly, the sticky flush mode may not have been my best design decision!  But that's the way it works for now).

ste/mgr-JobPartMgr.go
zezha-msft: This code is indeed obsolete. Let's remove it.

---
In reply to: [448049859](https://github.com/Azure/azure-storage-azcopy/pull/1058#discussion_r448049859) [](ancestors = 448049859)

cmd/jobsList.go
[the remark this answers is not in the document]
nakulkar-msft: This is not an exhaustive list in the comment. In the case described, the list used to display Completed - but will now display CompletedWithErrors. And Failed = Zero completed. I'll update the comment.

ste/mgr-JobMgr.go
nakulkar-msft: Done

ste/mgr-JobPartMgr.go
nakulkar-msft: With confirmation that is Obsolete, I'll 'panic' if we reach those statements.

ste/mgr-JobPartMgr.go
```diff
@@ -637,15 +646,32 @@ func (jpm *jobPartMgr) deleteSnapshotsOption() common.DeleteSnapshotsOption {
...
+
+	//Add a safety count-check
+
 	if jpm.ShouldLog(pipeline.LogInfo) {
 		plan := jpm.Plan()
 		jpm.Log(pipeline.LogInfo, fmt.Sprintf("JobID=%v, Part#=%d, TransfersDone=%d of %d", plan.JobID, plan.PartNum, transfersDone, plan.NumTransfers))
 	}
 	if transfersDone == jpm.planMMF.Plan().NumTransfers {
-		jpm.jobMgr.ReportJobPartDone()
+		jpm.jobMgr.ReportJobPartDone(jpm.progressInfo)
```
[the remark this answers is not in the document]
nakulkar-msft: Now I see it - there were parts of code which accessed this atomic* variables in multithreaded ctx (updateJobPartProgress) and few others in a single thread(reportTransferDoneHandler). And I did not separate this code properly. 
I'll not make the updateJobPartProgress() return progressInfo, instead I'll compute at completion of last jptm(for a given jpm). Saves a unnecessary atomic loads. Should be good right?

cmd/jobsClean.go
```diff
@@ -90,7 +90,7 @@ func handleCleanJobsCommand(givenStatus common.JobStatus) error {
 
 	// we must query the jobs and find out which one to remove
 	resp := common.ListJobsResponse{}
-	Rpc(common.ERpcCmd.ListJobs(), nil, &resp)
+	Rpc(common.ERpcCmd.ListJobs(), common.EJobStatus.All(), &resp)
```
nakulkar-msft, line 93: Filtering Jobs was incorrect earlier (skipped json o/p). Review ListJobs() as well.

ste/mgr-JobMgr.go
```diff
@@ -467,30 +472,46 @@ func (jm *jobMgr) ResetAllTransfersScheduled() {
...
+		jobProgressInfo.transfersCompleted += partProgressInfo.transfersCompleted
+		jobProgressInfo.transfersSkipped += partProgressInfo.transfersSkipped
+		jobProgressInfo.transfersFailed += partProgressInfo.transfersFailed
+
+		// If the last part is still awaited or other parts all still not complete,
+		// JobPart 0 status is not changed (unless we are cancelling)
+		haveFinalPart = atomic.LoadInt32(&jm.atomicFinalPartOrderedIndicator) == 1
+		allKnownPartsDone := partsDone == jm.jobPartMgrs.Count()
+		isCancelling := jobStatus == common.EJobStatus.Cancelling()
+		shouldComplete := allKnownPartsDone && (haveFinalPart || isCancelling)
```
nakulkar-msft, line 502: Reorder shouldComplete and shouldLog

ste/JobsAdmin.go
```diff
@@ -455,6 +455,7 @@ func (ja *jobsAdmin) transferProcessor(workerID int) {
 			if jptm.ShouldLog(pipeline.LogInfo) {
 				jptm.Log(pipeline.LogInfo, fmt.Sprintf(" is not picked up worked %d because transfer was cancelled", workerID))
 			}
+			jptm.SetStatus(common.ETransferStatus.Failed())
```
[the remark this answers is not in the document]
nakulkar-msft: 1. I see that we do not specially handle cancelled as different status from Failed(). These changes were necessary because, in the existing workflow we leave the status to be Started(), which interferes with the bookkeeping and technically wrong. Other reason, with the existing code, there can be cancelled jptms with both Failed() and Started() status, depending on where they failed. Marking the Started() jptms Failed() would make it uniform. This might be a good time to introduce Cancelled() status - it can inform us the reason for failure of this file (similar to other status like BlobTierFailure()), but I'm not sure it might be useful.
2. Now assuming Started() is not a terminal/Static state for jptm, I expected the the transferStatus terminal state (intutively) before we ReportTransferDone(). And the places where I've set the status to Failed(), violated this condition. There were two places this violation could have been fixed: 
	a. Before calling ReportTransferDone(), set the status appropriately
	b. At the places where the status actually changes - like the Started() when scheduling, Completed() on success in Epilogue, Failed() on failure, and *Failed() when the jptm is cancelled*
3. I totally ignored the point (b) above until your comments(KMN). The new places where we've set the status dees not impact Cancel() and WasCancelled(). But all the other forms of isDead() invoked after SetJobStatus() line would behave differently (Though they still return the same result - Need to verify this again). 

I'm inclined towards moving jptm.TransferStatus to a terminal status with this change. I'm looking at 2a and 2b bit more. I think both don't impact Is*, though 2a appears to.

cmd/jobsClean.go
nakulkar-msft: Tracking this in above comment.

ste/JobsAdmin.go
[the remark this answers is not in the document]
nakulkar-msft: Update:
I looked at the code to understand a bit more about how we cancel a job and what might be a good place to set the status to Failed().

How do we cancel the job? This is required because our jptm.isDead() checks the context to decide.
All the jptm are given a new context (jptm.Ctx) whose parent is JobMgr's context(jm.Ctx). The jptm.Ctx is released either by invoking jptm.Cancel() or jm.Cancel(), whichever happens first. When transfer of a file is complete(failure or success), the jptm.Ctx is released by invoking jptm.Cancel. When we cancel a job, jptm.Ctx is released because parent ctx is released.

Where can we set the transfer Status to be Failed()
1. Reactive: Before calling ReportTransferDone(), if we find the status is not terminal, we set it appropriately. This is what I'd implemented before this discussion. In terms of code, we'd have to change the status to Failed() in commonSenderCompletion(), commonDownloaderCompletion(). Not a fan of this method but appears clean, because we set the Success() state here.

2. reactive: When we release jptm.Ctx, because the parent context - jm.Ctx is released. This would require us to loop through all the jptms while we are cancelling jobs, should not be done. 

3. First Detection: Set the status to Failed() when the jptm can detect its context is released.
Now, when can we detect the jptm.Ctx is released:
a. Before we schedule chunks - remoteToLocal(), anyToRemote()
b. After we schedule chunks - epilogueWithCleanupDownload(), epilogueWithCleanupSendToRemote()

In any case, we'd have to set the status in transferProcessor() because of a possible race. Also, jptm.Is* methods behave similarly in all the cases. As soon as the context is released, jptm is considered dead.

Issue #1028, opened 2020-06-01
The attribute filter does not work if it contains an asterisk in the path


basty0722:
##### Note: The version is visible when running AzCopy without any argument
 10.4.3

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

### What command did you run?

example

.\azcopy.exe copy "D:\dummy\2020*" "SAS" --include-attributes A --recursive --overwrite ifSourceNewer --log-level ERROR

.\azcopy.exe copy "D:\dummy\2019*" "SAS" --include-attributes A --recursive --overwrite ifSourceNewer --log-level ERROR

.\azcopy.exe copy "D:\dummy\2018*" "SAS" --include-attributes A --recursive --overwrite ifSourceNewer --log-level ERROR

-There are many files under the D:\dummy\YYYYMMDD format folder.
-Details are given below
-- Transfer the last 5 years of data to Azure
-- I want to transfer only files with the archive attribute
-- Clear archive attributes after transfer

### What problem was encountered?
When the above command was executed, the following alert occurred.

INFO: Skipping file attribute filter for file 20200105/50004512-00340967/dummy-001.jpg due to error: The filename, directory name, or volume label syntax is incorrect.
INFO: Skipping file attribute filter for file 20200105/50004512-00340967/dummy-002.jpg due to error: The filename, directory name, or volume label syntax is incorrect.

### Have you found a mitigation/solution?
No error occurs if you do not use an asterisk in the source path
example:
 "D:\dummy\*"

basty0722:
Sorry.
Some statements were incorrect.

> Have you found a mitigation/solution?
No error occurs if you do not use an asterisk in the source path
example:
"D:\dummy*"

Correct
example:
"D:\dummy\\*"

basty0722:
Thank you for your support!
I will test it.

mohsha-msft:
Issue has been in v10.5.0
Please reach out to me if the issue persists.

PR #1063, opened 2020-06-22
Fixed Issue:1028

mohsha-msft:
Issue [Link](https://github.com/Azure/azure-storage-azcopy/issues/1028)

Whenever a file path contains wildcard character such as *  (F:\202* ) and file attribute filter --include-attributes A  is specified, the code skips file attribute filter. 

This is because when we create filters inside initModularFilters() function, we get the file path from cca.source.ValueLocal() which contains "F:\202*". And when we try to get the file attributes by generating full path inside doesPath method [zc_attr_filter_windows.go], it appends F:\202* with 2020\temp.txt to get "F:\202*\2020\temp.txt"  but it should be **F:\2020\temp.txt**

To solve this, first get the base file path without wildcard and then append it with the relative paths.

Another possible way to solve this is add fix inside initModularFilters but filter.filepath maybe getting used somewhere in the code and it might  break other functionalities.

### DIFF
--- a/cmd/zc_attr_filter_windows.go
+++ b/cmd/zc_attr_filter_windows.go
@@ -45,7 +45,14 @@ func (f *attrFilter) appliesOnlyToFiles() bool {
 }
 
 func (f *attrFilter) doesPass(storedObject storedObject) bool {
-	fileName := common.GenerateFullPath(f.filePath, storedObject.relativePath)
+	fileName := ""
+	if strings.Index(f.filePath, "*") == -1 {
+		fileName = common.GenerateFullPath(f.filePath, storedObject.relativePath)
+	} else {
+		basePath := getPathBeforeFirstWildcard(f.filePath)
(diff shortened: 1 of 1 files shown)


cmd/zc_attr_filter_windows.go
```diff
@@ -45,7 +45,14 @@ func (f *attrFilter) appliesOnlyToFiles() bool {
 }
 
 func (f *attrFilter) doesPass(storedObject storedObject) bool {
-	fileName := common.GenerateFullPath(f.filePath, storedObject.relativePath)
+	fileName := ""
+	if strings.Index(f.filePath, "*") == -1 {
```
adreed-msft, line 49: >strings.Index(f.filePath, "*") == -1 [](start = 4, length = 36)

strings.Contains also works here.

adreed-msft:
Also, you can link the issue like this: #1028, and github will shorthand the link for you.

PR #1064, opened 2020-06-22
Clarify credential validation message

JohnRusk:
To reduce risk of folks using trusted-microsoft-suffixes inappropriately. (Inspired by a support case)


zezha-msft:
A minor update in the unit test is needed.

JohnRusk:
@zezha-msft I've just done the unit test fix

PR #1072, opened 2020-06-26
Declarative test initial framework

JohnRusk:
This PR adds a declarative testing layer on top of the core of the new test framework (which was introduced in https://github.com/Azure/azure-storage-azcopy/pull/842).  The declarative layer adds:

* A declarative style, in which we come closer to specifying _what_ should be tested without having to to write code (in the test itself) for _how_
* The ability to cover multiple scenarios in a single test.  Specifically, a single test can cover both copy and sync (when applicable), and can cover all source-destination pairs.  The declarative layer expands each test out into a set of subtests, to cover copy/sync and all the specified source-destination pairs.
* "Pseudo-suites", in which the framework enforces that each test must begin with a suite name, and that the suite name must match the filename that contains the tests (for flexibility, it must match one word in the filename).  E.g. TestFilter_IncludePattern is a valid test name, if it appears in a file named "enumeration_filter_test.go" because the suite name, which is the bit that appears between "Test" and the first _ matches a word in the filename.  This approach is used because the more detailed suite features in gocheck and testify/suite were both, for different reasons, unsuitable.
* As a bonus feature, it allows semi-automated debugging into AzCopy when running inside the test harness. See azcopyDebugHelper.go

## Goals

1. To have one declarative syntax that covers most scenarios that we need to test.
1. Each test focusses on one thing that AzCopy can do, e.g. filter by filename pattern, and tests that one feature in _all_ scenarios where is applies.
1. To have a syntax that emphasizes what its unique about each test, with minimal boiler-plate.
1. The same verifier code is used by all tests so there is _one_ source of truth about what constitutes a successful job, that one source of truth contains _all_ our rules, and does not have to copied or or-re-written in each test. (Instead, the dev just specifies which files should succeed and fail, and the verifier does the rest.)
1. The tests, and the declarative framework, should explain to devs what they are doing, through their code structure and comments.
1. To organize our tests into an understandable structure (hence the "pseudo-suites").
1. To stay true to the intent of the underlying imperative layer, so that the benefits of that layer (as described here https://github.com/Azure/azure-storage-azcopy/pull/842) can be achieved.
2. (Not in the initial draft of this PR) To have clear strategy for how to approach tests that don't fit this framework. The primary intention is to have this framework be enhanced so that it _can_ cover most cases that we want to test.  With falling back to the underlying imperative layer (from https://github.com/Azure/azure-storage-azcopy/pull/842) being a last resort.


cmd/benchmark.go
```diff
@@ -61,7 +61,7 @@ const (
 	sizeStringDescription = "a number immediately followed by K, M or G. E.g. 12k or 200G"
 )
 
-func parseSizeString(s string, name string) (int64, error) {
+func ParseSizeString(s string, name string) (int64, error) {
```
JohnRusk, line 64: Reminder to self: set base to dev after the AzCopy changes are merged there

adreed-msft:
> 	var name, key string

I think we should add service principal auth, so that we have the ability to test OAuth end-to-end.

---
Refers to: e2etest/config.go:38 in 8c8b7e8. [](commit_id = 8c8b7e87ae6bf0420c5dfe7e50d3fb6ba310419f, deletion_comment = False)

e2etest/declarativeHelpers.go
```diff
@@ -0,0 +1,528 @@
...
+	"reflect"
+	"strings"
+	"testing"
+)
+
+///////////
+
+// folder is syntactic sugar to make folders stand out in our file lists
+func folder(s string) string {
+	return strings.TrimRight(s, "/") + "/" // ensure it ends in one slash
```
adreed-msft: Minor: Might align with the 'f' file shortcut if this is 'd' for directory

e2etest/declarativeHelpers.go
```diff
@@ -0,0 +1,528 @@
...
+	Assert(obtained interface{}, comp comparison, expected interface{}, comment ...string)
+	AssertNoErr(err error, comment ...string)
+	Error(reason string)
+	Skip(reason string)
+	Failed() bool
+
+	// ScenarioName piggy-backs on this interface, in a context-value-like way (ugly, but it works)
+	CompactScenarioName() string
+}
+
```
adreed-msft, line 72: I would suggest modeling our assertion somewhat around Gocheck's assertion interface.

This would allow for custom comparisons (and for us to log a diff or something else of that nature).

e2etest/declarativeHelpers.go
JohnRusk: Maybe.  I'm hesitant because we've used the word "folder" so predominantly in the code. And also, because having them different lengths makes the folders look more distinct visually, which is not a bad thing IMHO, when it comes to reading and understanding the tests.

JohnRusk: I'll leave it as-is for now, but you could change it one day if that's the consensus opinion.

e2etest/declarativeHelpers.go
JohnRusk: It is somewhat similar to that interface, with two notable differences that come to mind:
1. No checker object.  The rationale for that is that (a) as soon as we commit to a particular checker object, that ties us (somewhat) to a particular framework, which is the thing this interface was added to avoid.  And (b), it sees like we don't need to actually write many assertions. E.g. the typical test, created with this framework, contains no assertions at all in the test itself. (And that's consistent with the goal of a declarative style).  For this reason, I feel it's not clear that more advanced comparisons would give us much benefit.
2. Our assert method does not cause the executing test to stop. I.e. our "Assert" == their "Check".  Maybe we should rename ours, or maybe not. (AFAICT Assert doesn't mean AssertAndStopIfFailed in other Go test frameworks)  In any case , I think it's correct that our main assertion/check method does allow the test to carry on. So you can see, for instance, information about all files that were transferred wrong, instead of just one.  Or, if the lengths of two lists don't match, we don't stop at the length check, but instead also output more info about the individual differences. (e.g. specific files that aren't found)

I'm going to leave it as-is for now, but you could easily do renames later.

JohnRusk:
After discussion with @zezha-msft  and @nakulkar-msft , we've decided to merge this now.  There e2etests themselves are not yet mature, but merging them now doesn't impact on AzCopy itself.

Work will be performed in new branches to flesh out the test harness and add tests.

PR #1080, opened 2020-07-01
E2E Scenario helpers - Implement getAllProperties for filestore

nakulkar-msft:


e2etest/scenario_helpers.go
```diff
@@ -541,11 +541,82 @@ func (scenarioHelper) generateAzureFilesFromList(c asserter, shareURL azfile.Sha
...
+	for i := 0; i < len(dirQ); i++ {
+		currentDirURL := dirQ[i]
+		for marker := (azfile.Marker{}); marker.NotDone(); {
+			lResp, err := currentDirURL.ListFilesAndDirectoriesSegment(context.TODO(), marker, azfile.ListFilesAndDirectoriesOptions{})
+			a.AssertNoErr(err)
+
+			// Process the files and folders we listed
+			for _, fileInfo := range lResp.FileItems {
+				fileURL := currentDirURL.NewFileURL(fileInfo.Name)
+				fgpr, err := fileURL.GetProperties(context.TODO())
```
JohnRusk, line 558: Minor: in terms of coding style, fgpr is not a "Go-ish" name. Idiomatic Go code would probably just use "p" as the name of this variable.  But, here in AzCopy, we tend to be a bit more verbose.  So "props" might be a good AzCopy idiomatic name.  

BTW You'll often also see us use two-word names, such as fileSize.  I think it could be argued that our two word names are too long by Go standards in some cases (e.g. when the meaning would be obvious with one word) but justifiable in some others (IMHO when they help with readability of non-obvious code).

But one thing that you would never see in AzCopy or in conventional idiomatic Go with shorter names, would be something like fgpr.  So I'd recommend avoiding names of that style.

e2etest/scenario_helpers.go
```diff
@@ -541,11 +541,82 @@ func (scenarioHelper) generateAzureFilesFromList(c asserter, shareURL azfile.Sha
...
+				filePermissions := fgpr.FilePermissionKey()
+
+				props := objectProperties{
+					isFolder:           false, // no folders in Blob
+					size:               &fileSize,
+					nameValueMetadata:  fgpr.NewMetadata(),
+					contentHeaders:     &h,
+					creationTime:       &creationTime,
+					lastWriteTime:      &lastWriteTime,
+					smbAttributes:      &fileAttrs, //TODO
```
JohnRusk, line 586: For this, I actually never figured out how attributes should be represented in the test harness code.  There seem to be at least three possibilities:
1. Use an unsigned integer that corresponds to what is natively used (as a set of bitflags) by Windows.
2. Use an array of strings (whole words) which corresponds to what is used by Azure Files.
3. Use our own format in the test harness, which might be a set of single letters (specifically the same set that we use as our command line filter parameters when filtering by attributes) or maybe some other format of our choosing.  I like the conciseness of the command-line format.

Since you're taking over the test harness, this is probably something that you can decide and implement. Basically you'll need to change the type of objectProperties.smbAttributes to match your choice. Then you'll need to ensure the appropriate conversions to and/or from that format when that property is used to interface with Windows and Azure Files. (Windows and Azure Files are the only places that have such attributes and the only places we support the attributes).

If you do _not_ use the same format as on the command line, then you'll need to look at with.smbAttributes in file declarativeWithPropertyProviders.go.  There's a "todo" comment there.

e2etest/scenario_helpers.go
```diff
@@ -541,11 +541,82 @@ func (scenarioHelper) generateAzureFilesFromList(c asserter, shareURL azfile.Sha
...
+				h := contentHeaders{
+					cacheControl:       &contentHeader.CacheControl,
+					contentDisposition: &contentHeader.ContentDisposition,
+					contentEncoding:    &contentHeader.ContentEncoding,
+					contentLanguage:    &contentHeader.ContentLanguage,
+					contentType:        &contentHeader.ContentType,
+					contentMD5:         contentHeader.ContentMD5,
+				}
+				fileAttrs := fgpr.FileAttributes()
+				filePermissions := fgpr.FilePermissionKey()
```
JohnRusk, line 577: In answer to your offline question: the key is not the thing that we need, in the test harness. What we actually need, is the SDDL string that the key is the key for.  Basically the way AzCopy works with permissions is like this:
* When uploading from Windows to Azure Files, ask Windows for the SDDL string that describes the security info or the file or directory.
* Do some search and replace operations on that for certain short names and replace them with real SIDs that belong to the source location. 
* Send the resulting SDDL to Azure Files.  
* Azure Files Service does a thing where it takes the SDDL and assigns a key to it, and then when you read, its the key that you get back. (That's what you are doing here).  Within a given File Share, all uploads of the same SDDL string will map to the same key at download time.  So typically, when downloading as you are doing here, you'll get many items with the same key. So in theory, you can do some caching to look up the SDDL for each key just once.
* (In the main AzCopy code) if we want to set the permissions on a downloaded file, that we are saving to disk, we just pass Windows the SDDL that we got by looking up the key.

You can see examples of all this if you search for SDDL in the main AzCopy code.

e2etest/scenario_helpers.go
```diff
@@ -541,11 +541,82 @@ func (scenarioHelper) generateAzureFilesFromList(c asserter, shareURL azfile.Sha
...
+
+			for _, dirInfo := range lResp.DirectoryItems {
+				dirURL := currentDirURL.NewDirectoryURL(dirInfo.Name)
+				dgpr, err := dirURL.GetProperties(context.TODO())
+				a.AssertNoErr(err)
+
+				// Construct the properties object
+				creationTime, err := time.Parse(azfile.ISO8601, dgpr.FileCreationTime())
+				a.AssertNoErr(err)
+				lastWriteTime, err := time.Parse(azfile.ISO8601, dgpr.FileLastWriteTime())
```
JohnRusk, line 601: IIRC, we have a policy in AzCopy of NOT using or populating or transferring lastWriteTimes for directories. That's because it doesn't have any meaninginful value in most cases. I'll send you a link to our document about that decision.  So I think you should leave lastWriteTime nil.

e2etest/scenario_helpers.go
```diff
@@ -541,11 +541,82 @@ func (scenarioHelper) generateAzureFilesFromList(c asserter, shareURL azfile.Sha
...
+				a.AssertNoErr(err)
+
+				props := objectProperties{
+					isFolder:          true,
+					nameValueMetadata: dgpr.NewMetadata(),
+					creationTime:      &creationTime,
+					lastWriteTime:     &lastWriteTime,
+				}
```
JohnRusk, line 609: We do need the attributes and SDDL for folders.

e2etest/scenario_helpers.go
```diff
@@ -541,11 +541,82 @@ func (scenarioHelper) generateAzureFilesFromList(c asserter, shareURL azfile.Sha
...
+					contentEncoding:    &contentHeader.ContentEncoding,
+					contentLanguage:    &contentHeader.ContentLanguage,
+					contentType:        &contentHeader.ContentType,
+					contentMD5:         contentHeader.ContentMD5,
+				}
+				fileAttrs := fgpr.FileAttributes()
+				filePermissions := fgpr.FilePermissionKey()
+
+				props := objectProperties{
+					isFolder:           false, // no folders in Blob
```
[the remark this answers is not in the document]
nakulkar-msft: Fixed

PR #1089, opened 2020-07-07
V10.5.0 Changelog + Resolve conflicts with master branch

zezha-msft:

PR #1091, opened 2020-07-07
Fix FileLastWriteTime for SMB properties

zezha-msft:


zezha-msft:
Merged as part of #1089

PR #1117, opened 2020-07-28
Always log command even at ERROR level

zezha-msft:
Some random fixes discovered while working on a support case.

### DIFF
--- a/cmd/jobsList.go
+++ b/cmd/jobsList.go
@@ -75,7 +75,7 @@ func init() {
 
 	jobsCmd.AddCommand(lsCmd)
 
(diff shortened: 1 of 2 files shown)


cmd/jobsList.go
```diff
@@ -75,7 +75,7 @@ func init() {
 
 	jobsCmd.AddCommand(lsCmd)
 
-	jobsCmd.PersistentFlags().StringVar(&commandLineInput.withStatus, "with-status", "All",
+	lsCmd.PersistentFlags().StringVar(&commandLineInput.withStatus, "with-status", "All",
```
zezha-msft, line 78: This was showing up as a flag that applied to every sub-command. Instead, the `jobs show` command actually had its definition (which has the same flag for transfers not jobs) overwritten.

PR #1123, opened 2020-08-01
Support download/delete of blob versions

mohsha-msft:


cmd/copy.go
```diff
@@ -1532,7 +1553,7 @@ func init() {
 		"In the cases that setting access tier is not supported, please use s2sPreserveAccessTier=false to bypass copying access tier. (default true). ")
 	cpCmd.PersistentFlags().BoolVar(&raw.s2sSourceChangeValidation, "s2s-detect-source-changed", false, "Detect if the source file/blob changes while it is being read. (This parameter only applies to service to service copies, because the corresponding check is permanently enabled for uploads and downloads.)")
 	cpCmd.PersistentFlags().StringVar(&raw.s2sInvalidMetadataHandleOption, "s2s-handle-invalid-metadata", common.DefaultInvalidMetadataHandleOption.String(), "Specifies how invalid metadata keys are handled. Available options: ExcludeIfInvalid, FailIfInvalid, RenameIfInvalid. (default 'ExcludeIfInvalid').")
-
+	cpCmd.PersistentFlags().StringVar(&raw.listOfVersionIDs, "list-of-version-ids", "", "Provides a list of version ids of a blob to download")
```
zezha-msft, line 1556: >list-of-version-ids [](start = 59, length = 19)

I'd suggest to keep the flag name concise: list-of-versions

zezha-msft, line 1556: >Provides a list of version ids of a blob to download [](start = 86, length = 52)

Please add that when this flag is specified, it points to a file where each version id is listed on a separate line, and that the source should point to a single blob.

cmd/copy.go
```diff
@@ -272,6 +273,24 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
...
+		}
+		defer filePtr.Close()
+		scanner := bufio.NewScanner(filePtr)
+		scanner.Split(bufio.ScanLines)
+		var versionIDs []string
+		for scanner.Scan() {
+			versionIDs = append(versionIDs, scanner.Text())
+		}
+		cooked.listOfVersionIDs = versionIDs
+	}
```
zezha-msft, line 292: please follow the same workflow as the list-of-files, i.e. not read out the entire file right away, but pass a channel to the traverser. Also, we need to skip empty lines, and UTFBOM, both can become sources of bugs.

cmd/copyEnumeratorInit.go
```diff
@@ -497,16 +497,19 @@ func (cca *cookedCopyCmdArgs) makeEscapedRelativePath(source bool, dstIsDir bool
 		if source {
 			relativePath = ""
 		} else {
+			if object.versionID != "" {
+				relativePath += "/" + object.versionID
+			}
```
zezha-msft, line 502: This does not look correct, I thought the blob name would become blobname.versionid. Also, the cx might have specified a folder as the destination.

cmd/remove.go
```diff
@@ -97,4 +97,5 @@ func init() {
 	deleteCmd.PersistentFlags().BoolVar(&raw.forceIfReadOnly, "force-if-read-only", false, "When deleting an Azure Files file or folder, force the deletion to work even if the existing object is has its read-only attribute set")
 	deleteCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "Defines the location of a file which contains the list of files and directories to be deleted. The relative paths should be delimited by line breaks, and the paths should NOT be URL-encoded.")
 	deleteCmd.PersistentFlags().StringVar(&raw.deleteSnapshotsOption, "delete-snapshots", "", "By default, the delete operation fails if a blob has snapshots. Specify 'include' to remove the root blob and all its snapshots; alternatively specify 'only' to remove only the snapshots but keep the root blob.")
+	deleteCmd.PersistentFlags().StringVar(&raw.listOfVersionIDs, "list-of-version-ids", "", "Provides a list of version ids of a blob to delete")
```
zezha-msft, line 100: >Provides a list of version ids of a blob to delete [](start = 90, length = 50)

same as copy

cmd/zc_traverser_blob.go
```diff
@@ -129,6 +129,7 @@ func (t *blobTraverser) traverse(preprocessor objectMorpher, processor objectPro
 			blobPropertiesResponseAdapter{blobProperties},
 			common.FromAzBlobMetadataToCommonMetadata(blobProperties.NewMetadata()), // .NewMetadata() seems odd to call, but it does actually retrieve the metadata from the blob properties.
 			blobUrlParts.ContainerName,
+			"", // versionID field is an attribute of blobs only.
```
zezha-msft, line 132: >versionID field is an attribute of blobs only. [](start = 10, length = 46)

minor: fix comment since this is the blob traverser

cmd/zc_traverser_blob_versions.go
```diff
@@ -0,0 +1,130 @@
+// Copyright © 2017 Microsoft <wastore@microsoft.com>
```
zezha-msft, line 1: >2017 [](start = 15, length = 4)

Please remove the year 2017

cmd/zc_traverser_blob_versions.go
```diff
@@ -0,0 +1,130 @@
...
+	props, err = blobURL.GetProperties(t.ctx, azblob.BlobAccessConditions{})
+
+	if err != nil {
+		return props, err
+	}
+
+	// if there was no problem getting the properties, it implies that we are looking at a single blob
+	if gCopyUtil.doesBlobRepresentAFolder(props.NewMetadata()) {
+		return props, errors.New("This is not a blob")
+	}
```
zezha-msft, line 74: we don't have to worry about the directory stubs in this traverser

cmd/zc_traverser_blob_versions.go
```diff
@@ -0,0 +1,130 @@
...
+	return nil
+}
+
+func newBlobVersionsTraverser(rawURL *url.URL, p pipeline.Pipeline, ctx context.Context, recursive, includeDirectoryStubs bool,
+	incrementEnumerationCounter enumerationCounterFunc, listOfVersionIds []string) (t *blobVersionsTraverser) {
+	return &blobVersionsTraverser{
+		rawURL:                      rawURL,
+		p:                           p,
+		ctx:                         ctx,
+		recursive:                   recursive,
```
zezha-msft, line 125: >recursive [](start = 31, length = 9)

recursive doesn't apply to this traverser, right?

cmd/zc_traverser_blob_versions.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	return props, nil
+
+}
+
+func (t *blobVersionsTraverser) traverse(preprocessor objectMorpher, processor objectProcessor, filters []objectFilter) (err error) {
+	blobURLParts := azblob.NewBlobURLParts(*t.rawURL)
+
+	for _, versionID := range t.listOfVersionIds {
+		blobProperties, err := t.getBlobProperties(versionID)
```
zezha-msft, line 83: >blobProperties, err := t.getBlobProperties(versionID) [](start = 2, length = 53)

while getting the properties for every version is the technically right thing to do, but in both of our scenarios (download and remove) we do not need the properties, so I suggest to not get the properties for now, since it's a perf hit (1 i/o per version in a serial chain of gets). We should have a check somewhere to verify that if --list-of-versions is specified, the destination has to be local or trash.

ste/JobPartPlan.go
```diff
@@ -45,6 +45,7 @@ type JobPartPlanHeader struct {
 	SourceRootLength       uint16            // The length of the source root path
 	SourceRoot             [1000]byte        // The root directory of the source
 	SourceExtraQueryLength uint16
+	SourceVersionIDs       []string
```
zezha-msft, line 48: >[]string [](start = 24, length = 8)

as we said, array of string cannot be used here. Make the version part of each transfer instead.

cmd/copy.go
mohsha-msft: Done

mohsha-msft: Done

cmd/remove.go
mohsha-msft: Done

cmd/zc_traverser_blob.go
mohsha-msft: Done

cmd/zc_traverser_blob_versions.go
mohsha-msft: And replace it with?

cmd/zc_traverser_blob_versions.go
mohsha-msft: Ohh correct. Since the versions are already specified

cmd/zc_traverser_blob_versions.go
mohsha-msft: When we construct stored object, we also create a field called  blobProperties.ContentLength(). 
How do you suggest we get that value? If it's not mandatory it get this value, we can easily remove getProperties call.

cmd/zc_traverser_blob_versions.go
mohsha-msft: Done

ste/JobPartPlan.go
mohsha-msft: Done

cmd/copyEnumeratorInit.go
mohsha-msft: Awaiting response from StgExp folks.

cmd/copy.go
mohsha-msft: Done

cmd/copy.go
```diff
@@ -436,6 +438,68 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
...
+		}
+	}
+
+	go func() {
+		defer close(versionsChan)
+
+		addToChannel := func(v string, paramName string) {
+			// empty strings should be ignored, otherwise the source root itself is selected
+			if len(v) > 0 {
+				raw.warnIfHasWildcard(includeWarningOncer, paramName, v)
```
zezha-msft, line 459: >raw.warnIfHasWildcard(includeWarningOncer, paramName, v) [](start = 4, length = 56)

this check does not apply to versions

cmd/copy.go
```diff
@@ -436,6 +438,68 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
...
+					} else {
+						const jsonStart = "{\"Files\":"
+						jsonStartNoBrace := strings.TrimPrefix(jsonStart, "{")
+						isJSON := cleanedLine == jsonStart || firstLineIsCurlyBrace && cleanedLine == jsonStartNoBrace
+						if isJSON {
+							glcm.Error("The format for list-of-files has changed. The old JSON format is no longer supported")
+						}
+					}
+					headerLineNum++
+				}
```
zezha-msft, line 492: this entire check as doesn't apply to versions. This was checking whether the old list-of-files format (JSON) is mistakenly used by the user.

cmd/zc_traverser_blob_versions.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/Azure/azure-pipeline-go/pipeline"
+	"github.com/Azure/azure-storage-azcopy/common"
+	"github.com/Azure/azure-storage-blob-go/azblob"
+)
+
+type blobVersionsTraverser struct {
+	rawURL                      *url.URL
+	p                           pipeline.Pipeline
+	ctx                         context.Context
+	recursive                   bool
```
zezha-msft, line 37: >recursive                   bool [](start = 1, length = 32)

remove recursive here too

cmd/zc_enumerator.go
```diff
@@ -388,6 +388,8 @@ func initResourceTraverser(resource common.ResourceString, location common.Locat
 			}
 
 			output = newBlobAccountTraverser(resourceURL, *p, *ctx, includeDirectoryStubs, incrementEnumerationCounter)
+		} else if listOfVersionIds != nil {
+			output = newBlobVersionsTraverser(resourceURL, *p, *ctx, recursive, includeDirectoryStubs, incrementEnumerationCounter, listOfVersionIds)
```
zezha-msft, line 392: >output = newBlobVersionsTraverser(resourceURL, *p, *ctx, recursive, includeDirectoryStubs, incrementEnumerationCounter, listOfVersionIds) [](start = 3, length = 137)

Can we enforce that the source is a single blob?

cmd/zc_traverser_blob_versions.go
```diff
@@ -0,0 +1,121 @@
...
+			common.EEntityType.File(),
+			blobProperties.LastModified(),
+			blobProperties.ContentLength(),
+			blobProperties,
+			blobPropertiesResponseAdapter{blobProperties},
+			common.FromAzBlobMetadataToCommonMetadata(blobProperties.NewMetadata()),
+			blobURLParts.ContainerName,
+		)
+
+		storedObject.Metadata["versionID"] = versionID
```
zezha-msft, line 97: >storedObject.Metadata["versionID"] = versionID [](start = 2, length = 46)

this works but it's a hacky solution. Metadata has a very specific meaning in the context of blob service; it is a set of key/value pairs specified by the user; we should use a dedicated field to store the version ID. If we sneak in the version ID this way, it'll be hard for the future devs to notice this and we might accidentally overwrite the metadata when blob -> blob copy is supported for versions.

cmd/zc_enumerator.go
zezha-msft: and also the destination can only be local or trash

---
In reply to: [465411289](https://github.com/Azure/azure-storage-azcopy/pull/1123#discussion_r465411289) [](ancestors = 465411289)

cmd/copy.go
mohsha-msft: Yes. I was thinking the same. I wanted to confirm with you once.

cmd/zc_traverser_blob_versions.go
mohsha-msft: Yes that's entirely correct. In fact Narasimha raised the same concern. I'm also thinking of a unique field but there is always a chance that user may use that dedicated field to set metadata.

cmd/zc_traverser_blob_versions.go
mohsha-msft: Done

cmd/copy.go
mohsha-msft: Done

cmd/zc_traverser_blob_versions.go
mohsha-msft: Done

cmd/copyEnumeratorInit.go
mohsha-msft: Done. Going ahead with "-" in the name. 
"2020-08-05T06-59-20.9890140Z-myblob"

cmd/zc_enumerator.go
mohsha-msft: Done.

ste/JobPartPlan.go
```diff
@@ -309,6 +312,7 @@ type JobPartPlanTransfer struct {
 	SourceSize int64
 	// CompletionTime represents the time at which transfer was completed
 	CompletionTime uint64
+	VersionID      string
```
nakulkar-msft, line 315: Flag: This is not being used.

ste/mgr-JobPartTransferMgr.go
```diff
@@ -272,6 +272,20 @@ func (jptm *jobPartTransferMgr) Info() TransferInfo {
...
 	}
 
+	if versionID != "" {
+		versionID = "versionId=" + versionID
+		sURL, e := url.Parse(src)
+		if e != nil {
+			panic(e)
+		}
+		if len(sURL.RawQuery) > 0 {
+			sURL.RawQuery += "&" + versionID
```
nakulkar-msft, line 282: Minor: Use query.Add() instead. As in [here](https://golang.org/pkg/net/url/#example_URL)

cmd/copy.go
mohsha-msft: Done

cmd/copyEnumeratorInit.go
```diff
@@ -65,7 +67,9 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 
 	// Check if the destination is a directory so we can correctly decide where our files land
 	isDestDir := cca.isDestDirectory(cca.destination, &ctx)
-
+	if cca.listOfVersionIDs != nil && (!(cca.fromTo.To() == cca.fromTo.To().Local() || cca.fromTo.To() == cca.fromTo.To().Unknown()) || !isDestDir) {
```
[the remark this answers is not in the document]
mohsha-msft: In the line 63: `isSourceDir := traverser.isDirectory(true)`, we are checking if source is a directory or not. `isDirectory() ` function internally calls getProperties() which will fail if given source is not a valid blob URL. So we're already checking the source. Checking only destination should do.  So that's why I didn't add check for source. Am I missing something?

cmd/copy.go
```diff
@@ -436,6 +438,46 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
...
 
+	versionsChan := make(chan string)
+	var filePtr *os.File
+	// Get file path from user which would contain list of all versionIDs
+	// Process the file line by line and then prepare a list of all version ids of the blob.
+	if raw.listOfVersionIDs != "" {
+		filePtr, err = os.Open(raw.listOfVersionIDs)
+
+		if err != nil {
+			return cooked, fmt.Errorf("cannot open %s file passed with the list-of-file flag", raw.listOfFilesToCopy)
```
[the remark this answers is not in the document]
mohsha-msft: Oh silly copy-paste mistake. Changed.

cmd/copy.go
```diff
@@ -1532,7 +1576,7 @@ func init() {
 		"In the cases that setting access tier is not supported, please use s2sPreserveAccessTier=false to bypass copying access tier. (default true). ")
 	cpCmd.PersistentFlags().BoolVar(&raw.s2sSourceChangeValidation, "s2s-detect-source-changed", false, "Detect if the source file/blob changes while it is being read. (This parameter only applies to service to service copies, because the corresponding check is permanently enabled for uploads and downloads.)")
 	cpCmd.PersistentFlags().StringVar(&raw.s2sInvalidMetadataHandleOption, "s2s-handle-invalid-metadata", common.DefaultInvalidMetadataHandleOption.String(), "Specifies how invalid metadata keys are handled. Available options: ExcludeIfInvalid, FailIfInvalid, RenameIfInvalid. (default 'ExcludeIfInvalid').")
-
+	cpCmd.PersistentFlags().StringVar(&raw.listOfVersionIDs, "list-of-versions", "", "Specifies a file where each version id is listed on a separate line. Ensure that the source must point to a single blob and all the version ids specified in the file using this flag must belong to the source blob only.")
```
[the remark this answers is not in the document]
mohsha-msft: Yeah okay. Also, why not add it inside `helpMessages.go ` as well? See `const copyCmdExample ` variable in `helpMessages.go`

ste/mgr-JobPartTransferMgr.go
mohsha-msft: Yeah we can follow this practice as well. I just copied the same way as written in line 262 where we are appending SAS in the source URL string to keep things similar. 

```
if len(srcSAS) > 0 {
		sUrl, e := url.Parse(src)
		if e != nil {
			panic(e)
		}
		if len(sUrl.RawQuery) > 0 {
			sUrl.RawQuery += "&" + srcSAS
		} else {
			sUrl.RawQuery = srcSAS
		}
		src = sUrl.String()
	}
```

cmd/copyEnumeratorInit.go
```diff
@@ -501,12 +505,16 @@ func (cca *cookedCopyCmdArgs) makeEscapedRelativePath(source bool, dstIsDir bool
...
 				// and so relativePath needs the _name_ of the source.
-				relativePath = "/" + object.name
+				processedVID := ""
+				if len(object.blobVersionID) > 0 {
+					processedVID = strings.ReplaceAll(object.blobVersionID, ":", "-") + "-"
+				} 
+				relativePath += "/" + processedVID + object.name
 			} else {
-				relativePath = ""
+				relativePath += ""
```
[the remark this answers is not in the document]
mohsha-msft: Done

ste/JobPartPlan.go
mohsha-msft: Done

cmd/copyEnumeratorInit.go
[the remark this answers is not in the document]
nakulkar-msft: How does this code behave if I have a filestore as source and a local destination?

cmd/copy.go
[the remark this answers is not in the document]
mohsha-msft: Done

cmd/copyEnumeratorInit.go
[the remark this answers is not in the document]
mohsha-msft: Rectified the if statement

ste/xfer-anyToRemote-file.go
```diff
@@ -171,10 +171,18 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
...
-			jptm.LogSendError(info.Source, info.Destination, "Transfer source and destination are the same, which would cause data loss. Aborting transfer.", 0)
-			jptm.SetStatus(common.ETransferStatus.Failed())
-			jptm.ReportTransferDone()
-			return
+
+			srcRQ := srcURL.Query()
+			dstRQ := dstURL.Query()
+			if (len(srcRQ["versionId"]) > 0 || len(srcRQ["versionid"]) > 0) && !(len(dstRQ["versionId"]) > 0 || len(dstRQ["versionid"]) > 0) {
+				// Case: Replacing the current version of the blob with the previous version.
+				// In this particular case, source URL should contain version id and destination URL should not have any version id specified
```
zezha-msft, line 179: suggestion: if destRQ has a version ID, we should throw a meaningful error to tell the user that they cannot copy into a version explicitly.

zezha-msft:
Looking good apart from that one comment. @mohsha-msft please squash&merge appropriately. 👍

ste/xfer-anyToRemote-file.go
mohsha-msft: Ohh Okay. I'll add that.

PR #1149, opened 2020-08-20
Fixed issue with stdin and sockets

johanburati:
Seem Korn shell uses sockets


cmd/copy.go
```diff
@@ -1463,7 +1463,7 @@ func isStdinPipeIn() (bool, error) {
 	// if the stdin is a named pipe, then we assume there will be data on the stdin
 	// the reason for this assumption is that we do not know when will the data come in
 	// it could come in right away, or come in 10 minutes later
-	return info.Mode()&os.ModeNamedPipe != 0, nil
+	return info.Mode()&(os.ModeNamedPipe|os.ModeSocket) != 0, nil
```
vibhansa-msft, line 1466: Shall put a check to allow socket only on KSH.

vibhansa-msft (approved):

PR #1153, opened 2020-08-25
Minor fixes 

nakulkar-msft:
1. Always perform GetRemoteProperties if  --overwrite=IfSourceNewer
2. Discard Charset while deciding mime type


ste/mgr-JobPartMgr.go
```diff
@@ -579,8 +579,8 @@ func (jpm *jobPartMgr) resourceDstData(fullFilePath string, dataFileToXfer []byt
 var builtinTypes = map[string]string{
 	".css":  "text/css;",
```
mohsha-msft, line 580: remove ; from `".css":  "text/css;"`, and `".xml":  "text/xml;"`

mohsha-msft (approved):
LGTM.
Minor Comment: I recommend linking PR with the issue. To do that we can use #issueNumber in the title of the PR and Github will autolink it.

PR #1158, opened 2020-08-28
[AzCopyV10][Perf] Set tier directly at time of upload (instead of a separate set tier call)

mohsha-msft:
[Setting access tier at the time of Upload](https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob#request-headers-all-blob-types), instead of invoking ``SetTier()`` method separately, will improve the performance.

Note 1: **``block-blob-tier``** flag is used to set access tier of block blob. Valid tiers are - **``Hot/Cool/Archive``**
Note 2: **``page-blob-tier``** flag is used to set access tier of page blob. Valid tiers are - **``P10/P15/P20/P30/P4/P40/P50/P6/P60/P70/P80``**. 

e.g. `"cp" "https://[accountName].blob.core.windows.net/[sourceContainerName]/[blobname]?[SASQuery]" "https://[accountName].blob.core.windows.net/[destinationContainerName]/[blobname]?[SASQuery]" --block-blob-tier="[valid-tier]"`

For more information on access tiers, refer [this](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-storage-tiers?tabs=azure-portal).

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -648,12 +648,12 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 		if cooked.blobType != common.EBlobType.Detect() && cooked.fromTo.To() != common.ELocation.Blob() {
 			return cooked, fmt.Errorf("blob-type is not supported for the scenario (%s)", cooked.fromTo.String())
 		}
-		// Disabling blob tier override, when copying block -> block blob or page -> page blob, blob tier will be kept,
-		// For s3 and file, only hot block blob tier is supported.
-		//if cooked.blockBlobTier != common.EBlockBlobTier.None() ||
-		//	cooked.pageBlobTier != common.EPageBlobTier.None() {
-		//	return cooked, fmt.Errorf("blob-tier is not supported while copying from service to service")
-		//}
+
+		// Setting blob tier is supported only when destination is a blob storage. Disabling it for all the other transfer scenarios.
+		if (cooked.blockBlobTier != common.EBlockBlobTier.None() || cooked.pageBlobTier != common.EPageBlobTier.None()) &&
+			cooked.fromTo.To() != common.ELocation.Blob() {
+			return cooked, fmt.Errorf("blob-tier is not supported for the scenario (%s)", cooked.fromTo.String())
+		}
 		if cooked.noGuessMimeType {
 			return cooked, fmt.Errorf("no-guess-mime-type is not supported while copying from service to service")
 		}
--- a/ste/xfer-anyToRemote-file.go
+++ b/ste/xfer-anyToRemote-file.go
@@ -40,7 +40,7 @@ import (
 // This code for blob tier safety is _not_ safe for multiple jobs at once.
 // That's alright, but it's good to know on the off chance.
 // This sync.Once is present to ensure we output information about a S2S access tier preservation failure to stdout once
-var s2sAccessTierFailureLogStdout sync.Once
+var tierNotAllowedFailure sync.Once
 var checkLengthFailureOnReadOnlyDst sync.Once
 
 // This sync.Once and string pair ensures that we only get a user's destination account kind once when handling set-tier
@@ -131,12 +131,12 @@ func ValidateTier(jptm IJobPartTransferMgr, blobTier azblob.AccessTierType, blob
 		tierAvailable := BlobTierAllowed(blobTier)
 
 		if tierAvailable {
-			if tierSetPossibleFail {
-				jptm.LogTransferInfo(pipeline.LogWarning, jptm.Info().Source, jptm.Info().Destination, "Cannot set destination block blob to the pending access tier ("+string(blobTier)+"), because either the destination account or blob type does not support it. The transfer will still succeed.")
-			}
 			return true
 		} else {
-			jptm.LogTransferInfo(pipeline.LogWarning, jptm.Info().Source, jptm.Info().Destination, "The intended tier ("+string(blobTier)+") isn't available on the destination blob type or storage account, so it was left as the default.")
+			tierNotAllowedFailure.Do(func() {
+				glcm := common.GetLifecycleMgr()
+				glcm.Info("Destination could not accommodate the tier " + string(blobTier) + ". Going ahead with the default tier. In case of service to service transfer, consider setting the flag --s2s-preserve-access-tier=false.")
+			})
 			return false
 		}
 	} else {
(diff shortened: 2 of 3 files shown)


ste/sender-blockBlobFromLocal.go
```diff
@@ -92,7 +92,7 @@ func (u *blockBlobUploader) generatePutWholeBlob(id common.ChunkID, blockIndex i
 		jptm.LogChunkStatus(id, common.EWaitReason.Body())
 		var err error
 		if jptm.Info().SourceSize == 0 {
-			_, err = u.destBlockBlobURL.Upload(jptm.Context(), bytes.NewReader(nil), u.headersToApply, u.metadataToApply, azblob.BlobAccessConditions{})
+			_, err = u.destBlockBlobURL.Upload(jptm.Context(), bytes.NewReader(nil), u.headersToApply, u.metadataToApply, azblob.BlobAccessConditions{}, u.destBlobTier)
```
nakulkar-msft, line 95: The verification and logging performed by `AttemptSetBlobTier()` should be performed here. There are restrictions on what tier can be applied to a blob. `u.dstBlobTier` should be set if the tier clears those conditions.  Also, the log messages wrt. to `SetTier` failures should be preserved.

nakulkar-msft (changes requested):

cmd/copy.go
```diff
@@ -650,10 +650,10 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 		}
 		// Disabling blob tier override, when copying block -> block blob or page -> page blob, blob tier will be kept,
 		// For s3 and file, only hot block blob tier is supported.
-		if cooked.blockBlobTier != common.EBlockBlobTier.None() ||
-			cooked.pageBlobTier != common.EPageBlobTier.None() {
-			return cooked, fmt.Errorf("blob-tier is not supported while copying from service to service")
-		}
```
zezha-msft, line 656: is this meant to be commented out?

cmd/zt_scenario_helpers_for_test.go
```diff
@@ -285,7 +285,7 @@ func (scenarioHelper) generateBlobsFromList(c *chk.C, containerURL azblob.Contai
 	for _, blobName := range blobList {
 		blob := containerURL.NewBlockBlobURL(blobName)
 		cResp, err := blob.Upload(ctx, strings.NewReader(data), azblob.BlobHTTPHeaders{},
-			nil, azblob.BlobAccessConditions{})
+			nil, azblob.BlobAccessConditions{}, azblob.DefaultAccessTier)
```
zezha-msft, line 288: >azblob.DefaultAccessTier [](start = 39, length = 24)

To confirm, the default value is no value, right? In other words we don't send any value to the service?

I'm asking since the cx's account has a default tier, so we should let the service decide the default.

cmd/copy.go
mohsha-msft: S2S blob tier copy is now supported.

cmd/zt_scenario_helpers_for_test.go
mohsha-msft: Yes.  
```var DefaultAccessTier AccessTierType = ""```
That's how it is declared and it is equivalent to AccessTierNone

File: ste/xfer-anyToRemote-file.go [hunk 4]
```diff
@@ -129,29 +131,16 @@ func AttemptSetBlobTier(jptm IJobPartTransferMgr, blobTier azblob.AccessTierType
...
-				}
-
-				// If we know the destination tier is possible, something's wrong and we should error out.
-				if tierSetPossibleFail {
-					jptm.LogTransferInfo(pipeline.LogWarning, jptm.Info().Source, jptm.Info().Destination, "Cannot set destination block blob to the pending access tier ("+string(blobTier)+"), because either the destination account or blob type does not support it. The transfer will still succeed.")
-				} else {
-					jptm.FailActiveSendWithStatus("Setting tier", err, common.ETransferStatus.BlobTierFailure())
-				}
+			if tierSetPossibleFail {
+				jptm.LogTransferInfo(pipeline.LogWarning, jptm.Info().Source, jptm.Info().Destination, "Cannot set destination block blob to the pending access tier ("+string(blobTier)+"), because either the destination account or blob type does not support it. The transfer will still succeed.")
```
zezha-msft, line 135: I'm not sure that this log makes sense anymore, since we didn't perform the set tier operation here.

zezha-msft (changes requested):
The new error handling may need some analysis: 

1. If the user specified a tier, and we weren't able to persist it, we should fail the transfer
2. If the user didn't specify a tier for S2S, we should attempt to set tier based on the source blob, but if it fails, we should not fail the transfer, since it was optional.

Are we respecting the same behavior as before?

cmd/copy.go
mohsha-msft: FromTo scenarios where I think tier can be set using block-blob-tier or page-blob-tier flags are LocalBlob(), BlobBlob(), FileBlob(), & S3Blob(). Is that correct? If yes, I'll add the below if statement to check the error.

```
if (cooked.blockBlobTier != common.EBlockBlobTier.None() || cooked.pageBlobTier != common.EPageBlobTier.None()) &&
   cooked.fromTo.To() != common.ELocation.Blob() {
   return cooked, fmt.Errorf("blob-tier is not supported while copying from service to service")
}
```

mohsha-msft:
> The new error handling may need some analysis:
> 
> 1. If the user specified a tier, and we weren't able to persist it, we should fail the transfer

To answer your first question, I am setting default blob tier in case I'm unable to set tier specified by user. This is totally because I think it would be bad customer experience to fail the AzCopy request after user has invested his network & time in uploading the blocks using Put Block call. If Put Block requests have succeeded but Put Block List fails, data uploaded by Put Block calls would be rendered useless. Earlier, we used to do Set Tier call separately which means even if the call fails the upload was successful. 

For example: Suppose the customer is trying to upload 10GB blob. We first do Put Block calls and upload the data. Next we validate the tier but fails the request because we couldn't validate the tier. Data uploaded by Put Block calls would be rendered useless.

What do you think?

ste/xfer-anyToRemote-file.go
mohsha-msft: Yeah this needs to be removed. Thanks for pointing.

zezha-msft:
@mohsha-msft to be clear, we need to be extra careful when making a breaking change, since existing users would have made assumptions that we would fail the transfer if some part of the transfer wasn't done correctly.

Here is my question: how does the user notice that their tier wasn't set properly?

ste/xfer-anyToRemote-file.go
zezha-msft: It's still here though? We didn't perform the set tier operation here, so it doesn't make much sense.

Issue #971, opened 2020-04-21
Impossibility to scan a directory in a public storage account


NicolasDorier:
My account is public.
```
./azcopy cp "https://$AZURE_ACCOUNT_NAME.blob.core.windows.net/$AZURE_STORAGE_CONTAINER/$DIRECTORY_NAME/*" "dist"
INFO: Scanning...

failed to perform copy command due to error: no SAS token or OAuth token is present and the resource is not public
```
Version `azcopy version 10.4.0`.

It worked long ago... unsure when it got broken.

## How to reproduce

This works:
https://btcpayservertravis.blob.core.windows.net/dist?prefix=dist-618031009&restype=container&comp=list

This does not work:
```
./azcopy cp "https://btcpayservertravis.blob.core.windows.net/dist/dist-618031009/*" "dist"
```

JohnRusk:
Thanks for logging this.

nakulkar-msft:
Fixed in 10.6.1

PR #1165, opened 2020-09-01
[AzCopyV10][StgExp][Github971] Cannot scan public folder with /* when performing copy

mohsha-msft:
Issue:
https://github.com/Azure/azure-storage-azcopy/issues/971
https://github.com/microsoft/AzureStorageExplorer/issues/3394

Unable to scan virtual directories of a public container.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -205,7 +205,7 @@ func (raw rawCopyCmdArgs) stripTrailingWildcardOnRemoteSource(location common.Lo
 			return
 		}
 
-		// Return immediately here because we know this'll be safe.
+		// Return immediately here because we know this will be safe.
 		return
 	}
 
@@ -242,7 +242,7 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 	tempDest := raw.dst
 
 	if strings.EqualFold(tempDest, common.Dev_Null) && runtime.GOOS == "windows" {
-		tempDest = common.Dev_Null // map all capitalizations of "NUL"/"nul" to one because (on Windows) they all mean the same thing
+		tempDest = common.Dev_Null // map all capitalization of "NUL"/"nul" to one because (on Windows) they all mean the same thing
 	}
 
 	// Check if source has a trailing wildcard on a URL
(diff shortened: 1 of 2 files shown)


cmd/credentialUtil.go
```diff
@@ -125,6 +126,18 @@ func getBlobCredentialType(ctx context.Context, blobResourceURL string, canBePub
 				isPublicResource = true
 			}
 		} else {
+			// Scenario 2: When resourceURL points to a virtual directory.
+			// Check if the virtual directory is accessible by doing GetProperties on container.
+			// Virtual directory can be accessed/scanned only when its parent container is public.
+			bURLParts := azblob.NewBlobURLParts(*resourceURL)
+			bURLParts.BlobName = ""
+			containerURL := azblob.NewContainerURL(bURLParts.URL(), p)
+			if _, err := containerURL.GetProperties(ctx, azblob.LeaseAccessConditions{}); err == nil {
```
nakulkar-msft, line 135: This statement is repeated in both the blocks. Scenario1 and scenario2 are the same, and should be merged.  A virtual dir can be public if container is public.

mohsha-msft: Done. I concur that code duplicity can be reduced here. I've extracted out the common part of both if...else flow and for scenario 3 only I'm calling blob.GetProperties().

cmd/copy.go
```diff
@@ -242,7 +242,7 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 	tempDest := raw.dst
 
 	if strings.EqualFold(tempDest, common.Dev_Null) && runtime.GOOS == "windows" {
-		tempDest = common.Dev_Null // map all capitalizations of "NUL"/"nul" to one because (on Windows) they all mean the same thing
+		tempDest = common.Dev_Null // map all capitalization of "NUL"/"nul" to one because (on Windows) they all mean the same thing
```
adreed-msft, line 245: "capitalizations" was correct here. On a side note, perhaps this is not the PR to run around correcting comment grammar in unrelated files... Though, we should certainly go through the code and ensure all comments make sense at some point.

adreed-msft (approved):

adreed-msft:
Nice job, this closely matches the semantics of the service side (where it's either the entire container, or just the blobs that are public).

cmd/copy.go
nakulkar-msft: Adding to adreed's comment, changing the comments will alter the change description of the line. Change descriptions are helpful to know the history and reason for a change. I'd prefer if this file were unchanged.

nakulkar-msft (approved):

PR #1167, opened 2020-09-01
Auto login for Blob and BlobFS endpoints

nakulkar-msft:


nakulkar-msft:
Question: Should I just get rid of AutoLogin() cred type.?  Might help in a bit of debugging I assumed.

cmd/jobsResume.go
```diff
@@ -286,7 +286,8 @@ func (rca resumeCmdArgs) process() error {
 		destinationSAS: rca.DestinationSAS,
 	}); err != nil {
 		return err
-	} else if credentialInfo.CredentialType == common.ECredentialType.OAuthToken() {
+	} else if credentialInfo.CredentialType == common.ECredentialType.OAuthToken() ||
+		credentialInfo.CredentialType == common.ECredentialType.Anonymous() {
```
mohsha-msft, line 290: Why Anonymous?

cmd/credentialUtil.go
```diff
@@ -240,7 +284,7 @@ func GetCredTypeFromEnvVar() common.CredentialType {
 	}
 
 	// Remove the env var after successfully fetching once,
-	// in case of env var is further spreading into child processes unexpectedly.
+	// in case of env var is further spreading into child processes unexpectly.
```
mohsha-msft, line 287: It should be unexpectedly

mohsha-msft:
I don't see any use case of `common.ECredentialType.AutoLogin()` as we can do the same with `common.ECredentialType.OAuthToken()`. Do we have any other way of finding out whether it was one-shot login or not? Perhaps a log. I may be wrong about this. What do you think @nakulkar-msft ?

cmd/credentialUtil.go
```diff
@@ -186,7 +229,8 @@ func getBlobFSCredentialType(ctx context.Context, blobResourceURL string, standa
 	if name != "" && key != "" { // TODO: To remove, use for internal testing, SharedKey should not be supported from commandline
 		return common.ECredentialType.SharedKey(), nil
 	} else {
-		return common.ECredentialType.Unknown(), errors.New("OAuth token, SAS token, or shared key should be provided for Blob FS")
+		/* We dont return error here. The caller can either fail or attempt auto-login */
+		return common.ECredentialType.Unknown(), nil
```
adreed-msft, line 233: This is also called in cmd/make.go-- To avoid breaking compatibility, perhaps we should return an additional flag, or make the error static and check for it. Alternatively, we can implement auto-login in create as well (Which probably has the higher engineering cost of the two options here)

For instance, `var blobFSNoAuthError = errors.New("Not enough authentication detail for blobFS")` (not withholding the actual phrasing here), then when we are attempting auto-login, we should check if the resulting error is this one.

cmd/credentialUtil.go
```diff
@@ -140,7 +182,8 @@ func getBlobCredentialType(ctx context.Context, blobResourceURL string, canBePub
 
 		// No forms of auth are present.
 		if !isPublicResource {
-			return common.ECredentialType.Unknown(), isPublicResource, errors.New("no SAS token or OAuth token is present and the resource is not public")
+			/* We dont return error here, the caller can either auto-login or fail */
```
adreed-msft, line 185: Ditto to the last get*CrendentialType comment potentially as well

cmd/jobsResume.go
adreed-msft: I think this smells a bit fishy as well... SAS tokens are anonymous, and get upset when you combine them with OAuth.

adreed-msft: @nakulkar-msft did you by chance mean AutoLogin here?

cmd/credentialUtil.go
adreed-msft: I noticed you did change make.go to appropriately handle this-- I do think though, that the marked error would be best, that way, it's not a bit of "mental overhead" when we write new code with this.

common/environment.go
```diff
@@ -64,6 +64,15 @@ func (EnvironmentVariable) UserDir() EnvironmentVariable {
 	}
 }
 
+func (EnvironmentVariable) ApplicationID() EnvironmentVariable {
```
adreed-msft, line 67: I'm somewhere in between "this should be a environment variable" and "this should be additional command line parameters under cmd/root.go".

In my imagination, when the thought of "auto login" came along, my assumption was that we'd perform the login in root.go, that way it'd be automatically implemented for whatever new feature, a la "Write once, use it a million times".

adreed-msft (dismissed):

cmd/jobsResume.go
nakulkar-msft: Yes. I blame IDE :(

cmd/copy.go
```diff
@@ -1053,7 +1053,8 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 
 	// For OAuthToken credential, assign OAuthTokenInfo to CopyJobPartOrderRequest properly,
 	// the info will be transferred to STE.
-	if cca.credentialInfo.CredentialType == common.ECredentialType.OAuthToken() {
+	if cca.credentialInfo.CredentialType == common.ECredentialType.OAuthToken() ||
+		cca.credentialInfo.CredentialType == common.ECredentialType.AutoLogin() {
```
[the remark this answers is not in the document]
nakulkar-msft: I sort of convinced myself that the new enum is just noise. All we're doing is OAuth, except for persisting on disk. Enough logging should be okay instead of repeating case statements twice for OAuth and Auto all through the code.

cmd/credentialUtil.go
nakulkar-msft: AzCopy errors are not helpful to compare. I took an easy way out I guess. Changin this.

common/environment.go
nakulkar-msft: I'm sure this has to be set in the env. I'd say set these var, and your script will not have to change or fight with keyctl for login. I did not want to change anything in FE for this. Also, this is more a static variable, unlike 'cap-mbps' or 'log-level' which are tuned for  every new instance of AzCopy.

root.go seems tempting. But at that point I'm sure we'd  not differentiate source/dst  or if one of them have a SAS. Not sure about this though, Auto should be last attempt at Authentication right? After we've explored all of SAS, saved OAuth...

common/azError.go
```diff
@@ -0,0 +1,52 @@
+// Copyright © 2020 Microsoft <wastore@microsoft.com>
```
zezha-msft, line 1: >2020 [](start = 15, length = 4)

Minor: no need to include year for new files.

common/environment.go
```diff
@@ -80,6 +95,31 @@ func (EnvironmentVariable) CertificatePassword() EnvironmentVariable {
 	}
 }
 
+// For MSI login
+func (EnvironmentVariable) IdentityClientID() EnvironmentVariable {
```
zezha-msft, line 99: >IdentityClientID [](start = 27, length = 16)

Minor: consider prefixing these with "Managed", e.g. ManagedIdentityClientID

common/environment.go
```diff
@@ -80,6 +95,31 @@ func (EnvironmentVariable) CertificatePassword() EnvironmentVariable {
...
+		Description: "Client ID for User-assigned identity",
+		Hidden:      true,
+	}
+}
+
+func (EnvironmentVariable) IdentityObjectID() EnvironmentVariable {
+	return EnvironmentVariable{
+		Name:        "AZCOPY_MSI_OBJECT_ID",
+		Description: "Object ID for user-assigned identity",
+		Hidden:      true,
```
zezha-msft, line 111: >true [](start = 15, length = 4)

Why hidden? Just too many variables?

zezha-msft (approved):
:shipit:

common/environment.go
nakulkar-msft: Copy paste error

mohsha-msft (approved):
Looks Good 👍

cmd/credentialUtil.go
```diff
@@ -69,6 +71,47 @@ func GetUserOAuthTokenManagerInstance() *common.UserOAuthTokenManager {
...
+		if glcm.GetEnvironmentVariable(common.EEnvironmentVariable.AutoLoginType()) == "" {
+			err = errors.New("no login type specified")
+			return
+		}
+		// Fill up lca
+		switch glcm.GetEnvironmentVariable(common.EEnvironmentVariable.AutoLoginType()) {
+		case "SPN":
+			lca.applicationID = glcm.GetEnvironmentVariable(common.EEnvironmentVariable.ApplicationID())
+			lca.certPath = glcm.GetEnvironmentVariable(common.EEnvironmentVariable.CertificatePassword())
+			lca.clientSecret = glcm.GetEnvironmentVariable(common.EEnvironmentVariable.ClientSecret())
```
adreed-msft, line 91: Oops! This isn't correct. CertificatePassword is not the path of the certificate. We'll need a new env var.

Note: CertificatePassword is a separate parameter, in case the user passes in a encrypted certificate.

adreed-msft: Try looking into the normal flow for a Service Principal login, it should help you understand what needs to happen here.

---
In reply to: [517100301](https://github.com/Azure/azure-storage-azcopy/pull/1167#discussion_r517100301) [](ancestors = 517100301)

cmd/credentialUtil.go
```diff
@@ -403,15 +448,31 @@ func doGetCredentialTypeForLocation(ctx context.Context, location common.Locatio
 		case common.ELocation.Local(), common.ELocation.Benchmark():
 			credType = common.ECredentialType.Anonymous()
 		case common.ELocation.Blob():
-			if credType, isPublic, err = getBlobCredentialType(ctx, resource, isSource, resourceSAS != ""); err != nil {
+			credType, isPublic, err = getBlobCredentialType(ctx, resource, isSource, resourceSAS != "")
+			if azErr, ok := err.(common.AzError); ok && azErr.Equals(common.EAzError.LoginCredMissing()) {
```
adreed-msft, line 452: >if azErr, ok := err.(common.AzError); ok && azErr.Equals(common.EAzError.LoginCredMissing()) { [](start = 3, length = 94)

I like this enumeration of errors. It'll help us compare errors a bit more. IIRC, there were a couple of errors regarding fixes that I made that I think should be merged into this model.

common/environment.go
adreed-msft: I think that this might potentially be confusing for the end-user to have both the Application ID environment variable and the login parameter application ID. At least make them interchangeable on login and I think it'll be OK.

---
In reply to: [484205933](https://github.com/Azure/azure-storage-azcopy/pull/1167#discussion_r484205933) [](ancestors = 484205933)

cmd/credentialUtil.go
```diff
@@ -69,6 +71,47 @@ func GetUserOAuthTokenManagerInstance() *common.UserOAuthTokenManager {
...
+ */
+func GetOAuthTokenManagerInstance() (*common.UserOAuthTokenManager, error) {
+	var err error
+	autoOAuth.Do(func() {
+		var lca loginCmdArgs
+		if glcm.GetEnvironmentVariable(common.EEnvironmentVariable.AutoLoginType()) == "" {
+			err = errors.New("no login type specified")
+			return
+		}
+		// Fill up lca
```
adreed-msft, line 86: I think it would be safe to assume off environment variables (such as ApplicationID, CertificatePath,  or something else. Note that no environment variables are shared between SPN, MSI, or device (which uses none anyway).

adreed-msft (changes requested):
:clock1:

adreed-msft:
There's a pretty big oopsie in there and something that I feel is bound to cause end user confusion.

common/environment.go
zezha-msft: @nakulkar-msft  maybe we can add in the description of these env variables some thing like "This only takes effect when AZCOPY_LOGIN_TYPE is set.

PR #1179, opened 2020-09-18
[AzCopyV10] Content-type for auto detected types should be set in the create call as well for 0 length files

mohsha-msft:
AzCopy always sets content-type of a zero-bytes file to upload as “application/octet-stream” regardless of its extension name while other tools or APIs don’t.  Content-type is supposed to be determined by an extension name of a file to upload if it is failed to be detected by a body of the file. Even for an empty file, the content-type should be set based on the extension name.

### DIFF
--- a/ste/sender-pageBlob.go
+++ b/ste/sender-pageBlob.go
@@ -223,11 +223,9 @@ func (s *pageBlobSenderBase) Prologue(ps common.PrologueState) (destinationModif
 		destinationModified = true
 	}
 
-	if ps.CanInferContentType() {
-		// sometimes, specifically when reading local files, we have more info
-		// about the file type at this time than what we had before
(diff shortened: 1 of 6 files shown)


ste/sender-pageBlob.go
```diff
@@ -223,11 +223,9 @@ func (s *pageBlobSenderBase) Prologue(ps common.PrologueState) (destinationModif
 		destinationModified = true
 	}
 
-	if ps.CanInferContentType() {
```
zezha-msft, line 226: CanInferContentType seems useless now, right? Please remove it if appropriate.

zezha-msft (approved):

ste/sender-pageBlob.go
mohsha-msft: Done

PR #1184, opened 2020-09-24
Figuring out source/destination on the basis of from-to flag rather than stdPipeIn()

mohsha-msft:
If there is one location present

- If from-to flag is present
   - PipeBlob (Upload)
       - Check for input pipe. If input pipe is missing or parsing input fails, throw error. Else set source/destination.
   - BlobPipe (Download)
      - If output pipe is missing, content will be echoed on the terminal. Else set source/destination.
- Else try to figure out source and destination from os.Stdin.Stat()/os.Stdout.Stat() and proceed like existing flow.


cmd/copy.go
```diff
@@ -1480,17 +1480,31 @@ func init() {
...
+						return fmt.Errorf("fatal: incorrect from-to argument passed: %s", raw.fromTo)
 					}
 				}
+
+				stdinPipeIn, err := isStdinPipeIn()
+				if err != nil {
+					return fmt.Errorf("fatal: failed to read from Stdin due to error: %s", err)
+				}
+
+				if (userFromTo == common.EFromTo.PipeBlob() || userFromTo == common.EFromTo.Unknown()) && stdinPipeIn == true {
```
adreed-msft, line 1498: >if (userFromTo == common.EFromTo.PipeBlob() || userFromTo == common.EFromTo.Unknown()) && stdinPipeIn == true { [](start = 4, length = 111)

I don't believe this covers the "if both pipes are in with no fromto, error out" scenario.

cmd/copy.go
```diff
@@ -1480,17 +1480,31 @@ func init() {
...
+					raw.src = pipeLocation
+					raw.dst = args[0]
+				} else if userFromTo == common.EFromTo.BlobPipe() || userFromTo == common.EFromTo.Unknown() {
+					// In case of BlobPage, if pipe is missing, content will be echoed on the terminal
```
adreed-msft, line 1502: >BlobPage [](start = 19, length = 8)

BlobPipe*

cmd/copy.go
adreed-msft: Although, since it doesn't look like we're actually checking for stdout as a pipe... I don't think we can do that.

---
In reply to: [494506908](https://github.com/Azure/azure-storage-azcopy/pull/1184#discussion_r494506908) [](ancestors = 494506908)

adreed-msft: I think the best option is to just error out if we don't know the fromto in this case, and publicly document BlobPipe and PipeBlob.

---
In reply to: [494507871](https://github.com/Azure/azure-storage-azcopy/pull/1184#discussion_r494507871) [](ancestors = 494507871,494506908)

adreed-msft: though that might be a breaking change...

---
In reply to: [494508421](https://github.com/Azure/azure-storage-azcopy/pull/1184#discussion_r494508421) [](ancestors = 494508421,494507871,494506908)

mohsha-msft: Enforcing `from-to` flag. 
Validated cases
-  Upload/Download when no `from-to` flag provided => `fatal: from-to flag required`
-  Upload/Download when invalid `from-to` flag provided (BlobTrash, LocalBlob) => `fatal: invalid from-to flag argument passed: BlobTrash`
- Upload/Download when valid `from-to` flag provided (BlobPipe, PipeBlob) => Success
- Upload when invalid stdPipeIn provided => `fatal: failed to read from Stdin due to error` 
- Download when no stdPipeOut provided => Success. Echoed content on the terminal

cmd/copy.go
mohsha-msft: Done

PR #1187, opened 2020-09-25
Piping fix

mohsha-msft:


azure-pipelines.yml
```diff
@@ -76,7 +76,7 @@ jobs:
           # -timeout 25m as our tests typically run between 10-15m and I'd rather have some headroom.
           # "-check.v" (must be after package list) outputs timings
           set -e
-          go test -timeout 25m -race -short -cover ./cmd ./common ./common/parallel ./ste ./azbfs ./sddl "-check.v"
+          go test -timeout 35m -race -short -cover ./cmd ./common ./common/parallel ./ste ./azbfs ./sddl "-check.v"
```
adreed-msft, line 79: >35m [](start = 27, length = 3)

What prompted the increase in test run time?

mohsha-msft: @nakulkar-msft I know this in bits and pieces. Can you answer this?

cmd/copy.go
```diff
@@ -1480,16 +1480,28 @@ func init() {
 		Example:    copyCmdExample,
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 1 { // redirection
-				if stdinPipeIn, err := isStdinPipeIn(); stdinPipeIn == true {
+				// Enforce the usage of from-to flag when pipes are involved
+				if raw.fromTo == "" {
+					return fmt.Errorf("fatal: from-to argument required")
```
adreed-msft, line 1485: >return fmt.Errorf("fatal: from-to argument required") [](start = 5, length = 53)

I'd say to avoid confusion, we should add "when piping" to the end.

mohsha-msft: Sure Adele. Private drop is already out there so we cannot change that but since we are planning to do these changes in the actual flow as well, I'll add "while piping" to the message. 👍

cmd/zt_copy_s2smigration_test.go
```diff
@@ -637,6 +637,7 @@ func (s *cmdIntegrationSuite) TestS2SCopyFromContainerToContainerNoPreserveBlobT
 
 //Attempt to copy from a page blob to a block blob
 func (s *cmdIntegrationSuite) TestS2SCopyFromPageToBlockBlob(c *chk.C) {
+	c.Skip("Enable after setting Account to non-HNS")
 	bsu := getBSU()
```
adreed-msft, line 641: Did we change the test storage account to HNS? I recall that accessing account.dfs.core.windows.net from an account without HNS enabled functioned similarly.

Additionally, I feel like it's not a great idea to disable these things in this fashion-- It's easy to forget about.

PR #1193, opened 2020-10-06
Enable OAuth on blob piping

adreed-msft:
Previously, piped blob upload and download only allowed for anonymous forms of auth-- e.g. public auth or SAS

With this PR existing OAuth support for the destination is added. Note that because getCredentialTypeForLocation calls checkAuthSafeForTarget, it isn't necessary to add the additional check here.


mohsha-msft (approved):
LGTM

PR #1194, opened 2020-10-06
Fix OAuth on source of sync

adreed-msft:
@nakulkar-msft identified a regression in Sync at 10.5.0 where users could no longer do an OAuth download sync as of #970. This PR fixes said regression.


nakulkar-msft (approved):

mohsha-msft (approved):
LGTM

PR #1212, opened 2020-10-16
Retry copy to Archived Blob after deleting it

nakulkar-msft:


zezha-msft (approved):
Lgtm, although I must say the jptm reset func makes me a bit nervous. Please double check that the job reporting logic still works as expected.

zezha-msft:
As discussed offline this feature will not be published.

PR #1221, opened 2020-10-21
UpdateProgessPanic: Reduce severity to logging

nakulkar-msft:
There can be a race such that the file is transferred, but before we report the jptm to be done, the context of this jptm can be cancelled. This is to acknowledge the race.

### DIFF
--- a/ste/mgr-JobPartMgr.go
+++ b/ste/mgr-JobPartMgr.go
@@ -670,7 +670,7 @@ func (jpm *jobPartMgr) updateJobPartProgress(status common.TransferStatus) {
 		atomic.AddUint32(&jpm.atomicTransfersSkipped, 1)
 	case common.ETransferStatus.Cancelled():
 	default:
-		panic("Unexpected status")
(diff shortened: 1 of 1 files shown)


zezha-msft (approved):

ste/mgr-JobPartMgr.go
```diff
@@ -670,7 +670,7 @@ func (jpm *jobPartMgr) updateJobPartProgress(status common.TransferStatus) {
 		atomic.AddUint32(&jpm.atomicTransfersSkipped, 1)
 	case common.ETransferStatus.Cancelled():
 	default:
-		panic("Unexpected status")
+		jpm.Log(pipeline.LogError, fmt.Sprintf("Unexpected status: %v", status.String()))
```
zezha-msft, line 673: This does avoid the panic, but is this log even useful?

nakulkar-msft: If there are other races or changes which would lead us to this scenario, we'd want to know.

PR #1228, opened 2020-10-26
Adding frontend logging

zezha-msft:
Two open questions:

1. Do we want to prompt the user about the scanning log file?
2. Do we want to have a separate flag to control the scanning log level? Right now it's 1 knob for both the frontend and backend logging.


cmd/copyEnumeratorInit.go
```diff
@@ -54,7 +54,7 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 	jobPartOrder.DestLengthValidation = cca.CheckLength
 	jobPartOrder.S2SInvalidMetadataHandleOption = cca.s2sInvalidMetadataHandleOption
 
-	traverser, err = initResourceTraverser(cca.source, cca.fromTo.From(), &ctx, &srcCredInfo, &cca.followSymlinks, cca.listOfFilesChannel, cca.recursive, getRemoteProperties, cca.includeDirectoryStubs, func(common.EntityType) {}, cca.listOfVersionIDs)
+	traverser, err = initResourceTraverser(cca.source, cca.fromTo.From(), &ctx, &srcCredInfo, &cca.followSymlinks, cca.listOfFilesChannel, cca.recursive, getRemoteProperties, cca.includeDirectoryStubs, func(common.EntityType) {}, cca.listOfVersionIDs, cca.logVerbosity.ToPipelineLogLevel())
```
mohsha-msft, line 57: I small suggestion:
I think we should consider creating an option bag similar to track2 here. The number of parameters are increasing and with more features like versioning and CPK coming to AzCopy, we cannot keep on updating the definition.
Also, while passing some attributes make sense while calling initResourceTraverser() 
with one operation (copy for example), passing those attributes with some other operation(list for example) may not even make sense (even passing a nil).

Alternatively, we can pass the whole `cooked` struct and get rid of changing function signature in one go.

What do you think?

mohsha-msft (approved):
Looks good to me.

nakulkar-msft (approved):

nakulkar-msft:
> Two open questions:
> 
> 1. Do we want to prompt the user about the scanning log file?
> 2. Do we want to have a separate flag to control the scanning log level? Right now it's 1 knob for both the frontend and backend logging.

I'm of opinion that we should not prompt. But allow them to change, like we're doing now. Also, 1 knob for both should be fine. It'll be too verbose, and might slow down, but the usual recommendation si to turn them off.

mohsha-msft:
> > Two open questions:
> > 
> > 1. Do we want to prompt the user about the scanning log file?
> > 2. Do we want to have a separate flag to control the scanning log level? Right now it's 1 knob for both the frontend and backend logging.
> 
> I'm of opinion that we should not prompt. But allow them to change, like we're doing now. Also, 1 knob for both should be fine. It'll be too verbose, and might slow down, but the usual recommendation si to turn them off.

Yep, I concur with the idea as well. Nicely put @nakulkar-msft 👍.
As a customer, I may not care about components of AzCopy (front-end, back-end) but AzCopy as a whole. Having one more prompt or additional flag to control the logging will attribute to bad customer experience and also more documentation explaining how the logging works in front-end & back-end.

cmd/copyEnumeratorInit.go
zezha-msft: Ok let me follow up with a separate PR proposal.

PR #1234, opened 2020-10-30
go fmt and removed unnecessary else blocks (if block ends with a retu…

SimonWaldherr:
go fmt and removed unnecessary else blocks (if block ends with a return statement, so drop the else and outdent its block)


zezha-msft (approved):

cmd/jobsResume.go
```diff
@@ -291,9 +289,8 @@ func (rca resumeCmdArgs) process() error {
 		// Get token from env var or cache.
 		if tokenInfo, err := uotm.GetTokenInfo(ctx); err != nil {
 			return err
-		} else {
-			credentialInfo.OAuthTokenInfo = *tokenInfo
 		}
+		credentialInfo.OAuthTokenInfo = *tokenInfo
```
mohsha-msft, line 293: This one right here is a syntax error. `tokenInfo` is a local variable whose scope resides in the `if` statement above. it cannot be used outside `if`

mohsha-msft: Hey @Simon, 

You can run `go build` & `go build ./e2etest` to check for syntax errors.

PR #1236, opened 2020-11-03
Content Type Bug (Preserve content type in S2S transfer + Upload 0B files and infer content type based on extension)

mohsha-msft:
Scenarios to consider
- Remote source: Content Type should always be preserved.
- Local File:
    -  Files with size less than ``512 bytes`` including ``0 byte``.
    -  File with size greater than or equal to ``512 bytes``.

FYI: How Infer Content Type works for local files
1. Try to infer content type on the basis of file extension. 
    1. [Get file extension](https://golang.org/pkg/path/filepath/#Ext).
    2. [Get MIME type using file extension](https://golang.org/pkg/mime/#TypeByExtension)
2. If above step fails, try to [get the content type by considering at most the first 512 bytes of data](https://golang.org/pkg/net/http/#DetectContentType).

### DIFF
--- a/ste/mgr-JobPartTransferMgr.go
+++ b/ste/mgr-JobPartTransferMgr.go
@@ -86,6 +86,7 @@ type IJobPartTransferMgr interface {
 	SecurityInfoPersistenceManager() *securityInfoPersistenceManager
 	FolderDeletionManager() common.FolderDeletionManager
 	GetDestinationRoot() string
+	CanInferContentType() bool
 }
 
 type TransferInfo struct {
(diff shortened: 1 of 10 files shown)


ste/mgr-JobPartTransferMgr.go
```diff
@@ -881,3 +882,10 @@ func (jptm *jobPartTransferMgr) GetDestinationRoot() string {
 	p := jptm.jobPartMgr.Plan()
 	return string(p.DestinationRoot[:p.DestinationRootLength])
 }
+
+func (jptm *jobPartTransferMgr) CanInferContentType() bool {
```
zezha-msft, line 886: Minor: it's more like ShouldInferContentType

zezha-msft (approved):

ste/mgr-JobPartTransferMgr.go
mohsha-msft: Yup you're right. Will do the changes

adreed-msft (approved):
:shipit:

PR #1258, opened 2020-11-30
Fixed share snapshot->share copy with smb permissions

zezha-msft:
Missing: need to add an e2etest


mohsha-msft (approved):
Changes look good. Nice addition to E2E Test Framework

nakulkar-msft (approved):

PR #1263, opened 2020-12-03
Offer option to disable parallel blob listing

zezha-msft:


nakulkar-msft (approved):

mohsha-msft (approved):
Changes look good to me. 👍

PR #1265, opened 2020-12-03
Adding support for component governance tool

zezha-msft:


zezha-msft:
CP tool does not work with Go components

PR #1280, opened 2020-12-14
[AzCopyV10][StgExp] Support preservation of tags when copying blobs

mohsha-msft:
- Support preservation of blob index tags when transferring blobs from one blob storage to another.
- To preserve tags, use the following syntax 
    - `azcopy cp "https://[account].blob.core.windows.net/[source_container]/[path/to/directory]?[SAS]" "https://[account].blob.core.windows.net/[destination_container]/[path/to/directory]?[SAS]" --s2s-preserve-blob-tags=true`
- To read more about blob indexing, refer the public documentation [here](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-index-how-to?tabs=azure-portal).
- While setting tags on the blobs, there are additional permissions('t' for tags) in SAS without which the service will give authorization error back.


cmd/syncEnumerator.go
```diff
@@ -62,7 +62,7 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 		if entityType == common.EEntityType.File() {
 			atomic.AddUint64(&cca.atomicSourceFilesScanned, 1)
 		}
-	}, nil)
+	}, nil, false)
```
zezha-msft, line 65: >false [](start = 9, length = 5)

We should perhaps support preserving tags for sync.

cmd/zc_traverser_blob.go
```diff
@@ -50,6 +50,9 @@ type blobTraverser struct {
 
 	// a generic function to notify that a new stored object has been enumerated
 	incrementEnumerationCounter enumerationCounterFunc
+
+	// When user pro
```
zezha-msft, line 54: >// When user pro [](start = 1, length = 16)

unfinished comment?

cmd/zc_traverser_blob.go
```diff
@@ -95,6 +98,26 @@ func (t *blobTraverser) getPropertiesIfSingleBlob() (props *azblob.BlobGetProper
...
+func (t *blobTraverser) getBlobTags(requestID *string, snapshot *string, versionID *string, ifTags *string) (common.BlobTags, error) {
+	// trim away the trailing slash before we check whether it's a single blob
+	// so that we can detect the directory stub in case there is one
+	blobUrlParts := azblob.NewBlobURLParts(*t.rawURL)
+	blobUrlParts.BlobName = strings.TrimSuffix(blobUrlParts.BlobName, common.AZCOPY_PATH_SEPARATOR_STRING)
+
+	// perform the check
+	blobURL := azblob.NewBlobURL(blobUrlParts.URL(), t.p)
+	blobTagsMap := make(common.BlobTags)
+	blobGetTagsResp, err := blobURL.GetTags(t.ctx, nil, requestID, snapshot, versionID, ifTags)
```
zezha-msft, line 110: >snapshot, versionID [](start = 64, length = 19)

shouldn't snapshot and version ID be part of the blob URL already? Why do we need to pass them separately?

cmd/zc_traverser_blob.go
```diff
@@ -134,6 +157,18 @@ func (t *blobTraverser) traverse(preprocessor objectMorpher, processor objectPro
 			blobUrlParts.ContainerName,
 		)
 
+		if t.s2sPreserveSourceTags {
+			blobTagsMap := common.BlobTags{}
```
zezha-msft, line 161: >blobTagsMap := common.BlobTags{} [](start = 3, length = 32)

this is a duplication of what getBlobTags already does

cmd/zc_traverser_blob.go
```diff
@@ -180,7 +215,7 @@ func (t *blobTraverser) parallelList(containerURL azblob.ContainerURL, container
 		currentDirPath := dir.(string)
 		for marker := (azblob.Marker{}); marker.NotDone(); {
 			lResp, err := containerURL.ListBlobsHierarchySegment(t.ctx, marker, "/", azblob.ListBlobsSegmentOptions{Prefix: currentDirPath,
-				Details: azblob.BlobListingDetails{Metadata: true}})
+				Details: azblob.BlobListingDetails{Metadata: true, Tags: true}})
```
zezha-msft, line 218: >Tags: true [](start = 55, length = 10)

Including tags may incur a perf hit, maybe we should only do it if we need the tag info.

cmd/zc_traverser_list.go
```diff
@@ -104,7 +104,7 @@ func newListTraverser(parent common.ResourceString, parentType common.Location,
 		}
 
 		// Construct a traverser that goes through the child
-		traverser, err := initResourceTraverser(source, parentType, ctx, credential, &followSymlinks, nil, recursive, getProperties, includeDirectoryStubs, incrementEnumerationCounter, nil)
+		traverser, err := initResourceTraverser(source, parentType, ctx, credential, &followSymlinks, nil, recursive, getProperties, includeDirectoryStubs, incrementEnumerationCounter, nil, false)
```
zezha-msft, line 107: >false [](start = 184, length = 5)

we may need to pass the value, in case of account->account transfer

cmd/syncEnumerator.go
mohsha-msft: That's a big TODO. I'm working on it.

mohsha-msft: Turned out it wasn't such a big change. Done

cmd/zc_traverser_blob.go
```diff
@@ -95,6 +97,26 @@ func (t *blobTraverser) getPropertiesIfSingleBlob() (props *azblob.BlobGetProper
 	return nil, false, false, err
 }
 
+func (t *blobTraverser) getBlobTags(requestID *string, ifTags *string) (common.BlobTags, error) {
+	// trim away the trailing slash before we check whether it's a single blob
+	// so that we can detect the directory stub in case there is one
```
zezha-msft, line 102: correct comment

cmd/zc_traverser_blob.go
```diff
@@ -95,6 +97,26 @@ func (t *blobTraverser) getPropertiesIfSingleBlob() (props *azblob.BlobGetProper
...
+	blobUrlParts := azblob.NewBlobURLParts(*t.rawURL)
+	blobUrlParts.BlobName = strings.TrimSuffix(blobUrlParts.BlobName, common.AZCOPY_PATH_SEPARATOR_STRING)
+
+	// perform the check
+	blobURL := azblob.NewBlobURL(blobUrlParts.URL(), t.p)
+	blobTagsMap := make(common.BlobTags)
+	blobGetTagsResp, err := blobURL.GetTags(t.ctx, nil, requestID, nil, nil, ifTags)
```
zezha-msft, line 109: >blobGetTagsResp, err := blobURL.GetTags(t.ctx, nil, requestID, nil, nil, ifTags) [](start = 1, length = 80)

Oh wait, GetTags is taking in timeout *int32, requestID *string, snapshot *string, versionID *string, this doesn't look right, since the timeout/requestID are not supposed to be exposed, and snapshot/versionID are on the URL.

zezha-msft: This should be fixed in the SDK.

---
In reply to: [569209305](https://github.com/Azure/azure-storage-azcopy/pull/1280#discussion_r569209305) [](ancestors = 569209305)

cmd/zc_traverser_blob.go
```diff
@@ -95,6 +97,26 @@ func (t *blobTraverser) getPropertiesIfSingleBlob() (props *azblob.BlobGetProper
 	return nil, false, false, err
 }
 
+func (t *blobTraverser) getBlobTags(requestID *string, ifTags *string) (common.BlobTags, error) {
```
zezha-msft, line 100: >(requestID *string, ifTags *string [](start = 35, length = 34)

we don't ever pass in any request ID or ifTags right?

cmd/zc_traverser_blob.go
mohsha-msft: Yeah the enlightenment about not exposing RequestID/Timeout came to me pretty recently. I'll fix this in SDK.

e2etest/zt_blob_index_tags_test.go
```diff
@@ -0,0 +1,176 @@
...
+		eValidate.AutoPlusContent(),
+		params{
+			recursive: true,
+			blobTags:  blobTagsStr,
+		},
+		nil,
+		testFiles{
+			defaultSize: "1M",
+			shouldTransfer: []interface{}{
+				//folder("", ),
```
zezha-msft, line 44: >//folder("", ), [](start = 4, length = 15)

remove commented code

e2etest/zt_blob_index_tags_test.go
```diff
@@ -0,0 +1,176 @@
...
+				f("file1.txt", with{blobTags: "foo=bar&baz=blah&YeAr=2020"}),
+				f("fdlr1/file2.txt", with{blobTags: "temp123=321pmet&zab=halb&rAeY=0202"}),
+			},
+		})
+}
+
+func TestTags_PreserveTagsSpecialCharactersMultipleBlobs(t *testing.T) {
+	RunScenarios(
+		t,
+		eOperation.Copy(),
```
zezha-msft, line 160: >eOperation.Copy(), [](start = 2, length = 18)

enable tests for sync as well

zezha-msft (changes requested):
Looks good in general, some minor changes needed.

e2etest/zt_blob_index_tags_test.go
mohsha-msft: We have it `TestTags_PreserveTagsSpecialCharactersDuringSync`  below this one.

e2etest/zt_blob_index_tags_test.go
mohsha-msft: Done

cmd/zc_traverser_blob.go
mohsha-msft: https://github.com/Azure/azure-storage-blob-go/pull/252 - Here is the fix for Blob SDK

mohsha-msft:
[AzBlob](https://github.com/Azure/azure-storage-blob-go) has [dependencies](https://github.com/Azure/azure-storage-blob-go/blob/fb7ebad3966326dd47a778c3ba8e1a3b10759d67/azblob/zc_policy_unique_request_id.go#L24) on Go Version `1.15` . So upgrading from `1.13` -> `1.15`
[AzFile](https://github.com/Azure/azure-storage-file-go) is also using Go version `1.15`

nakulkar-msft (approved):

cmd/copy.go
```diff
@@ -509,6 +511,19 @@ func (raw rawCopyCmdArgs) cookWithId(jobId common.JobID) (cookedCopyCmdArgs, err
 	}
 	cooked.blobTags = blobTags
 
+	// Check if user has provided `s2s-preserve-blob-tags` flag. If yes, we have to ensure that
+	// 1. Both source and destination must be blob storages.
+	// 2. `blob-tags` is not present as they create conflicting scenario of whether to preserve blob tags from the source or set user defined tags on the destination
+	if raw.s2sPreserveBlobTags {
+		if cooked.fromTo.From() != common.ELocation.Blob() || cooked.fromTo.To() != common.ELocation.Blob() {
+			return cooked, errors.New("either source or destination is not a blob storage. blob index tags is a property of blobs only therefore both source and destination must be blob storage")
```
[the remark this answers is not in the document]
adreed-msft: I don't feel like that's necessary, however this can be rephrased though-- "s2s-preserve-blob-tags is only compatible with a blob to blob S2S transfer."

---
In reply to: [568035725](https://github.com/Azure/azure-storage-azcopy/pull/1280#discussion_r568035725) [](ancestors = 568035725)

adreed-msft (approved):
:shipit:

PR #1286, opened 2020-12-16
[AzCopyV10][Feature] Add support for CPK-N and CPK-V

mohsha-msft:
- Clients making requests against Azure Blob storage have the option to provide an encryption key on a per-request basis.
- Including the encryption key on the request provides granular control over encryption settings for Blob storage operations.
-  Customer-provided keys can be stored in Azure Key Vault or in another key store.

[Public Documentation - 1](https://docs.microsoft.com/en-us/azure/storage/blobs/encryption-customer-provided-keys)
[Public Documentation - 2](https://azure.microsoft.com/en-in/blog/customer-provided-keys-with-azure-storage-service-encryption/)

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -144,6 +144,8 @@ type rawCopyCmdArgs struct {
 
 	// whether to include blobs that have metadata 'hdi_isfolder = true'
 	includeDirectoryStubs bool
+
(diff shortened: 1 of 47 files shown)


zezha-msft:
Closing since you have a new PR.

PR #1311, opened 2021-01-17
added date to list output

SharonBrizinov:
Printing also the modified date


zezha-msft (approved):

mohsha-msft (approved):
Looks Fine. 
Last modified time may not be useful to all. 
For better customer experience, we should try to add flag in the command through users can to specify what other fields do they want. Default nature should be to just include file path as we're doing right now.

stonebig:
to comment on @mohsha-msft : for me pushing to azure by script, being able to retreive the "Date Modified" to check the file has been updated, is important.

PR #1036, opened 2020-06-03
Feature/gcp connector

dheeraj-msft:
This will add feature to transfer Objects from Google Cloud Platform to Azure Block Blob Storage
To test, export GOOGLE_APPLICATION_CREDENTIALS="path/to/serviceAccountkey.json"
Eg cmd:
azcopy copy 'https://storage.cloud.google.com/bucket-name/path/to/object' 'destinationURL'

PR #1318, opened 2021-01-21
GCP source for AzCopy

nakulkar-msft:


go.sum
```diff
@@ -1,101 +1,401 @@
...
+cloud.google.com/go v0.57.0 h1:EpMNVUorLiZIELdMZbCYX/ByTFCdoYopYAGxaGVz9ms=
+cloud.google.com/go v0.57.0/go.mod h1:oXiQ6Rzq3RAkkY7N6t3TcE6jE+CIBBbA36lwQ1JyzZs=
+cloud.google.com/go/bigquery v1.0.1/go.mod h1:i/xbL2UlR5RvWAURpBYZTtm/cXjCha9lbfbpx4poX+o=
+cloud.google.com/go/bigquery v1.3.0/go.mod h1:PjpwJnslEMmckchkHFfq+HTD2DmtT67aNFKH1/VBDHE=
+cloud.google.com/go/bigquery v1.4.0/go.mod h1:S8dzgnTigyfTmLBfrtrhyYhwRxG72rYxvftPBK2Dvzc=
+cloud.google.com/go/bigquery v1.5.0/go.mod h1:snEHRnqQbz117VIFhE8bmtwIDY80NLUZUMb4Nv6dBIg=
+cloud.google.com/go/bigquery v1.7.0/go.mod h1://okPTzCYNXSlb24MZs83e2Do+h+VXtc4gLoIoXIAPc=
+cloud.google.com/go/datastore v1.0.0/go.mod h1:LXYbyblFSglQ5pkeyhO+Qmw7ukd3C+pD7TKLgZqpHYE=
+cloud.google.com/go/datastore v1.1.0/go.mod h1:umbIZjpQpHh4hmRpGhH4tLFup+FVzqBi1b3c64qFpCk=
+cloud.google.com/go/pubsub v1.0.1/go.mod h1:R0Gpsv3s54REJCy4fxDixWD93lHJMoZTyQ2kNxGRt3I=
```
zezha-msft, line 22: >cloud.google.com/go/pubsub [](start = 0, length = 26)

changes are required for the NOTICE file

ste/sender-blockBlobFromURL.go
```diff
@@ -163,6 +178,32 @@ func (c *urlToBlockBlobCopier) generatePutBlockFromURL(id common.ChunkID, blockI
 	})
 }
 
+func (c *urlToBlockBlobCopier) generateStartCopyBlobFromURL(id common.ChunkID, blockIndex int32, adjustedChunkSize int64) chunkFunc {
+	return createSendToRemoteChunkFunc(c.jptm, id, func() {
+
+		encodedBlockID := c.generateEncodedBlockID()
+
+		c.setBlockID(blockIndex, encodedBlockID)
```
zezha-msft, line 186: block ID is not needed if we are just calling CopyFromURL, right?

ste/sourceInfoProvider-GCP.go
```diff
@@ -0,0 +1,167 @@
...
+	resolvedMetadata, err := p.handleInvalidMetadataKeys(srcProperties.SrcMetadata)
+
+	if err != nil {
+		return nil, err
+	}
+	srcProperties.SrcMetadata = resolvedMetadata
+	return &srcProperties, nil
+}
+
+func (p *gcpSourceInfoProvider) handleInvalidMetadataKeys(m common.Metadata) (common.Metadata, error) {
```
zezha-msft, line 121: This code looks identical to the S3 source info provider, consider consolidating the two methods if possible.

cmd/zt_generic_traverser_test.go
```diff
@@ -170,6 +172,75 @@ func (s *genericTraverserSuite) TestS3GetProperties(c *chk.C) {
...
+		ContentEncoding:    "gzip",
+		ContentLanguage:    "en",
+		ContentDisposition: "inline",
+		CacheControl:       "no-cache",
+	}
+
+	bucketName := generateBucketName()
+	objectName := generateObjectName()
+	bkt := client.Bucket(bucketName)
+	err = bkt.Create(context.Background(), os.Getenv("GOOGLE_CLOUD_PROJECT"), &storage.BucketAttrs{})
```
zezha-msft, line 195: >os.Getenv("GOOGLE_CLOUD_PROJECT") [](start = 40, length = 33)

use lifecycle manager instead to fetch the env var

cmd/zt_copy_s2smigration_test.go
```diff
@@ -247,6 +306,47 @@ func (s *cmdIntegrationSuite) TestS2SCopyFromS3ToBlobWithWildcardInSrcAndBucketN
 	})
 }
 
+func (s *cmdIntegrationSuite) TestS2SCopyFromGCPToBlobWithWildcardInSrcAndBucketNameNeedBeResolved(c *chk.C) {
+	skipIfS3Disabled(c)
```
zezha-msft, line 310: we can perhaps rename skipIfS3Disabled since it controls both S3 and GCP tests

cmd/helpMessages.go
```diff
@@ -28,6 +28,7 @@ Copies source data to a destination location. The supported directions are:
   - Azure Files (SAS) -> Azure Files (SAS)
   - Azure Files (SAS) -> Azure Blob (SAS or OAuth authentication)
   - AWS S3 (Access Key) -> Azure Block Blob (SAS or OAuth authentication)
+  - Google Cloud Storage (Service Account Key) -> Azure Block Blob (SAS or OAuth authentication)
```
zezha-msft, line 31: > Google Cloud Storage (Service Account Key) -> Azure Block Blob (SAS or OAuth authentication) [](start = 3, length = 93)

Make a note here to point out GCP -> Blob is in preview.

cmd/copyEnumeratorInit.go
```diff
@@ -21,6 +21,10 @@ import (
 	"github.com/Azure/azure-storage-azcopy/ste"
 )
 
+type RemoteResourceToAzureResourcesResolver interface {
+	ResolveName(bucketName string) (string, error)
```
adreed-msft, line 25: Perhaps a better name would be "BucketNameResolver"?

cmd/copyEnumeratorInit.go
```diff
@@ -171,8 +183,9 @@ func (cca *cookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 					// this will probably never be reached
 					return nil, fmt.Errorf("failed to get container name from source (is it formatted correctly?)")
 				}
+				var resName string
```
adreed-msft, line 187: Was there any logical impact of this change?

cmd/credentialUtil.go
```diff
@@ -409,7 +409,26 @@ func checkAuthSafeForTarget(ct common.CredentialType, resource, extraSuffixesAAD
...
+		host := "<unparseable url>"
 
+		u, err := url.Parse(resource)
+		if err == nil {
+			host = u.Host
+			parts, err := common.NewGCPURLParts(*u)
+			if err == nil {
+				_, err := url.Parse("https://" + parts.Endpoint)
+				ok = err == nil
+			}
```
adreed-msft, line 427: This is unsafe.

Is there a general format for GCP URIs? Does the GCP SDK have a way to validate a URI is in fact, GCP?

ste/sender-blockBlobFromURL.go
nakulkar-msft: Removed the blockID generation

cmd/zt_generic_traverser_test.go
nakulkar-msft: This is for the testcase and not built into binary. We've preferred fetching directly from Env in testcases.

cmd/zt_copy_s2smigration_test.go
nakulkar-msft: What if we specialize for GCP so that way we can control S3 tests and GCP individually?

cmd/copyEnumeratorInit.go
nakulkar-msft: Discarded refactor. Edited it back.
https://github.com/Azure/azure-storage-azcopy/pull/1036#discussion_r436445994

mohsha-msft (commented):
Looks good in general. I have just some comments on URL parsing which I felt is error-prone.

cmd/credentialUtil.go
nakulkar-msft: Unlike for Azure and AWS where we generate requests depending on the Endpoint provided to us by the user, the EP for GCP is generated by the GCP SDK, based on the bucket Name and objectName (aka object key). Which means that  bucket and object names are only worthwhile info from user input.

In the current code as well, once we've the names, we interact with GCP through the signed URLs generated by SDK. Given that all we need are bucket and object names, appears the SDK does not have a validate an URI.

It is acceptable restriction that we expect the resource to be in format https://storage.cloud.google.com/<bucketName>/<path/to/object>. Taking out EP refs in GCPURLParts, since EP does not make sense here. This restriction means Custom domains and Signed URLs (though used internally) cannot be used. They should be taken at a later stage.

ste/sourceInfoProvider-GCP.go
nakulkar-msft: Probably gonna leave this. This could be made a method of common.Metadata, but that is logically wrong. Suggestions?

mohsha-msft:
Running test pipeline with S3 tests enabled to check the regression, if any.
https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=3848&view=results

common/credentialFactory.go
```diff
@@ -21,6 +21,7 @@
 package common
 
 import (
+	"cloud.google.com/go/storage"
 	"context"
```
adreed-msft, line 25: Consider using an aliased import instead of the stock storage package name-- Perhaps `gcp` would be a good fit?

common/credentialFactory.go
```diff
@@ -285,3 +285,44 @@ func (f *S3ClientFactory) GetS3Client(ctx context.Context, credInfo CredentialIn
...
+// ====================================================================
+func CreateGCPClient(ctx context.Context) (*storage.Client, error) {
+	client, err := storage.NewClient(ctx)
+	return client, err
+}
+
+type GCPClientFactory struct {
+	gcpClients map[CredentialInfo]*storage.Client
+	lock       sync.RWMutex
+}
```
adreed-msft, line 300: Would sync.Map be preferable here?

cmd/zt_copy_s2smigration_test.go
nakulkar-msft: Separating GCP and S3 tests

azure-pipelines.yml
```diff
@@ -111,6 +118,8 @@ jobs:
           ACCOUNT_KEY: $(ACCOUNT_KEY)
           AWS_ACCESS_KEY_ID: $(AWS_ACCESS_KEY_ID)
           AWS_SECRET_ACCESS_KEY: $(AWS_SECRET_ACCESS_KEY)
+          GOOGLE_CLOUD_PROJECT: $(GOOGLE_CLOUD_PROJECT)
+          GOOGLE_APPLICATION_CREDENTIALS: $(ciGCSServiceAccountKey.secureFilePath)
```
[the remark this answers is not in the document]
nakulkar-msft: Skipping this

testSuite/scripts/test_google_cloud_storage_copy.py
```diff
@@ -0,0 +1,796 @@
...
+    
+    def test_copy_single_1kb_file_from_gcp_to_blob(self):
+        if 'GCP_TESTS_OFF' in os.environ and os.environ['GCP_TESTS_OFF'] != "":
+            self.skipTest('GCS testing is disabled for this smoke test run')
+        src_bucket_url = util.get_object_without_sas(util.test_s2s_src_gcp_service_url, self.bucket_name)
+        dst_container_url = util.get_object_sas(util.test_s2s_dst_blob_account_url, self.bucket_name)
+        self.util_test_copy_single_file_from_x_to_x(src_bucket_url, "GCP", dst_container_url, "Blob", 1)
+
+    def test_copy_single_0kb_file_from_gcp_to_blob(self):
+        if 'GCP_TESTS_OFF' in os.environ and os.environ['GCP_TESTS_OFF'] != "":
```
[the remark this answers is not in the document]
nakulkar-msft: Separating out GCP and S3 variables. This'll be fixed as part.

go.sum
nakulkar-msft: Followed by @aishmanoh

common/credentialFactory.go
nakulkar-msft: Changing this to gcpUtils. Storage is indeed confusing in this code.

common/credentialFactory.go
nakulkar-msft: Yes, but not gonna impact perf. This is help per jptm, and is upper bound. I'll skip this, makes this look different from S3 code for no tangible gain.

adreed-msft (approved):
:shipit:

zezha-msft (approved):
:shipit:

zezha-msft:
lgtm @nakulkar-msft looks like there's just a minor merge conflict left

mohsha-msft (approved):
Let's go with it 👍

Issue #1107, opened 2020-07-21
AzCopy reports conflict for file share folder with name ";"


JasonYeMSFT:
### Which version of the AzCopy was used? 
10.5.0

### Which platform are you using? (ex: Windows, Mac, Linux)
Mac

### What command did you run?
copy

### What problem was encountered?
AzCopy reports a conflict, which doesn't seem right.

### How can we reproduce the problem in the simplest way?
1. Create a new file share (e.g. "f1")
2. Create a folder named ";" in it
3. Create a new file share (e.g. "f2")
4. Copy everything in "f1" to "f2" using the container url (e.g. cp "https://acct.file.core.windows.net/f1?sas_token" "https://acct.file.core.windows.net/f2?sas_token"

AzCopy prompts for a conflict for the ";" folder. Whether you choose to overwrite or not determines whether the folder properties will be copied or not.

### Have you found a mitigation/solution?
Choose to overwrite seems to bring the expected behavior. Why is it reporting conflict? Folders with other names don't have this problem. They just get themselves and the properties copied without a prompt.

PR #1335, opened 2021-02-16
Conflict for file share folder with name ";"

mohsha-msft:
Fix for [Issue 1107](https://github.com/Azure/azure-storage-azcopy/issues/1107)

Reason:

While creating file directories, we maintain a file creation tracker(which basically is a map).
When we create a folder, folder URL looks like - 
`https://<accountName>.file.core.windows.net/<fileShare>/<folder>`

In this case, where folder name contains a special character `;`, folder URL will look like -
`https://<accountName>.file.core.windows.net/<fileShare>/;`
After we're done with creating the folder, we make entry in the file creation tracker.

However the issue occurs when we try to lookup the created folders for transferring properties. During lookup, we're searching for folder `%3B` which is percentage encoding equivalent of `;` i.e. `https://<accountName>.file.core.windows.net/<fileShare>/%3B`

Fixed the issue by performing URL encoding on the directory name containing ";" before making an entry in the file tracker..

### DIFF
--- a/common/folderCreationTracker.go
+++ b/common/folderCreationTracker.go
@@ -65,8 +65,13 @@ type simpleFolderTracker struct {
 func (f *simpleFolderTracker) RecordCreation(folder string) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
+	decodedFolderName, err := url.QueryUnescape(folder)
+	if err == nil {
+		f.contents[decodedFolderName] = struct{}{}
+	} else {
+		f.contents[folder] = struct{}{}
+	}
 
-	f.contents[folder] = struct{}{}
 }
 
 func (f *simpleFolderTracker) ShouldSetProperties(folder string, overwrite OverwriteOption, prompter prompter) bool {
@@ -81,6 +86,12 @@ func (f *simpleFolderTracker) ShouldSetProperties(folder string, overwrite Overw
 		defer f.mu.Unlock()
 
 		_, exists := f.contents[folder] // should only set properties if this job created the folder (i.e. it's in the map)
+		if !exists {
+			decodedFolderName, err := url.QueryUnescape(folder)
+			if err == nil {
+				_, exists = f.contents[decodedFolderName]
+			}
+		}
 
 		// prompt only if we didn't create this folder
 		if overwrite == EOverwriteOption.Prompt() && !exists {
@@ -111,7 +122,10 @@ func (f *simpleFolderTracker) ShouldSetProperties(folder string, overwrite Overw
 func (f *simpleFolderTracker) StopTracking(folder string) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
-
+	decodedFolderName, err := url.QueryUnescape(folder)
+	if err == nil {
+		delete(f.contents, decodedFolderName)
+	}
 	delete(f.contents, folder)
 }
 
--- a/common/folderCreationTracker.go
+++ b/common/folderCreationTracker.go
@@ -65,13 +65,8 @@ type simpleFolderTracker struct {
 func (f *simpleFolderTracker) RecordCreation(folder string) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
-	decodedFolderName, err := url.QueryUnescape(folder)
-	if err == nil {
-		f.contents[decodedFolderName] = struct{}{}
-	} else {
-		f.contents[folder] = struct{}{}
-	}
 
+	f.contents[folder] = struct{}{}
 }
 
 func (f *simpleFolderTracker) ShouldSetProperties(folder string, overwrite OverwriteOption, prompter prompter) bool {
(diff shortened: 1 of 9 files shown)


ste/sender-azureFile.go
```diff
@@ -446,7 +446,8 @@ func (d AzureFileParentDirCreator) CreateDirToRoot(ctx context.Context, dirURL a
 					// Must do it here, in the routine that is shared by both the folder and the file code,
 					// because due to the parallelism of AzCopy, we don't know which will get here first, file code, or folder code.
 					dirUrl := curDirURL.URL()
-					t.RecordCreation(dirUrl.String())
+					dirUrlStr := strings.Replace(dirUrl.String(), ";", url.QueryEscape(";"), -1)
+					t.RecordCreation(dirUrlStr)
```
adreed-msft, line 450: Is `;` the only one of these "unsafe" characters? Should we instead use the encoding rules from the `cmd` package?

mohsha-msft: I noticed that `url.String()` method, which we have in file SDK, doesn't encode `;` but encode other dir/file with special characters. I tried that with general case and it seems to work.

common/folderCreationTracker.go
```diff
@@ -65,8 +65,13 @@ type simpleFolderTracker struct {
 func (f *simpleFolderTracker) RecordCreation(folder string) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
+	decodedFolderName, err := url.QueryUnescape(folder)
+	if err == nil {
+		f.contents[decodedFolderName] = struct{}{}
+	} else {
+		f.contents[folder] = struct{}{}
```
nakulkar-msft, line 72: This is potentially a bug. We're kind of silently ignoring error without the knowledge of surrounding code.

This is a Map lookup failure - we're using dirURL().String to lookup while we created the entry with info.Destination. Instead of specializing it for escaped characters, we should change lookup/creation to use the same string.

nakulkar-msft (changes requested):
Lookup using same keys.

common/folderCreationTracker.go
mohsha-msft: Did the changes by making some adjustments in the design as we discussed. I think this should solve the bug.

adreed-msft (approved):
:shipit:

e2etest/zt_naming_file_and_folder_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"github.com/Azure/azure-storage-azcopy/common"
+	"testing"
+)
+
+// Upload, Download, S2S transfer of folders/files with special characters. Required for avoiding regression.
+func TestNaming_ShareFileFoldersSpecialChar(t *testing.T) {
+	files := []string{"file1.txt", "fi,le2.pdf", "fil%e3.mp3", "file 4.jpg", "file;a5.csv", "file_a6.cpp", "file+a7.mp4"}
+	folders := []string{";", ";;", "%", "_", "+", "test%folder1", "test+folder2", "test%folder3", "test folder4", "test_folder5"}
```
zezha-msft, line 31: maybe include a case for comma, since that that one of the reported issues as well.

e2etest/zt_naming_file_and_folder_test.go
```diff
@@ -0,0 +1,53 @@
...
+			transfers = append(transfers, f(folders[i]+"/"+files[j]))
+		}
+	}
+	RunScenarios(
+		t,
+		eOperation.CopyAndSync(),
+		eTestFromTo.Other(common.EFromTo.FileFile(), common.EFromTo.FileLocal(), common.EFromTo.LocalFile()),
+		eValidate.Auto(),
+		params{
+			recursive: true,
```
zezha-msft, line 46: From the issue, it looks like the overwrite flag should be false to repro this problem.

ste/sender-azureFile.go
```diff
@@ -383,6 +383,11 @@ func (u *azureFileSenderBase) SetFolderProperties() error {
 	return err
 }
 
+func (u *azureFileSenderBase) DirUrlToString() string {
+	dirUrl := (common.FileURLPartsExtension{FileURLParts: azfile.NewFileURLParts(u.dirURL().URL())}).URL()
```
zezha-msft, line 387: The URL() is on FileURLParts, we don't really need to involve FileURLPartsExtension right?

zezha-msft (approved):
The approach looks good to me.

Please double check that the test cases cover all the scenarios.

e2etest/zt_naming_file_and_folder_test.go
mohsha-msft: Alright. Added folder with `;` and `,` as well

e2etest/zt_naming_file_and_folder_test.go
mohsha-msft: Alright. Added folder with `;` and `,` as well

e2etest/zt_naming_file_and_folder_test.go
mohsha-msft: Ohh, we don't support overwrite flag in E2E test cases. I'll add it.

ste/sender-azureFile.go
mohsha-msft: Done.

PR #1340, opened 2021-02-20
Return appropriate error if we cannot determine if given path is a directory

nakulkar-msft:


cmd/copyUtil.go
```diff
@@ -137,13 +137,13 @@ func (util copyHandlerUtil) urlIsBFSFileSystemOrDirectory(ctx context.Context, u
 		}
 	}
 
-	return isDir
+	return isDir, err
```
mohsha-msft, line 140: Earlier scenario: 
If err is not nil, we return `isDir` and we silently assume that destination is a file but this will not fail the job.

Current Scenario:
If err is not nil, we return `isDir` and `err`. Wherever we're calling `isDirectory`, we're also performing `err != nil` check which will basically fail the job.
```
isSourceDir, err := sourceTraverser.isDirectory(true)
if err != nil {
	return nil, err
}
```

Is this the expected scenario we're trying to create?

nakulkar-msft:
Replaced by #1412

PR #1357, opened 2021-03-03
Improve Job Reporting infra

nakulkar-msft:
AzCopy reports the total number of transfers, # of transfers completed and percentage transfer while the transfer is in progress. These are fetched by looping through all the jobparts individually. All this data will be saved in mem and reported to the frontend whenever queried.

1. A separate Summary struct is introduced to hold all stats to be reported - ListJobSummaryResponse
2. A separate Goroutine is introduced to interact with this struct. This shall be only thread which modifies above struct. Change 482196a
3. At time where each part is created, total number of transfers are updated (change 6dab8c7 & 482196a)
4. At end of each transfer, completed or failure count in updated. (Change 
d941b28)
5. FE requests status to the go-routine, which then creates a copy of statement at that point. 
(2811adf)
6. Using buffered channels to reduce contention. In a very large job, a small offset should not matter. In a small job, the channels are not full.

### DIFF
--- a/ste/jobStatusManager.go
+++ b/ste/jobStatusManager.go
@@ -99,63 +99,6 @@ func (jm *jobMgr) handleStatusUpdateMessage() {
 			/* Display stats */
 			js.Timestamp = time.Now().UTC()
 			jstm.respChan <- *js
-
-		case <-time.After(2 * time.Second):
-			part0, ok := jm.JobPartMgr(0)
-			if !ok {
-				break
-			}
-			part0PlanStatus := part0.Plan().JobStatus()
-
-			// Add on byte count from files in flight, to get a more accurate running total
-			js.TotalBytesTransferred += JobsAdmin.SuccessfulBytesInActiveFiles()
-			if js.TotalBytesExpected == 0 {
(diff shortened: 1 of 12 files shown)


ste/jobStatusManager.go
```diff
@@ -0,0 +1,104 @@
...
+			case common.ETransferStatus.SkippedEntityAlreadyExists(),
+				common.ETransferStatus.SkippedBlobHasSnapshots():
+				js.TransfersSkipped++
+				js.SkippedTransfers = append(js.SkippedTransfers, common.TransferDetail(msg))
+			}
+
+		case <-jstm.listReq:
+			/* Display stats */
+			js.Timestamp = time.Now().UTC()
+			jstm.respChan <- *js
```
zezha-msft, line 101: Nice, very elegant.

Should there be a default sleep if nothing came through? To avoid hogging CPU

zezha-msft (approved):
:shipit:

mohsha-msft (approved):

mohsha-msft:
Rebase with dev and resolve the conflicts :shipit:

PR #1365, opened 2021-03-12
[AzCopyV10][Feature] Offer knob to disable application logging (Syslog/Windows Evert Log)

mohsha-msft:
By default, AzCopy logs errors and warnings in Syslog/Windows Event Log in addition to logging in AzCopy log file. 
It can be useful for monitoring warnings/errors when the job is very large. But it can also create noise by polluting system logs.

Here is a knob to disable logging in these channels (Syslog/Windows Event Log). We fetch environment variable  `AZCOPY_DISABLE_SYSLOG`. If the value of `AZCOPY_DISABLE_SYSLOG` is set to true, we stop force logging in these channels.

### DIFF
--- a/azbfs/zc_policy_request_log.go
+++ b/azbfs/zc_policy_request_log.go
@@ -22,6 +22,7 @@ type RequestLogOptions struct {
 	// LogWarningIfTryOverThreshold logs a warning if a tried operation takes longer than the specified
 	// duration (-1=no logging; 0=default threshold).
 	LogWarningIfTryOverThreshold time.Duration
+	ForceLog                     bool
 }
 
 func (o RequestLogOptions) defaults() RequestLogOptions {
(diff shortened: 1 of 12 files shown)


common/lifecyleMgr.go
```diff
@@ -558,6 +562,19 @@ func (lcm *lifecycleMgr) E2EEnableAwaitAllowOpenFiles(enable bool) {
 	}
 }
 
+func (lcm *lifecycleMgr) SetForceLogging() {
```
zezha-msft, line 565: please add a comment to explain why we fetch this value only once and save it here in lcm, for future reference.

zezha-msft (approved):
lgtm, to confirm, you've tried causing errors and observed that we log nothing to the syslog right? I'm not 100% sure that we always use our own pipeline created inside the AzCopy package, it'd be good to double check (do a search to see where we use the default SDK pipeline).

common/lifecyleMgr.go
mohsha-msft: Sure. I'll add that.
As for the test run, I did test the functionality by performing a transfer with error and it didn't write anything in the linux syslog. So I guess handling `forcelog` was the key to handle it.

zezha-msft: It'd be good to do a quick search to see whether we use the default pipeline provided by the SDK.

azbfs/zc_policy_request_log.go
```diff
@@ -22,6 +22,7 @@ type RequestLogOptions struct {
 	// LogWarningIfTryOverThreshold logs a warning if a tried operation takes longer than the specified
 	// duration (-1=no logging; 0=default threshold).
 	LogWarningIfTryOverThreshold time.Duration
+	ForceLog                     bool
```
nakulkar-msft, line 25: Did you have to make this change in other sdks as well? 
Minor: ForceLog is a misnomer. EnableSyslog/EnableEventLog or something which says what we're doing.

nakulkar-msft (approved):
Consider changing ForceLog to more appropriate name.

azbfs/zc_policy_request_log.go
mohsha-msft: Will rename it. 
As for your other question of changes in SDK, yes I am on it.

mohsha-msft:
Changes in SDK to provide Knob :-
[Blob SDK PR](https://github.com/Azure/azure-storage-blob-go/pull/268)
[File SDK PR](https://github.com/Azure/azure-storage-file-go/pull/52)

PR #1366, opened 2021-03-12
[AzCopyV10][Feature] Display more fields for list command

mohsha-msft:
Added flag `--properties` in the list command which takes `;` separated list of properties and will display these fields in list command output. 
Currently list of supported properties is `"LastModifiedTime", "VersionId", "BlobType", "BlobAccessTier", "ContentType", and "ContentEncoding"` . Based on requirement, we can add more parameters in the future.

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -45,7 +45,15 @@ type rawListCmdArgs struct {
 }
 
 var validProperties = []string{
-	"LastModifiedTime", "VersionId", "BlobType", "BlobAccessTier", "ContentType", "ContentEncoding",
+	"LastModifiedTime",
+	"VersionId",
+	"BlobType",
+	"BlobAccessTier",
+	"ContentType",
+	"ContentEncoding",
+	"LeaseState",
+	"LeaseDuration",
+	"LeaseStatus",
 }
 
 func (raw *rawListCmdArgs) parseProperties(rawProperties string) []string {
@@ -158,6 +166,12 @@ func (cooked cookedListCmdArgs) processProperties(object storedObject) string {
 			builder.WriteString(property + ": " + object.contentType + "; ")
 		case "ContentEncoding":
 			builder.WriteString(property + ": " + object.contentEncoding + "; ")
+		case "LeaseState":
+			builder.WriteString(property + ": " + string(object.leaseState) + "; ")
+		case "LeaseStatus":
+			builder.WriteString(property + ": " + string(object.leaseStatus) + "; ")
+		case "LeaseDuration":
+			builder.WriteString(property + ": " + string(object.leaseDuration) + "; ")
 		}
 	}
 	return builder.String()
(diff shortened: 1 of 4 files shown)


cmd/list.go
```diff
@@ -178,17 +255,17 @@ func HandleListContainerCommand(unparsedSource string, location common.Location)
...
+//func printListContainerResponse(lsResponse *common.ListContainerResponse) {
+//	if len(lsResponse.Blobs) == 0 {
+//		return
+//	}
+//	// TODO determine what's the best way to display the blobs in JSON
+//	// TODO no partner team needs this functionality right now so the blobs are just outputted as info
+//	for index := 0; index < len(lsResponse.Blobs); index++ {
+//		glcm.Info(lsResponse.Blobs[index])
+//	}
+//}
```
zezha-msft, line 268: we can remove this if it's not used.

cmd/list.go
```diff
@@ -26,15 +26,80 @@ import (
...
+	// obtained from argument
+	sourcePath string
+
+	Properties      string
+	MachineReadable bool
+	RunningTally    bool
+	MegaUnits       bool
+}
+
+var validProperties = []string{
```
zezha-msft, line 47: perhaps you could consider using another data structure to make the comparison easier, and also ignore the casing.

zezha-msft:
> 		Example: listCmdExample,

Update examples.

---
Refers to: cmd/list.go:110 in b73c6dc. [](commit_id = b73c6dcf77ac1f4372d0c84e5fbb76a359f72cf7, deletion_comment = False)

cmd/list.go
```diff
@@ -26,15 +26,80 @@ import (
...
+	"LastModifiedTime", "VersionId", "BlobType", "BlobAccessTier", "ContentType", "ContentEncoding",
```
zezha-msft, line 48: >LastModifiedTime [](start = 2, length = 16)

Sorry I should have given you this info earlier, I found the original ask and it included lease status too.

mohsha-msft: No problem. Changes weren't that big. Added `LeaseStatus`, `LeaseState`, and `LeaseDuration` in the `storedObject` as well.

cmd/list.go
mohsha-msft: I am using `strings.EqualFolds()` for string comparison so it is case-insensitive. As for the data-structure, I've converted it to `const` to keep the behavior similar as other similar types (`BlobType`, `BlobAccessTier`, etc.).

cmd/list.go
```diff
@@ -208,13 +296,13 @@ func byteSizeToString(size int64) string {
 		"GiB",
 		"TiB",
 		"PiB",
-		"EiB", // Let's face it, a file, account, or container probably won't be more than 1000 exabytes in YEARS. (and int64 literally isn't large enough to handle too many exbibytes. 128 bit processors when)
+		"EiB", // Let's face it, a file, account, or container probably won't be more than 1000 exabytes in YEARS. (and int64 literally isn't large enough to handle too many exabytes. 128 bit processors when)
```
nakulkar-msft, line 299: Change this back to exbiytes. The correct term here is exbibytes.

nakulkar-msft (changes requested):

cmd/list.go
mohsha-msft: Sure.

nakulkar-msft (approved):

PR #1368, opened 2021-03-15
[AzCopyV10][Feature] Add support for CPK-N and CPK-V

mohsha-msft:


cmd/jobsResume.go
```diff
@@ -284,7 +284,7 @@ func (rca resumeCmdArgs) process() error {
 		destination:    getJobFromToResponse.Destination,
 		sourceSAS:      rca.SourceSAS,
 		destinationSAS: rca.DestinationSAS,
-	}); err != nil {
+	}, common.CpkOptions{}); err != nil {
```
nakulkar-msft, line 287: Nitpick: Formatting

common/fe-ste-models.go
```diff
@@ -1368,3 +1368,94 @@ func (p PreservePermissionsOption) IsTruthy() bool {
...
+
+func (csi CpkScopeInfo) Marshal() (string, error) {
+	result, err := json.Marshal(csi)
+	if err != nil {
+		return "", err
+	}
+	return string(result), nil
+}
+
+//func UnmarshalToCpkScopeInfo(cpkScopeInfoStr string) (CpkScopeInfo, error) {
```
nakulkar-msft, line 1387: Check if this commented code can be removed

ste/JobPartPlanFileName.go
```diff
@@ -189,6 +189,9 @@ func (jpfn JobPartPlanFileName) Create(order common.CopyJobPartOrderRequest) {
 			MetadataLength:           uint16(len(order.BlobAttributes.Metadata)),
 			BlockSize:                blockSize,
 			BlobTagsLength:           uint16(len(order.BlobAttributes.BlobTagsString)),
+			CpkInfo:                  order.CpkOptions.CpkInfo,
```
nakulkar-msft, line 192: Maj: This is a secret and should not be written to disk, much like a SAS key. Instead we should pass it on jptm/jpm and then reconstruct it from environment during resume.

mohsha-msft: Hey @nakulkar-msft  , 
It is just a boolean variable.
If `order.CpkOptions.CpkInfo` is true, that means we'll fetch the secrets from environment variables at the time of use only. We will not save any secrets whatsoever on the disk.

nakulkar-msft: (y) I took it to cpkOptions

nakulkar-msft (approved):
Fix the offline discussed comments.

common/fe-ste-models.go
mohsha-msft: Can I remove it later once CPK goes live? I want to keep it for future reference.

cmd/copy.go
```diff
@@ -1718,6 +1765,13 @@ func init() {
...
 	cpCmd.PersistentFlags().BoolVar(&raw.s2sGetPropertiesInBackend, "s2s-get-properties-in-backend", true, "get S3 objects' or Azure files' properties in backend, if properties need to be accessed. Properties need to be accessed if s2s-preserve-properties is true, and in certain other cases where we need the properties for modification time checks or MD5 checks")
 
+	// Public Documentation: https://docs.microsoft.com/en-us/azure/storage/blobs/encryption-customer-provided-keys
+	// Clients making requests against Azure Blob storage have the option to provide an encryption key on a per-request basis.
+	// Including the encryption key on the request provides granular control over encryption settings for Blob storage operations.
+	// Customer-provided keys can be stored in Azure Key Vault or in another key store linked to storage account.
+	cpCmd.PersistentFlags().StringVar(&raw.cpkScopeInfo, "cpk-by-name", "", "Client provided key by name let clients making requests against Azure Blob storage an option to provide an encryption key on a per-request basis. Provided key name will be fetched from Azure Key Vault and will be used to encrypt the data")
```
zezha-msft, line 1772: >encrypt [](start = 295, length = 7)

it works for decryption too right? When downloading for example.

mohsha-msft: Correct

cmd/copy.go
```diff
@@ -525,6 +534,38 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
...
+		// Due to service limitation, S2S transfer is not supported for source encrypted by user key.
+		if cooked.fromTo.IsDownload() {
+			glcm.Info("Client Provided Key (CPK) for encryption/decryption is provided for download scenario. " +
+				"Assuming source is encrypted.")
+			cpkOptions.IsSourceEncrypted = true
+		}
+
+		// TODO: Remove these warnings once service starts supporting it
+		if cooked.blockBlobTier != common.EBlockBlobTier.None() || cooked.pageBlobTier != common.EPageBlobTier.None() {
+			glcm.Info("Tier is provided by user explicitly. Ignoring it because Azure Service currently does" +
```
zezha-msft, line 562: >Azure Service [](start = 71, length = 13)

Minor: Storage Service

cmd/copy.go
```diff
@@ -1115,7 +1162,7 @@ func (cca *cookedCopyCmdArgs) processRedirectionUpload(blobResource common.Resou
 	}
 
 	// getCredentialInfoForLocation populates oauth token fields... so, it's very easy.
-	credInfo, _, err := getCredentialInfoForLocation(ctx, common.ELocation.Blob(), blobResource.Value, blobResource.SAS, false)
+	credInfo, _, err := getCredentialInfoForLocation(ctx, common.ELocation.Blob(), blobResource.Value, blobResource.SAS, false, cca.cpkOptions)
```
zezha-msft, line 1165: > cca.cpkOptions [](start = 124, length = 15)

we didn't call `common.GetClientProvidedKey` yet, any reason for that?

common/credentialFactory.go
```diff
@@ -326,3 +326,39 @@ func (f *GCPClientFactory) GetGCPClient(ctx context.Context, credInfo Credential
 		return gcpClient, nil
 	}
 }
+
+// Default Encryption Algorithm Supported
+const EncryptionAlgorithmAES256 string = "AES256"
```
zezha-msft, line 331: we have azblob.EncryptionAlgorithmAES256 already right?

common/fe-ste-models.go
```diff
@@ -1368,3 +1368,94 @@ func (p PreservePermissionsOption) IsTruthy() bool {
...
+//		if err != nil {
+//			return result, err
+//		}
+//	}
+//
+//	return result, nil
+//}
+
+func ToClientProvidedKeyOptions(cpkInfo CpkInfo, cpkScopeInfo CpkScopeInfo) azblob.ClientProvidedKeyOptions {
+	return azblob.ClientProvidedKeyOptions{
```
zezha-msft, line 1435: >azblob.ClientProvidedKeyOptions [](start = 8, length = 31)

minor style issue: wouldn't it be easier to use azblob.ClientProvidedKeyOptions directly to hold the data in the first place?

mohsha-msft: This is inline with track2 changes which we're going to bring in and inline with other SDKs as well. There CpkScopeInfo and CpkInfo are treated as separate entities. To avoid that change, I've made sure to use two objects instead of one,

cmd/copy.go
mohsha-msft: Yup. We call `common.GetClientProvidedKey(cpkOptions)` when we need the credentials just before using them. 
If you go inside the definition of this function, we're calling it inside `getBlobCredentialType()`

```
if !isContainer {
	clientProvidedKey := azblob.ClientProvidedKeyOptions{}
	if cpkOptions.IsSourceEncrypted {
		clientProvidedKey = common.GetClientProvidedKey(cpkOptions)
	}
	// Scenario 3: When resourceURL points to a blob
	blobURL := azblob.NewBlobURL(*resourceURL, p)
	if _, err := blobURL.GetProperties(ctx, azblob.BlobAccessConditions{}, clientProvidedKey); err == nil {
		return true
	}
}

```

zezha-msft (approved):
:shipit:

zezha-msft:
It's be great to have some E2E tests. 😉

mohsha-msft:
> It's be great to have some E2E tests. 😉

I'm planning to write them with sync changes. There are other minor structural changes in E2E which I need to make before I can do that.

PR #1375, opened 2021-03-25
Create SAS tokens for blob->* transfers on the fly in the STE

adreed-msft:
Pasted from the old PR:

Some background:

Previous iterations of this PR have done this creation in the frontend. This would cause problems if a user attempted to resume a transfer of this style, because the SAS has to come in externally in that scenario. Furthermore, this would have caused issues with S2S transfers, as S2S transfers may potentially contain multiple source containers, and UDKs can only create SAS tokens for containers or blobs, not the entire service.

Therefore, I moved it all to the STE. It lives under a single-instance struct named UserDelegationAuthenticationManager, which does two things:

If a UDK was created, it maps container names to SAS tokens.
If a UDK was not created, it creates empty SAS tokens.
Furthermore, rather than just grabbing the URL once from SIP, we grab it every time we make a call to enable refreshing the underlying UDK and user delegated SAS token

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -1165,9 +1165,15 @@ func (cca *cookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 		return err
 	}
 
+	srcCredType, srcPublic, err := getCredentialTypeForLocation(ctx, cca.fromTo.From(), cca.source.Value, cca.source.SAS, true)
+
+	if err != nil {
+		return err
+	}
+
 	// For OAuthToken credential, assign OAuthTokenInfo to CopyJobPartOrderRequest properly,
 	// the info will be transferred to STE.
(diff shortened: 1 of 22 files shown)


zezha-msft:
@adreed-msft tests need some fixing.

mohsha-msft:
Please fix the conflicts and failing UTs.

doanduyhai:
Any update ?

zezha-msft:
Closing for now as the service has a new feature that makes this scenario a lot easier.

seanralexander:
> Closing for now as the service has a new feature that makes this scenario a lot easier.

What feature is this which makes it easier?

karimdabbagh:
> Hi @seanralexander, the new feature allows us to pass an additional bearer token that the destination account can pass to the source account, thus eliminating the need to generate user delegation SAS which requires more permission.
> 
> We'll adopt this feature soon.

any updates on this?

PR #1376, opened 2021-03-25
[AzCopyV10][Feature] Enable Local <-> File and Blob <-> File for sync

mohsha-msft:


nakulkar-msft (approved):
Approved, subject to testing. Also verify if E2E runs on these combinations.

cmd/syncEnumerator.go
```diff
@@ -45,9 +45,9 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 		if cca.fromTo.From() != common.ELocation.S3() {
 			// Adding files here seems like an odd case, but since files can't be public
 			// the second half of this if statement does not hurt.
-			if cca.fromTo.From() != cca.fromTo.To() {
-				panic(fmt.Sprintf("The semantics of authorization for an S2S transfer (within sync) is unknown for %s->%s transfers", cca.fromTo.From(), cca.fromTo.To()))
-			}
+			//if cca.fromTo.From() != cca.fromTo.To() {
+			//	panic(fmt.Sprintf("The semantics of authorization for an S2S transfer (within sync) is unknown for %s->%s transfers", cca.fromTo.From(), cca.fromTo.To()))
+			//}
```
zezha-msft, line 50: remove since not needed any more.

zezha-msft (approved):
The code looks good, but don't we have to enable tests in E2E?

mohsha-msft:
> The code looks good, but don't we have to enable tests in E2E?

I have to write test cases for couple of things. I'll raise a separate PR for that.

PR #1384, opened 2021-03-31
Explicitly mark that remove destination cannot be public

nakulkar-msft:
For remove command, we overload copy with source as remove target.
getCredentialTypeForLocation() for source does not authenticate to public containers and blobs, as a public read access is enough on regular source. However this is incorrect for remove command, as azcopy needs delete permissions, among others. This change fixes this issue.


zezha-msft (approved):

mohsha-msft (approved):
LGTM

PR #1385, opened 2021-03-31
Update #813: Enable the creation of dummy transfers to deliver errors earlier in the process

adreed-msft:
Previously, when we skipped containers due to an inability to resolve the name or traverse the source, it came out via stdout and the job logs as a warning... this wasn't the most "visible" of things, especially in automated transfers.

This now outputs actual transfers with an embedded failure status, so that we have a direct failure listing in the job part plan, and a proper failure output from azcopy. This should be more alarming to users who would not otherwise have known their jobs weren't running completely.

### DIFF
--- a/cmd/zc_enumerator.go
+++ b/cmd/zc_enumerator.go
@@ -82,6 +82,8 @@ type storedObject struct {
 	Metadata      common.Metadata
 	blobVersionID string
 	blobTags      common.BlobTags
+	// injected error data
+	transferFailureReason string
 }
 
 const (
@@ -155,6 +157,7 @@ func (s *storedObject) ToNewCopyTransfer(
 	t := common.CopyTransfer{
 		Source:             Source,
 		Destination:        Destination,
+		FailureReason:      s.transferFailureReason,
 		EntityType:         s.entityType,
 		LastModifiedTime:   s.lastModifiedTime,
 		SourceSize:         s.size,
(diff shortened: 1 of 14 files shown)


ste/xfer.go
```diff
@@ -64,6 +66,60 @@ type newJobXfer func(jptm IJobPartTransferMgr, pipeline pipeline.Pipeline, pacer
...
+			// Send the right type of error
+			if fromTo.IsDownload() {
+				jptm.LogDownloadError(shortSrc, shortDst, info.FailureReason, 0)
+			} else if fromTo.IsUpload() {
+				jptm.LogUploadError(shortSrc, shortDst, info.FailureReason, 0)
+			} else if fromTo.IsS2S() {
+				jptm.LogS2SCopyError(shortSrc, shortDst, info.FailureReason, 0)
+			} else if fromTo.To() == common.ELocation.Unknown() {
+				jptm.LogError(shortSrc, "DELETE ERROR", errors.New(info.FailureReason))
+			}
```
zezha-msft, line 111: We have scanning logs now, I think that these scanning errors should be logged there instead.

cmd/zc_enumerator.go
```diff
@@ -210,6 +213,25 @@ type blobPropsProvider interface {
 	AccessTier() azblob.AccessTierType
 }
 
+func newForcedErrorStoredObject(err, name, relativePath, containerName string) storedObject {
```
zezha-msft, line 216: I see the goal of the change, but I feel a bit uneasy about how we represent the scanning failures. So we are representing each scanning failure as an individual transfer failure, isn't that a bit confusing for regular users?

adreed-msft: I don't feel like this is confusing for the end user-- If any single unit of a transfer can't be handled, shouldn't it just be a single, tangible error? Do you feel like we should do this for every potential transfer in a container? I feel like that would pollute the job log files and the job plan files unnecessarily.

Alternatively, if you feel that we should not output any failed transfer, In my opinion, in the case of automated users, it makes sense to present an actual failed transfer, because they may depend on `jobs status`.

Is there some alternative thought you have that would satisfy these well?

adreed-msft: A second note is that this does actively affect the return code of azcopy, unlike logging straight to the scanning log

PR #1392, opened 2021-04-07
[AzCopyV10][UX] Provide flag to set try timeout 

siminsavani-msft:
Adding optional environment variable called AZCOPY_UPLOAD_TRY_TIMEOUT to set time (in minutes) for how long AzCopy should try to upload files before AzCopy times out.


common/environment.go
```diff
@@ -326,6 +327,12 @@ func (EnvironmentVariable) UserAgentPrefix() EnvironmentVariable {
 	}
 }
 
+func (EnvironmentVariable) UploadTryTimeout() EnvironmentVariable {
+	return EnvironmentVariable{
+		Name:        "AZCOPY_UPLOAD_TRY_TIMEOUT",
+		Description: "Set time (in minutes) for how long AzCopy should try to upload files before AzCopy times out.",
```
zezha-msft, line 333: Clarify that the try is for each request, not in general for the entire file.

cmd/root.go
```diff
@@ -64,7 +63,7 @@ var rootCmd = &cobra.Command{
 	Short:   rootCmdShortDescription,
 	Long:    rootCmdLongDescription,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-
+		ste.UploadTryTimeout, _ = time.ParseDuration(glcm.GetEnvironmentVariable(common.EEnvironmentVariable.UploadTryTimeout()) + "m")
```
zezha-msft, line 66: So if the environment variable is empty or not set incorrectly, what happens?

We need a default value.

zezha-msft (changes requested):

cmd/root.go
siminsavani-msft: There is a default value for UploadTryTimeout in xfer.go (line 39) but I added error validation for wrong or empty environment variable.

Issue #1408, opened 2021-04-16
AZCOPY Bench command is not able to Cleanup


RedOhne:
##### Note: The version is visible when running AzCopy without any argument
10.09 / 10.10

### Which platform are you using? (ex: Windows, Mac, Linux) 
Linux

##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.

azcopy bench "https://sahec44neetransfer0001.blob.core.windows.net/mig..."

### What problem was encountered?
After the bench run was successfully finished, AZCOPY tries to cleanup (remove the used files from the Blob), but is "hanging". I can still see the azcopy process running.
The log is piling up with the following entries:

2021/04/16 07:25:46 PERF: primary performance constraint is Unknown. States: X:  0, O:  0, M:  0, L:  0, R:  0, D:  0, W:  0, F:  0, B:  0, E:  0, T:  0, GRs: 64
2021/04/16 07:25:46 Cleanup 200/200

The log for cleanup is looking the following:
2021/04/16 06:51:26 AzcopyVersion  10.10.0
2021/04/16 06:51:26 OS-Environment  linux
2021/04/16 06:51:26 OS-Architecture  amd64
2021/04/16 06:51:26 Log times are in UTC. Local time is 16 Apr 2021 06:51:26
2021/04/16 06:51:26 AzcopyVersion  10.10.0
2021/04/16 06:51:26 OS-Environment  linux
2021/04/16 06:51:26 OS-Architecture  amd64
2021/04/16 06:51:26 Log times are in UTC. Local time is 16 Apr 2021 06:51:26
2021/04/16 06:52:53 ==> REQUEST/RESPONSE (Try=1/24.700659ms, OpTime=174.508359ms) -- RESPONSE SUCCESSFULLY RECEIVED
   HEAD https://sahec44neetransfer0001.blob.core.windows.net/mig/benchmark-aefdf877-0ddf-3d4b-6f04-545398f2f135?.....&timeout=901
   X-Ms-Request-Id: [08ecd991-701e-007a-1d8d-323a83000000]

2021/04/16 06:52:53 ==> REQUEST/RESPONSE (Try=1/41.489766ms, OpTime=41.750408ms) -- RESPONSE SUCCESSFULLY RECEIVED
[...]

PR #1409, opened 2021-04-16
Retain reportJobPartDoneHandler() for follow-up jobs

nakulkar-msft:
The thread reportJobPartDoneHandler() was introduced for bookkeeping sometime in 10.5. This thread is responsible to make azcopy exit after completing all the transfers. But, in case of benchmark, after transfers we start another followup-job to delete all the dummy files we uploaded. After the second job completes we should ideally exit the process.  Because reportJobPartDoneHandler() has already exited, the second job would hang indefinitely. The new change will make this thread permanent and will not close it, so that follow-up jobs are serviced as well.

### DIFF
--- a/ste/mgr-JobMgr.go
+++ b/ste/mgr-JobMgr.go
@@ -514,38 +514,41 @@ func (jm *jobMgr) reportJobPartDoneHandler() {
 		isCancelling := jobStatus == common.EJobStatus.Cancelling()
 		shouldComplete := allKnownPartsDone && (haveFinalPart || isCancelling)
 		if shouldComplete {
-			break
+			jobPart0Mgr, _ := jm.jobPartMgrs.Get(0)
+			part0Plan := jobPart0Mgr.Plan() // status of part 0 is status of job as whole.
+
+			partDescription := "all parts of entire Job"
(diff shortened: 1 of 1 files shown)


ste/mgr-JobMgr.go
```diff
@@ -514,38 +514,41 @@ func (jm *jobMgr) reportJobPartDoneHandler() {
 		isCancelling := jobStatus == common.EJobStatus.Cancelling()
 		shouldComplete := allKnownPartsDone && (haveFinalPart || isCancelling)
 		if shouldComplete {
-			break
-		} //Else log and wait for next part to complete
+			jobPart0Mgr, _ := jm.jobPartMgrs.Get(0)
+			part0Plan := jobPart0Mgr.Plan() // status of part 0 is status of job as whole.
```
mohsha-msft, line 518: JobPart0Mgr and part0Plan are redundant.

mohsha-msft (approved):
Looks good to me. 😄 
Please just verify this [issue](https://github.com/Azure/azure-storage-azcopy/issues/1408) linked with PR is resolved.

zezha-msft (approved):

PR #1410, opened 2021-04-16
Implement OAuth in the modern test framework

adreed-msft:
Makes use of SPN and One-Shot login to enable OAuth in the modern OAuth test framework.

This works by generating valid credential type combos with a requested source and destination set of credentials.

### DIFF
--- a/e2etest/declarativeRunner.go
+++ b/e2etest/declarativeRunner.go
@@ -24,6 +24,8 @@ import (
 	"fmt"
 	"os"
 	"testing"
+
+	"github.com/Azure/azure-storage-azcopy/v10/common"
 )
 
 // This declarative test runner adds a layer on top of e2etest/base. The added layer allows us to test in a declarative style,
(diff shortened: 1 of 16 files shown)


e2etest/declarativeRunner.go
```diff
@@ -24,26 +24,90 @@ import (
...
 )
 
 // This declarative test runner adds a layer on top of e2etest/base. The added layer allows us to test in a declarative style,
 // saying what to do, but not how to do it.
 // In particular, it lets one test cover a range of different source/dest types, and even cover both sync and copy.
 // See first test in zt_enumeration for an annotated example.
 
+var validCredTypesPerLocation = map[common.Location][]common.CredentialType {
+	common.ELocation.Unknown(): {common.ECredentialType.Unknown()},
+	common.ELocation.File(): {common.ECredentialType.Anonymous()},
```
nakulkar-msft, line 38: Filerest now supports AAD. This will need changes in azcopy as well.

e2etest/declarativeRunner.go
```diff
@@ -54,34 +118,43 @@ func RunScenarios(
...
-			subtestName := fmt.Sprintf("%s-%s", op, fromTo)
-
-			hsToUse := hooks{}
-			if hs != nil {
-				hsToUse = *hs
+			credentialTypes := getValidCredCombinationsForFromTo(fromTo, requestedCredentialTypesSrc, requestedCredentialTypesDst)
+
+			scenarioList := make([]scenario, 0)
+
+			for _, credTypes := range credentialTypes {
```
nakulkar-msft, line 128: It might be an overkill to execute both credential types for all testcases. Can we run oauth on selective scenarios? Like the tests which focus on Frontend filters - execute same path on both credentials

e2etest/declarativeScenario.go
```diff
@@ -43,6 +43,7 @@ type scenario struct {
 	operation           Operation
 	validate            Validate
 	fromTo              common.FromTo
+	credTypes           [2]common.CredentialType
```
nakulkar-msft, line 46: Could have had an option bag like thing here. I dont think there will be many changes here. But any change here, we'llhave to change a lot of files.

nakulkar-msft (changes requested):
We should not be running all combinations on all testcases.

mohsha-msft (changes requested):
Changes do look good to me. However I noticed that we're running all the existing test cases with new login  which I think is an overkill. It doubled the running time of E2E tests. Can we have just some test cases that can validate login flow without implementing it for the existing test cases?

adreed-msft:
I agree that it's overkill. I'll limit OAuth to some simpler test cases.

mohsha-msft:
Can we also change the base branch from `master` to `dev`?

e2etest/declarativeRunner.go
```diff
@@ -33,15 +33,80 @@ import (
...
+	common.ELocation.Local():   {common.ECredentialType.Anonymous()},
+	common.ELocation.Pipe():    {common.ECredentialType.Anonymous()},
+	common.ELocation.S3():      {common.ECredentialType.S3AccessKey()},
+	common.ELocation.GCP():     {common.ECredentialType.GoogleAppCredentials()},
+}
+
+var allCredentialTypes []common.CredentialType = nil
+var anonymousAuthOnly = []common.CredentialType{common.ECredentialType.Anonymous()}
+
+func getValidCredCombinationsForFromTo(fromTo common.FromTo, requestedCredentialTypesSrc, requestedCredentialTypesDst []common.CredentialType) [][2]common.CredentialType {
```
mohsha-msft, line 50: I'm a bit confused with this implementation.
1. If we have `fromTo`, then `validCredTypesPerLocation[fromTo.from]` and `validCredTypesPerLocation[fromTo.To]` will give me respective lists of source and destination supported credTypes.
2. We'll just have to filter out all elements of `requestedCredentialTypesSrc` that area not supported. (`validCredTypesPerLocation[fromTo.from]`). Same step for destination.
3. Return back the filtered output but doing a cross product of both filtered lists.

e2etest/declarativeRunner.go
```diff
@@ -73,44 +138,54 @@ func RunScenarios(
...
+					validate:            validate,
+					p:                   p, // copies them, because they are a struct. This is what we need, since they may be morphed while running
+					hs:                  hsToUse,
+					fs:                  fs.DeepCopy(),
+					needResume:          operations&eOperation.Resume() != 0,
+					stripTopDir:         p.stripTopDir,
+					isSourceAcc:         isSourceAcc,
+				}
+
+				scenarioList = append(scenarioList, s)
```
mohsha-msft, line 185: Nit: Append to scenarios only and not create a 2D array, maybe? There's no grouping logic required when we're running all tests in parallel.

mohsha-msft: Or have few comments explaining what 2D array of scenarios contain.

e2etest/declarativeScenario.go
```diff
@@ -186,21 +187,28 @@ func (s *scenario) runAzCopy() {
...
+		s.state.source.getParam(s.stripTopDir, s.credTypes[0] == common.ECredentialType.Anonymous(), tf.objectTarget),
+		s.state.dest.getParam(false, s.credTypes[1] == common.ECredentialType.Anonymous(), common.IffString(tf.destTarget != "", tf.destTarget, tf.objectTarget)),
+		s.credTypes[0] == common.ECredentialType.OAuthToken() || s.credTypes[1] == common.ECredentialType.OAuthToken(), // needsOAuth
 		afterStart, s.chToStdin)
 
 	if !wasClean {
 		s.a.AssertNoErr(err, "running AzCopy")
 	}
 
+	// // Generally, a cancellation is done when auth fails.
```
mohsha-msft, line 203: Nit: Remove comments if not rquired.

e2etest/declarativeRunner.go
```diff
@@ -33,15 +33,80 @@ import (
 // In particular, it lets one test cover a range of different source/dest types, and even cover both sync and copy.
 // See first test in zt_enumeration for an annotated example.
 
+var validCredTypesPerLocation = map[common.Location][]common.CredentialType{
+	common.ELocation.Unknown(): {common.ECredentialType.Unknown(), common.ECredentialType.Anonymous(), common.ECredentialType.OAuthToken()}, // Delete!
+	common.ELocation.File():    {common.ECredentialType.Anonymous()},
+	common.ELocation.Blob():    {common.ECredentialType.Anonymous(), common.ECredentialType.OAuthToken()},
+	common.ELocation.BlobFS():  {common.ECredentialType.Anonymous(), common.ECredentialType.OAuthToken()}, // todo: currently, account key auth isn't even supported in e2e tests.
```
mohsha-msft, line 40: Yeah `SharedKey` isn't supported.

mohsha-msft (commented):
Changes look straightforward.  :shipit: 
Can you check why the time taken by CI has increase from avg 18min to 24-25min if you have some time? We haven't added new tests.

mohsha-msft:
Triggering CI few times for the sake of consistency check.

mohsha-msft:
Wait, why is run_e2e_tests under Test_On_Ubuntu?
image
Perhaps a problem in YAML syntax?

adreed-msft:
> Wait, why is run_e2e_tests under Test_On_Ubuntu?

You're right, there's no need for that.

e2etest/declarativeRunner.go
adreed-msft: I... have questions about it myself. I wrote this almost a year ago now, so it's rather fuzzy in my head.

I *believe* the intent was to prevent a test author from inputting nonsensical auth types. I think the last bit where we directly check against _every_ credential type is not good code upon coming back to this. I'll rewrite this function a bit.

adreed-msft: Upon a bit more thought, the original implementation may be better. The set of possibilities available to either end of the from/to is always smaller than the requested set (assuming the requested set is allCredentialTypes).

mohsha-msft (approved):
🟢 LGTM

nakulkar-msft (approved):

PR #1414, opened 2021-04-19
[Week of Quality] Adding more tests and fixing issues

mohsha-msft:


cmd/copy.go
```diff
@@ -526,7 +526,8 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	// 2. `blob-tags` is not present as they create conflicting scenario of whether to preserve blob tags from the source or set user defined tags on the destination
 	if raw.s2sPreserveBlobTags {
 		if cooked.fromTo.From() != common.ELocation.Blob() || cooked.fromTo.To() != common.ELocation.Blob() {
```
adreed-msft, line 528: Simplification of if statement, `cooked.fromTo != EFromTo.BlobBlob()`

cmd/copy.go
```diff
@@ -526,7 +526,8 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
...
-			return cooked, errors.New("either source or destination is not a blob storage. blob index tags is a property of blobs only therefore both source and destination must be blob storage")
+			return cooked, errors.New("either source or destination is not a blob storage. " +
```
adreed-msft, line 529: Perhaps we should remove "a" here? Or just this entire first sentence.

cmd/sync.go
```diff
@@ -65,6 +65,8 @@ type rawSyncCmdArgs struct {
 	deleteDestination string
 
 	s2sPreserveAccessTier bool
+
+	blobTags string
```
adreed-msft, line 69: I don't know how I feel about blob tags on sync.

It would make sense if it were a way to filter the source. But, in this case, it looks like it's not. Which feels against the spirit of sync.

adreed-msft (changes requested):

cmd/sync.go
zezha-msft: I would normally advise against adding info for dest that src didn't have, but I just realized that tags might be a special case.

If we allow tags in sync, then the user can tag blobs that get copied with a specific value, thus allowing an interesting scenario: if they want, they can distinguish the batches of blobs that got moved over, and do processing on them later. I have no idea if this is an actual scenario for cx, but it sounds feasible.

PR #1422, opened 2021-04-22
`azcopy jobs clean` exit silently if no JobIDs are present

mohsha-msft:


nakulkar-msft (approved):
Verify that other users of ListJobs can handle zero jobs being returned.

PR #1436, opened 2021-05-06
Interpret DFS endpoints as blob for Sync command

nakulkar-msft:


cmd/sync.go
```diff
@@ -127,7 +127,18 @@ func (raw *rawSyncCmdArgs) cook() (cookedSyncCmdArgs, error) {
 	// this if statement ladder remains instead of being separated to help determine valid combinations for sync
 	// consider making a map of valid source/dest combos and consolidating this to generic source/dest setups, akin to the lower if statement
 	// TODO: if expand the set of source/dest combos supported by sync, update this method the declarative test framework:
-	// TODO: add support for account-to-account operations (for those from-tos that support that)
+
+	/* We support DFS by using blob end-point of the account. We replace dfs by blob in src and dst */
+	if loc := inferArgumentLocation(raw.src); loc == common.ELocation.BlobFS() {
+		raw.src = strings.Replace(raw.src, ".dfs", ".blob", 1)
```
adreed-msft, line 133: Since we're choosing to "cleverly" change endpoints in sync, perhaps we should consider enabling such behavior for blob S2S as well?

cmd/zt_sync_cmd_test.go
```diff
@@ -0,0 +1,39 @@
...
+func (s *cmdIntegrationSuite) TestSyncS2SWithDFS(c *chk.C) {
+	src := "https://myaccount1.dfs.core.windows.net/container1/"
+	dst := "https://myaccount2.dfs.core.windows.net/container2/"
+
+	raw := getDefaultSyncRawInput(src, dst)
+
+	cooked, err := raw.cook()
+
+	c.Assert(err, chk.IsNil)
+	c.Assert(cooked.fromTo, chk.Equals, common.EFromTo.BlobBlob())
```
adreed-msft, line 38: Our testing should really be more in depth than this for something like this (even though such operations by all means should work as expected). You should set up a job like this in the declarative test suite, just on the off-chance something is not happy.

adreed-msft (changes requested):

cmd/zt_sync_cmd_test.go
nakulkar-msft: That would mean we're not confident in other code, testing or both. Unless there is a reason other than to be on safer side, I'd be wary against adding test to a suite which is already overloaded.

cmd/sync.go
nakulkar-msft: I'm skeptical even on this. @Ze had reasons to avoid copy. I'll reconfirm

cmd/copy.go
```diff
@@ -255,6 +255,20 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 		azcopyScanningLogger.CloseLog()
 	})
 
+	/* We support DFS by using blob end-point of the account. We replace dfs by blob in src and dst */
+	if src,dst := inferArgumentLocation(raw.src), inferArgumentLocation(raw.dst);
+				src == common.ELocation.BlobFS() || dst == common.ELocation.BlobFS() {
+		if src == common.ELocation.BlobFS() && dst != common.ELocation.Local() {
```
zezha-msft, line 261: minor: add a comment to explain that we are trying to phase out the usage of dfs endpoint in general, hence why we are using blob endpoint even when the source is dfs.

cmd/zt_copy_s2smigration_test.go
```diff
@@ -1271,3 +1273,84 @@ func (s *cmdIntegrationSuite) TestS2SCopyFromSingleAzureFileToBlobContainer(c *c
...
+	c.Assert(err, chk.IsNil)
+
+	rawSync := getDefaultSyncRawInput(dirURLWithSASSource.String(), dirURLWithSAS.String())
+	runSyncAndVerify(c, rawSync, func(err error) {
+		c.Assert(err, chk.IsNil)
+
+		// validate that the right number of transfers were scheduled
+		c.Assert(len(mockedRPC.transfers), chk.Equals, 2)
+
+		//c.Assert(mockedRPC.transfers[0].Destination, chk.Equals, "/file2")
```
zezha-msft, line 1353: minor: remove commented out code please

zezha-msft (approved):
lgtm

+1 on adding a test to the new E2E framework (can be done after merging PR), the new framework is super fast and would give more confidence on this scenario.

cmd/zt_sync_cmd_test.go
zezha-msft: +1 on adding a test to the new E2E framework (can be done after merging PR), the new framework is super fast and would give more confidence on this scenario.

mohsha-msft: Agreed. Let's do an AllS2S() or AllPairs() test for DFS endpoint for more confidence in the scenario.

cmd/zt_copy_s2smigration_test.go
nakulkar-msft: E2E tests require implementation of resource manages for Gen2. That would be a significant feature in test-infra, and I dont want to include it here. The one in this file will give us similar confidence. Resource managers for Gen2 will be taken care of.

cmd/zt_sync_cmd_test.go
nakulkar-msft: E2E tests require implementation of resource manages for Gen2. That would be a significant feature in test-infra, and I dont want to include it here. The one in this file will give us similar confidence, we still execute whole copy/sync. Resource managers for Gen2 will be taken care of.

mohsha-msft (approved):

mohsha-msft (approved):

PR #1439, opened 2021-05-10
[AzCopyV10][Feature] Add mirror mode to sync

mohsha-msft:


cmd/sync.go
```diff
@@ -64,6 +64,9 @@ type rawSyncCmdArgs struct {
 	// otherwise the user is prompted to make a decision
 	deleteDestination string
 
+	// this flag is to disable comparator and overwrite files at destination irrespective
+	mirrorMode bool
```
mohsha-msft, line 68: I'll happily change this name to something else.

nakulkar-msft (approved):

adreed-msft (approved):

cmd/sync.go
nakulkar-msft: Mirror mode - as far as this change goes- does copy alone, and you'll have to additionally use delete-destination with --mirrorMode to achieve what we think is intuitional.

Kapanther:
I guess the difference between `mirror` and `overwrite` is context between a "`sync`" and a "`copy`"

Does `overwrite=true` skip files with the same Date modified Stamps? I'm guessing not... But not sure.

Where as mirror-mode in a `sync` context still performs the date modified check, but will not overwrite files that have the same file stamp. Just files with older or newer stamps.

phrak:
As a long-time robocopy user, "Mirror" mode should be consistent with the robocopy behaviour.
Mirror = Copy + Delete from destination + Overwrite if different.

AZCopy still has a major functionality gap, not having an "Overwrite if different" capability. 
Hopefully the new "Mirror" mode will be able to handle this scenario.

Issue #964, opened 2020-04-17
AzCopy fail to copy multiple items from a file share snapshot


JasonYeMSFT:
### Which version of the AzCopy was used? 
10.4.0

### Which platform are you using?
macOS

### What command did you run?
./azcopy copy "https://easybugs110.file.core.windows.net/fs10/*?sv=2018-03-28&ss=f&srt=co&sp=rl&se=2020-04-23T23%3A18%3A35Z&sig=<REDACTED>&sharesnapshot=2020-03-05T18%3A06%3A58.0000000Z" "https://easybugs110.file.core.windows.net/fs10/?se=2020-05-16T23%3A19%3A38Z&sp=rwl&sv=2018-03-28&sr=s&sig=LP5STGXpBMabAHQVw1zd8%2BxVVOFyWai8wjaecS92AEY%3D" --overwrite=prompt --recursive --list-of-files "/var/folders/ww/bk24s62x4n3g9vhzt6czwndr0000gn/T/stg-exp-azcopy-cbf7ea31-f07f-417f-a784-4028e1558ec5.txt" --cap-mbps 123 --s2s-preserve-access-tier=false;

List of file contains:
512B copy 100.txt
blobs2.txt

### What problem was encountered?
AzCopy fails with the log file
[c0910ca2-36ce-5845-5893-cef4abf7c4c7.log](https://github.com/Azure/azure-storage-azcopy/files/4494775/c0910ca2-36ce-5845-5893-cef4abf7c4c7.log).

### How can we reproduce the problem in the simplest way?
Try copying two items from a file share snapshot to its root.

### Have you found a mitigation/solution?
Use Azure Storage SDK.

nakulkar-msft:
Fixed with #1442

PR #1442, opened 2021-05-11
[AzCopyV10][StgExp] Cannot copy file from sharesnapshot to current share

siminsavani-msft:
Sharesnapshot can be copied to current share with this small fix.

### DIFF
--- a/ste/xfer-anyToRemote-file.go
+++ b/ste/xfer-anyToRemote-file.go
@@ -168,7 +168,7 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
 			dstRQ := dstURL.Query()
 
 			// note that query is now all lower case at this point
-			if len(srcRQ["snapshot"]) == 0 && len(srcRQ["versionid"]) == 0 && len(dstRQ["snapshot"]) == 0 && len(dstRQ["versionid"]) == 0 {
+			if len(srcRQ["sharesnapshot"]) == 0 && len(srcRQ["versionid"]) == 0 && len(dstRQ["sharesnapshot"]) == 0 && len(dstRQ["versionid"]) == 0 {
 				jptm.LogSendError(info.Source, info.Destination, "Transfer source and destination are the same, which would cause data loss. Aborting transfer.", 0)
 				jptm.SetStatus(common.ETransferStatus.Failed())
 				jptm.ReportTransferDone()
--- a/ste/xfer-anyToRemote-file.go
+++ b/ste/xfer-anyToRemote-file.go
@@ -167,8 +167,13 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
 			srcRQ := srcURL.Query()
 			dstRQ := dstURL.Query()
 
-			// note that query is now all lower case at this point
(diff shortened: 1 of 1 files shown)


nakulkar-msft (approved):
This change warrants a testcase. Check if you can add one.

mohsha-msft (approved):

ste/xfer-anyToRemote-file.go
```diff
@@ -168,7 +168,7 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
 			dstRQ := dstURL.Query()
 
 			// note that query is now all lower case at this point
-			if len(srcRQ["snapshot"]) == 0 && len(srcRQ["versionid"]) == 0 && len(dstRQ["snapshot"]) == 0 && len(dstRQ["versionid"]) == 0 {
+			if len(srcRQ["sharesnapshot"]) == 0 && len(srcRQ["versionid"]) == 0 && len(dstRQ["sharesnapshot"]) == 0 && len(dstRQ["versionid"]) == 0 {
```
adreed-msft, line 171: Something feels fishy here. As far as I know, users can't write to old snapshots and versions. But, out of caution (in case the service were ever to allow such functionality), I feel like we should be comparing these items, INSTEAD of just checking they're nonzero.

adreed-msft: What do other contributors think?

adreed-msft (changes requested):

adreed-msft:
We _absolutely_ should have a test case, @nakulkar-msft.

ste/xfer-anyToRemote-file.go
mohsha-msft: Hey @siminsavani-msft ,
I think @adreed-msft is right. This if-else condition is never going to execute because -
1. Snapshots are immutable copies taken at a point of time. We can read copy from one snapshot but not change them.
2. Similarly we cannot edit a particular version. Think of versions as a Singly LinkedList with head pointer pointing to the latest version. The only way we can update a version is to move the header back to that version except in this case, we create new version by discarding changes.

mohsha-msft:
Is there any linked issue with this PR?

ste/xfer-anyToRemote-file.go
nakulkar-msft: Snapshot being immutable is universally accepted, and we *should* assume same as well. Comments on line 165/166 explain the rest. If the source and destination are same container/filestore, and if we're not reading from snapshot(or versionid) - throw that source and destination are same. Else, we'll let the service dictate error.

siminsavani-msft:
> Is there any linked issue with this PR?

@mohsha-msft I believe that this is the issue: https://github.com/Azure/azure-storage-azcopy/issues/964

ste/xfer-anyToRemote-file.go
siminsavani-msft: This makes complete sense! The sharesnapshot & versionID of both source and destination should be compared instead of the current behavior where their length is compared to 0. Will update this! Thanks y'all!

siminsavani-msft:
> This change warrants a testcase. Check if you can add one.

So this scenario warrants an e2e test but is not currently supported. Instead, I've manually tested going from (a) ShareSnapshot to Current Share and (b) ShareSnapshot to itself, ensured that snapshots are immutable

PR #1445, opened 2021-05-13
Fixed Pipe Upload: Added Metadata, Blob index tags, CPK

mohsha-msft:


cmd/copy.go
```diff
@@ -1182,9 +1182,17 @@ func (cca *cookedCopyCmdArgs) processRedirectionUpload(blobResource common.Resou
...
 	blockBlobUrl := azblob.NewBlockBlobURL(*u, p)
+	metadata, err := common.UnMarshalToCommonMetadata(cca.metadata)
+	if err != nil {
+		return err
+	}
+	blobTags := cca.blobTags
 	_, err = azblob.UploadStreamToBlockBlob(ctx, os.Stdin, blockBlobUrl, azblob.UploadStreamToBlockBlobOptions{
-		BufferSize: int(blockSize),
-		MaxBuffers: pipingUploadParallelism,
+		BufferSize:               int(blockSize),
```
adreed-msft, line 1191: Should we consider the additional similar simple to support parameters, e.g. blob tier, content headers (content type, disposition, encoding, etc.), cache control?

mohsha-msft: Ohh my bad. I don't know why I missed that

cmd/zc_traverser_list.go
```diff
@@ -61,7 +61,7 @@ func (l *listTraverser) traverse(preprocessor objectMorpher, processor objectPro
 		}
 
 		// listTraverser will only ever execute on the source
-		if childTraverser.isDirectory(true) && !l.recursive {
+		if !l.recursive && childTraverser.isDirectory(true) {
```
mohsha-msft, line 64: Minor perf improvement.

siminsavani-msft (commented):
LGTM!

siminsavani-msft (approved):

zezha-msft (approved):

PR #1448, opened 2021-05-20
[AzCopyV10][StgExp] Add flag to disable automatic decoding of illegal chars on Windows

siminsavani-msft:
- Added new copy command variable called **disableAutoDecoding** that is set by the flag **disable-auto-decoding** 
- The **disable-auto-decoding** flag is false by default
- Added **disableAutoDecoding** as a boolean argument for the _pathEncodeRules_ where the bool is used to verify if flag has been set
- Updated usages of _pathEncodeRules_ to have the **disableAutoDecoding**
- Wrote a test that verifies that encoded file names are not decoded when **disableAutoDecoding** is set to false.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -147,6 +147,9 @@ type rawCopyCmdArgs struct {
 	// whether to include blobs that have metadata 'hdi_isfolder = true'
 	includeDirectoryStubs bool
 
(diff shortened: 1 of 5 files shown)


adreed-msft (approved):

mohsha-msft (approved):
LGTM

PR #1378, opened 2021-03-25
Adding PutBlobFromUrl for AzCopy

siminsavani-msft:
- Refactoring AzCopy to use PutBlobFromUrl

PR #1452, opened 2021-05-26
[AzCopyV10][perf] Add support for put blob from URL - small blob perf

siminsavani-msft:
New Clean PR for **PutBlobFromUrl PR** (https://github.com/Azure/azure-storage-azcopy/pull/1378)

- Replaced **CopyBlobFromURL API** with **PutBlobFromURL API** and updated the _generateStartPutBlobFromURL_ method accordingly.
- Updated the _GenerateCopyFunc_ with new _generateStartPutBlobFromURL_
- Retained _generateStartPutBlobFromURL_ from files (Files should use PutBlockFromUrl + CommitList)

### DIFF
--- a/ste/sender-blockBlobFromURL.go
+++ b/ste/sender-blockBlobFromURL.go
@@ -184,6 +184,11 @@ func (c *urlToBlockBlobCopier) generateStartPutBlobFromURL(id common.ChunkID, bl
 			return
 		}
 
+		if separateSetTagsRequired {
+			if _, err := c.destBlockBlobURL.SetTags(c.jptm.Context(), nil, nil, nil, c.blobTagsToApply); err != nil {
+				c.jptm.Log(pipeline.LogWarning, err.Error())
+			}
+		}
 	})
 }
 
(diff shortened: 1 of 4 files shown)


ste/sender-blockBlobFromURL.go
```diff
@@ -68,20 +68,15 @@ func (c *urlToBlockBlobCopier) GenerateCopyFunc(id common.ChunkID, blockIndex in
 		setPutListNeed(&c.atomicPutListIndicator, putListNotNeeded)
 		return c.generateCreateEmptyBlob(id)
 	}
-
-	if c.NumChunks() == 1 && adjustedChunkSize <= int64(azblob.BlockBlobMaxUploadBlobBytes) &&
-		c.blockBlobSenderBase.jptm.FromTo() == common.EFromTo.GCPBlob() {
+	// Small blobs from all sources will be copied over to destination using PutBlobFromUrl with the exception of files
+	fromTo := c.blockBlobSenderBase.jptm.FromTo()
+	if c.NumChunks() == 1 && adjustedChunkSize <= int64(azblob.BlockBlobMaxUploadBlobBytes) && fromTo.From() != common.ELocation.File() {
```
mohsha-msft, line 73: Just curious: Are we only disabling `fromTo.From() != ELocation.File()` ? I thought it was `fromTo.To() != ELocation.File()` as well. Please confirm once.

nakulkar-msft: This code executes only when destination is blobstore.

nakulkar-msft (approved):

ste/sender-blockBlobFromURL.go
mohsha-msft: Ohh you're right.

mohsha-msft (approved):

PR #1455, opened 2021-05-27
[AzCopyV10][UX] Tolerate enumeration errors for Azure Files + [AzCopyV10] [Github] AzCopy is trying to scan excluded path

mohsha-msft:


siminsavani-msft (approved):
LGTM! Is there a test written for this or are you manually testing this change?

PR #1464, opened 2021-06-03
Include and exclude regex flag 

t-cguruceaga-msft:
copy.go & sync.go -- added includeRegex and excludeRegex for raw and cook
copyEnumeratorInit.go & syncEnumerator.go -- calls zc_filter.go to append filters
created tests in both zt_sync_blob_blob_test.go & zt_copy_blob_download_test.go

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -72,8 +72,8 @@ type rawCopyCmdArgs struct {
 	exclude               string
 	includePath           string // NOTE: This gets handled like list-of-files! It may LOOK like a bug, but it is not.
 	excludePath           string
-	includeRegex	   	  string
-	excludeRegex 		  string
+	includeRegex          string
+	excludeRegex          string
 	includeFileAttributes string
 	excludeFileAttributes string
 	includeBefore         string
@@ -984,8 +984,8 @@ type cookedCopyCmdArgs struct {
 	includeAfter          *time.Time
 
 	// include/exclude filters with regular expression (also for sync)
-	includeRegex		  []string
-	excludeRegex		  []string
+	includeRegex []string
+	excludeRegex []string
 
 	// list of version ids
 	listOfVersionIDs chan string
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -72,6 +72,8 @@ type rawCopyCmdArgs struct {
 	exclude               string
 	includePath           string // NOTE: This gets handled like list-of-files! It may LOOK like a bug, but it is not.
 	excludePath           string
+	includeRegex	   	  string
+	excludeRegex 		  string
 	includeFileAttributes string
 	excludeFileAttributes string
 	includeBefore         string
@@ -804,6 +806,9 @@ func (raw rawCopyCmdArgs) cook() (cookedCopyCmdArgs, error) {
 	cooked.includeFileAttributes = raw.parsePatterns(raw.includeFileAttributes)
 	cooked.excludeFileAttributes = raw.parsePatterns(raw.excludeFileAttributes)
 
+	cooked.includeRegex = raw.parsePatterns(raw.includeRegex)
+	cooked.excludeRegex = raw.parsePatterns(raw.excludeRegex)
+
 	return cooked, nil
 }
 
@@ -977,6 +982,10 @@ type cookedCopyCmdArgs struct {
 	excludeFileAttributes []string
 	includeBefore         *time.Time
 	includeAfter          *time.Time
+
+	// include/exclude filters with regular expression (also for sync)
+	includeRegex		  []string
+	excludeRegex		  []string
 
 	// list of version ids
 	listOfVersionIDs chan string
@@ -1755,6 +1764,8 @@ func init() {
 		"This option does not support wildcard characters (*). Checks relative path prefix (For example: myFolder;myFolder/subDirName/file.pdf).")
 	cpCmd.PersistentFlags().StringVar(&raw.excludePath, "exclude-path", "", "Exclude these paths when copying. "+ // Currently, only exclude-path is supported alongside account traversal.
 		"This option does not support wildcard characters (*). Checks relative path prefix(For example: myFolder;myFolder/subDirName/file.pdf). When used in combination with account traversal, paths do not include the container name.")
+	cpCmd.PersistentFlags().StringVar(&raw.includeRegex, "include-regex", "", "Include only the files that align with regular expressions. Separate regular expressions with ';'.")
+	cpCmd.PersistentFlags().StringVar(&raw.excludeRegex, "exclude-regex", "", "Exclude all the files that align with regular expressions. Separate regular expressions with ';'.")
 	// This flag is implemented only for Storage Explorer.
 	cpCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "Defines the location of text file which has the list of only files to be copied.")
 	cpCmd.PersistentFlags().StringVar(&raw.exclude, "exclude-pattern", "", "Exclude these files when copying. This option supports wildcard characters (*)")
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -1764,8 +1764,8 @@ func init() {
 		"This option does not support wildcard characters (*). Checks relative path prefix (For example: myFolder;myFolder/subDirName/file.pdf).")
 	cpCmd.PersistentFlags().StringVar(&raw.excludePath, "exclude-path", "", "Exclude these paths when copying. "+ // Currently, only exclude-path is supported alongside account traversal.
 		"This option does not support wildcard characters (*). Checks relative path prefix(For example: myFolder;myFolder/subDirName/file.pdf). When used in combination with account traversal, paths do not include the container name.")
-	cpCmd.PersistentFlags().StringVar(&raw.includeRegex, "include-regex", "", "Include only the files that align with regular expressions. Separate regular expressions with ';'.")
-	cpCmd.PersistentFlags().StringVar(&raw.excludeRegex, "exclude-regex", "", "Exclude all the files that align with regular expressions. Separate regular expressions with ';'.")
+	cpCmd.PersistentFlags().StringVar(&raw.includeRegex, "include-regex", "", "Include only the relative path of the files that align with regular expressions. Separate regular expressions with ';'.")
+	cpCmd.PersistentFlags().StringVar(&raw.excludeRegex, "exclude-regex", "", "Exclude all the relative path of the files that align with regular expressions. Separate regular expressions with ';'.")
 	// This flag is implemented only for Storage Explorer.
 	cpCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "Defines the location of text file which has the list of only files to be copied.")
 	cpCmd.PersistentFlags().StringVar(&raw.exclude, "exclude-pattern", "", "Exclude these files when copying. This option supports wildcard characters (*)")
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -711,8 +711,8 @@ func init() {
 		"This option does not support wildcard characters (*). Checks relative path prefix(For example: myFolder;myFolder/subDirName/file.pdf).")
 	syncCmd.PersistentFlags().StringVar(&raw.includeFileAttributes, "include-attributes", "", "(Windows only) Include only files whose attributes match the attribute list. For example: A;S;R")
 	syncCmd.PersistentFlags().StringVar(&raw.excludeFileAttributes, "exclude-attributes", "", "(Windows only) Exclude files whose attributes match the attribute list. For example: A;S;R")
-	syncCmd.PersistentFlags().StringVar(&raw.includeRegex, "include-regex", "", "Include files that match with the regular expressions. Separate regular expressions with ';'.")
-	syncCmd.PersistentFlags().StringVar(&raw.excludeRegex, "exclude-regex", "", "Exclude files that match with the regular expressions. Separate regular expressions with ';'.")
+	syncCmd.PersistentFlags().StringVar(&raw.includeRegex, "include-regex", "", "Include the relative path of the files that match with the regular expressions. Separate regular expressions with ';'.")
+	syncCmd.PersistentFlags().StringVar(&raw.excludeRegex, "exclude-regex", "", "Exclude the relative path of the files that match with the regular expressions. Separate regular expressions with ';'.")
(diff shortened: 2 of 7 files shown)


cmd/sync.go
```diff
@@ -51,6 +51,8 @@ type rawSyncCmdArgs struct {
 	excludeFileAttributes string
 	legacyInclude         string // for warning messages only
 	legacyExclude         string // for warning messages only
+	includeRegex		  string
+	excludeRegex		  string
```
siminsavani-msft, line 55: minor: double check if spacing is consistent

cmd/sync.go
```diff
@@ -329,6 +334,8 @@ type cookedSyncCmdArgs struct {
 	excludePaths          []string
 	includeFileAttributes []string
 	excludeFileAttributes []string
+	includeRegex		  []string
+	excludeRegex		  []string
```
siminsavani-msft, line 338: minor: double check if spacing is consistent

cmd/zt_copy_blob_download_test.go
```diff
@@ -455,3 +455,211 @@ func (s *cmdIntegrationSuite) TestDownloadBlobContainerWithPattern(c *chk.C) {
...
+	rawContainerURLWithSAS.Path = path.Join(rawContainerURLWithSAS.Path, string([]byte{0x00}))
+	containerString := strings.ReplaceAll(rawContainerURLWithSAS.String(), "%00", "*")
+	raw := getDefaultCopyRawInput(containerString, dstDirName)
+	raw.recursive = true
+	raw.includeRegex = "es{4,}"
+
+ 	runCopyAndVerify(c, raw, func(err error) {
+		c.Assert(err, chk.IsNil)
+		// validate that the right number of transfers were scheduled
+		c.Assert(len(mockedRPC.transfers), chk.Equals, len(blobsToInclude))
```
siminsavani-msft, line 494: Is there a way to check that the actual values of the transfers match the values of blobsToInclude? 

Length is a bit unreliable.

cmd/zt_copy_blob_download_test.go
```diff
@@ -455,3 +455,211 @@ func (s *cmdIntegrationSuite) TestDownloadBlobContainerWithPattern(c *chk.C) {
...
+	rawContainerURLWithSAS.Path = path.Join(rawContainerURLWithSAS.Path, string([]byte{0x00}))
+	containerString := strings.ReplaceAll(rawContainerURLWithSAS.String(), "%00", "*")
+	raw := getDefaultCopyRawInput(containerString, dstDirName)
+	raw.recursive = true
+	raw.includeRegex = "es{4,};^zxc"
+
+	runCopyAndVerify(c, raw, func(err error) {
+		c.Assert(err, chk.IsNil)
+		// validate that the right number of transfers were scheduled
+		c.Assert(len(mockedRPC.transfers), chk.Equals, len(blobsToInclude))
```
siminsavani-msft, line 536: check actual values here

cmd/zt_copy_blob_download_test.go
```diff
@@ -455,3 +455,211 @@ func (s *cmdIntegrationSuite) TestDownloadBlobContainerWithPattern(c *chk.C) {
...
+	containerString := strings.ReplaceAll(rawContainerURLWithSAS.String(), "%00", "*")
+	raw := getDefaultCopyRawInput(containerString, dstDirName)
+	raw.recursive = true
+	raw.includeRegex = ""
+	raw.excludeRegex = ""
+
+	runCopyAndVerify(c, raw, func(err error) {
+		c.Assert(err, chk.IsNil)
+		// validate that the right number of transfers were scheduled
+		c.Assert(len(mockedRPC.transfers) , chk.Equals, len(blobsToInclude))
```
siminsavani-msft, line 576: check actual values here

cmd/zt_copy_blob_download_test.go
```diff
@@ -455,3 +455,211 @@ func (s *cmdIntegrationSuite) TestDownloadBlobContainerWithPattern(c *chk.C) {
...
+	rawContainerURLWithSAS.Path = path.Join(rawContainerURLWithSAS.Path, string([]byte{0x00}))
+	containerString := strings.ReplaceAll(rawContainerURLWithSAS.String(), "%00", "*")
+	raw := getDefaultCopyRawInput(containerString, dstDirName)
+	raw.recursive = true
+	raw.excludeRegex = "es{4,}"
+
+	runCopyAndVerify(c, raw, func(err error) {
+		c.Assert(err, chk.IsNil)
+		// validate that only blobsTo
+		c.Assert(len(mockedRPC.transfers), chk.Equals, len(blobsToInclude) )
```
siminsavani-msft, line 618: check actual values here

cmd/zt_copy_blob_download_test.go
```diff
@@ -455,3 +455,211 @@ func (s *cmdIntegrationSuite) TestDownloadBlobContainerWithPattern(c *chk.C) {
...
+	rawContainerURLWithSAS.Path = path.Join(rawContainerURLWithSAS.Path, string([]byte{0x00}))
+	containerString := strings.ReplaceAll(rawContainerURLWithSAS.String(), "%00", "*")
+	raw := getDefaultCopyRawInput(containerString, dstDirName)
+	raw.recursive = true
+	raw.excludeRegex = "es{4,};o(g)"
+
+	runCopyAndVerify(c, raw, func(err error) {
+		c.Assert(err, chk.IsNil)
+		// validate that the right number of transfers were scheduled
+		c.Assert(len(mockedRPC.transfers), chk.Equals, len(blobsToInclude) )
```
siminsavani-msft, line 660: check actual values here

cmd/zt_sync_blob_blob_test.go
```diff
@@ -681,3 +681,114 @@ func (s *cmdIntegrationSuite) TestSyncS2SADLSDirectory(c *chk.C) {
...
+	// construct the raw input to simulate user input
+	srcContainerURLWithSAS := scenarioHelper{}.getRawContainerURLWithSAS(c, srcContainerName)
+	dstContainerURLWithSAS := scenarioHelper{}.getRawContainerURLWithSAS(c, dstContainerName)
+	raw := getDefaultSyncRawInput(srcContainerURLWithSAS.String(), dstContainerURLWithSAS.String())
+	raw.includeRegex = includeString
+
+	// verify that only the blobs specified by the include flag are synced
+	runSyncAndVerify(c, raw, func(err error) {
+		c.Assert(err, chk.IsNil)
+		validateS2SSyncTransfersAreScheduled(c, "", "", blobsToInclude, mockedRPC)
```
siminsavani-msft, line 716: You should check if the values of blobsToInclude match the results.

cmd/zt_sync_blob_blob_test.go
```diff
@@ -681,3 +681,114 @@ func (s *cmdIntegrationSuite) TestSyncS2SADLSDirectory(c *chk.C) {
...
+	// construct the raw input to simulate user input
+	srcContainerURLWithSAS := scenarioHelper{}.getRawContainerURLWithSAS(c, srcContainerName)
+	dstContainerURLWithSAS := scenarioHelper{}.getRawContainerURLWithSAS(c, dstContainerName)
+	raw := getDefaultSyncRawInput(srcContainerURLWithSAS.String(), dstContainerURLWithSAS.String())
+	raw.excludeRegex = excludeString
+
+	// make sure the list doesn't include the blobs specified by the exclude flag
+	runSyncAndVerify(c, raw, func(err error) {
+		c.Assert(err, chk.IsNil)
+		validateS2SSyncTransfersAreScheduled(c, "", "", blobList, mockedRPC)
```
siminsavani-msft, line 751: Check if the values of blobsToInclude match the results.

cmd/zt_sync_blob_blob_test.go
```diff
@@ -681,3 +681,114 @@ func (s *cmdIntegrationSuite) TestSyncS2SADLSDirectory(c *chk.C) {
...
+	srcContainerURLWithSAS := scenarioHelper{}.getRawContainerURLWithSAS(c, srcContainerName)
+	dstContainerURLWithSAS := scenarioHelper{}.getRawContainerURLWithSAS(c, dstContainerName)
+	raw := getDefaultSyncRawInput(srcContainerURLWithSAS.String(), dstContainerURLWithSAS.String())
+	raw.includeRegex = includeString
+	raw.excludeRegex = excludeString
+
+	// verify that only the blobs specified by the include flag are synced
+	runSyncAndVerify(c, raw, func(err error) {
+		c.Assert(err, chk.IsNil)
+		validateS2SSyncTransfersAreScheduled(c, "", "", blobsToInclude, mockedRPC)
```
siminsavani-msft, line 792: You should check if the values of blobsToInclude match the results.

siminsavani-msft (changes requested):
Great job on coming up with test cases! Please review the comments and reach out with any questions!

cmd/sync.go
adreed-msft: `go fmt` may help.

cmd/copy.go
```diff
@@ -72,6 +72,8 @@ type rawCopyCmdArgs struct {
 	exclude               string
 	includePath           string // NOTE: This gets handled like list-of-files! It may LOOK like a bug, but it is not.
 	excludePath           string
+	includeRegex	   	  string
```
adreed-msft, line 75: Hmmm... I am trying to think of why we _shouldn't_ just accept regex in the normal include and exclude filter.

Possibly because we use it in the blob traverser as a performance option... BUT, we already have a complexity limitation on it, which would completely mitigate this question.

@zezha-msft and @siminsavani-msft any thoughts as to why we couldn't merge these? I'm mostly just trying to reduce the number of flags we have.

cmd/zc_filter.go
```diff
@@ -230,6 +231,107 @@ func (fs filterSet) GetEnumerationPreFilter(recursive bool) string {
...
+		if pattern != "" {
+			validPatterns = append(validPatterns, pattern)
+		}
+	}
+
+	return []objectFilter{&includeRegex{patterns: validPatterns}}
+}
+
+// excludeRegex does not include files that match with the regex expression
+type excludeRegex struct {
```
adreed-msft, line 285: You do not need two types for this. Just add a flag in includeRegex to set it up as an exclude filter. This is otherwise duplicated code.

adreed-msft: Perhaps even have it be like,

```go
type includeRegex struct {
    patterns []string
    matchResponse bool // true = include, false = exclude
}
```

and return `matchResponse` when there's a regex match, but `!matchResponse` when there's a hit

cmd/zt_copy_blob_download_test.go
adreed-msft: You could convert to maps and cross-check. I want to say there was a function for that SOMEWHERE in the test suite... But don't quote me on it.

adreed-msft (changes requested):

cmd/copy.go
siminsavani-msft: Not sure if combining the the regex with normal in/ex filter would work smoothly as a regex library is being used to match the files (zc_filter.go, ln 257). @t-cguruceaga-msft Could you comment more on this?

cmd/zt_copy_blob_download_test.go
t-cguruceaga-msft: I was not able to find the specific function instead I am looping through each mockedRPC.transfers and checking if they are equal to blobsToInclude

cmd/copy.go
```diff
@@ -1755,6 +1764,8 @@ func init() {
 		"This option does not support wildcard characters (*). Checks relative path prefix (For example: myFolder;myFolder/subDirName/file.pdf).")
 	cpCmd.PersistentFlags().StringVar(&raw.excludePath, "exclude-path", "", "Exclude these paths when copying. "+ // Currently, only exclude-path is supported alongside account traversal.
 		"This option does not support wildcard characters (*). Checks relative path prefix(For example: myFolder;myFolder/subDirName/file.pdf). When used in combination with account traversal, paths do not include the container name.")
+	cpCmd.PersistentFlags().StringVar(&raw.includeRegex, "include-regex", "", "Include only the files that align with regular expressions. Separate regular expressions with ';'.")
```
zezha-msft, line 1767: Clarify that the regex is matching the relative path of the files.

cmd/sync.go
```diff
@@ -704,6 +711,8 @@ func init() {
 		"This option does not support wildcard characters (*). Checks relative path prefix(For example: myFolder;myFolder/subDirName/file.pdf).")
 	syncCmd.PersistentFlags().StringVar(&raw.includeFileAttributes, "include-attributes", "", "(Windows only) Include only files whose attributes match the attribute list. For example: A;S;R")
 	syncCmd.PersistentFlags().StringVar(&raw.excludeFileAttributes, "exclude-attributes", "", "(Windows only) Exclude files whose attributes match the attribute list. For example: A;S;R")
+	syncCmd.PersistentFlags().StringVar(&raw.includeRegex, "include-regex", "", "Include files that match with the regular expressions. Separate regular expressions with ';'.")
```
zezha-msft, line 714: Clarify that the regex is matching the relative path of the files.

cmd/zc_filter.go
```diff
@@ -230,6 +231,57 @@ func (fs filterSet) GetEnumerationPreFilter(recursive bool) string {
...
+}
+
+func (f *regexFilter) doesSupportThisOS() (msg string, supported bool) {
+	msg = ""
+	supported = true
+	return
+}
+
+func (f *regexFilter) appliesOnlyToFiles() bool {
+	return true
```
zezha-msft, line 247: Why can't the filter apply to folders?

cmd/zc_filter.go
```diff
@@ -230,6 +231,57 @@ func (fs filterSet) GetEnumerationPreFilter(recursive bool) string {
...
+
+func (f *regexFilter) doesPass(storedObject storedObject) bool {
+	if len(f.patterns) == 0 {
+		return true
+	}
+	for _, pattern := range f.patterns {
+		matched := false
+		var err error
+
+		matched, err = regexp.MatchString(pattern, storedObject.name)
```
zezha-msft, line 258: We should match on the relative path instead of just the name.

cmd/zc_filter.go
```diff
@@ -230,6 +231,57 @@ func (fs filterSet) GetEnumerationPreFilter(recursive bool) string {
...
+
+		if err != nil {
+			return false
```
zezha-msft, line 261: If the regex parsing failed, perhaps we should let it be matchResponse instead of always false?

siminsavani-msft (approved):
LGTM!

adreed-msft (approved):

PR #1469, opened 2021-06-09
Snapshot Fix

siminsavani-msft:
- Fix logic to check source snapshot id/sharesnapshot id/version id

### DIFF
--- a/ste/xfer-anyToRemote-file.go
+++ b/ste/xfer-anyToRemote-file.go
@@ -166,7 +166,7 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
 			// if dst has snapshot or versionId specified, do not error and let the service fail the request with clear message
 			srcRQ := srcURL.Query()
 
-			if !(len(srcRQ["sharesnapshot"]) > 0 || len(srcRQ["snapshot"]) > 0 || len(srcRQ["versionid"]) > 0) {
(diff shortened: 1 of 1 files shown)


ste/xfer-anyToRemote-file.go
```diff
@@ -165,15 +165,8 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
...
-			dstRQ := dstURL.Query()
 
-			// note that query is now all lower case at this point, extract strings from query
-			srcShareSnapshot := strings.Join(srcRQ["sharesnapshot"], "")
-			dstShareSnapshot := strings.Join(dstRQ["sharesnapshot"], "")
-			srcVID := strings.Join(srcRQ["versionid"], "")
-			dstVID := strings.Join(dstRQ["versionid"], "")
-
-			if srcShareSnapshot == dstShareSnapshot && srcVID == dstVID {
+			if !(len(srcRQ["sharesnapshot"]) > 0 || len(srcRQ["snapshot"]) > 0 || len(srcRQ["versionid"]) > 0) {
```
zezha-msft, line 169: Minor style issue: for readability it might be best to avoid the leading ! and instead do:

len(srcRQ["sharesnapshot"]) == 0 && len(srcRQ["snapshot"])  == 0 && len(srcRQ["versionid"])  == 0

zezha-msft (approved):

ste/xfer-anyToRemote-file.go
adreed-msft, line 169: ditto

adreed-msft (approved):

PR #1477, opened 2021-06-15
Record folder creation state in the job plan to properly handle resumes

adreed-msft:
Historically, folder property persistence jobs have _not_ had proper functionality on resume. If a file created it's parent directory, and the user cancelled the job before the properties were persisted, the properties would not be persisted on resume.

This adds a new transfer status-- FolderCreated, sitting between Started and one of the many Finished statuses, indicating the job has created the folder, but properties have not been persisted. It is treated like Started in all other respects than folder properties persistence, however.

Additionally, we piggyback off the existing FolderCreationTracker interface and hold onto a dictionary of transfer IDs to achieve this without re-traversing the job plan every time. This causes the memory footprint of the folder tracker to exist a little longer (from folder scheduling or creation to persistence, as opposed to from creation to persistence)


cmd/root.go
```diff
@@ -179,6 +193,7 @@ func init() {
 	// special E2E testing flags
 	rootCmd.PersistentFlags().BoolVar(&azcopyAwaitContinue, "await-continue", false, "Used when debugging, to tell AzCopy to await `continue` on stdin before starting any work. Assists with debugging AzCopy via attach-to-process")
 	rootCmd.PersistentFlags().BoolVar(&azcopyAwaitAllowOpenFiles, "await-open", false, "Used when debugging, to tell AzCopy to await `open` on stdin, after scanning but before opening the first file. Assists with testing cases around file modifications between scanning and usage")
+	rootCmd.PersistentFlags().StringVar(&debugSkipFiles, "debug-skip-files", "", "Used when debugging, to tell AzCopy to cancel the job midway. List of relative paths to skip in the STE.")
```
zezha-msft, line 196: hide flag

common/fe-ste-models.go
```diff
@@ -627,6 +625,10 @@ func (TransferStatus) Started() TransferStatus { return TransferStatus(1) }
 // Transfer successfully completed
 func (TransferStatus) Success() TransferStatus { return TransferStatus(2) }
 
+// Folder was created, but properties have not been persisted yet. Equivalent to Started, but never intended to be set on anything BUT folders.
+// Reviewers: Can you think of a better choice? Initially, the plan was to use a boolean in JobPartPlanTransfer instead, but there's no atomic form of a boolean in Go, so I thought piggybacking off status was a better plan.
+func (TransferStatus) FolderCreated() TransferStatus { return TransferStatus(3) }
```
zezha-msft, line 630: I don't feel strongly about this, looks fine to me

zezha-msft (commented):
lgtm on first read, I'll take a second look at the tests

e2etest/declarativeHelpers.go
```diff
@@ -174,26 +185,27 @@ var eOperation = Operation(0)
 type Operation uint8
 
 func (Operation) Copy() Operation        { return Operation(1) }
-func (Operation) Sync() Operation        { return Operation(2) }
-func (Operation) CopyAndSync() Operation { return eOperation.Copy() & eOperation.Sync() }
-func (Operation) Remove() Operation      { return Operation(3) }
+func (Operation) Sync() Operation        { return Operation(1<<1) }
+func (Operation) CopyAndSync() Operation { return eOperation.Copy() | eOperation.Sync() }
+func (Operation) Remove() Operation      { return Operation(1<<2) }
+func (Operation) Resume() Operation 	 { return Operation(1<<7) } // Resume should only ever be combined with Copy or Sync, and is a mid-job cancel/resume.
```
mohsha-msft, line 191: just curious: Why 7?

e2etest/declarativeHelpers.go
```diff
@@ -481,6 +493,9 @@ type hookHelper interface {
 	// CreateFiles creates the specified files (overwriting any that are already there of the same name)
 	CreateFiles(fs testFiles, atSource bool)
 
+	// CreateFile creates a specified file (overwriting what was already there of the same name)
+	CreateFile(f *testObject, atSource bool)
```
mohsha-msft, line 497: I see that `CreateFile()` is creating file/folder in the middle of transfer. Can you please write proper comment to explain the behaviour here as it may create confusion with `CreateFiles()`. 

Another thing: Can we not refactor `CreateFiles()` to adjust the required creation of files/folders? 
I think adding one more interface in `testFiles` could have made the design clearer.
```go
type testFiles struct {
     shouldCreateWhileTrasfer []interface{}
}
```

Right now we're redundantly rewriting the exact function for all the resource types. i.e. Local, Azure File, Blob, Dummy.

cmd/root.go
adreed-msft: Meant to hide it, accidentally didn't.

e2etest/declarativeHelpers.go
adreed-msft: Last bit on a uint8, gives us plenty of room for future operations we may want to implement or test (e.g. CLFSLoad)

e2etest/declarativeHelpers.go
adreed-msft: CreateFile reuses the same underlying code as CreateFiles. I don't think there's really any code duplication to fix here.

Regarding shouldCreateDuringTransfer-- I think that's reinventing the wheel and creating something less useful at that. The hooks struct is already intended for this kinda stuff.

zezha-msft (approved):

mohsha-msft (approved):

PR #1494, opened 2021-07-13
[AzCopyV10][StgExp][Feature] Set from-to for rm

siminsavani-msft:
- Added from-to flag in remove.go
- Added tests for BlobTrash, FileTrash, BlobFSTrash

### DIFF
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -22,7 +22,6 @@ package cmd
 
 import (
 	"fmt"
-
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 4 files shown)


nakulkar-msft (approved):

zezha-msft (approved):

cmd/remove.go
```diff
@@ -45,18 +44,19 @@ func init() {
...
-				raw.fromTo = common.EFromTo.BlobTrash().String()
-			} else if srcLocationType == common.ELocation.File() {
-				raw.fromTo = common.EFromTo.FileTrash().String()
-			} else if srcLocationType == common.ELocation.BlobFS() {
-				raw.fromTo = common.EFromTo.BlobFSTrash().String()
-			} else {
-				return fmt.Errorf("invalid source type %s to delete. azcopy support removing blobs/files/adls gen2", srcLocationType.String())
+			if raw.fromTo == "" {
+				// infer the location of the delete
+				if srcLocationType == common.ELocation.Blob() {
```
mohsha-msft, line 50: Minor: I advice to use `switch` in place of multiple `if-else`. `switch` is faster and elegant as compared to `if-else-if` ladder.

cmd/remove.go
```diff
@@ -45,18 +44,19 @@ func init() {
...
+					raw.fromTo = common.EFromTo.BlobTrash().String()
+				} else if srcLocationType == common.ELocation.File() {
+					raw.fromTo = common.EFromTo.FileTrash().String()
+				} else if srcLocationType == common.ELocation.BlobFS() {
+					raw.fromTo = common.EFromTo.BlobFSTrash().String()
+				} else {
+					return fmt.Errorf("invalid source type %s to delete. azcopy support removing blobs/files/adls gen2", srcLocationType.String())
+				}
```
mohsha-msft, line 58: If `raw.fromTo != ""`,  i.e. from-to is provided by the user, it makes sense to verify that the destination is a trash. Otherwise we throw appropriate error saying `fromto.To()` isn't trash. Let's have that check in frontend layer for that as well.

cmd/zt_remove_adls_test.go
```diff
@@ -244,3 +244,178 @@ func (s *cmdIntegrationSuite) TestRemoveListOfALDSFilesWithIncludeExclude(c *chk
...
+	// invoke the interceptor so lifecycle manager does not shut down the tests
+	mockedRPC := interceptor{}
+	mockedRPC.init()
+	ctx := context.Background()
+
+	// get service SAS for raw input
+	serviceURLWithSAS := scenarioHelper{}.getRawAdlsServiceURLWithSAS(c)
+
+	// set up the filesystem to be deleted
+	bfsServiceURL := GetBFSSU()
```
mohsha-msft, line 258: The code between line 258-267 is redundant in all the ADLS Gen2 UTs you've written. Let's wrap these lines in one common function and call that function instead of writing the same lines again and again.

mohsha-msft (changes requested):

PR #1495, opened 2021-07-14
Dry Run Mode Sync

t-cguruceaga-msft:
sync.go -- added dry run flag 
syncProcessor.go -- made DeleteTransfer struct to return for json format and called Dryrun() for remove
zc_processor.go -- added copy for sync to Dryrun() 

sample output for sync:
DRYRUN: copy https://carlotatest.blob.core.windows.net/testcontainertwo/test2/technology.txt to https://carlotatest.blob.core.windows.net/testcontainer/test2/technology.txt
DRYRUN: remove https://carlotatest.blob.core.windows.net/testcontainer/to/excelling.xlsx

### DIFF
--- a/cmd/syncProcessor.go
+++ b/cmd/syncProcessor.go
@@ -22,12 +22,14 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"github.com/Azure/azure-storage-file-go/azfile"
 	"net/url"
 	"os"
 	"path"
-
-	"github.com/Azure/azure-storage-file-go/azfile"
+	"strings"
+	"time"
 
 	"github.com/Azure/azure-pipeline-go/pipeline"
 	"github.com/Azure/azure-storage-azcopy/v10/common"
@@ -95,6 +97,28 @@ type interactiveDeleteProcessor struct {
 
 	// count the deletions that happened
 	incrementDeletionCount func()
+
+	//dryrunMode
+	dryrunMode bool
+}
+
+// struct represents a removal
+type DeleteTransfer struct {
+	Source           string
+	EntityType       common.EntityType
+	LastModifiedTime time.Time
+	SourceSize       int64
+
+	ContentType        string
+	ContentEncoding    string
+	ContentDisposition string
+	ContentLanguage    string
+	CacheControl       string
+	Metadata           common.Metadata
+
+	BlobType      azblob.BlobType
+	BlobVersionID string
+	BlobTags      common.BlobTags
 }
 
 func (d *interactiveDeleteProcessor) removeImmediately(object storedObject) (err error) {
@@ -103,6 +127,26 @@ func (d *interactiveDeleteProcessor) removeImmediately(object storedObject) (err
 	}
 
 	if !d.shouldDelete {
+		return nil
+	}
+
+	if d.dryrunMode {
+		glcm.Dryrun(func(format common.OutputFormat) string {
+			if format == common.EOutputFormat.Json() {
+				jsonOutput, err := json.Marshal(newDeleteTransfer(object))
+				common.PanicIfErr(err)
+				return string(jsonOutput)
+			} else { // remove for sync
+				if d.objectTypeToDisplay == "local file" { //removing from local src
+					return fmt.Sprintf("DRYRUN: remove %v\\%v",
+						common.ToShortPath(d.objectLocationToDisplay),
+						strings.ReplaceAll(object.relativePath, "/", "\\"))
+				}
+				return fmt.Sprintf("DRYRUN: remove %v/%v",
+					d.objectLocationToDisplay,
+					object.relativePath)
+			}
+		})
 		return nil
 	}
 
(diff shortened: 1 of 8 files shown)


cmd/syncProcessor.go
```diff
@@ -106,6 +130,26 @@ func (d *interactiveDeleteProcessor) removeImmediately(object storedObject) (err
...
+		glcm.Dryrun(func(format common.OutputFormat) string {
+			if format == common.EOutputFormat.Json() {
+				jsonOutput, err := json.Marshal(newDeleteTransfer(object))
+				common.PanicIfErr(err)
+				return string(jsonOutput)
+			} else { // remove for sync
+				if d.objectTypeToDisplay == "local file" { //removing from local src
+					return fmt.Sprintf("DRYRUN: remove %v\\%v",
+						common.ToShortPath(d.objectLocationToDisplay),
+						strings.ReplaceAll(object.relativePath, "/", "\\"))
```
zezha-msft, line 143: Not sure about this slash swapping, the same code will run on Linux and MacOS

cmd/syncProcessor.go
```diff
@@ -106,6 +130,26 @@ func (d *interactiveDeleteProcessor) removeImmediately(object storedObject) (err
...
+				}
+				return fmt.Sprintf("DRYRUN: remove %v/%v",
```
zezha-msft, line 145: Could you please post a sample output in the PR description?

cmd/syncProcessor.go
```diff
@@ -117,6 +161,24 @@ func (d *interactiveDeleteProcessor) removeImmediately(object storedObject) (err
 	return
 }
 
+func newDeleteTransfer(object storedObject) (newDeleteTransfer DeleteTransfer) {
```
zezha-msft, line 164: Style: put constructor next to the struct for readability.

cmd/syncProcessor.go
```diff
@@ -95,6 +97,28 @@ type interactiveDeleteProcessor struct {
 
 	// count the deletions that happened
 	incrementDeletionCount func()
+
+	//dryrunMode
+	dryrunMode bool
+}
+
+// struct represents a removal
+type DeleteTransfer struct {
```
zezha-msft, line 106: Why do we need a new struct? Is storedObject too verbose?

cmd/zc_processor.go
```diff
@@ -90,10 +91,34 @@ func (s *copyTransferProcessor) scheduleCopyTransfer(storedObject storedObject)
 				common.PanicIfErr(err)
```
mohsha-msft, line 91: Why are we panicking here instead of just throwing appropriate error and terminating gracefully? We already have well-defined error here which we can return. I was going through the AzCopy code to find the usage of `json.Marshal()` and found that we use both - panic and throwing error.
```go
jsonOutput, err := json.Marshal(summary)
common.PanicIfErr(err)
```
and 

```go
b, err := json.Marshal(m)
if err != nil {
	return "", err
}
```
It makes sense to panic when unexpected wrong thing happens. In this case, we have well defined error which we can throw. Notice the return type of `scheduleCopyTransfer` method.

`func (s *copyTransferProcessor) scheduleCopyTransfer(storedObject storedObject) (err error)`

@zezha-msft your thoughts here ?

cmd/zc_processor.go
```diff
@@ -90,10 +91,34 @@ func (s *copyTransferProcessor) scheduleCopyTransfer(storedObject storedObject)
...
 				return string(jsonOutput)
 			} else {
-				// formatting string for remove
-				return fmt.Sprintf("DRYRUN: remove %v/%v",
-					s.copyJobTemplate.SourceRoot.Value,
-					srcRelativePath)
+				// if remove then To() will equal to common.ELocation.Unknown()
+				if s.copyJobTemplate.FromTo.To() == common.ELocation.Unknown() { //remove
```
mohsha-msft, line 95: I don't have full context. Can you tell me -
1. Why "Unknown" and not "Trash" too? 
2. Are we supporting dry-run for sync and remove only (and not copy)?

mohsha-msft (changes requested):

cmd/syncProcessor.go
mohsha-msft: Yup, same question. Fields are redundant in `DeleteTransfer` struct.

cmd/syncProcessor.go
```diff
@@ -106,6 +130,26 @@ func (d *interactiveDeleteProcessor) removeImmediately(object storedObject) (err
 		return nil
 	}
 
+	if d.dryrunMode {
+		glcm.Dryrun(func(format common.OutputFormat) string {
+			if format == common.EOutputFormat.Json() {
+				jsonOutput, err := json.Marshal(newDeleteTransfer(object))
+				common.PanicIfErr(err)
```
mohsha-msft, line 137: Same as [this](https://github.com/Azure/azure-storage-azcopy/pull/1495/files#r672995443)

mohsha-msft (changes requested):

cmd/syncProcessor.go
t-cguruceaga-msft: @zezha-msft @mohsha-msft  If I were to use storedObject instead of DeleteTransfer struct I would need to capitalize the fields that we want to json.Marshal() and since storedObject is used throughout the code base I would have to refactor by capitalizing all the references. 
Another alternative to making a DeleteTransfer struct is using the copyTransfer struct to marshal.

cmd/zc_processor.go
t-cguruceaga-msft: @mohsha-msft I used common.panicIfError(err) because I am calling glcm.Dryrun() and returning the string that would be printed in cmd. This is similarly to how errors are handled in glcm.Progress()

cmd/zc_processor.go
t-cguruceaga-msft: @mohsha-msft  There is currently no "Trash" location. If a file is removed the FromTo.To() will read as unknown since the removed file will not have a destination location.
We are supporting dryrun for all copy, remove and sync. The logistic for copy are in copyEnumeratorInit.go

mohsha-msft (approved):
LGTM.

adreed-msft (approved):

PR #1496, opened 2021-07-20
Login Status Check

t-cguruceaga-msft:
Added login status sub-command -- new file: loginStatus.go
login.go --made lgCmd a global var 

subcommand has two flags:
--tenant 
--endpoint

### DIFF
--- a/cmd/helpMessages.go
+++ b/cmd/helpMessages.go
@@ -305,7 +305,14 @@ Set the environment variable AZCOPY_SPA_CERT_PASSWORD to the certificate's passw
 
    Please treat /path/to/my/cert as a path to a PEM or PKCS12 file-- AzCopy does not reach into the system cert store to obtain your certificate.
    --certificate-path is mandatory when doing cert-based service principal auth.
+
+Subcommand for login to check the login status of your current session.
+	- azcopy login status 
 `
+
+const loginStatusShortDescription = "Prints if you are currently logged in to your Azure Storage account."
+
+const loginStatusLongDescription = "This command will let you know if you are currently logged in to your Azure Storage account."
 
 // ===================================== LOGOUT COMMAND ===================================== //
 const logoutCmdShortDescription = "Log out to terminate access to Azure Storage resources."
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -103,6 +103,8 @@ type loginCmdArgs struct {
 	certPass      string
 	clientSecret  string
 	persistToken  bool
+	tenantName    bool
+	endpoint      bool
 }
 
 type argValidity struct {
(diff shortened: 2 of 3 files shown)


cmd/loginStatus.go
```diff
@@ -0,0 +1,60 @@
...
+			}
+			return nil
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			// getting login token info
+			ctx := context.WithValue(context.TODO(), ste.ServiceAPIVersionOverride, ste.DefaultServiceApiVersion)
+			uotm := GetUserOAuthTokenManagerInstance()
+			tokenInfo, err := uotm.GetTokenInfo(ctx)
+
+			if err == nil && tokenInfo != nil {
```
nakulkar-msft, line 49: We should verify the token is still valid, something similar to tokenInfo.IsExpired()

cmd/helpMessages.go
```diff
@@ -305,8 +305,15 @@ Set the environment variable AZCOPY_SPA_CERT_PASSWORD to the certificate's passw
 
    Please treat /path/to/my/cert as a path to a PEM or PKCS12 file-- AzCopy does not reach into the system cert store to obtain your certificate.
    --certificate-path is mandatory when doing cert-based service principal auth.
+
+Subcommand for login to check the login status of your current session.
+	- azcopy login status
```
nakulkar-msft, line 310: What are your thoughts on having this as additional flag? 'azcopy login --status', this would be much smaller change in that case. Adding @manumsft here

cmd/login.go
```diff
@@ -29,57 +29,57 @@ import (
...
-		},
-		RunE: func(cmd *cobra.Command, args []string) error {
-			loginCmdArgs.certPass = glcm.GetEnvironmentVariable(common.EEnvironmentVariable.CertificatePassword())
-			loginCmdArgs.clientSecret = glcm.GetEnvironmentVariable(common.EEnvironmentVariable.ClientSecret())
-			loginCmdArgs.persistToken = true
-
-			if loginCmdArgs.certPass != "" || loginCmdArgs.clientSecret != "" {
-				glcm.Info(environmentVariableNotice)
-			}
+var loginCmdArg = loginCmdArgs{tenantID: common.DefaultTenantID}
```
nakulkar-msft, line 32: Changes in this file are not required.

cmd/helpMessages.go
mohsha-msft: I agree with this. It'll reduce the amount of code changes to bare minimum.

manumsft: I agree @nakulkar-msft . It makes sense. thanks.

siminsavani-msft (approved):
LGTM!

cmd/login.go
nakulkar-msft: Changes in this file are cosmetic, and should be reverted.

nakulkar-msft (approved):

PR #1500, opened 2021-07-26
[AzCopyV10][Feature] Offer permanent delete to remove soft-deleted versions/snapshots

siminsavani-msft:
- Implemented a permanent delete flag (--permanent-delete) for azcopy remove (“snapshots”, “versions”, “SnapshotsAndVersions”, “none”). "None" is the default value.
- Using this flag will permanently delete soft deleted snapshots and versions.
- Added filters to retrieve soft-deleted snapshots/versions.


common/fe-ste-models.go
```diff
@@ -145,6 +145,43 @@ func (d DeleteSnapshotsOption) ToDeleteSnapshotsOptionType() azblob.DeleteSnapsh
 	return azblob.DeleteSnapshotsOptionType(strings.ToLower(d.String()))
 }
 
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+var EPermanentDeleteOption = PermanentDeleteOption(3) // Default to "None"
+
+type PermanentDeleteOption uint8
+
+func (PermanentDeleteOption) Snapshots() PermanentDeleteOption { return PermanentDeleteOption(0) }
```
adreed-msft, line 153: I think making these bitflags would be better here because...

cmd/copyEnumeratorInit.go
```diff
@@ -409,6 +409,15 @@ func (cca *CookedCopyCmdArgs) InitModularFilters() []ObjectFilter {
 		}
 	}
 
+	switch cca.permanentDeleteOption {
```
adreed-msft, line 412: This wouldn't be so complicated.

cmd/zc_enumerator.go
```diff
@@ -303,12 +306,28 @@ type enumerationCounterFunc func(entityType common.EntityType)
...
+	credential *common.CredentialInfo, followSymlinks *bool, listOfFilesChannel chan string, recursive, getProperties,
+	includeDirectoryStubs bool, permanentDeleteOption common.PermanentDeleteOption, incrementEnumerationCounter enumerationCounterFunc, listOfVersionIds chan string,
 	s2sPreserveBlobTags bool, logLevel pipeline.LogLevel, cpkOptions common.CpkOptions) (ResourceTraverser, error) {
 	var output ResourceTraverser
 	var p *pipeline.Pipeline
 
+	var includeDeleted bool
+	var includeSnapshot bool
+	var includeVersion bool
+	switch permanentDeleteOption {
```
adreed-msft, line 318: This also wouldn't be necessary.

cmd/zc_filter.go
```diff
@@ -357,6 +357,54 @@ func (_ IncludeBeforeDateFilter) FormatAsUTC(t time.Time) string {
 	return formatAsUTC(t)
 }
 
+type softDeleteFilter struct {
+	deleteSnapshots bool
+	deleteVersions  bool
+}
+
+func (s *softDeleteFilter) DoesSupportThisOS() (msg string, supported bool) {
+	msg = ""
```
adreed-msft, line 366: You can do `return "", true`.

A bit of non-PR commentary; perhaps we should migrate DoesSupportThisOS into it's own interface, since it doesn't apply to all filters.

cmd/zc_filter.go
```diff
@@ -357,6 +357,54 @@ func (_ IncludeBeforeDateFilter) FormatAsUTC(t time.Time) string {
...
+	supported = true
+	return
+}
+
+func (s *softDeleteFilter) AppliesOnlyToFiles() bool {
```
adreed-msft, line 371: Ditto.

cmd/zc_filter.go
```diff
@@ -357,6 +357,54 @@ func (_ IncludeBeforeDateFilter) FormatAsUTC(t time.Time) string {
...
+	return false
+}
+
+func (s *softDeleteFilter) DoesPass(storedObject StoredObject) bool {
+	if s.deleteVersions && s.deleteSnapshots {
```
adreed-msft, line 376: This seems like it could all be simplified down into a single conditional.

e.g. `return (s.deleteVersions || s.deleteSnapshots) && ((s.deleteVersions && storedObject.blobVersionID != "") || (s.deleteSnapshots && storedObject.blobSnapshotID != ""))`

cmd/zc_filter.go
```diff
@@ -357,6 +357,54 @@ func (_ IncludeBeforeDateFilter) FormatAsUTC(t time.Time) string {
...
+			return true
+		}
+		return false
+	}
+	return false
+}
+
+func buildIncludeSoftDeleted(permanentDeleteOption common.PermanentDeleteOption) []ObjectFilter {
+	filters := make([]ObjectFilter, 0)
+	switch permanentDeleteOption {
```
adreed-msft, line 397: Ditto on the bitflags.

cmd/zc_traverser_blob.go
```diff
@@ -55,6 +56,12 @@ type blobTraverser struct {
 	s2sPreserveSourceTags bool
 
 	cpkOptions common.CpkOptions
+
+	includeDeleted bool
```
adreed-msft, line 60: Ditto on bitflags.

adreed-msft:
> func (t *blobTraverser) Traverse(preprocessor objectMorpher, processor objectProcessor, filters []ObjectFilter) (err error) {

Perhaps we could opt to glean the permanent delete setting from the filters here, since they're not really used outside of Traverse?

---
Refers to: cmd/zc_traverser_blob.go:132 in 803536e. [](commit_id = 803536eaa1544d6d9e51eb70c102e9def96541fb, deletion_comment = False)

cmd/zt_remove_blob_test.go
```diff
@@ -714,3 +718,101 @@ func (s *cmdIntegrationSuite) TestRemoveBlobsUnderVirtualDirWithFromTo(c *chk.C)
 		}
 	})
 }
+
+func (s *cmdIntegrationSuite) TestPermDeleteSnapshotsVersionsUnderSingleBlob(c *chk.C) {
```
adreed-msft, line 722: Nice tests!

cmd/zt_remove_blob_test.go
```diff
@@ -714,3 +718,101 @@ func (s *cmdIntegrationSuite) TestRemoveBlobsUnderVirtualDirWithFromTo(c *chk.C)
...
+	u, _ := url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net/", accountName))
+
+	credential, err := azblob.NewSharedKeyCredential(accountName, accountKey)
+	if err != nil {
+		panic(err)
+	}
+
+	sasQueryParams, err := azblob.AccountSASSignatureValues{
+		Protocol:      azblob.SASProtocolHTTPS,
+		ExpiryTime:    time.Now().UTC().Add(168 * time.Hour), // week long sas
```
adreed-msft, line 799: We actually have a 30m or so maximum length on testing; 6-12 hours may actually be a much better fit here in case some bizarre leakage of the SAS token happens in testing.

cmd/zt_remove_blob_test.go
```diff
@@ -714,3 +718,101 @@ func (s *cmdIntegrationSuite) TestRemoveBlobsUnderVirtualDirWithFromTo(c *chk.C)
...
+		Permissions:   azblob.AccountSASPermissions{Read: true, List: true, Write: true, Create: true, PermanentDelete: true, Delete: true, DeletePreviousVersion: true}.String(),
+		Services:      azblob.AccountSASServices{Blob: true}.String(),
+		ResourceTypes: azblob.AccountSASResourceTypes{Service: true, Container: true, Object: true}.String(),
+	}.NewSASQueryParameters(credential)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	qp := sasQueryParams.Encode()
+	urlToSendToSomeone := fmt.Sprintf("https://%s.blob.core.windows.net?%s", accountName, qp)
```
adreed-msft, line 809: I'm not sure this variable name makes sense here.

cmd/zt_scenario_helpers_for_test.go
```diff
@@ -133,12 +133,45 @@ func (s scenarioHelper) generateCommonRemoteScenarioForLocal(c *chk.C, dirPath s
...
-// 10 of them in sub dir "sub1"
-// 10 of them in sub dir "sub2"
-// 10 of them in deeper sub dir "sub1/sub3/sub5"
-// 10 of them with special characters
+func (scenarioHelper) generateCommonRemoteScenarioForSoftDelete(c *chk.C, containerURL azblob.ContainerURL, prefix string) (string, []azblob.BlockBlobURL, []string) {
+	blobList := make([]azblob.BlockBlobURL, 3)
+	blobNames := make([]string, 3)
+	var listOfTransfers []string
+
+	blobURL1, blobName1 := createNewBlockBlob(c, containerURL, prefix+"top")
```
adreed-msft, line 141: A for loop would save the hassle and allow for generating larger scenarios w/ less hassle.

cmd/zt_scenario_helpers_for_test.go
```diff
@@ -133,12 +133,45 @@ func (s scenarioHelper) generateCommonRemoteScenarioForLocal(c *chk.C, dirPath s
...
+
+	blobList[0] = blobURL1
+	blobNames[0] = blobName1
+	blobList[1] = blobURL2
+	blobNames[1] = blobName2
+	blobList[2] = blobURL3
+	blobNames[2] = blobName3
+
+	for i := 0; i < len(blobList); i++ {
+		for j := 0; j < 3; j++ {
```
adreed-msft, line 153: >3 [](http://example.com/codeflow?start=18&length=1)

Magic number should have a descriptive name & explanation.

cmd/zt_scenario_helpers_for_test.go
```diff
@@ -133,12 +133,45 @@ func (s scenarioHelper) generateCommonRemoteScenarioForLocal(c *chk.C, dirPath s
...
+			snapshotBlob := blobList[i].WithSnapshot(snapResp.Snapshot())
+			_, err = snapshotBlob.Delete(ctx, azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
+			c.Assert(err, chk.IsNil)
+
+			listOfTransfers = append(listOfTransfers, blobNames[i])
+		}
+	}
+
+	// sleep a bit so that the blobs' lmts are guaranteed to be in the past
+	time.Sleep(time.Millisecond * 1050)
```
adreed-msft, line 171: >time.Sleep(time.Millisecond * 1050) [](http://example.com/codeflow?start=1&length=35)

Why is time.Second unacceptable here? What makes the extra 50ms mandatory?

adreed-msft (changes requested):
:clock1:

adreed-msft:
I fundamentally approve; I just think there's room for improvement in some places.

mohsha-msft:
>  Implemented a permanent delete flag (--permanent-delete) for azcopy remove (“snapshots”, “versions”, “BothSnapshotsAndVersions”, “none”). "None" is the default value.

It should be "SnapshotsAndVersions" instead of "BothSnapshotsAndVersions"

cmd/zt_scenario_helpers_for_test.go
siminsavani-msft: This is just something I noticed in other scenario helpers (i.e. generateLocalFilesFromList, etc)

cmd/copyEnumeratorInit.go
```diff
@@ -335,7 +335,7 @@ func (cca *CookedCopyCmdArgs) isDestDirectory(dst common.ResourceString, ctx *co
 	}
 
 	rt, err := InitResourceTraverser(dst, cca.FromTo.To(), ctx, &dstCredInfo, nil,
-		nil, false, false, false,
+		nil, false, false, false, common.PermanentDeleteOption(3),
```
zezha-msft, line 338: common.PermanentDeleteOption(3)

doesn't look like the right way to use enums

cmd/copyEnumeratorInit.go
```diff
@@ -409,6 +409,15 @@ func (cca *CookedCopyCmdArgs) InitModularFilters() []ObjectFilter {
 		}
 	}
 
+	switch cca.permanentDeleteOption {
+	case common.PermanentDeleteOption(0):
+		filters = append(filters, &softDeleteFilter{deleteSnapshots: true})
+	case common.PermanentDeleteOption(1):
```
zezha-msft, line 415: It should look like something similar to common.EPermanentDeleteOption.DeleteSnapshots()

cmd/remove.go
```diff
@@ -110,4 +110,5 @@ func init() {
 	deleteCmd.PersistentFlags().StringVar(&raw.listOfVersionIDs, "list-of-versions", "", "Specifies a file where each version id is listed on a separate line. Ensure that the source must point to a single blob and all the version ids specified in the file using this flag must belong to the source blob only. Specified version ids of the given blob will get deleted from Azure Storage.")
 	deleteCmd.PersistentFlags().BoolVar(&raw.dryrun, "dry-run", false, "Prints the path files that would be removed by the command. This flag does not trigger the removal of the files.")
 	deleteCmd.PersistentFlags().StringVar(&raw.fromTo, "from-to", "", "Optionally specifies the source destination combination. For Example: BlobTrash, FileTrash, BlobFSTrash")
+	deleteCmd.PersistentFlags().StringVar(&raw.permanentDeleteOption, "permanent-delete", "none", "Permanent-delete permanently deletes soft-deleted snapshots/versions. Specify 'snapshots', 'versions', or 'snapshotsandversions' to remove soft-deleted snapshots/versions. Possible values include 'snapshots', 'versions', 'snapshotsandversions', 'none'.")
```
zezha-msft, line 113: Please be more concise and avoid redundancy in the description.

common/environment.go
```diff
@@ -188,8 +188,9 @@ func (EnvironmentVariable) EnumerationPoolSize() EnvironmentVariable {
 
 func (EnvironmentVariable) DisableHierarchicalScanning() EnvironmentVariable {
 	return EnvironmentVariable{
-		Name:        "AZCOPY_DISABLE_HIERARCHICAL_SCAN",
-		Description: "Applies only when Azure Blobs is the source. Concurrent scanning is faster but employs the hierarchical listing API, which can result in more IOs/cost. Specify 'true' to sacrifice performance but save on cost.",
+		Name:         "AZCOPY_DISABLE_HIERARCHICAL_SCAN",
+		Description:  "Applies only when Azure Blobs is the source. Concurrent scanning is faster but employs the hierarchical listing API, which can result in more IOs/cost. Specify 'true' to sacrifice performance but save on cost.",
+		DefaultValue: "false",
```
zezha-msft, line 193: any reason for the default value?

zezha-msft (changes requested):

common/fe-ste-models.go
mohsha-msft: I see where Adele is coming from. I think it'll make comparisons easier instead of writing switch statement every time. 
None = 0,
Snapshot = 1
Version = 2
SnapshotAndVersion = Snapshot|Version = 0b01 | 0b10 = ob11 = 3

We'll not have to create 3 variables in the code that way.

cmd/zc_enumerator.go
mohsha-msft: +1

common/fe-ste-models.go
mohsha-msft:
```golang
(o PermanentDeleteOption) DeleteSnapshot() {
    return o & EPermanentDeleteOption.Snapshot();
}

(o PermanentDeleteOption) DeleteVersion() {
    return o & EPermanentDeleteOption.Version();
}
```

These functions can act as flags anywhere you may want.

ste/xfer-deleteBlob.go
```diff
@@ -60,7 +61,14 @@ func doDeleteBlob(jptm IJobPartTransferMgr, p pipeline.Pipeline) {
 
 	// note: if deleteSnapshotsOption is 'only', which means deleting all the snapshots but keep the root blob
 	// we still count this delete operation as successful since we accomplished the desired outcome
-	_, err := srcBlobURL.Delete(jptm.Context(), jptm.DeleteSnapshotsOption().ToDeleteSnapshotsOptionType(), azblob.BlobAccessConditions{})
+	err := error(nil)
+	if jptm.PermanentDeleteOption().ToPermanentDeleteOptionType() == azblob.BlobDeletePermanent {
+		// Change service version before performing permanent delete
+		ctx := context.WithValue(jptm.Context(), ServiceAPIVersionOverride, "2020-08-04")
```
mohsha-msft, line 67: We've already updated AZCOPY_DEFAULT_SERVICE_API_VERSION=2020-08-04. So we can remove the hardcoded string.

ste/mgr-JobPartMgr.go
```diff
@@ -799,6 +799,10 @@ func (jpm *jobPartMgr) deleteSnapshotsOption() common.DeleteSnapshotsOption {
 	return jpm.Plan().DeleteSnapshotsOption
 }
 
+func (jpm *jobPartMgr) permanentDeleteOption() common.PermanentDeleteOption {
```
mohsha-msft, line 802: I'm wondering why some functions in this file are public while others are private.

common/version.go
```diff
@@ -1,6 +1,6 @@
 package common
 
-const AzcopyVersion = "10.13.0"
+const AzcopyVersion = "10.13.0-permanent-delete"
```
mohsha-msft, line 3: Nit: We're merging directly to `dev` branch (no private build) so this can go.

cmd/zt_remove_blob_test.go
```diff
@@ -714,3 +719,103 @@ func (s *cmdIntegrationSuite) TestRemoveBlobsUnderVirtualDirWithFromTo(c *chk.C)
 		}
 	})
 }
+
+func (s *cmdIntegrationSuite) TestPermDeleteSnapshotsVersionsUnderSingleBlob(c *chk.C) {
+	serviceURL := setUpAccountPermDelete(c)
+	os.Setenv("AZCOPY_DISABLE_HIERARCHICAL_SCAN", "true")
+
+	// From FE, 30 seconds is guaranteed to be enough.
+	time.Sleep(time.Second * 30)
```
mohsha-msft, line 728: Can we reduce the time to smaller values? It increases the CI run-time.

cmd/zt_remove_blob_test.go
```diff
@@ -714,3 +719,103 @@ func (s *cmdIntegrationSuite) TestRemoveBlobsUnderVirtualDirWithFromTo(c *chk.C)
...
+		c.Assert(len(mockedRPC.transfers), chk.Equals, 3)
+	})
+}
+
+func (s *cmdIntegrationSuite) TestPermDeleteSnapshotsVersionsUnderContainer(c *chk.C) {
+	serviceURL := setUpAccountPermDelete(c)
+	os.Setenv("AZCOPY_DISABLE_HIERARCHICAL_SCAN", "true")
+
+	// From FE, 30 seconds is guaranteed to be enough.
+	time.Sleep(time.Second * 30)
```
mohsha-msft, line 764: ditto for this one.

cmd/zc_traverser_blob.go
```diff
@@ -418,7 +431,12 @@ func (t *blobTraverser) serialList(containerURL azblob.ContainerURL, containerNa
 	return nil
 }
 
-func newBlobTraverser(rawURL *url.URL, p pipeline.Pipeline, ctx context.Context, recursive, includeDirectoryStubs bool, incrementEnumerationCounter enumerationCounterFunc, s2sPreserveSourceTags bool, cpkOptions common.CpkOptions) (t *blobTraverser) {
+func newBlobTraverser(rawURL *url.URL, p pipeline.Pipeline, ctx context.Context, recursive, includeDirectoryStubs bool, incrementEnumerationCounter enumerationCounterFunc, s2sPreserveSourceTags bool, cpkOptions common.CpkOptions, includeDeleted, includeSnapshot, includeVersion bool) (t *blobTraverser) {
+	if strings.ToLower(glcm.GetEnvironmentVariable(common.EEnvironmentVariable.DisableHierarchicalScanning())) == "false" &&
+		includeDeleted && (includeSnapshot || includeVersion) {
+		os.Setenv("AZCOPY_DISABLE_HIERARCHICAL_SCAN", "true")
```
mohsha-msft, line 437: Will this setting gets enabled right away? In case of tests, we set the time.sleep(30) for this, right?

mohsha-msft:
Logically the code looks good 👍 . I think do some improvements in the code. I'll do the testing for functional changes now.

cmd/zc_traverser_blob.go
siminsavani-msft: Yes, this gets enabled right away. I've tested this locally and it'll enable it immediately.

zezha-msft (approved):
lgtm, please call out that the feature is preview, and the delete will remove the snapshots/versions **PERMANENTLY**.

adreed-msft (approved):

PR #1504, opened 2021-08-02
[AzCopyV10][Github1226] Fix Infinite Sync

mohsha-msft:


cmd/syncEnumerator.go
```diff
@@ -211,6 +213,14 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 	}
 }
 
+func IsLocationCaseInsensitive(fromTo common.FromTo) bool {
+	if fromTo.IsDownload() && runtime.GOOS == "windows" {
```
zezha-msft, line 217: I tried it on Mac and it has the same problem, have you tried on Linux just to be safe?

mohsha-msft: It shouldn't work with UNIX based systems because, in UNIX, path is case-sensitive.
So a user may have two distinct paths -
`/home/mohsha/go_projects/src/azure-storage-azcopy`
`/home/mohsha/go_projects/SRC/Azure-Storage-Azcopy`

Notice they have same name but different casing and it is still allowed to create such paths.

With our approach, we lowercase the relative path and store it in the map. So we may end up thinking two or more different paths are actually the same.

Try that on Windows.

nakulkar-msft: Apple File System can be configured to be case sensitive or insensitive.
https://support.apple.com/en-in/guide/disk-utility/dsku19ed921c/mac

cmd/syncEnumerator.go
```diff
@@ -211,6 +213,14 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 	}
 }
 
+func IsLocationCaseInsensitive(fromTo common.FromTo) bool {
```
nakulkar-msft, line 216: This function should be a method of common.Location or have common.Location as argument instead of FromTo. We're interested in destination alone, and the name is a misnomer.

cmd/syncEnumerator.go
nakulkar-msft: We may have to deduce if the APFS is case-insensitive and then apply this logic.

nakulkar-msft (changes requested):

cmd/syncEnumerator.go
adreed-msft: For a really dumb simple test, we could create two test files in the azcopy folder. This would be OS agnostic.

adreed-msft (approved):

PR #1511, opened 2021-08-09
Added features to azbfs to support azfuse 

gapra-msft:
Added support for create with metadata, rename, setAccessControl and list filesystem paths


azbfs/datalake_storage_transforms.go
```diff
@@ -0,0 +1,31 @@
+package azbfs
+
+import (
+	"encoding/base64"
+	"strings"
+)
+
+// Converts metadata into a string of format "key1=value1, key2=value2" and Base64 encodes the values.
+func BuildMetadataString(md map[string]string) *string {
```
gapra-msft, line 9: Not sure what the best path is for how to handle metadata. It looks like it was autogenerated in blobs, should I handwrite something similar here?

azbfs/zt_test.go
```diff
@@ -55,7 +55,7 @@ func getBfsServiceURL() azbfs.ServiceURL {
 func generateName(prefix string) string {
 	// These next lines up through the for loop are obtaining and walking up the stack
 	// trace to extract the test name, which is stored in name
-	pc := make([]uintptr, 10)
+	pc := make([]uintptr, 20)
```
vibhansa-msft, line 58: What is this for ?

azbfs/datalake_storage_transforms.go
vibhansa-msft: Whatever was auto generated for block-blob, can't same be used here if they are in some common file. If its there why shall we create duplicate code.

azbfs/url_file.go
```diff
@@ -198,3 +204,50 @@ func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byt
...
+		nil, nil, nil, nil, nil, nil, nil)
+
+	if err != nil {
+		return FileURL{}, err
+	}
+
+	return destinationFileURL, nil
+}
+
+func (f FileURL) SetAccessControlList(ctx context.Context, accessControlList *string, group *string, owner *string) (*PathUpdateResponse, error) {
```
vibhansa-msft, line 236: Can you check if mode and additional principals are set in acl through portal, calling a setACL removes those or not. In blobfuse I am hitting this so better to handle it here from begining.

azbfs/datalake_storage_transforms.go
gapra-msft: azcopy seems to have different types for different services, so I dont think we should mix azblob namespaces in the azbfs code. Maybe someone on Ze's team can clarify too.

azbfs/url_directory.go
```diff
@@ -69,21 +69,26 @@ func (d DirectoryURL) NewDirectoryURL(dirName string) DirectoryURL {
 
 // Create creates a new directory within a File System
 func (d DirectoryURL) Create(ctx context.Context, recreateIfExists bool) (*DirectoryCreateResponse, error) {
+	return d.CreateWithMetadata(ctx, recreateIfExists, nil)
+}
+
+// Create creates a new directory within a File System
+func (d DirectoryURL) CreateWithMetadata(ctx context.Context, recreateIfExists bool, metadata map[string]string) (*DirectoryCreateResponse, error) {
```
gapra-msft, line 76: Is there some naming convention I should be using here or like an options bag approach possibly in case we want to add more options?

go.sum
```diff
@@ -10,40 +10,33 @@ cloud.google.com/go v0.52.0/go.mod h1:pXajvRH/6o3+F9jDHZWQ5PbGhn+o8w9qiu/CffaVdO
 cloud.google.com/go v0.53.0/go.mod h1:fp/UouUEsRkN6ryDKNW/Upv/JBKnv6WDthjR6+vze6M=
 cloud.google.com/go v0.54.0/go.mod h1:1rq2OEkV3YMf6n/9ZvGWI3GWw0VoqH/1x2nd8Is/bPc=
 cloud.google.com/go v0.56.0/go.mod h1:jr7tqZxxKOVYizybht9+26Z/gUq7tiRzu+ACVAMbKVk=
```
gapra-msft, line 12: Should I revert the go sum changes? Not sure why it changed, looks like the hashes changed

azbfs/zt_test.go
gapra-msft: This is no longer necessary. I was having some issues with the pipeline passing and I thought it was cause this pc stack wasnt big enough. I resolved the issue in another way.  Will revert this

azbfs/url_directory.go
gapra-msft: For example: We may need to expose permissions as well soon

go.sum
siminsavani-msft: Yeah, you should revert them as they shouldn't make too much of a difference.

mohsha-msft: +1 for revert

azbfs/README.md
```diff
@@ -0,0 +1,26 @@
+# Azcopy - azbfs
```
adreed-msft, line 1: I know we discussed offline that generation issues oddly occurred with a README.md with generation settings, but it'd be nice if we could imitate [azblob](https://github.com/Azure/azure-storage-blob-go/blob/master/swagger/README.md)

azbfs/datalake_storage_transforms.go
adreed-msft: Unfortunately our SDKs do contain duplicate types (e.g. SharedKeyCredential) for the sake of preventing breaking each other with each release. That is a more historical decision, from before the days of `go mod`.

I think with the more modern go.mod, we _could_ definitely do for a azure-storage-core-go instead in our track 2 SDKs. @zezha-msft what do you think?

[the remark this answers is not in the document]
adreed-msft: Maps are inherently nullable. I fought this decision on track 2 because it's rather poor UX to have a pointer to something that's already nullable to make it "optional".

azbfs/datalake_storage_transforms.go
```diff
@@ -0,0 +1,31 @@
...
+	if len(md) == 0 {
```
adreed-msft, line 10: for reference for other reviewers: maps are nullable, and when nil, maps and lists return 0 on len. [This is safe.](https://play.golang.org/p/XaFWmYwgVCB)

azbfs/url_directory.go
adreed-msft: In Track 2, we use Options bags containing pointers for optionals, instead of adding "overloads".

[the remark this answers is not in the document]
adreed-msft: Ditto. Maps are nullable. empty map = empty metadata, no map = no change on metadata.

azbfs/url_directory.go
```diff
@@ -157,3 +162,31 @@ func (d DirectoryURL) IsDirectory(ctx context.Context) (bool, error) {
 func (d DirectoryURL) NewFileUrl() FileURL {
 	return NewFileURL(d.URL(), d.directoryClient.Pipeline())
 }
+
+// Renames the directory to the provided destination
+func (d DirectoryURL) Rename(ctx context.Context, destinationFileSystem *string, destinationPath *string) (DirectoryURL, error) {
```
adreed-msft, line 167: Options bag, maybe?

azbfs/url_file.go
```diff
@@ -198,3 +204,50 @@ func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byt
...
+// Renames the file to the provided destination
+func (f FileURL) Rename(ctx context.Context, destinationFileSystem *string, destinationPath *string) (FileURL, error) {
+
+	// If the destinationFileSystem is not provided, use the current filesystem
+	fileSystemName := destinationFileSystem
+	if fileSystemName == nil || *fileSystemName == "" {
+		fileSystemName = &f.fileSystemName
+	}
+
+	renameSource := "/" + f.fileSystemName + "/" + f.path
```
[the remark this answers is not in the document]
adreed-msft: Incorrect. f.fileSystemName is a `string`, not `*string`.

azbfs/url_file.go
```diff
@@ -198,3 +204,50 @@ func (f FileURL) FlushData(ctx context.Context, fileSize int64, contentMd5 []byt
 		md5InBase64, nil, nil, nil, nil, nil, nil, nil,
 		nil, nil, &overrideHttpVerb, nil, nil, nil, nil)
 }
+
+// Renames the file to the provided destination
+func (f FileURL) Rename(ctx context.Context, destinationFileSystem *string, destinationPath *string) (FileURL, error) {
```
adreed-msft, line 209: ditto to options bags

azbfs/url_file.go
adreed-msft: Admittedly, this is pulled straight from my gen 2 ACLs PR-- But would you mind adding options bags?

adreed-msft (changes requested):

azbfs/datalake_storage_transforms.go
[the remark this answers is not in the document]
gapra-msft: Made the method private, since maps are nullable I will leave the input as is.

azbfs/README.md
gapra-msft: I moreso added this because I was struggling with figuring out how to generate without asking Adele for instructions. However, I never ended up generating for this branch or PR, so I think this is slightly out of scope. I'm happy to make an item for this though. Same with the comment below.

azbfs/README.md
```diff
@@ -0,0 +1,26 @@
...
+npm install -g autorest
+# run using command 'autorest' to check if installation worked
+autorest --help
+``` 
+
+## Generation Instructions.
+From the root of azure-storage-azcopy, run the following commands
+```bash
+cd azbfs
+sudo autorest --use=@microsoft.azure/autorest.go@v3.0.63 --input-file=./azure_dfs_swagger_manually_edited.json --go=true --output-folder=./ --namespace=azbfs --go-export-clients=false --file-prefix=zz_generated_
```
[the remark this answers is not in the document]
gapra-msft: See above comment. I am happy to create an issue for the team to investigate later on but I feel like its slightly out of scope.

azbfs/datalake_storage_transforms.go
```diff
@@ -0,0 +1,31 @@
...
+		return nil
+	}
+	var sb strings.Builder
+	var i int
+	for key, value := range md {
+		sb.WriteString(key)
```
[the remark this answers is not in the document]
gapra-msft: What do you mean by query escaped? I am just copying the Java implementation here: https://github.com/Azure/azure-sdk-for-java/blob/975e9793aee64d9f29000d37f0269ffc48a43695/sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakePathAsyncClient.java#L153

azbfs/url_directory.go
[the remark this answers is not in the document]
gapra-msft: This seems to be resolved as is.

azbfs/url_file.go
gapra-msft: I'll just wait for your acls PR to merge and I can just remove this.

gapra-msft: @vibhansa-msft I believe this will set the ACL (not update) just as it does in blobfuse so we will have to prior getACL and or it

azbfs/url_directory.go
gapra-msft: Created an options bag

azbfs/url_file.go
gapra-msft: Done

azbfs/url_directory.go
gapra-msft: Done

azbfs/url_filesystem.go
```diff
@@ -74,3 +74,8 @@ func (s FileSystemURL) Delete(ctx context.Context) (*FilesystemDeleteResponse, e
 func (s FileSystemURL) GetProperties(ctx context.Context) (*FilesystemGetPropertiesResponse, error) {
 	return s.fileSystemClient.GetProperties(ctx, s.name, nil, nil, nil)
 }
+
+// ListPaths returns a list of paths in the file system.
+func (s FileSystemURL) ListPaths(ctx context.Context, path *string, recursive bool, upnReturned *bool, maxResults int32, continuation *string) (*PathList, error) {
```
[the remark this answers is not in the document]
gapra-msft: Created an options bag and added comments there. This should be resolved as a result.

go.sum
```diff
@@ -777,4 +777,4 @@ honnef.co/go/tools v0.0.1-2020.1.3/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9
 honnef.co/go/tools v0.0.1-2020.1.4/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9vFzvIQ3k=
 rsc.io/binaryregexp v0.2.0/go.mod h1:qTv7/COck+e2FymRvadv62gMdZztPaShugOCi3I+8D8=
 rsc.io/quote/v3 v3.1.0/go.mod h1:yEA65RcK8LyAZtP9Kv3t0HmxON59tX3rD+tICJqUlj0=
-rsc.io/sampler v1.3.0/go.mod h1:T1hPZKmBbMNahiBKFy5HrXp6adAjACjK9JXDnKaTXpA=
+rsc.io/sampler v1.3.0/go.mod h1:T1hPZKmBbMNahiBKFy5HrXp6adAjACjK9JXDnKaTXpA=
```
gapra-msft, line 780: I tried to revert my go sum changes and it still thinks this is a change when it isnt.

go.sum
gapra-msft: I attempted to resolve this. I think there is one more problem with it regarding whitespace but the hash is the same. any idea on how to fix that?

adreed-msft (approved):

vibhansa-msft (approved):

PR #1514, opened 2021-08-10
S3-> Blob: Warn on detecting invalid S3 names

nakulkar-msft:


cmd/zc_traverser_s3.go
```diff
@@ -64,13 +64,25 @@ func (t *s3Traverser) isDirectory(isSource bool) bool {
 }
 
 func (t *s3Traverser) traverse(preprocessor objectMorpher, processor objectProcessor, filters []objectFilter) (err error) {
+	invalidAzureBlobName := func (objectKey string) bool {
+		/* S3 object name is invalid if it ends with period or
+		   one of virtual directories in path ends with period.
+		   This list is not exhaustive
+		 */
+		return strings.HasSuffix(objectKey, ".") ||
+		       strings.Contains(objectKey, "/.")
```
siminsavani-msft, line 73: Would it make sense to handle these invalid names by removing the invalid chars? Just a thought!

siminsavani-msft (approved):
LGTM!

cmd/zc_traverser_s3.go
nakulkar-msft: Azure cannot compose a PutBlockFromURL if the name matches these patterns. Changing the name will still result in invalid URL.

siminsavani-msft: Makes sense!

cmd/zc_traverser_s3.go
```diff
@@ -122,6 +134,13 @@ func (t *s3Traverser) traverse(preprocessor objectMorpher, processor objectProce
 			continue
 		}
 
+		if invalidAzureBlobName(objectInfo.Key) {
+			//Throw a warning on console and continue
+			msg := fmt.Sprintf("Skipping S3 object %s, as it is not a valid Blob name. Rename the object and retry the transfer", objectInfo.Key)
```
adreed-msft, line 139: Log this into the scanning logger as well.

adreed-msft (changes requested):

cmd/zc_traverser_s3.go
adreed-msft: I feel like we _should_ eventually prompt for a rename instead of straight rejecting it. It'd be a nicer UX imo. @zezha-msft thoughts?

zezha-msft: It's a great question @adreed-msft, unfortunately as @nakulkar-msft found out, our service API is unable to target the source object with the invalid chars. It is a service limitation.

zezha-msft (approved):

zezha-msft:
We don't support S3 in E2E yet, correct? @nakulkar-msft

cmd/zc_traverser_s3.go
nakulkar-msft: Add this to scanner logs.

mohsha-msft (approved):
LGTM

adreed-msft (approved):

cmd/zc_traverser_s3.go
adreed-msft: Duh-doy! I knew this was related, but forgot that it was the source not the destination.

PR #1516, opened 2021-08-11
Fixing Version Check

siminsavani-msft:
The aka.ms link was sending back the cached request ID rather than the real request ID so the versionMetadataUrl is updated to resolve the issue.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -196,7 +196,7 @@ func init() {
 func beginDetectNewVersion() chan struct{} {
 	completionChannel := make(chan struct{})
 	go func() {
-		const versionMetadataUrl = "https://aka.ms/azcopyv10-version-metadata"
+		const versionMetadataUrl = "https://azcopyvnextrelease.blob.core.windows.net/releasemetadata/latest_version.txt"
 
 		// step 0: check the Stderr before checking version
 		_, err := os.Stderr.Stat()


cmd/root.go
```diff
@@ -196,7 +196,7 @@ func init() {
 func beginDetectNewVersion() chan struct{} {
 	completionChannel := make(chan struct{})
 	go func() {
-		const versionMetadataUrl = "https://aka.ms/azcopyv10-version-metadata"
+		const versionMetadataUrl = "https://azcopyvnextrelease.blob.core.windows.net/releasemetadata/latest_version.txt"
```
adreed-msft, line 199: @zezha-msft Perhaps we should figure out what went wrong with the aka.ms shortener instead of referencing the storage account directly.

adreed-msft (changes requested):

cmd/root.go
zezha-msft: IIRC it had to do with the client request ID validation. The aka.ms link was sending back the cached request ID rather than the real request ID. 

@siminsavani-msft please update the PR description to be more precise.

zezha-msft (approved):

mohsha-msft (approved):
LGTM

adreed-msft (approved):

PR #1525, opened 2021-08-23
Exported APIs required to create custom Enumerator.

nitin-deamon:
- Exported function/struct from cmd package. So that user
  can write custome enumerator.
  Some of exported Structure/functions :-
  - cookedCopyCmdArgs
  - InitResourceTraverser
  - StoredObject
  - ObjectFilter
  - CopyEnumerator
- Apart from that changes also done in test files.
At this moment minimal changes done required to create custom
enumerator. In future we need to export more APIs depending on requirement.


zezha-msft:
@nitin-deamon please target dev instead, and rebase

PR #1538, opened 2021-08-31
Fix always receiving prompt on azure files folders

adreed-msft:


zezha-msft (approved):
Please perform some manual tests to double check the folder property preservation and overwrite bahavior.

zezha-msft:
Please also add the change log and bump version to 10.12.1

Issue #1536, opened 2021-08-31
`azcopy jobs resume` doesn't work with public URL


jiasli:
### Which version of the AzCopy was used? 

10.11.0

### Which platform are you using? (ex: Windows, Mac, Linux)

Windows

### To repro

Download from a public URL:

```powershell
./azcopy cp https://azcliprod.blob.core.windows.net/msi/azure-cli-2.27.2.msi output.msi
```

Cancel the task with <kbd>Ctrl</kbd>+<kbd>C</kbd> (https://github.com/Azure/azure-storage-azcopy/issues/1476).

It is not possible to resume the job:

```powershell
> ./azcopy jobs list

Existing Jobs
JobId: 12cfef28-8655-3f4a-64a7-fcd203a15a7d
Start Time: Tuesday, 31-Aug-21 16:51:47 CST
Status: Cancelled
Command: cp https://azcliprod.blob.core.windows.net/msi/azure-cli-2.27.2.msi output.msi


> ./azcopy jobs resume 12cfef28-8655-3f4a-64a7-fcd203a15a7d

cannot resume job with JobId 12cfef28-8655-3f4a-64a7-fcd203a15a7d. The source-sas switch must be provided to resume the job
```

But this is a public URL, without shared access signatures (SAS).

### Have you found a mitigation/solution?

No

maptz:
Hi, I was just came across this issue as I was looking at using AzCopy to download from a public source. 

I'm unable to `cp` from a public Uri (not Blob storage). I'm on v10.16.0. Does this no longer work, or is thee some config issue?

gapra-msft:
Hi @jiasli it looks like the original issue was resolved with the PR linked above and should be shipped by now so I will go ahead and close this issue. Please feel free to reopen in case you happen to be still hitting it.

@maptz could you please open a new issue if you are still encountering the other problem you descibed?

PR #1543, opened 2021-09-08
Fix resuming with a public source and add tests

adreed-msft:
Related bug: #1536

### DIFF
--- a/e2etest/declarativeResourceManagers.go
+++ b/e2etest/declarativeResourceManagers.go
@@ -23,6 +23,9 @@ package e2etest
 import (
 	"net/url"
 	"os"
+	"path"
+	"runtime"
+	"strings"
 
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 	"github.com/Azure/azure-storage-blob-go/azblob"
@@ -78,7 +81,7 @@ type resourceManager interface {
 	cleanup(a asserter)
 
 	// gets the azCopy command line param that represents the resource.  withSas is ignored when not applicable
-	getParam(stripTopDir bool, withSas bool) string
+	getParam(stripTopDir bool, withSas bool, withFile string) string
 
 	getSAS() string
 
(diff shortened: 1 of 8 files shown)


zezha-msft (approved):
lgtm, some conflicts need resolving.

e2etest/declarativeResourceManagers.go
```diff
@@ -169,7 +183,7 @@ type resourceBlobContainer struct {
 }
 
 func (r *resourceBlobContainer) createLocation(a asserter, s *scenario) {
-	cu, _, rawSasURL := TestResourceFactory{}.CreateNewContainer(a, r.accountType)
+	cu, _, rawSasURL := TestResourceFactory{}.CreateNewContainer(a, s.GetTestFiles().sourcePublic, r.accountType)
```
mohsha-msft, line 186: Let's do the refactoring of factory functions the same way we did in track2.

e2etest/zt_resume_test.go
```diff
@@ -83,4 +85,54 @@ func TestResume_LargeGeneric(t *testing.T) {
...
+			recursive:      true,
+			debugSkipFiles: []string{"a.txt"}, // skip the only file is ;
+		},
+		nil,
+		testFiles{
+			defaultSize:  "1K",
+			sourcePublic: azblob.PublicAccessContainer,
+
+			shouldTransfer: []interface{}{
+				f("a.txt"),
```
mohsha-msft, line 133: Let's add a virtual directory too. Obviously, the container in this case will be public.

```golang
folder("testfolder1")
file("testfolder1/a.txt")
```

ste/init.go
```diff
@@ -316,6 +318,42 @@ func ResumeJobOrder(req common.ResumeJobRequest) common.CancelPauseResumeRespons
...
+					return true
+				}
+
+				// since that failed, maybe it doesn't exist and is public to list?
+				blobParts.BlobName = ""
+			}
+
+			cURL := azblob.NewContainerURL(blobParts.URL(), azblob.NewPipeline(azblob.NewAnonymousCredential(), azblob.PipelineOptions{}))
+
+			_, err := cURL.ListBlobsFlatSegment(steCtx, azblob.Marker{}, azblob.ListBlobsSegmentOptions{})
```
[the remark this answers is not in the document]
adreed-msft: This handles virtual directories and tests listing from the root of the container. First it assumes any virtual directory is actually a blob and tries getproperties. If that fails, it continues along and removes the vdir name from the URI and lists against the raw container.

GetProperties is _marginally_ cheaper than ListBlobs. That said, they're priced per 10k ops, and _one_ operation for a resume isn't going to hurt.

Issue #1551, opened 2021-09-13
AzCopy from S3 to Blob fails with 401 code after update


AlexDHoffer:
### Which version of the AzCopy was used? 

10.12.1

### Which platform are you using? (ex: Windows, Mac, Linux)

Linux

### What command did you run?

````
azcopy copy 'https://s3.us-west-2.amazonaws.com/data-science/real-time/ids_noEncoding/aaaaaa-bbbbb-ccccc-dddd/graph_latest.pickle' 'https://storageaccountname.blob.core.windows.net/container-name/subdirectory/13958/aaaaaa-bbbbb-ccccc-dddd/graph_latest.pickle' --overwrite=true --check-md5 "FailIfDifferent" --disable-auto-decoding
````

We authenticate to both AWS S3 and Azure Blob via the recommended environment variables. We log in to Azure using a service principal.

### What problem was encountered?

````
   RESPONSE Status: 401 Server failed to authenticate the request. Please refer to the information in the www-authenticate header.

   Content-Length: [302]

   Content-Type: [application/xml]

   Date: [Mon, 13 Sep 2021 17:46:07 GMT]

   Server: [Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0]

   Www-Authenticate: [Bearer authorization_uri=https://login.microsoftonline.com/[my-tenant]/oauth2/authorize resource_id=https://storage.azure.com]

   X-Ms-Client-Request-Id: [03e650bb-c33e-....]

   X-Ms-Error-Code: [NoAuthenticationInformation]

   X-Ms-Request-Id: [f4e19b20-b01e-....]

   X-Ms-Version: [2020-08-04]

Response Details: ﻿<Code>NoAuthenticationInformation</Code><Message>Server failed to authenticate the request. Please refer to the information in the www-authenticate header. </Message>



2021/09/13 17:46:08 ERR: [P#0-T#0] COPYFAILED: https://s3.us-west-2.amazonaws.com/data-science/real-time/ids_noEncoding/aaaaaa-bbbbb-ccccc-dddd/graph_latest.pickle : 401 : 401 Server failed to authenticate the request. Please refer to the information in the www-authenticate header.. When Put Blob from URL. X-Ms-Request-Id: f4e19b20-b01e-003b-37c7-a871b0000000



[...]

PR #1556, opened 2021-09-16
Fixing "AzCopy from S3 to Blob fails with 401 code after update"

siminsavani-msft:
Fix for https://github.com/Azure/azure-storage-azcopy/issues/1551

### DIFF
--- a/cmd/credentialUtil.go
+++ b/cmd/credentialUtil.go
@@ -562,8 +562,6 @@ func GetCredentialInfoForLocation(ctx context.Context, location common.Location,
 func getCredentialType(ctx context.Context, raw rawFromToInfo, cpkOptions common.CpkOptions) (credType common.CredentialType, err error) {
 
 	switch {
-	case raw.fromTo == common.EFromTo.S3Blob(): // S3 Bucket
-		credType, _, err = getCredentialTypeForLocation(ctx, raw.fromTo.From(), raw.source, raw.sourceSAS, true, common.CpkOptions{})
 	case raw.fromTo.To().IsRemote():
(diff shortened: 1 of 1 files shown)


cmd/credentialUtil.go
```diff
@@ -562,8 +562,6 @@ func GetCredentialInfoForLocation(ctx context.Context, location common.Location,
 func getCredentialType(ctx context.Context, raw rawFromToInfo, cpkOptions common.CpkOptions) (credType common.CredentialType, err error) {
 
 	switch {
-	case raw.fromTo == common.EFromTo.S3Blob(): // S3 Bucket
-		credType, _, err = getCredentialTypeForLocation(ctx, raw.fromTo.From(), raw.source, raw.sourceSAS, true, common.CpkOptions{})
```
mohsha-msft, line 566: Was this code added in 10.12?
If not, was this scenario working in 10.11?

siminsavani-msft: Yes, it was added in 10.12! 

I also tested 10.11 and this scenario was working.

zezha-msft (commented):
So this was a regression and we don't have a test, can the e2e package accommodate such a test right now?

siminsavani-msft:
> So this was a regression and we don't have a test, can the e2e package accommodate such a test right now?

I will look into adding tests for this scenario!

mohsha-msft:
> So this was a regression and we don't have a test, can the e2e package accommodate such a test right now?

We currently don't have support for both S3 and GCP in our E2E test framework. Shall we pick it up for 10.14 or next cleanup week? @zezha-msft

PR #1563, opened 2021-09-21
Add Arc VM support for auth via managed identity

Strikerzee:
AzCopy v10 in its current state does not support logging into AzCopy via managed identities on Arc server VMs. 

This PR is an attempt to enable Arc server users to log onto AzCopy via managed identities and provide a similar experience as to the Azure VM users.

Tested on:

Windows 2019 Server
Linux Ubuntu 20.04 LTS

Golang v1.16


common/oauthTokenManager.go
```diff
@@ -713,15 +718,81 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
 	resp, err := msiTokenHTTPClient.Do(req)
+
+	return req, resp, err
+}
+
+// Dated 15th Sep 2021.
+// Token JSON returned by ARC-server endpoint API currently does not set a valid integral value for "not_before" key.
+// Fix will be available in future ARC versions, till then use the current time to fixup the byte slice before we pass it to json.Unmarshal().
+// If the token JSON already has "not_before" correctly set, this will be a no-op.
+// fixupTokenJson corrects the value of JSON field "not_before" in the Byte slice from blank to a valid value and returns the corrected Byte slice.
```
linuxsmiths, line 738: In accordance with godoc, have this as the very first line in the comment.
then leave a blank line and have rest of the comment lines.

common/oauthTokenManager.go
```diff
@@ -713,15 +718,81 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+		challengeToken, fileErr := reader.ReadString('\n')
+		if fileErr != nil && fileErr != io.EOF {
+			return nil, fmt.Errorf("error occurred while reading file %s: %v", challengeTokenPath, fileErr)
+		}
+
+		req.Header.Set("Authorization", "Basic "+challengeToken)
+
+		resp, err = msiTokenHTTPClient.Do(req)
+		if err != nil {
+			return nil, fmt.Errorf("failed to query token from Arc IMDS endpoint. Please report the issue to xxx@microsoft.com: %v", err)
```
linuxsmiths, line 794: Reminder: Need to fix xxx@microsoft.com to a valid support channel id.

common/oauthTokenManager.go
```diff
@@ -713,15 +718,85 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+
+// Dated 15th Sep 2021.
+// Token JSON returned by ARC-server endpoint API currently does not set a valid integral value for "not_before" key.
+// If the token JSON already has "not_before" correctly set, this will be a no-op.
+func fixupTokenJson(bytes []byte) []byte {
+	byteSliceToString := string(bytes)
+	separatorString := `"not_before":"`
+	stringSlice := strings.Split(byteSliceToString, separatorString)
+
+	if stringSlice[1][0] != '"' {
```
zezha-msft, line 744: Minor: is "not_before" always part of the json? Perhaps it's best to check for it before invoking stringSlice[1] just to make sure.

common/oauthTokenManager.go
zezha-msft: Is this actually actionable for us? Perhaps the message should be about asking the user to verify their configuration first. They'll reach out via Github or Azure Support eventually if things don't work out.

common/oauthTokenManager.go
```diff
@@ -713,15 +718,85 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+	notBeforeTime := strconv.FormatUint(notBeforeTimeInteger, 10)
+	return []byte(stringSlice[0] + separatorString + notBeforeTime + stringSlice[1])
+}
+
+// GetNewTokenFromMSI gets token from Azure Instance Metadata Service identity endpoint. It first checks if the VM is registered with Azure Arc. Failing that case, it checks if it is an Azure VM.
+// For details, please refer to https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview
+// Note: Currently the msiTokenHTTPClient timeout is configured for 30 secs. Should be reduced to 5 sec as IMDS endpoint is local to the machine.
+// Without this change, if some router is configured to not return "ICMP unreachable" then it will take 30 secs to timeout and fallback to ARC.
+// For now, this has been mitigated by checking Arc first, and then Azure
+// TODO: Handle the case where the VM has some process already listening to port 40342.
```
zezha-msft, line 759: I think this TODO should perhaps be addressed right away, we should make sure to fail fast if the port 40342 is not what we expected.

zezha-msft (approved):
lgtm in general. Please target dev branch instead.

common/oauthTokenManager.go
Strikerzee: Yes, it is always a part of the JSON - Be it a response from the Arc IMDS endpoint (with an empty string value), or Azure VM (with a non-empty string value)

common/oauthTokenManager.go
linuxsmiths: Yes, this need not be a support email but whatever we follow in other places, basically at this point it's not likely a user config error since we did get the proper challenge token file location in our first call to the ARC IMDS endpoint at L762 and then we could read the challenge token file at L772.
Since this is an ARC bug we want user to tell us that. Where shall we direct the user?

common/oauthTokenManager.go
linuxsmiths: yes, we should fix it.
@Strikerzee, can you look at the return err from msiTokenHTTPClient.Do() at L729 for the following cases.
1. Nothing listening on 127.0.0.1:40342
2. Some process listening on that port but not responding correctly to the GET request.

shalinijoshi19 (commented):
Hey @Strikerzee could we have a more descriptive PR description? We should just add details around what the change is broadly and how we tested it out for regressions etc. (Also should refer to Azure ARC Connected Machine Agents specifically).

mohsha-msft (approved):

common/oauthTokenManager.go
```diff
@@ -713,16 +724,134 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
 	}
+
 	req.URL.RawQuery = params.Encode()
 	req.Header.Set("Metadata", "true")
+
 	// Set context.
 	req.WithContext(ctx)
 
 	// Send request
+	msiTokenHTTPClient.Timeout = 10 * time.Second
```
linuxsmiths, line 735: add a comment why we are adding this timeout.

common/oauthTokenManager.go
```diff
@@ -713,16 +724,134 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
 	resp, err := msiTokenHTTPClient.Do(req)
+	msiTokenHTTPClient.Timeout = 0
+	return req, resp, err
+}
+
+// checkIfWWWAuthenticateUnavailable checks if the key "Www-Authenticate" is unavailable in the header of an http response
+func checkIfWWWAuthenticateUnavailable(resp *http.Response) bool {
```
linuxsmiths, line 742: A more generic name like isValidArcResponse() would be better.
Inside that you can search for this specific header explaining with a comment why this header qualified for a valid Arc response.

common/oauthTokenManager.go
```diff
@@ -693,17 +703,18 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromTokenStore(ctx context.Context) (
 	return &(tokenInfo.Token), nil
 }
 
-// GetNewTokenFromMSI gets token from Azure Instance Metadata Service identity endpoint.
-// For details, please refer to https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview
-func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.Token, error) {
+// queryIMDS sends a token request to the IMDS endpoint passed by the caller. This IMDS endpoint will be different for Azure and Arc VMs.
+func (credInfo *OAuthTokenInfo) queryIMDS(msiEndpoint string, resource string, imdsAPIVersion string, ctx context.Context) (*http.Request, *http.Response, error) {
```
linuxsmiths, line 707: nit. as a convention ctx should be the first parameter.

common/oauthTokenManager.go
```diff
@@ -713,16 +724,134 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+// For now, this has been mitigated by checking Arc first, and then Azure
+func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.Token, error) {
+	// Try Arc VM
+	req, resp, err := credInfo.queryIMDS(MSIEndpointArcVM, Resource, IMDSAPIVersionArcVM, ctx)
 	if err != nil {
-		return nil, fmt.Errorf("please check whether MSI is enabled on this PC, to enable MSI please refer to https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/qs-configure-portal-windows-vm#enable-system-assigned-identity-on-an-existing-vm. (Error details: %v)", err)
+		// Try Azure VM since there was an error in trying Arc VM
+		reqAzureVM, respAzureVM, errAzureVM := credInfo.queryIMDS(MSIEndpointAzureVM, Resource, IMDSAPIVersionAzureVM, ctx)
+		var serr syscall.Errno
+		errorConverted := errors.As(err, &serr)
```
linuxsmiths, line 784: move L783-784 inside "if errAzureVM != nil"

common/oauthTokenManager.go
```diff
@@ -713,16 +724,134 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+		if errAzureVM != nil {
+			if errorConverted {
+				var econnrefusedValue int
```
linuxsmiths, line 787: nit, 
econnrefusedValue  := -1

and get rid of L792-794

common/oauthTokenManager.go
```diff
@@ -713,16 +724,134 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+				return nil, fmt.Errorf("unknown process running at the endpoint: %v", err)
+			} else {
+				// If there is some other issue
+				return nil, fmt.Errorf("failed to process request at MSI endpoint: %v", err)
+			}
+		} else {
+			req, resp = reqAzureVM, respAzureVM
+		}
+	} else if checkIfWWWAuthenticateUnavailable(resp) {
+		reqAzureVM, respAzureVM, errAzureVM := credInfo.queryIMDS(MSIEndpointAzureVM, Resource, IMDSAPIVersionAzureVM, ctx)
```
linuxsmiths, line 816: write a comment above this

// Not valid response from ARC IMDS endpoint. Some other process listening on it?

common/oauthTokenManager.go
```diff
@@ -713,16 +724,134 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+		if errAzureVM != nil {
+			return nil, fmt.Errorf("please check if IMDS is running at the endpoint: %v", err)
+		} else {
+			req, resp = reqAzureVM, respAzureVM
+		}
+	} else {
+		challengeTokenPath := strings.Split(resp.Header["Www-Authenticate"][0], "=")[1]
```
linuxsmiths, line 823: and a comment above this

// Valid response received from ARC IMDS endpoint. Proceed with step#2.

common/oauthTokenManager.go
```diff
@@ -713,16 +724,134 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+				// If there is some other issue
+				return nil, fmt.Errorf("failed to process request at MSI endpoint: %v", err)
+			}
+		} else {
+			req, resp = reqAzureVM, respAzureVM
+		}
+	} else if checkIfWWWAuthenticateUnavailable(resp) {
+		reqAzureVM, respAzureVM, errAzureVM := credInfo.queryIMDS(MSIEndpointAzureVM, Resource, IMDSAPIVersionAzureVM, ctx)
+		if errAzureVM != nil {
+			return nil, fmt.Errorf("please check if IMDS is running at the endpoint: %v", err)
```
linuxsmiths, line 818: comment at L818 (above return)

// Neither Arc nor Azure VM IMDS endpoint available. Can't use MSI.

common/oauthTokenManager.go
```diff
@@ -713,16 +724,134 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+		} else {
```
linuxsmiths, line 819: comment here

// Azure IMDS endpoint ok!

common/oauthTokenManager.go
```diff
@@ -713,16 +724,134 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+					econnrefusedValue = WSAECONNREFUSED
+				} else {
+					econnrefusedValue = -1
+				}
+				if int(serr) == econnrefusedValue {
+					// If connection to Arc endpoint was refused
+					return nil, fmt.Errorf("please check whether MSI is enabled on this PC, to enable MSI please refer to https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/qs-configure-portal-windows-vm#enable-system-assigned-identity-on-an-existing-vm: %v", errAzureVM)
+				} else if checkIfWWWAuthenticateUnavailable(resp) {
+					// If response was malformed
+					return nil, fmt.Errorf("unknown process running at the endpoint: %v", err)
```
linuxsmiths, line 800: log at L800 should read

invalid response received from Arc IMDS endpoint (%s), probably some unknown process listening:

common/oauthTokenManager.go
```diff
@@ -713,16 +724,134 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+				} else {
```
linuxsmiths, line 801: L801-804 is impossible

linuxsmiths (approved):
Thanks for doing this and diligently testing all scenarios!

common/oauthTokenManager.go
shalinijoshi19: @linuxsmiths typically we dont embed support details in production code if we can help it. What we do is return a specific/documented error code + message and then have separate support channels that we advertise separately from the code for customers to follow up on (_this could be github here for AzCopy customers OR our CSS routes for 1P services etc etc; But we don't need to make or embed the follow up route here in code - just the error code + actionable message_)

Strikerzee: @zezha-msft What would you suggest we put here instead of directing to the support mail? What should the actionable message be here?

Strikerzee:
Hello @mohsha-msft , I have re-requested a review on the latest PR. Please let know in case of any comments. Thanks!

common/oauthTokenManager.go
```diff
@@ -713,16 +722,135 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+	msiTokenHTTPClient.Timeout = 0
+	return req, resp, err
+}
+
+// isValidArcResponse checks if the key "Www-Authenticate" is unavailable in the header of an http response
+func isValidArcResponse(resp *http.Response) bool {
+	// Parameter for validity is whether "Www-Authenticaite" exists in the response header
+	// "Www-Authenticate" contains the path to the challenge token file for Arc VMs
+	wwwAuthenticateExists := false
+	if resp != nil {
```
zezha-msft, line 745: Minor: the two if statements can be combined into one line

common/oauthTokenManager.go
```diff
@@ -713,16 +722,135 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+
+	// If the value of not_before is blank, set to "now - 5 sec" and return the updated slice
+	notBeforeTimeInteger := uint64(time.Now().Unix() - 5)
+	notBeforeTime := strconv.FormatUint(notBeforeTimeInteger, 10)
+	return []byte(stringSlice[0] + separatorString + notBeforeTime + stringSlice[1])
+}
+
+// GetNewTokenFromMSI gets token from Azure Instance Metadata Service identity endpoint. It first checks if the VM is registered with Azure Arc. Failing that case, it checks if it is an Azure VM.
+// For details, please refer to https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview
+// Note: Currently the msiTokenHTTPClient timeout is configured for 30 secs. Should be reduced to 5 sec as IMDS endpoint is local to the machine.
```
zezha-msft, line 775: update comment

common/oauthTokenManager.go
```diff
@@ -713,16 +722,135 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+// Without this change, if some router is configured to not return "ICMP unreachable" then it will take 30 secs to timeout and fallback to ARC.
+// For now, this has been mitigated by checking Arc first, and then Azure
+func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.Token, error) {
+	// Try Arc VM
+	req, resp, err := credInfo.queryIMDS(ctx, MSIEndpointArcVM, Resource, IMDSAPIVersionArcVM)
```
zezha-msft, line 780: consider renaming err to errArcVM to increase readability

common/oauthTokenManager.go
```diff
@@ -80,8 +88,8 @@ func newAzcopyHTTPClient() *http.Client {
 			Proxy: GlobalProxyLookup,
 			// We use Dial instead of DialContext as DialContext has been reported to cause slower performance.
 			Dial /*Context*/ : (&net.Dialer{
-				Timeout:   30 * time.Second,
-				KeepAlive: 30 * time.Second,
+				Timeout:   10 * time.Second,
```
mohsha-msft, line 91: QQ: Why are we reducing the Timeout and KeepAlive time? Can you add a comment to explain it?

common/oauthTokenManager.go
```diff
@@ -713,16 +722,135 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
...
+// Without this change, if some router is configured to not return "ICMP unreachable" then it will take 30 secs to timeout and fallback to ARC.
+// For now, this has been mitigated by checking Arc first, and then Azure
+func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.Token, error) {
+	// Try Arc VM
+	req, resp, err := credInfo.queryIMDS(ctx, MSIEndpointArcVM, Resource, IMDSAPIVersionArcVM)
 	if err != nil {
-		return nil, fmt.Errorf("please check whether MSI is enabled on this PC, to enable MSI please refer to https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/qs-configure-portal-windows-vm#enable-system-assigned-identity-on-an-existing-vm. (Error details: %v)", err)
+		// Try Azure VM since there was an error in trying Arc VM
+		reqAzureVM, respAzureVM, errAzureVM := credInfo.queryIMDS(ctx, MSIEndpointAzureVM, Resource, IMDSAPIVersionAzureVM)
+		if errAzureVM != nil {
```
zezha-msft, line 784: a bit of a readability concern here: in general it'd be best to avoid multiple nested if/else statements

zezha-msft (approved):

common/oauthTokenManager.go
```diff
@@ -742,8 +870,11 @@ func (credInfo *OAuthTokenInfo) GetNewTokenFromMSI(ctx context.Context) (*adal.T
 	result := &adal.Token{}
 	if len(b) > 0 {
 		b = ByteSliceExtension{ByteSlice: b}.RemoveBOM()
+		// Unmarshal will give an error for Go version >= 1.14 for a field with blank values. Arc-server endpoint API returns blank for "not_before" field.
+		// TODO: Remove fixup once Arc team fixes the issue.
+		b = fixupTokenJson(b)
```
mohsha-msft, line 875: Just Curious: Is it a known issue? If not, have we filed the issue on Arc team?

mohsha-msft (approved):
LGTM! 
Have we tested the scenarios? I don't think we can add unit tests for this so manual testing should do.

common/oauthTokenManager.go
mohsha-msft: +1

common/oauthTokenManager.go
Strikerzee: We need to set the Timeout attribute to some small value in case there's some process listening on 40342 (arc imds port) which does cause the TCP connection to complete (no ECNONNREFUSED error) but it doesn't respond to the HTTP GET request. With the current client, we'll wait for ever.
Hence, it was proposed that we use a small timeout, 5 secs, since we are talking to local processes.

common/oauthTokenManager.go
Strikerzee: We have discussed this issue with the team handling managed identity auth on Arc and it will be worked upon.

common/oauthTokenManager.go
Strikerzee: There is a possibility of resp being null, in which case if I try to access resp.header in the same line - It will throw a null pointer reference error.

mohsha-msft: No, it won't. Conditions are checked from left to right. i.e. `if A && B { // do this }` If A is false, B won't be evaluated.

Strikerzee: Understood, and noted.

common/oauthTokenManager.go
Strikerzee: I tried to reduce the nesting and pushed the code again.

PR #1570, opened 2021-09-27
Fix incorrect status for Sync

nakulkar-msft:


mohsha-msft (approved):
Looks Straightforward. LGTM!

Issue #1555, opened 2021-09-16
azcopy crashes during sync or copy of a whole Premium FileStorage to one another


flootsk:
##### Note: The version is visible when running AzCopy without any argument
10.12.1

### Which platform are you using? (ex: Windows, Mac, Linux)
Linux, Windows

##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.
```
./azcopy sync 'https://<sourcestorageaccount>.file.core.windows.net/<fileshare>?<SAS>' 'https://<deststorageaccount>.file.core.windows.net/<fileshare>?<SAS>' --log-level INFO --recursive=true
```

### What problem was encountered?
`azcopy` crashes during the scan phase after a random number of files scanned (e.g. sometime 2000, sometime 60000). It also happens during the copy with `azcopy copy [...]`

Here is a sample of what I get on stderr:

```
fatal error: concurrent map read and map write

goroutine 70 [running]:
runtime.throw(0xf70464, 0x21)
        /opt/hostedtoolcache/go/1.16.0/x64/src/runtime/panic.go:1117 +0x72 fp=0xc013f0dc40 sp=0xc013f0dc10 pc=0x438fd2
runtime.mapaccess2_faststr(0xe12200, 0xc0119b1c20, 0xc002149100, 0x7f, 0xc002149100, 0xc013f0dcd8)
        /opt/hostedtoolcache/go/1.16.0/x64/src/runtime/map_faststr.go:116 +0x4a5 fp=0xc013f0dcb0 sp=0xc013f0dc40 pc=0x415e85
github.com/Azure/azure-storage-azcopy/v10/ste.(*jpptFolderTracker).RecordCreation(0xc0116bc920, 0xc002149100, 0x7f)
        /home/vsts/work/1/s/ste/folderCreationTracker.go:80 +0x94 fp=0xc013f0dd08 sp=0xc013f0dcb0 pc=0xc7a554
github.com/Azure/azure-storage-azcopy/v10/ste.AzureFileParentDirCreator.CreateDirToRoot(0x10ca038, 0xc013c682a0, 0xc008d1da00, 0x5, 0x0, 0x0, 0x0, 0xc008d1da08, 0x2c, 0xc008d1da34, ...)
        /home/vsts/work/1/s/ste/sender-azureFile.go:455 +0xa55 fp=0xc013f0f728 sp=0xc013f0dd08 pc=0xc9a9f5
[...]

PR #1577, opened 2021-09-30
Jepio - ARM Support + CI Changes

mohsha-msft:


jepio:
If you install qemu-user on the VM, then you should be able to run the tests using `qemu-aarch64` on linux, but you'll have to disable `-race`. This article is relevant: https://ctrl-c.us/posts/test-goarch.html.

jepio:
This is fascinating: I think the testing is working correctly, but due to the emulation you're managing to reliably reproduce this issue (race condition): https://github.com/Azure/azure-storage-azcopy/issues/1555 👍

mohsha-msft:
Hey @jepio ,

Thanks for the help with QEMU. I was able to test ARM64 with this but, as you might have seen, I found race condition. 
I, unknowingly, reproduced error #1555 . I'll talk to my team about it.

zezha-msft:
@mohsha-msft sorry it took me a bit of time to do this test on a real ARM device, it worked great.

@jepio thank you so much for this contribution!

mohsha-msft:
@zezha-msft ,

Great. I'll remove the ARM E2E tests from CI. Will leave just the executables. We'll add it once Microsoft DevOps starts supporting ARM64 Agent.

zezha-msft (approved):
lgtm

PR #1579, opened 2021-10-04
Fix copying ADLS Gen 2 ACLs

adreed-msft:
In 10.12, with the introduction of ADLS Gen 2 ACL copying, our method of obtaining the folder objects was incorrect, and would lead to an inability to grab and copy ACLs on directory stubs.

This PR extends the test present and fixes this issue.

### DIFF
--- a/e2etest/declarativeScenario.go
+++ b/e2etest/declarativeScenario.go
@@ -271,7 +271,7 @@ func (s *scenario) validateTransferStates() {
 		actualTransfers, err := s.state.result.GetTransferList(statusToTest)
 		s.a.AssertNoErr(err)
 
-		Validator{}.ValidateCopyTransfersAreScheduled(s.a, isSrcEncoded, isDstEncoded, srcRoot, dstRoot, expectedTransfers, actualTransfers, statusToTest)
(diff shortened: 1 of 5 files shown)


e2etest/declarativeScenario.go
```diff
@@ -271,7 +271,7 @@ func (s *scenario) validateTransferStates() {
 		actualTransfers, err := s.state.result.GetTransferList(statusToTest)
 		s.a.AssertNoErr(err)
 
-		Validator{}.ValidateCopyTransfersAreScheduled(s.a, isSrcEncoded, isDstEncoded, srcRoot, dstRoot, expectedTransfers, actualTransfers, statusToTest)
+		Validator{}.ValidateCopyTransfersAreScheduled(s.a, isSrcEncoded, isDstEncoded, srcRoot, dstRoot, expectedTransfers, actualTransfers, statusToTest, s.FromTo())
```
mohsha-msft, line 274: We need to write struct for these types of functions in E2E. The list of parameters will keep on increasing otherwise.

mohsha-msft (approved):
:shipit:

siminsavani-msft (approved):

PR #1582, opened 2021-10-06
Widen managed disk accounts to all md- accounts instead of just md-impexp- accounts

pra-msft:
Confidential VM Provisioning Service (CPS) works directly with DiskRP for creation of managed disk. DiskRP sends a read-write SAS of an empty managed disk blob (it could have prefix md- or md-ssd- and not md-impexp-) to CPS. CPS is then supposed to upload encrypted VHD file to this managed disk blob. DiskRP then uses this blob for creation of managed disk. 
As the managed disk SAS that CPS get is of type md- or md-ssd- and not md-impexp-, azcopy doesn’t treat these SAS as managed disk blob but as typical unmanaged blob. Therefore, during upload to these blobs PutBlob is called which result in deletion of the original blob and re-creation of a new one wiping properties and metadata of the original blob. It also messes up the caching that BlobService has put in place for managed disk usecases.
This PR aims to consider all md- account as managed disk account and just md-impexp-.

### DIFF
--- a/ste/downloader-blob.go
+++ b/ste/downloader-blob.go
@@ -97,12 +97,10 @@ func (bd *blobDownloader) GenerateDownloadFunc(jptm IJobPartTransferMgr, srcPipe
 		info := jptm.Info()
 		u, _ := url.Parse(info.Source)
 		srcBlobURL := azblob.NewBlobURL(*u, srcPipeline)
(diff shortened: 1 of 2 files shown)


zezha-msft (approved):

mohsha-msft (approved):
LGTM

PR #1589, opened 2021-10-15
Change from Set/GetNamedSecurityInfo to Nt{Query,Set}SecurityInfo

adreed-msft:
This PR changes the way that we were obtaining and setting ACLs to ensure accuracy. Set/GetNamedSecurityInfo would do some merging, which made it not _entirely_ accurate for backups. However, the Nt{Query,Set}SecurityInfo APIs do _not_ do merging, and allow for considerably more accurate preservation of ACLs.


siminsavani-msft (approved):
LGTM!

e2etest/scenario_os_helpers_for_windows.go
```diff
@@ -28,9 +28,11 @@ import (
 	"strings"
 	"syscall"
 	"time"
+	"unsafe"
 
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 	"github.com/Azure/azure-storage-azcopy/v10/ste"
+	"github.com/hillu/go-ntdll"
```
mohsha-msft, line 35: Please update the license file for this package. 
Also, Is this a known repository?

e2etest/scenario_os_helpers_for_windows.go
```diff
@@ -86,8 +88,27 @@ func (osScenarioHelper) getFileAttrs(c asserter, filepath string) *uint32 {
...
+	bufLen := uint32(len(buf))
+	status := ntdll.CallWithExpandingBuffer(func() ntdll.NtStatus {
+		return ntdll.NtQuerySecurityObject(
+			ntdll.Handle(fd),
+			windows.OWNER_SECURITY_INFORMATION|windows.GROUP_SECURITY_INFORMATION|windows.DACL_SECURITY_INFORMATION,
+			(*ntdll.SecurityDescriptor)(unsafe.Pointer(&buf[0])),
+			uint32(len(buf)),
+			&bufLen)
+	}, &buf, &bufLen)
+
```
mohsha-msft, line 109: Can we verify that changes are working as expected by any unit test?

mohsha-msft (commented):
Change looks straight forward. I have some concerns about the package we're using tho.

e2etest/scenario_os_helpers_for_windows.go
adreed-msft: We already use this package-- It was required for fixing up folder times.

adreed-msft: (Line 9371 on NOTICE.txt)

zezha-msft:
TODO: an automated test to double check

ste/sourceInfoProvider-Local_windows.go
```diff
@@ -18,15 +23,24 @@ import (
 
 // This file os-triggers the ISMBPropertyBearingSourceInfoProvider and CustomLocalOpener interfaces on a local SIP.
 
-func (f localFileSourceInfoProvider) Open(path string) (*os.File, error) {
+func (f localFileSourceInfoProvider) GetHandle(path string) (ntdll.Handle, error) {
```
mohsha-msft, line 26: 1. Nit: This method can be private unless you see a foreseeable need to export it.
2. From the looks of it, I see this function is returning a file descriptor for given path. Also, can you add a small comment to explain the function?

sddl/sddlString.go
```diff
@@ -1,23 +1,3 @@
-// Copyright © Microsoft <wastore@microsoft.com>
```
mohsha-msft, line 1: We need license text in the start of the file, right?

ste/downloader-azureFiles_windows.go
```diff
@@ -147,28 +148,34 @@ func (a *azureFilesDownloader) PutSDDL(sip ISMBPropertyBearingSourceInfoProvider
 	isProtectedAtSource := (ctl & windows.SE_DACL_PROTECTED) != 0
 	isAtTransferRoot := len(splitPath) == 1
 
-	if isProtectedAtSource || isAtTransferRoot || a.parentIsShareRoot(txInfo.Source) {
-		securityInfoFlags |= windows.PROTECTED_DACL_SECURITY_INFORMATION
+	parsedSDDL, err := sddl.ParseSDDL(sddlString)
+
+	if err != nil {
+		panic(fmt.Sprintf("Sanity check; SDDL failed to parse (downloader-azureFiles_windows.go), %s", err)) // We already parsed it. This is impossible.
```
mohsha-msft, line 154: QQ: Why are we not returning error and panicking instead? Does unable to parse SDDL mean we have to fail immediately fatal?

e2etest/zt_preserve_smb_properties_test.go
```diff
@@ -56,14 +56,15 @@ func TestProperties_SMBPermissionsSDDLPreserved(t *testing.T) {
...
 	}
 
-	folderSDDL, err := AdjustSDDLToLocal(FolderSampleSDDL, SampleSDDLPlaceHolder)
-	if err != nil {
-		t.Error(err)
-	}
+	// folderSDDL, err := AdjustSDDLToLocal(FolderSampleSDDL, SampleSDDLPlaceHolder)
+	// if err != nil {
+	// 	t.Error(err)
+	// }
```
mohsha-msft, line 62: Let's get rid of un-necessary comments.

mohsha-msft (approved):
Overall changes look good. Just some Nitpicking.

ste/downloader-azureFiles_windows.go
adreed-msft: The SDDL has successfully been parsed in the past-- it doesn't make sense for it to fail here, so, if it has, something is _severely_ wrong.

PR #1605, opened 2021-10-27
[azbfs] Fix Rename Dir/File Not Authorized

tasherif-msft:
This PR ensures that the SAS token is appended to the `x-ms-rename-source`. Destination SAS can also now be passed


gapra-msft (approved):

azbfs/url_directory.go
```diff
@@ -174,12 +174,16 @@ func (d DirectoryURL) Rename(ctx context.Context, options RenameDirectoryOptions
...
-	renameSource := "/" + d.filesystem + "/" + d.pathParameter
-
 	urlParts := NewBfsURLParts(d.directoryClient.URL())
 	urlParts.FileSystemName = *fileSystemName
 	urlParts.DirectoryOrFilePath = options.DestinationPath
+	// ensure we use our source's SAS token in the x-ms-rename-source header
+	renameSource := "/" + d.filesystem + "/" + d.pathParameter + "?" + urlParts.SAS.Encode()
 
+	// if we're changing our source SAS to a new SAS in the rename
+	if options.DestinationSas != nil && *options.DestinationSas != "" {
```
zezha-msft, line 184: why does the user need to swap the SAS? Could you please add a comment for context?

zezha-msft (approved):
lgtm

why does the user need to swap SAS btw? I'm not sure I fully understand the scenario

azbfs/url_directory.go
gapra-msft: @zezha-msft I think this is in case the user has separate directory level sas permissions on the destination and the source as opposed to one higher sas permission (container, account, etc) that cover both the destination and source

tasherif-msft: I'll write this in the comment: `In the case the user wants to substitute for example: a dir level sas to an account level sas to rename to the new dir`

PR #1611, opened 2021-11-04
Enable copying folder properties to the share root.

adreed-msft:
This is a PR that I expect to cause a little debate.

The functionality, I believe we're all in agreement on. However, the naming of the flag & description are questionable.

I went through a couple different ideas for the flag's name:

- `--rename`: Quite literal, but might mislead a user into thinking that it's a cut&paste type operation
- `--dest-is-target`: Wordy, doesn't deliver it's point
- `--copy-into-destination`: Similarly wordy, similarly doesn't get the point across.
- `--share-root`: At first glance, this seemed offputting-- This operation isn't ONLY for copying to the share root. However, thinking past the technical term, I realized that it's quite fitting-- The root is literally being shared, as in, this is just a merge into the destination folder, not a new folder.
- `--as-subdir=false`: The default (true) demonstrates present activity (placing directories as subdirectories) and the opposite functionality (treating the destination as the target) in the name

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -123,7 +123,7 @@ type rawCopyCmdArgs struct {
 	preserveSMBPermissions bool
 	preservePermissions    bool // Separate flag so that we don't get funkiness with two "flags" targeting the same boolean
 	preserveOwner          bool // works in conjunction with preserveSmbPermissions
-	// Opt-in "rename" flag.
+	// Default true; false indicates that the destination is the target directory, rather than something we'd put a directory under (e.g. a container)
 	asSubdir bool
 	// Opt-in flag to persist additional SMB properties to Azure Files. Named ...info instead of ...properties
 	// because the latter was similar enough to preserveSMBPermissions to induce user error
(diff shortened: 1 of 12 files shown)


Biyoni (approved):
All looks good.

adreed-msft:
Killed the sync functionality since it requires a bit more work for testing & I don't have the time to fix it up. We'll get it in another PR.

cmd/copy.go
```diff
@@ -123,6 +123,8 @@ type rawCopyCmdArgs struct {
 	preserveSMBPermissions bool
 	preservePermissions    bool // Separate flag so that we don't get funkiness with two "flags" targeting the same boolean
 	preserveOwner          bool // works in conjunction with preserveSmbPermissions
+	// Default true; false indicates that the destination is the target directory, rather than something we'd put a directory under (e.g. a container)
+	asSubdir bool
```
zezha-msft, line 127: could you please clarify the difference compared to the existing flag `internalOverrideStripTopDir bool` with a comment?

zezha-msft (approved):

nakulkar-msft (approved):

PR #1612, opened 2021-11-05
Dramatically reduce sync's memory footprint on larger jobs

adreed-msft:
This PR attempts to reduce the memory usage of sync through a couple of changes:

1. The original map of relative paths to stored objects is replaced with a trie.
2. The amount of data stored in the ObjectIndexer is reduced to only mandatory properties (LMT & Entity type), path & name are determined from the trie location when needed
3. The perf change made for uploads is removed in favor of reduced memory footprint

### DIFF
--- a/cmd/syncComparator.go
+++ b/cmd/syncComparator.go
@@ -20,59 +20,11 @@
 
 package cmd
 
-import "strings"
+import (
+	"strings"
+)
 
-// with the help of an objectIndexer containing the source objects
-// find out the destination objects that should be transferred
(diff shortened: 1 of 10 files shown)


mohsha-msft:
@zezha-msft  @adreed-msft 
Let's do a perf test for this one. I think it is of great importance to see the improvement in memory footprint and impact on performance.

zezha-msft:
1. very large job to make sure the new scheme scales
2. a small job to make sure the new scheme is equally efficient

zezha-msft:
To follow up with a new PR.

PR #1615, opened 2021-11-09
Handle auth error against single file

adreed-msft:
Handles the auth error up-front instead of in the deeper enumerators

### DIFF
--- a/cmd/zc_traverser_blob.go
+++ b/cmd/zc_traverser_blob.go
@@ -84,6 +84,11 @@ func (t *blobTraverser) getPropertiesIfSingleBlob() (props *azblob.BlobGetProper
 	blobUrlParts := azblob.NewBlobURLParts(*t.rawURL)
 	blobUrlParts.BlobName = strings.TrimSuffix(blobUrlParts.BlobName, common.AZCOPY_PATH_SEPARATOR_STRING)
 
+	if blobUrlParts.BlobName == "" {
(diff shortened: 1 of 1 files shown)


mohsha-msft (approved):

cmd/zc_traverser_blob.go
```diff
@@ -135,6 +140,14 @@ func (t *blobTraverser) Traverse(preprocessor objectMorpher, processor objectPro
 			return errors.New("this blob uses customer provided encryption keys (CPK). At the moment, AzCopy does not support CPK-encrypted blobs. " +
 				"If you wish to make use of this blob, we recommend using one of the Azure Storage SDKs")
 		}
+
+		if resp := stgErr.Response(); resp == nil {
+			return fmt.Errorf("cannot list files due to reason %s", stgErr)
```
mohsha-msft, line 145: Can you help me understand code between L144-150?
Also, can you please fix the message in L139

adreed-msft: 144-150 is supposed to catch an authorization error. The initial if statement checking if the response is nil will probably never be hit, but it's better safe than sorry to have a nil check.

149 doesn't need to be fixed imo-- Auth errors are pretty direct via their normal StorageError output. If you have an alternative suggestion, I'm happy to hear it out.

siminsavani-msft (approved):
LGTM!

PR #1629, opened 2021-11-30
[AzCopy][StgExp][GithubStgExp#3792] Fail to copy to classic blob container with ‘preserve blob access tier’ enabled

siminsavani-msft:
Added classic/standard accounts into account when determining if tiers are support or not.


e2etest/declarativeRunner.go
```diff
@@ -35,17 +35,18 @@ import (
...
 	testFromTo TestFromTo,
 	validate Validate, // TODO: do we really want the test author to have to nominate which validation should happen?  Pros: better perf of tests. Cons: they have to tell us, and if they tell us wrong test may not test what they think it tests
 	/*_ interface{}, // TODO if we want it??, blockBLobsOnly or specifc/all blob types
-	_ interface{}, // TODO if we want it??, default auth type only, or specific/all auth types*/
+	  _ interface{}, // TODO if we want it??, default auth type only, or specific/all auth types*/
 	p params,
 	hs *hooks,
 	fs testFiles,
 	// TODO: do we need something here to explicitly say that we expect success or failure? For now, we are just inferring that from the elements of sourceFiles
+	destIsClassic bool,
```
adreed-msft, line 49: Is it reasonable to add this as an AccountType?

e2etest/declarativeTestFiles.go
```diff
@@ -84,19 +84,20 @@ func (h *contentHeaders) String() string {
...
+	blobTags              common.BlobTags
+	blobType              common.BlobType
+	creationTime          *time.Time
+	lastWriteTime         *time.Time
+	smbAttributes         *uint32
+	smbPermissionsSddl    *string
+	adlsPermissionsACL    *string // TODO: Test owner and group; needs a good target though.
+	cpkInfo               *common.CpkInfo
+	cpkScopeInfo          *common.CpkScopeInfo
+	s2sPreserveAccessTier string
```
adreed-msft, line 100: I can't see anywhere this is used for the creation of files or testing the files.

e2etest/declarativeRunner.go
siminsavani-msft: Yeah, I think making a destAccountType would make sense!

e2etest/declarativeTestFiles.go
siminsavani-msft: It's being used in e2etest/zt_preserve_access_tier_test.go at ln 37 and ln 51.

zezha-msft (approved):

e2etest/declarativeTestFiles.go
adreed-msft: What I meant was, I don't see this actually _doing_ anything in the creation nor validation of files (in scenario_helpers.go:349 (scenarioHelper.generateBlobsFromList) or declarativeScenario.go:327 (scenario.validateProperties)). As such, I don't believe these tests are actually testing anything other than a normal copy with a classic account.

Additionally, `accessTier` may be a better name.

adreed-msft (changes requested):

adreed-msft (approved):

PR #1646, opened 2022-01-05
Enable lint

mohsha-msft:


siminsavani-msft (approved):

azure-pipelines.yml
```diff
@@ -25,8 +26,20 @@ jobs:
...
+        env:
+          GO111MODULE: 'on'
         inputs:
           version: '1.16'
+
+      - script: |
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.43.0
+          echo 'Installation complete' 
+          ./bin/golangci-lint --version 
+          ./bin/golangci-lint run e2etest
```
zezha-msft, line 38: is this basically the standard? since [golint](https://github.com/golang/lint) is deprecated.

zezha-msft: Have you considered adding `go vet` as well?

e2etest/scenario_helpers.go
```diff
@@ -836,10 +853,12 @@ func (s scenarioHelper) downloadFileContent(a asserter, options downloadContentO
 	defer retryReader.Close() // The client must close the response body when finished with it
 
 	destData, err := ioutil.ReadAll(retryReader)
+	a.AssertNoErr(err)
 	downloadResp.Body(azfile.RetryReaderOptions{})
 	return destData
 }
 
+//nolint
```
zezha-msft, line 861: there seems to be a ton of //nolint, what's the reason?

mohsha-msft: I don't want to remove code which we are going to use in the near future. For example supporting S3, GCP Clients. The functions for them are there but `//nolint` will stop `golint` from marking them.

azure-pipelines.yml
mohsha-msft: Yes you're right. It's not the standard golint. It is the same one used by OneEngg team in Azure SDK Go Pipelines.

mohsha-msft: As for the vet, I have not considered adding `go vet` but I can add it as long as it can be used for a particular directory.

e2etest/scenario_helpers.go
mohsha-msft: And some times it gives false-positive as well.

azure-pipelines.yml
mohsha-msft: Ideally, we should have [these steps](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/pipelines/templates/steps/analyze.yml) in CI as well.

PR #1662, opened 2022-01-27
Fix #Issue 1625

mohsha-msft:


cmd/zc_traverser_local.go
```diff
@@ -341,9 +341,12 @@ func (t *localTraverser) Traverse(preprocessor objectMorpher, processor objectPr
 
 				var entityType common.EntityType
 				if fileInfo.IsDir() {
-					fileInfo, err = WrapFolder(filePath, fileInfo)
+					newFileInfo, err := WrapFolder(filePath, fileInfo)
 					if err != nil {
 						WarnStdoutAndScanningLog(fmt.Sprintf("Failed to get last change of target at %s: %s", filePath, err))
+					} else {
+						// fileInfo becomes nil in case we fail to wrap folder.
+						fileInfo = newFileInfo
```
mohsha-msft, line 349: `fileInfo` is used in later steps to get `fileInfo.Mode()` so it shouldn't be nil

siminsavani-msft (approved):

cmd/zc_traverser_local.go
```diff
@@ -341,9 +341,12 @@ func (t *localTraverser) Traverse(preprocessor objectMorpher, processor objectPr
 
 				var entityType common.EntityType
 				if fileInfo.IsDir() {
-					fileInfo, err = WrapFolder(filePath, fileInfo)
+					newFileInfo, err := WrapFolder(filePath, fileInfo)
 					if err != nil {
 						WarnStdoutAndScanningLog(fmt.Sprintf("Failed to get last change of target at %s: %s", filePath, err))
```
zezha-msft, line 346: Should we warn that the time based filters won't work correctly? I think @adreed-msft wrote the WrapFolder so that we can apply time based filters properly. It'd be good to clarify the consequences, at least in the comments if not in the message itself.

zezha-msft (approved):

PR #1667, opened 2022-02-03
Env var to control downloading to temp path

nakulkar-msft:


ste/xfer-remoteToLocal-file.go
```diff
@@ -171,7 +171,7 @@ func remoteToLocal_file(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pac
 		// to correct name.
 		pseudoId := common.NewPseudoChunkIDForWholeFile(info.Source)
 		jptm.LogChunkStatus(pseudoId, common.EWaitReason.CreateLocalFile())
-		dstFile, err = createDestinationFile(jptm, info.getTempDownloadPath(), fileSize, writeThrough)
+		dstFile, err = createDestinationFile(jptm, info.getDownloadPath(), fileSize, writeThrough)
```
mohsha-msft, line 174: I liked the previous name better. It gives clear idea that we're talking about temp download directory and not the actual destination directory.

nakulkar-msft: The value of env var dictates if we're downloading to a temp or the actual path. We're not downloading to temp path in all cases.

mohsha-msft (approved):

common/lifecyleMgr.go
```diff
@@ -590,6 +591,15 @@ func (lcm *lifecycleMgr) IsForceLoggingDisabled() bool {
 	return lcm.disableSyslog
 }
 
+func (lcm *lifecycleMgr) DownloadToTempPath() bool {
+	ret, err := strconv.ParseBool(lcm.GetEnvironmentVariable(EEnvironmentVariable.DownloadToTempPath()))
```
zezha-msft, line 595: consider doing what SetForceLogging does above, and save the parsed result rather than parsing it every time.

zezha-msft (approved):

PR #1671, opened 2022-02-14
Bug#1630 Empty subfolders created when using --recursive and -preserve-permissions

tiverma-msft:
I checked if the size of the object is not 0, because all the folders have a size of 0.


mohsha-msft:
A file can be of 0B too, right? Will your fix work in that case?

tiverma-msft:
> A file can be of 0B too, right? Will your fix work in that case?
Oh yes, in fact this was my only concern with respect to this fix.
Basically the idea is, if we can **pinpoint which traversed objects are files and which are folders**, we can selectively add rootdir to the files and the code will work (according to me) fine.
I tried to run the code this way and it ran without the issue arising.

nakulkar-msft (approved):

mohsha-msft (approved):

Issue #1653, opened 2022-01-10
azcopy hangs for 2 seconds following every job


mkcello96:
(This is a repost of https://github.com/Azure/azure-storage-azcopy/issues/1624 since it got missed)

### Which version of the AzCopy was used? 

Note: The version is visible when running AzCopy without any argument
10.13.0

### Which platform are you using? (ex: Windows, Mac, Linux)

Linux

### What command did you run?

azcopy copy 'https://qag1ge1cassandrabackup.blob.core.usgovcloudapi.net/cassandra-backup-cct?' '/tmp/medusa-restore-94bfe653-12db-48be-838f-364ecbc9228b/system_auth/resource_role_permissons_index-5f2fbdad91f13946bd25d5da3a5c35ec' --include-path 'qag1ge1lcct001.ag1.taservs.net/data/system_auth/resource_role_permissons_index-5f2fbdad91f13946bd25d5da3a5c35ec/md-22-big-Digest.crc32'

### What problem was encountered?

There is a 2 second pause after every job, even for a simple single-file download from blob storage. We are trying to speed up database restores with azcopy, which means that many azcopy commands will run in sequence. We try and batch as many files as possible but we end up with a lot of small batches of small files - the 2 second pauses really add up.

Can a flag be added to disable whatever is happening during these 2 seconds? I assume it's the job summary at the end but logs do not tell me what is going on during that time (I've tried enabling DEBUG as well). I have added the logs to the end of this description.

NOTE: I've followed suggestions here for optimizing performance: https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-optimize. None seem to help this situation.

NOTE 2: The servers that these commands are running on are Azure VMs that are heavily locked down in terms of access to the internet. There is a proxy that only allows specific traffic.

### How can we reproduce the problem in the simplest way?

Do any small-file copy job

### Have you found a mitigation/solution?

No

Logs:

```
2021/11/17 21:58:15 OS-Environment  linux
2021/11/17 21:58:15 OS-Architecture  amd64
2021/11/17 21:58:15 Log times are in UTC. Local time is 17 Nov 2021 21:58:15
[...]

Issue #1188, opened 2020-09-29
azcopy sync always reports 0 files scanned


lesterw1:
### Which version of the AzCopy was used? 
10.6.0

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows 10

### What command did you run?
azcopy sync "m:\XBMC" "https://mystorageaccount.blob.core.windows.net/multimedia/CBB_NUC1/M$/XBMC" --put-md5 --delete-destination prompt

### What problem was encountered?
Output reports "**0 Files Scanned at Source, 0 Files Scanned at Destination**"
Just below though, it reports the correct information.

Example command and output:
```
C:\Users\lesterw>azcopy sync "M:\Wallpaper" "https://mystorageaccount.blob.core.windows.net/multimedia/CBB_NUC1/M$/Wallpaper" --put-md5 --delete-destination prompt
INFO: Authenticating to destination using Azure AD
INFO: Any empty folders will not be processed, because source and/or destination doesn't have full folder support

Job 37da4aa9-f177-1940-52c3-76b0f758a8c7 has started
Log file is located at: C:\Users\lesterw\.azcopy\37da4aa9-f177-1940-52c3-76b0f758a8c7.log

0 Files Scanned at Source, 0 Files Scanned at Destination

Job 37da4aa9-f177-1940-52c3-76b0f758a8c7 Summary
Files Scanned at Source: 290
Files Scanned at Destination: 290
Elapsed Time (Minutes): 0.0334
Number of Copy Transfers for Files: 0
Number of Copy Transfers for Folder Properties: 0
Total Number Of Copy Transfers: 0
Number of Copy Transfers Completed: 0
Number of Copy Transfers Failed: 0
Number of Deletions at Destination: 0
Total Number of Bytes Transferred: 0
Total Number of Bytes Enumerated: 0
Final Job Status: Completed
```
Note that M: is an SMB network drive from a QNAP device. Not sure if it matters.

### How can we reproduce the problem in the simplest way?
Apparently, any azcopy sync with storage account as the destination.
Note 

### Have you found a mitigation/solution?
The correct information is displayed immediately below.

nakulkar-msft:
Display status one last time before completing the job

PR #1674, opened 2022-02-17
Make LCM quit after completion of last part

nakulkar-msft:
This will fix issues #1653 & #1188

### DIFF
--- a/common/lifecyleMgr.go
+++ b/common/lifecyleMgr.go
@@ -539,9 +539,10 @@ func (lcm *lifecycleMgr) InitiateProgressReporting(jc WorkController) {
 				doCancel()
 				continue // to exit on next pass through loop
 			case <-lcm.doneChannel:
+				
 					newCount = jc.ReportProgressOrExit(lcm)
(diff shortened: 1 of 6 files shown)


mohsha-msft (approved):

common/lifecyleMgr.go
```diff
@@ -515,13 +524,6 @@ func (lcm *lifecycleMgr) InitiateProgressReporting(jc WorkController) {
 				}
 			}
 
-			// wait a bit before fetching job status again, as fetching has costs associated with it on the backend
-			select {
-			case <-lcm.cancelChannel:
-				doCancel()
-			case <-time.After(wait):
```
zezha-msft, line 522: If I'm understanding correctly, we are now doing a busy wait?

Have you considered giving a minimum wait period (like 10ms) so that we don't busy wait?

nakulkar-msft: Yes, we're doing a busy wait. I'll add a select channel instead of busy wait here.

adreed-msft (approved):

PR #1678, opened 2022-02-22
MinIO TraceOn in DEBUG mode

mohsha-msft:


cmd/zc_traverser_s3.go
```diff
@@ -63,6 +69,24 @@ func (t *s3Traverser) IsDirectory(isSource bool) bool {
...
+	// We'll read the buffer, send the data to the log file and reset the buffer for next HTTP Trace.
+	// We can dump the buffer to logfile in one go as well, but we're doing it 1 by 1
+	// 1. Size of buffer stays in control and doesn't grow huge
+	// 2. We want to write trace log before the error log to make sense of what is happening
+	if t.outputStream != nil {
+		traceLog := (*t.outputStream).String()
+		if len(traceLog) > 0 {
+			t.outputStream.Reset()
+			if ste.JobsAdmin != nil {
+				ste.JobsAdmin.LogToJobLog(strings.ReplaceAll(traceLog, "\n\n", "\n"), t.logLevel)
```
zezha-msft, line 84: anything to scrub?

zezha-msft (approved):
Please confirm that the necessary scrubbing is done when logging the trace.

cmd/zc_traverser_s3.go
```diff
@@ -214,6 +218,10 @@ func newS3Traverser(credentialType common.CredentialType, rawURL *url.URL, ctx c
 			LogError: glcm.Error,
 		})
 
+	if t.logLevel == pipeline.LogDebug {
```
nakulkar-msft, line 221: azcopyScanningLogger knows what level to log on, logLevel should need not be inherited from s3Client

ste/sourceInfoProvider-S3.go
```diff
@@ -82,6 +82,9 @@ func newS3SourceInfoProvider(jptm IJobPartTransferMgr) (ISourceInfoProvider, err
 			LogError: func(str string) { p.jptm.Log(pipeline.LogError, str) },
 			Panic:    func(err error) { panic(err) },
 		})
+	if jptm.MinimumLevelToLevel() == pipeline.LogDebug {
```
nakulkar-msft, line 85: Here as well. The appLogger which is part of jptm has enough context to filter logLevel, log level need not be inherited from jptm, you should just set pipeline.LogDebug as level in NewS3HTTPTraceLogger

nakulkar-msft (changes requested):
Inheriting logLevel from traverser or jptm is not necessary as we are always logging at pipeline.LogDebug. Consider defaulting logLevel in NewS3HTTPTraceLogger to pipeline.LogDebug.

nakulkar-msft (approved):

PR #1691, opened 2022-03-03
Merge main back to dev after 10.14

zezha-msft:


nakulkar-msft (approved):

mohsha-msft (approved):
Nit: We can remove `fmt.Println` with `glcm.Info()`

azure-pipelines.yml
```diff
@@ -67,6 +67,12 @@ jobs:
       - task: PublishBuildArtifacts@1
         displayName: 'Publish Artifacts'
         condition: succeededOrFailed()
+        
+      - task: ComponentGovernanceComponentDetection@0
```
mohsha-msft, line 71: Just Curious: Are we adding this task as part of both build steps (Windows, Unix)? Or should it be a separate task outside of build task? Does the output change when running from different hosts?

zezha-msft: I don't think it makes any difference, but perhaps we should confirm by running it on a Windows build as well

PR #1695, opened 2022-03-04
Fix memleak in chunkedFileWriter

nakulkar-msft:
This PR fixes memleaks in chunkedFileWriter which would be exposed in case of an error while transfer is in progress. While this is not a concern while running azcopy, the leak would hold up chunkProcessor and transferProcessor goroutines when runinng as a daemon.


siminsavani-msft (approved):
LGTM! Nice work on the comments 😄

adreed-msft (approved):

PR #1700, opened 2022-03-09
Fix for Perm Delete

siminsavani-msft:


cmd/zc_traverser_blob.go
```diff
@@ -377,10 +377,13 @@ func (t *blobTraverser) createStoredObjectForBlob(preprocessor objectMorpher, bl
...
 	)
 
-	object.blobSnapshotID = blobInfo.Snapshot
 	object.blobDeleted = blobInfo.Deleted
-	if blobInfo.VersionID != nil {
-		object.blobVersionID = *blobInfo.VersionID
+	if t.includeDeleted && t.includeSnapshot {
+		object.blobSnapshotID = blobInfo.Snapshot
+	} else if t.includeDeleted && t.includeVersion {
+		if blobInfo.VersionID != nil {
```
mohsha-msft, line 384: Nit: Can be included in the above else condition. Is this a regression?

siminsavani-msft: Yeah, it wasn't picked up by the CI unfortunately.

adreed-msft (approved):

adreed-msft:
Ditto on testing.

zezha-msft:
We should expand E2E test to include versioning enabled accounts, which are missing in our CI right now.

Issue #1404, opened 2021-04-13
azcopy has excessive delay when run without arguments


bmc-msft:
### Which platform are you using? (ex: Windows, Mac, Linux)

Linux (via WSL2)

### What command did you run?

`azcopy` without arguments

### What problem was encountered?

The command prints it's help message, then pauses before exiting.  `azcopy` takes 8 seconds to run each time.

Note, this is on an lightly used 32 core host.

### How can we reproduce the problem in the simplest way?

`time azcopy`

### Have you found a mitigation/solution?

None

bmc-msft:
Linked is an strace log of the command: [strace.log](https://github.com/Azure/azure-storage-azcopy/files/6304653/strace.log)

mohsha-msft:
Hey @bmc-msft ,

Whenever AzCopy command is executed, it makes a call to CDN to check for the latest version which may be the reason for delay.

bmc-msft:
I recognize the desire to check for the latest version.

The version checking performed on every launch does an HTTPS connection to `aka.ms`, then an HTTPS connection `azcopyvnext.azureedge.net` on every execution.   In network constrained areas (such as behind a firewall that drops these sessions rather than rejecting them), these delays can be extremely excessive with no value.

Can the latest version be cached on some regular window, such as once per day, perhaps saving the results in the already existing `~/.azcopy` directory?

mohsha-msft:
Hey @bmc-msft ,

This call timeouts in 8s and doesn't influence the AzCopy job. We'll try to optimize it.

PR #1707, opened 2022-03-14
Cache AzCopy Version

mohsha-msft:


zezha-msft:
Hmm I definitely see the advantage of caching the version info, i.e. no more waiting on the HTTP call and potential errors in locked-down environments, but how about the disadvantages?

mohsha-msft:
> Hmm I definitely see the advantage of caching the version info, i.e. no more waiting on the HTTP call and potential errors in locked-down environments, but how about the disadvantages?

Let's look at the reason behind moving towards Cached version file.
1. Our current `latest_version.txt file` is in West US and we don't have a CDN for fast download. So for a person running AzCopy from far locations (say NewZeland or Eastern Asia), the time to complete this REST request can be upto 1.5sec. 
2. It's not a problem for non-frequent AzCopy users but for users who are running AzCopy multiple times, it is a big problem. We got an issue where a person was trying to sync 18+M blobs. Since traversal can take a lot of time, he reasonably decided to break the sync job into smaller jobs. (~4500). When he ran the commands, it took 4500 * [200ms-1s] = [16mins-80mins]. 
3. I've set the cache expiry time to be 24 hrs. So, we check for version once every day. It reduced the number of REST request to check the version once every day.

nakulkar-msft (approved):

malthe:
Bump

demoray:
Does this PR handle when azcopy is called from within a constrained environment such that the version request fails?  In some environments, this has an even bigger impact.  Ref:  https://github.com/Azure/azure-storage-azcopy/issues/1404

cmd/root.go
```diff
@@ -212,15 +214,34 @@ func init() {
...
+		// Step 1: Fetch & validate the cached version and if it is updated, we'll return without making REST call.
+		filePath := filepath.Join(azcopyAppPathFolder, "latest_version.txt")
+		v2, err := ValidateCachedVersion(filePath)
+		if err == nil {
+			if v1.OlderThan(*v2) {
+				executablePathSegments := strings.Split(strings.Replace(os.Args[0], "\\", "/", -1), "/")
+				executableName := executablePathSegments[len(executablePathSegments)-1]
+
+				// output in info mode instead of stderr, as it was crashing CI jobs of some people
+				glcm.Info(executableName + ": A newer version " + v2.original + " is available to download\n")
```
gapra-msft, line 239: we should add the current version to this print string too

gapra-msft:
Closing this PR in favor of @siminsavani-msft 's

PR #1710, opened 2022-03-15
Dynamically change b/w

nakulkar-msft:


common/lcmInputs.go
```diff
@@ -0,0 +1,26 @@
...
+
+var EInputMsgType InputMsgType
+
+func(InputMsgType) Invalid()                                InputMsgType {return InputMsgType(0)}
+func(InputMsgType) CancelJob()                              InputMsgType {return InputMsgType(1)}
+func(InputMsgType) E2EInterrupts()                          InputMsgType {return InputMsgType(2)}
+func(InputMsgType) ThroughputAdjustment()                   InputMsgType {return InputMsgType(3)}
+
+
+var MsgTypeMap = map[string]InputMsgType{
```
zezha-msft, line 15: please check the other enums, they have a different pattern to achieve the "fromString" functionality

common/lifecyleMgr.go
```diff
@@ -108,24 +113,62 @@ func (lcm *lifecycleMgr) watchInputs() {
 			continue
 		}
 
-		// reads input until the first occurrence of \n in the input,
-		input, err := consoleReader.ReadString('\n')
-		timeReceived := time.Now()
+		//if we allow cancelFromStdin, the json messages will not be allowed
```
zezha-msft, line 116: why not? StgExp will use both most likely

common/lifecyleMgr.go
```diff
@@ -108,24 +113,62 @@ func (lcm *lifecycleMgr) watchInputs() {
...
+		decoder := json.NewDecoder(os.Stdin)
+		err := decoder.Decode(&msg)
 		if err != nil {
+			lcm.Info("Discarding incorrectly formatted input message.")
 			continue
 		}
 
-		// remove spaces before/after the content
-		msg := strings.TrimSpace(input)
+		lcm.Info(fmt.Sprintf("Received Msg %s: %s %s", msg.TimeStamp.String(), msg.MsgType, msg.Value))
```
zezha-msft, line 153: style: make output more formal

common/lcmInputs.go
```diff
@@ -0,0 +1,26 @@
...
+import "time"
+
+type InputMsgType uint16
+
+var EInputMsgType InputMsgType
+
+func(InputMsgType) Invalid()                                InputMsgType {return InputMsgType(0)}
+func(InputMsgType) CancelJob()                              InputMsgType {return InputMsgType(1)}
+func(InputMsgType) E2EInterrupts()                          InputMsgType {return InputMsgType(2)}
+func(InputMsgType) ThroughputAdjustment()                   InputMsgType {return InputMsgType(3)}
```
zezha-msft, line 12: Consider making ThroughputAdjustment more future proof, ex: PerformanceAdjustment

zezha-msft: PerformanceAdjustment's body could be a JSON that includes many fields

mohsha-msft (commented):
1. Please handle < 0 cases. Let's discuss why pacer uses `int64` instead of `uint64`
2. Please create documentation to help users understand the timestamp format, enums, and bandwidth units.  Changes in `helpMessage.go` or a Wiki section perhaps

common/lcmInputs.go
nakulkar-msft: Introduced a struct to handle this

nakulkar-msft: Introduced a struct to handle this

common/lcmInputs.go
```diff
@@ -0,0 +1,26 @@
+package common
```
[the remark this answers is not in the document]
nakulkar-msft: Removed the file. Appeared unnecessary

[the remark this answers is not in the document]
nakulkar-msft: Removed the file. Appeared unnecessary

common/lifecyleMgr.go
nakulkar-msft: We do not have an easy way to peek stdin to check if the message is cancel or a new json struct type. json decoder will not work with partial strings, and we would give up ease of using the library if we parse it ourselves. Given stegExp will be upgrading tto this feature, it would be reasonable to ask them to move to new format of input on cancel as well.

azure-pipelines.yml
```diff
@@ -95,7 +95,7 @@ jobs:
           set -e
           GOARCH=amd64 GOOS=linux go build -o azcopy_linux_amd64
           export AZCOPY_E2E_EXECUTABLE_PATH=$(pwd)/azcopy_linux_amd64
-          go test -timeout 20m -race -short -cover ./e2etest
```
nakulkar-msft, line 98: Changes in this file will be discarded.

mohsha-msft: +1. As long as we don't update the 60min timeout.

common/fe-ste-models.go
```diff
@@ -1495,3 +1495,33 @@ func GetClientProvidedKey(options CpkOptions) azblob.ClientProvidedKeyOptions {
...
+	}
+	return err
+}
+
+func (m *LCMMsgType) String() string {
+	return enum.StringInt(m, reflect.TypeOf(m))
+}
+
+type LCMMsg struct {
+	TimeStamp time.Time `json:"TimeStamp"`
```
mohsha-msft, line 1524: I am still a bit skeptic exposing the timestamp. But shouldn't be a major problem.

common/lifecyleMgr.go
```diff
@@ -124,6 +129,28 @@ func (lcm *lifecycleMgr) watchInputs() {
...
+			var m LCMMsg
+			err = json.Unmarshal([]byte(msg), &m)
+			if err != nil {
+				lcm.Info("Discarding incorrectly formatted input message.")
+				continue
+			}
+
+			lcm.Info(fmt.Sprintf("Received request for %s with timeStamp %s", m.MsgType, m.TimeStamp.String()))
+			var msgType LCMMsgType
+			if err := msgType.Parse(m.MsgType); err != nil {
```
mohsha-msft, line 142: Just Curious: Does it take care of casing (lower/upper) in param names?

ste/JobsAdmin.go
```diff
@@ -608,6 +606,10 @@ func (ja *jobsAdmin) BytesOverWire() int64 {
 	return ja.pacer.GetTotalTraffic()
 }
 
+func (ja *jobsAdmin) UpdateTargetBandwidth(newTarget int64) {
```
mohsha-msft, line 609: uint64? If not, can we check for <= condition

ste/pacer-tokenBucketPacer.go
```diff
@@ -69,6 +71,7 @@ type tokenBucketPacer struct {
 	atomicGrandTotal           int64
 	atomicWaitCount            int64
 	expectedBytesPerRequest    int64
+	newTargetBytesPerSecond    chan int64
```
mohsha-msft, line 74: ditto on `int64`

mohsha-msft (approved):
Minor `Nit`

siminsavani-msft (approved):
LGTM!

adreed-msft (approved):

PR #1711, opened 2022-03-15
[AzCopyV10]  last-write-time still getting persisted even when --preserve-smb-info is false

siminsavani-msft:


cmd/syncProcessor.go
```diff
@@ -49,7 +54,7 @@ func newSyncTransferProcessor(cca *cookedSyncCmdArgs, numOfTransfersPerPart int,
 
 		// flags
 		BlobAttributes: common.BlobTransferAttributes{
-			PreserveLastModifiedTime: true, // must be true for sync so that future syncs have this information available
+			PreserveLastModifiedTime: preserveLastModifiedTime,
```
mohsha-msft, line 57: Is it equivalent to: `!cca.preserveSMBInfo`?

siminsavani-msft: Oh yeah, it is!

siminsavani-msft: It would be equivalent to `cca.preserveSMBInfo` which makes it simpler. Good call out :)

mohsha-msft (approved):

nakulkar-msft (approved):

PR #1721, opened 2022-03-22
Introduce syslogger to be used with STE

nakulkar-msft:


common/logger_unix.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"runtime"
+	"log/syslog"
+
+	"github.com/Azure/azure-pipeline-go/pipeline"
+)
+
+//////////////////////////////////////////
+type sysLogger struct {
+	// maximum loglevel represents the maximum severity of log messages which can be logged to Job Log file.
```
zezha-msft, line 34: wait, it's the opposite, right? 

// **minimum** loglevel represents the **minimum** severity of log messages which can be logged to Job Log file.
// any message with severity **lower** than this will be ignored.

common/logger_unix.go
```diff
@@ -0,0 +1,115 @@
...
+	return &sysLogger{
+		jobID:             jobID,
+		minimumLevelToLog: minimumLevelToLog.ToPipelineLogLevel(),
+		logSuffix:         logSuffix, 
+		sanitizer:         NewAzCopyLogSanitizer(),
+	}
+}
+
+func (sl *sysLogger) OpenLog() {
+    writer, err := syslog.New(syslog.LOG_NOTICE, fmt.Sprintf("%s %s", sl.logSuffix, sl.jobID.String()))
```
zezha-msft, line 54: Are we missing the following block?

if sl.minimumLevelToLog == pipeline.LogNone {
		return
	}

Since we have it in CloseLog.

zezha-msft (approved):

PR #1735, opened 2022-04-04
Sync on par with Copy: Enable flags which can be supported

nakulkar-msft:


adreed-msft (dismissed):

zezha-msft (commented):
these are a lot of new flags, how do we test to make sure they all work as expected?

rsheptolut:
@nakulkar-msft Thanks for working on this! We're using sync and were kind of disappointed too that sync doesn't support all the flags when using sync. Makes no sense sometimes. 

Did you also think about making --cache-control work with sync? Our use case is we're syncing from our local folder to an Azure Storage container, but the cache-control headers that get applied by default are kind of meh, and we wanted to specify ours. But azcopy says that there's simply no parameter --cache-control! And looks like now there's no better workaround than to build a webjob that would be updating cache-control headers periodically on all blobs, which is not a pretty solution.

PR #1743, opened 2022-04-06
Set Properties / t-iverma

tiverma-msft:


cmd/setProperties.go
```diff
@@ -0,0 +1,120 @@
...
+	setPropCmd.PersistentFlags().StringVar(&raw.exclude, "exclude-pattern", "", "Exclude files where the name matches the pattern list. For example: *.jpg;*.pdf;exactName")
+	setPropCmd.PersistentFlags().StringVar(&raw.excludePath, "exclude-path", "", "Exclude these paths when removing. "+
+		"This option does not support wildcard characters (*). Checks relative path prefix. For example: myFolder;myFolder/subDirName/file.pdf")
+	setPropCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "Defines the location of text file which has the list of only files to be copied.")
+	setPropCmd.PersistentFlags().StringVar(&raw.blockBlobTier, "block-blob-tier", "None", "Changes the access tier of the blobs to the given tier")
+	setPropCmd.PersistentFlags().StringVar(&raw.pageBlobTier, "page-blob-tier", "None", "Upload page blob to Azure Storage using this blob tier. (default 'None').")
+	setPropCmd.PersistentFlags().BoolVar(&raw.recursive, "recursive", false, "Look into sub-directories recursively when uploading from local file system.")
+}
```
mohsha-msft, line 120: We're missing the variable to specify which properties to transfer too. It needs to be calculated here so that we can pass it along.

mohsha-msft (changes requested):

ste/xfer-setProperties.go
```diff
@@ -0,0 +1,224 @@
+package ste
```
mohsha-msft, line 1: Do we have unit tests in STE? Please add them here.

cmd/zt_set_properties_test.go
```diff
@@ -0,0 +1,540 @@
+// Copyright © 2017 Microsoft <wastore@microsoft.com>
```
mohsha-msft, line 1: Try adding test cases for cases when user passes wrong value of flag. Will the code be able to handle that?
For example: What if the user passes wrong fromTo?

mohsha-msft (changes requested):
Test review pending. I'll take one more look later.

common/fe-ste-models.go
```diff
@@ -119,6 +119,19 @@ func (DeleteSnapshotsOption) None() DeleteSnapshotsOption    { return DeleteSnap
 func (DeleteSnapshotsOption) Include() DeleteSnapshotsOption { return DeleteSnapshotsOption(1) }
 func (DeleteSnapshotsOption) Only() DeleteSnapshotsOption    { return DeleteSnapshotsOption(2) }
 
+type SetPropertiesAPIOption uint32 // [0000000000...32 times]
+
+var ESetPropertiesAPIOption = SetPropertiesAPIOption(0)
+
+// functions to set values
+func (SetPropertiesAPIOption) None()        { ESetPropertiesAPIOption |= 0 }
```
adreed-msft, line 127: Generally speaking, global variables are considered bad practice. Instead, return bitflags to be ORed into place, as this looks like it's supposed to be a bitflag enum.

E.g. I should be writing `ESetPropertiesAPIOption.SetTier | ESetPropertiesAPIOption.SetMetadata`

common/fe-ste-models.go
```diff
@@ -119,6 +119,19 @@ func (DeleteSnapshotsOption) None() DeleteSnapshotsOption    { return DeleteSnap
...
+func (SetPropertiesAPIOption) SetTier()     { ESetPropertiesAPIOption |= 1 }
+func (SetPropertiesAPIOption) SetMetadata() { ESetPropertiesAPIOption |= 2 }
+
+// functions to get values (to be used in sde)
+func (SetPropertiesAPIOption) TransferTier() bool     { return int32(ESetPropertiesAPIOption&1) == 1 }
```
adreed-msft, line 132: Having functions like this to check that it does one thing or another is a good idea for bitflag enums. However, I would advise 

1) Using a receiver variable (e.g. `func (e SetPropertiesAPIOption)`) and checking against that
2) Using the references to the actual flag names rather than magic numbers

cmd/zt_scenario_helpers_for_test.go
```diff
@@ -909,3 +909,28 @@ func getDefaultRemoveRawInput(src string) rawCopyCmdArgs {
 		includeDirectoryStubs:          true,
 	}
 }
+
+func getDefaultSetPropertiesRawInput(src string, accessTier common.BlockBlobTier) rawCopyCmdArgs {
+	fromTo := common.EFromTo.BlobNone()
+	srcURL, _ := url.Parse(src)
+
+	if strings.Contains(srcURL.Host, "file") {
+		fromTo = common.EFromTo.BlobFSNone()
```
adreed-msft, line 918: 1) Look into how to use the location detection from copy-- The URIs `https://myfileaccount.blob.core.windows.net` and `https://mydfsaccount.blob.core.windows.net` would fail this check.

2) BlobFS for Files? That smells wrong.

ste/xfer-setProperties.go
```diff
@@ -0,0 +1,224 @@
...
+		to := jptm.FromTo()
+		switch to.From() {
+		case common.ELocation.Blob():
+			setPropertiesBlob(jptm, p)
+		case common.ELocation.BlobFS():
+			setPropertiesBlobFS(jptm, p)
+		case common.ELocation.File():
+			setPropertiesFile(jptm, p)
+		default:
+			panic("Shouldn't have happened")
```
adreed-msft, line 76: This is a nondescript panic. Often, we receive reports of crashes without complete panic logs, which makes it very hard to debug if it's not easy to understand.

I would recommend using something like... `fmt.Sprintf("Sanity check: Attempted set properties on invalid location %s", to.String())`

ste/xfer-setProperties.go
```diff
@@ -0,0 +1,224 @@
...
+		jptm.ReportTransferDone()
+		return
+	}
+
+	// schedule the work as a chunk, so it will run on the main goroutine pool, instead of the
+	// smaller "transfer initiation pool", where this code runs.
+	id := common.NewChunkID(jptm.Info().Source, 0, 0)
+	cf := createChunkFunc(true, jptm, id, func() {
+		to := jptm.FromTo()
+		switch to.From() {
```
adreed-msft, line 68: Nitpicking: Generally speaking, our individual chunk functions don't operate across multiple different endpoints. This could be a function that returns a func() that has a single, dedicated purpose.

ste/xfer-setProperties.go
```diff
@@ -0,0 +1,224 @@
...
+	// Get the source blob url of blob to set properties on
+	u, _ := url.Parse(info.Source)
+	srcBlobURL := azblob.NewBlobURL(*u, p)
+
+	// Internal function which checks the transfer status and logs the msg respectively.
+	// Sets the transfer status and Reports Transfer as Done.
+	// Internal function is created to avoid redundancy of the above steps from several places in the api.
+	transferDone := func(status common.TransferStatus, err error) {
+		if status == common.ETransferStatus.Failed() {
+			jptm.LogError(info.Source, "SET-PROPERTIES ERROR ", err)
```
adreed-msft, line 93: Nitpick: When support engineers are assisting customers, they will grep the log for `FAIL` or `FAILED`.

common/fe-ste-models.go
adreed-msft: Upon second viewing, throughout your code, you choose to use SetPropertiesAPIOption two ways:

1) As a local variable in the STE
2) As a global variable in the frontend

ste/xfer-setProperties.go
```diff
@@ -0,0 +1,224 @@
...
+
+	blockBlobTier, pageBlobTier := jptm.BlobTiers()
+	srcBlobType := jptm.Info().SrcBlobType
+	_ = srcBlobType
+	SetPropertiesAPIOption := jptm.SetPropertiesAPIOption()
+
+	if SetPropertiesAPIOption.TransferTier() {
+		var err error = nil
+		switch srcBlobType {
+		case azblob.BlobBlockBlob:
```
adreed-msft, line 110: This can be a single line. In addition, there is tier safety checking in the STE. Please take a look at the blob senders for usage.

ste/xfer-setProperties.go
```diff
@@ -0,0 +1,224 @@
...
+			jptm.LogError(info.Source, "SET-PROPERTIES ERROR ", err)
+		} else {
+			jptm.Log(pipeline.LogInfo, fmt.Sprintf("SET-PROPERTIES SUCCESSFUL: %s", strings.Split(info.Destination, "?")[0]))
+		}
+
+		jptm.SetStatus(status)
+		jptm.ReportTransferDone()
+	}
+
+	_, err := srcBlobURL.SetAccessControl(jptm.Context(), azbfs.BlobFSAccessControl{})
```
adreed-msft, line 165: There is no option in the SetPropertiesAPIOption to restrict the use of ACLs.

ste/xfer-setProperties.go
```diff
@@ -0,0 +1,224 @@
...
+			jptm.LogError(info.Source, "SET-PROPERTIES ERROR ", err)
+		} else {
+			jptm.Log(pipeline.LogInfo, fmt.Sprintf("SET-PROPERTIES SUCCESSFUL: %s", strings.Split(info.Destination, "?")[0]))
+		}
+
+		jptm.SetStatus(status)
+		jptm.ReportTransferDone()
+	}
+
+	_, err := srcFileURL.SetMetadata(jptm.Context(), info.SrcMetadata.ToAzFileMetadata())
```
adreed-msft, line 205: Metadata is checked in blob; why not in Files?

ste/xfer-setProperties.go
adreed-msft: Metadata also still applies in BlobFS. Same potentially to blob tier.

adreed-msft (changes requested):

cmd/setProperties.go
```diff
@@ -0,0 +1,120 @@
...
+				switch srcLocationType {
+				case common.ELocation.Blob():
+					raw.fromTo = common.EFromTo.BlobNone().String()
+				case common.ELocation.BlobFS():
+					raw.fromTo = common.EFromTo.BlobFSTrash().String()
+				default:
+					return fmt.Errorf("invalid source type %s to delete. azcopy support removing blobs/files/adls gen2", srcLocationType.String())
+				}
+			} else if raw.fromTo != "" {
+				err := strings.Contains(raw.fromTo, "Trash")
```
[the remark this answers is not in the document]
tiverma-msft: Changed, strings will pass this preliminary check if you have "None" in them. Eg FileNone. However, some strings like "FileNonexyz" will pass too. This will be handled by the cook method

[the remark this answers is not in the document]
tiverma-msft: Here we're just creating the rawcopycmdargs. The cook method will take care of this. Here, we're just making sure that "none" is present (the given blob is indeed intended to be changed the properties of.

``` fromTo, err := ValidateFromTo(raw.src, raw.dst, raw.fromTo) // TODO: src/dst ```

"NoneBlob" fails here in raw.cook()

File: cmd/setProperties.go (hunk 1 above)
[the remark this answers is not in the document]
Comment (tiverma-msft) [reply cmd/setProperties.go:120]: done

Comment (tiverma-msft) [reply cmd/setProperties.go:120]: wouldn't that be calculated by the code itself? Why do we want to take the values from the user explicitly? Wouldn't it be simpler if we infer the value of that flag depending on what values the user has provided?

File: cmd/setPropertiesEnumerator.go [hunk 14]
```diff
@@ -0,0 +1,99 @@
...
+				return nil
+			} else if err == NothingScheduledError {
+				// No log file needed. Logging begins as a part of awaiting job completion.
+				return NothingToRemoveError
+			}
+
+			return err
+		}
+
+		// TODO: this appears to be obsolete due to the above err == NothingScheduledError. Review/discuss.
```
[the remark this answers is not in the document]
tiverma-msft: this todo is from remove.go itself. Since the code flow is similar, I chose to keep it in my code as well, just in case someone uses that TODO to do changes later

cmd/setPropertiesProcessor.go
```diff
@@ -0,0 +1,69 @@
...
+		FromTo:          cca.FromTo,
+		Fpo:             fpo,
+		SourceRoot:      cca.Source.CloneWithConsolidatedSeparators(), // TODO: why do we consolidate here, but not in "copy"? Is it needed in both places or neither? Or is copy just covering the same need differently?
+		CredentialInfo:  cca.credentialInfo,
+		ForceIfReadOnly: cca.ForceIfReadOnly,
+
+		// flags
+		LogLevel: cca.LogVerbosity,
+		BlobAttributes: common.BlobTransferAttributes{
+			DeleteSnapshotsOption: cca.deleteSnapshotsOption,
```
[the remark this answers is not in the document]
tiverma-msft: 👍

ste/xfer-setProperties.go
tiverma-msft: Totally makes sense, must have overlooked this while I was coding. Thanks

ste/xfer-setProperties.go
tiverma-msft: @mohsha-msft can you help me with this?

cmd/setProperties.go
```diff
@@ -0,0 +1,120 @@
...
+
+			cooked.commandString = copyHandlerUtil{}.ConstructCommandStringFromArgs()
+			err = cooked.process()
+
+			// TODO -> what is the purpose of these lines?
+			if err != nil {
+				glcm.Error("failed to perform set-properties command due to error: " + err.Error())
+			}
+
+			//if cooked.dryrunMode {
```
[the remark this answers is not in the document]
tiverma-msft: added, test pending

cmd/setPropertiesEnumerator.go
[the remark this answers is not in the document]
tiverma-msft: deleted

cmd/setProperties.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func init() {
+	raw := rawCopyCmdArgs{}
+
+	setPropCmd := &cobra.Command{
+		Use:        "set-properties",
+		Aliases:    []string{},
+		SuggestFor: []string{},
+		// TODO: t-iverma: short and long descriptions
```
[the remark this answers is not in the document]
tiverma-msft: will be added in the end when the final code is ready

cmd/setProperties.go
tiverma-msft: added

common/fe-ste-models.go
```diff
@@ -538,6 +554,9 @@ func (FromTo) BlobFile() FromTo    { return FromTo(fromToValue(ELocation.Blob(),
 func (FromTo) FileFile() FromTo    { return FromTo(fromToValue(ELocation.File(), ELocation.File())) }
 func (FromTo) S3Blob() FromTo      { return FromTo(fromToValue(ELocation.S3(), ELocation.Blob())) }
 func (FromTo) GCPBlob() FromTo     { return FromTo(fromToValue(ELocation.GCP(), ELocation.Blob())) }
+func (FromTo) BlobNone() FromTo    { return fromToValue(ELocation.Blob(), ELocation.None()) }
```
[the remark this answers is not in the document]
tiverma-msft: @mohsha-msft thoughts?

common/fe-ste-models.go
```diff
@@ -598,6 +617,10 @@ func (ft *FromTo) AreBothFolderAware() bool {
 	return ft.From().IsFolderAware() && ft.To().IsFolderAware()
 }
 
+func (ft *FromTo) IsPropertyOnlyTransfer() bool {
```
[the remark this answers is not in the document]
tiverma-msft: This function is not used anywhere. You added it right? What is it intended for?

ste/xfer-setProperties.go
tiverma-msft: added tier validation to code. [https://github.com/Azure/azure-storage-azcopy/pull/1743/commits/5ccf809af418ee4eb75e29136a7f7e0c2c115f02](url)

cmd/zt_scenario_helpers_for_test.go
tiverma-msft: `srcLocationType := InferArgumentLocation(src)` used.
Resolved for test helper function

ste/xfer-setProperties.go
tiverma-msft: added

common/fe-ste-models.go
```diff
@@ -1496,17 +1519,15 @@ func GetClientProvidedKey(options CpkOptions) azblob.ClientProvidedKeyOptions {
...
 type LCMMsgType uint16
 
 var ELCMMsgType LCMMsgType
 
-func(LCMMsgType) Invalid()                                LCMMsgType {return LCMMsgType(0)}
-func(LCMMsgType) CancelJob()                              LCMMsgType {return LCMMsgType(1)}
-func(LCMMsgType) E2EInterrupts()                          LCMMsgType {return LCMMsgType(2)}
-func(LCMMsgType) PerformanceAdjustment()                  LCMMsgType {return LCMMsgType(3)}
-
+func (LCMMsgType) Invalid() LCMMsgType               { return LCMMsgType(0) }
```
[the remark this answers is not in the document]
tiverma-msft: pending - discussion required.

common/fe-ste-models.go
tiverma-msft: resolved

cmd/copy.go
```diff
@@ -1443,6 +1443,14 @@ func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 	// case common.EFromTo.FileBlob():
 	// 	e := copyFileToNEnumerator(jobPartOrder)
 	// 	err = e.enumerate(cca)
+
+	case common.EFromTo.BlobNone(), common.EFromTo.BlobFSNone():
```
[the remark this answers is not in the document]
tiverma-msft: added

ste/xfer-setProperties.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+
+		jptm.SetStatus(status)
+		jptm.ReportTransferDone()
+	}
+
+	PropertiesToTransfer := jptm.PropertiesToTransfer()
+
+	if PropertiesToTransfer.ShouldTransferTier() {
+		err := fmt.Errorf("trying to change tier of file")
```
tiverma-msft, line 163: make error shift to frontend

tiverma-msft: in cook or other func

ste/JobPartPlanFileName.go
```diff
@@ -218,6 +218,7 @@ func (jpfn JobPartPlanFileName) Create(order common.CopyJobPartOrderRequest) {
 		DeleteSnapshotsOption:          order.BlobAttributes.DeleteSnapshotsOption,
 		PermanentDeleteOption:          order.BlobAttributes.PermanentDeleteOption,
 		RehydratePriority:              order.BlobAttributes.RehydratePriority,
+		Metadata:                       order.BlobAttributes.Metadata,
```
mohsha-msft, line 221: Pretty sure it's not required.

mohsha-msft (changes requested):

ste/JobPartPlanFileName.go
tiverma-msft: there's no existing way of retrieving metadata for setproperties

mohsha-msft: See how we set metadata. We already persist metadata in plan files. All we need is a way to extract is. I'm sure there are other occurrences in the code.

tiverma-msft: Okay I will. What I was trying to say was that what we do currently is that we have a flow to "preserve metadata" while copying file from one place to another. 
When we're copying from local, the user can define what metadata they want to upload with and we pass it to the common.CopyJobPartOrderRequest{}. However, this struct is passed to the enumerator only in case of the copy command and no other case, not even ours. Hence, I don't see a way of passing that metadata back if we don't do this

tiverma-msft: oh i get it now! You were right. Removing this.

mohsha-msft (changes requested):

PR #1754, opened 2022-04-22
quiet-mode-tiverma

tiverma-msft:


tiverma-msft:
problems with rebasing, will have to resolve that issue

mohsha-msft (changes requested):
Pull changes from azcopy:dev and rebase again. These commits shouldn't be part of the PR.

cmd/copy.go
```diff
@@ -853,6 +856,20 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
...
+		return cooked, err
+	}
+
+	if cooked.ForceWrite == common.EOverwriteOption.Prompt() && (cooked.quietMode == common.EQuietMode.Quiet() || cooked.quietMode == common.EQuietMode.ErrorsOnly()) {
+		err = fmt.Errorf("cannot set quiet mode '%s' with specified overwrite option '%s'", cooked.quietMode.String(), cooked.ForceWrite.String())
+	}
+	if err != nil {
+		return cooked, err
+	}
+	// TODO if we press ctrl c, terminate without seeking answer whether user wants to quit
```
tiverma-msft, line 871: does it really ask if we wanna quit? doesn't it just end?

cmd/copy.go
```diff
@@ -170,6 +170,9 @@ type rawCopyCmdArgs struct {
 
 	// Optional flag that permanently deletes soft-deleted snapshots/versions
 	permanentDeleteOption string
+
+	// Optional flag to specify the quiet-mode of the operation
+	quietMode string
```
nakulkar-msft, line 175: root.go would be a better place for this flag.

common/lifecyleMgr.go
```diff
@@ -72,8 +72,9 @@ type LifecycleMgr interface {
 	SetForceLogging()
 	IsForceLoggingDisabled() bool
 	DownloadToTempPath() bool
-	MsgHandlerChannel()   <-chan LCMMsg
+	MsgHandlerChannel() <-chan LCMMsg
 	ReportAllJobPartsDone()
+	SetQuietMode(mode QuietMode) //TODO add this to all the commands
```
nakulkar-msft, line 77: All the commands can inherit from root cmd.

PR #1762, opened 2022-05-05
[AzCopyV10][Bug] Uploading from top directory on linux fails when sub folder has the same name

siminsavani-msft:
Line 19 in cmd/copyEnumeratorHelper.go causes a bug in linux in which the root folder name is removed from path.

i.e. e.SourceRoot.Value = "/event"  and transfer.Source = "/event_stuff/foo.txt" --> incorrectly changes transfer.Source to "_stuff/foo.txt"

The transfer.Source is already a relative path and this line of code can be removed.

--- 
Refactoring AzCopy copy to use scheduleCopyTransfer instead of addTransfer.

Steps of scheduleCopyTransfer:

1. Gets source and destination relative path from stored object (transfer)
2. Sets up copy transfer and figures out if stored object should be sent to STE
3. If destination is none, then blobtier, metadata, and blob tags are set accordingly
4. If the stored object should not be sent to STE, it skips this object
5. If transfer has dry run mode on, it logs info and skips the object
6. If the length of transfers is same as number of transfers per part, then reset transfers buffer
7. Append transfer to list after checking and dispatching a part
8. If transfer type is file, increment file count or else increment folder count

Steps of addTransfer:

1. Trims prefix for source and destination url (to get relative paths for both)
2. Dispatch transfers once number of files reaches the number of files per dispatch job part (10000) and reset transfers buffer.
3. Append transfer to transfer list after a part has been checked and dispatched
4. If transfer type is file, increment file count or else increment folder count

### DIFF
--- a/cmd/copyEnumeratorHelper.go
+++ b/cmd/copyEnumeratorHelper.go
@@ -15,8 +15,10 @@ var EnumerationParallelStatFiles = false
 
 // addTransfer accepts a new transfer, if the threshold is reached, dispatch a job part order.
 func addTransfer(e *common.CopyJobPartOrderRequest, transfer common.CopyTransfer, cca *CookedCopyCmdArgs) error {
-	// Remove the source and destination roots from the path to save space in the plan files
-	transfer.Source = strings.TrimPrefix(transfer.Source, e.SourceRoot.Value)
+	// Remove the destination roots from the path to save space in the plan files
+
+	// TODO: Remove this code because transfer.Source will already have relative path
+	// transfer.Source = strings.TrimPrefix(transfer.Source, e.SourceRoot.Value)
 	transfer.Destination = strings.TrimPrefix(transfer.Destination, e.DestinationRoot.Value)
 
 	// dispatch the transfers once the number reaches NumOfFilesPerDispatchJobPart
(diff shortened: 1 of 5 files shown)


cmd/copyEnumeratorHelper.go
```diff
@@ -13,10 +13,13 @@ import (
...
 
+// TODO: potentially remove addTransfer and replace this with scheduleCopyTransfer
 // addTransfer accepts a new transfer, if the threshold is reached, dispatch a job part order.
 func addTransfer(e *common.CopyJobPartOrderRequest, transfer common.CopyTransfer, cca *CookedCopyCmdArgs) error {
-	// Remove the source and destination roots from the path to save space in the plan files
-	transfer.Source = strings.TrimPrefix(transfer.Source, e.SourceRoot.Value)
+	// Remove the destination roots from the path to save space in the plan files
+
+	// TODO: Remove this code because transfer.Source will already have relative path
+	// transfer.Source = strings.TrimPrefix(transfer.Source, e.SourceRoot.Value)
```
adreed-msft, line 22: is transfer.Destination not a similar case? Does this bug end up existing on the destination?

adreed-msft (approved):

cmd/copyEnumeratorHelper.go
siminsavani-msft: Good call out! I removed the transfer.Destination (in the new draft PR) as well because transfer.Destination will always have relative path.

PR #1772, opened 2022-05-10
output-level-tiverma

tiverma-msft:
The quiet mode.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -853,8 +853,12 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 
 	cooked.dryrunMode = raw.dryrun
 
-	if cooked.ForceWrite == common.EOverwriteOption.Prompt() && (azcopyOutputVerbosity == common.EOutputVerbosity.Quiet() || azcopyOutputVerbosity == common.EOutputVerbosity.Essential()) { //TODO add to other commands
-		err = fmt.Errorf("cannot set output level '%s' with overwrite option '%s'", azcopyOutputVerbosity.String(), cooked.ForceWrite.String())
+	if azcopyOutputVerbosity == common.EOutputVerbosity.Quiet() || azcopyOutputVerbosity == common.EOutputVerbosity.Essential() {
+		if cooked.ForceWrite == common.EOverwriteOption.Prompt() {
+			err = fmt.Errorf("cannot set output level '%s' with overwrite option '%s'", azcopyOutputVerbosity.String(), cooked.ForceWrite.String())
+		} else if cooked.dryrunMode {
+			err = fmt.Errorf("cannot set output level '%s' with dry-run mode", azcopyOutputVerbosity.String())
+		}
 	}
 	if err != nil {
 		return cooked, err
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -853,6 +853,13 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 
 	cooked.dryrunMode = raw.dryrun
 
+	if cooked.ForceWrite == common.EOverwriteOption.Prompt() && (azcopyOutputVerbosity == common.EOutputVerbosity.Quiet() || azcopyOutputVerbosity == common.EOutputVerbosity.Essential()) { //TODO add to other commands
+		err = fmt.Errorf("cannot set output level '%s' with overwrite option '%s'", azcopyOutputVerbosity.String(), cooked.ForceWrite.String())
+	}
+	if err != nil {
+		return cooked, err
+	}
+
 	return cooked, nil
 }
 
@@ -1811,6 +1818,7 @@ func init() {
 			if err != nil {
 				glcm.Error("failed to parse user input due to error: " + err.Error())
 			}
+			glcm.SetOutputVerbosity(azcopyOutputVerbosity) // TODO add this to all commands
 
 			glcm.Info("Scanning...")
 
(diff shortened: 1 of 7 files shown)


cmd/root.go
```diff
@@ -185,6 +192,7 @@ func init() {
 
 	rootCmd.PersistentFlags().Float64Var(&cmdLineCapMegaBitsPerSecond, "cap-mbps", 0, "Caps the transfer rate, in megabits per second. Moment-by-moment throughput might vary slightly from the cap. If this option is set to zero, or it is omitted, the throughput isn't capped.")
 	rootCmd.PersistentFlags().StringVar(&outputFormatRaw, "output-type", "text", "Format of the command's output. The choices include: text, json. The default value is 'text'.")
+	rootCmd.PersistentFlags().StringVar(&outputVerbosityRaw, "output-level", "default", "Define the output verbosity, available levels: ESSENTIAL(Job ID, Log File Location, Job Summary), and QUIET(no output).")
```
nakulkar-msft, line 195: Output levels should be in regular case. Check the values for output type.

common/lifecyleMgr.go
```diff
@@ -276,6 +278,12 @@ func (lcm *lifecycleMgr) Info(msg string) {
 }
 
 func (lcm *lifecycleMgr) Prompt(message string, details PromptDetails) ResponseOption {
+
+	if shouldQuietMessage(outputMessage{msgType: eOutputMessageType.Prompt()}, lcm.OutputVerbosityType) {
```
nakulkar-msft, line 282: Why is this required after you've made sure conflicting options are not allowed?

common/lifecyleMgr.go
```diff
@@ -400,15 +408,19 @@ func (lcm *lifecycleMgr) processOutputMessage() {
...
 		msgToPrint := <-lcm.msgQueue
 
-		switch lcm.outputFormat {
-		case EOutputFormat.Json():
-			lcm.processJSONOutput(msgToPrint)
-		case EOutputFormat.Text():
-			lcm.processTextOutput(msgToPrint)
-		case EOutputFormat.None():
+		if shouldQuietMessage(msgToPrint, lcm.OutputVerbosityType) {
 			lcm.processNoneOutput(msgToPrint)
```
nakulkar-msft, line 412: You should discard the message here and continue rather than passing message to processNoneOutput

nakulkar-msft (changes requested):

cmd/root.go
tiverma-msft: okay, changed

common/lifecyleMgr.go
tiverma-msft: Conflicting options are not allowed = the user can't pass in quiet or essential mode with overwrite=prompt.
What I'm doing here is checking if the current output-level is quiet or essential. If so, I'm passing yes as the answer to all prompts. This will not be the case when output-level=default.
This is logically the same as "if output-level is not 'default', return yes as the answer to the prompt"

common/lifecyleMgr.go
tiverma-msft: No actually that would lead to our code never ending. When the code ends, it sends a message with eOutputMessageType.Error() or eOutputMessageType.EndOfJob(), and that's how we know when to call 
`lcm.closeFunc();
os.Exit(int(msgToOutput.exitCode))`

If we just discard the message, that would never be triggered, and our code would hang

tiverma-msft: For all other eOutputMessageType (s) I could very well have discarded the message, but I have kept it for the times when it is exit or error

tiverma-msft: `func (lcm *lifecycleMgr) processNoneOutput(msgToOutput outputMessage)`
This function will automatically discard the rest of the types

cmd/copy.go
```diff
@@ -853,6 +853,13 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 
 	cooked.dryrunMode = raw.dryrun
 
+	if cooked.ForceWrite == common.EOverwriteOption.Prompt() && (azcopyOutputVerbosity == common.EOutputVerbosity.Quiet() || azcopyOutputVerbosity == common.EOutputVerbosity.Essential()) { //TODO add to other commands
```
tiverma-msft, line 856: don't allow with dry run either

common/lifecyleMgr.go
tiverma-msft: closing this now

common/lifecyleMgr.go
tiverma-msft: closing this

cmd/copy.go
tiverma-msft: done #

siminsavani-msft (approved):
I would also suggest writing tests to check the output. Take a look at how dry-run is tested to get a foundation for testing output level. Otherwise, LGTM!

common/lifecyleMgr.go
nakulkar-msft: Unresolving. This is unreachable code.

tiverma-msft:
https://github.com/Azure/azure-storage-azcopy/pull/1754/files/0c3337f7104826194dae19c0dd3cba7267b336cf#r864458485
Previous PR

nakulkar-msft (approved):

common/lifecyleMgr.go
tiverma-msft: makes sense, changing this.

mohsha-msft (approved):

PR #1780, opened 2022-05-12
Linux/POSIX properties persistence

adreed-msft:
This PR implements a new flag, --preserve-posix-properties that allows a user to persist the results of a [statx(2)](https://man7.org/linux/man-pages/man2/statx.2.html) (or [stat(2)](https://man7.org/linux/man-pages/man2/lstat.2.html) if statx is unavailable) syscall on upload.

Download is not implemented in this PR. S2S is handled by metadata, and in S2S --preserve-posix-properties acts equivalently to --include-directory-stubs.


shalinijoshi19:
Hey @adreed-msft  ! Thanks for putting this up - question - does this also include the metadata side-car'ing to Az Blobs in particular?

cmd/copy.go
```diff
@@ -638,6 +640,22 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 		cooked.preserveSMBInfo = false
 	}
 
+	if raw.preservePOSIXProperties == "" && areBothLocationsPOSIXAware(cooked.FromTo) {
+		if cooked.FromTo.IsUpload() {
+			raw.preservePOSIXProperties = common.EPosixPropertiesOption.FullFidelity().String()
+		} else if cooked.FromTo.IsDownload() {
+			raw.preservePOSIXProperties = common.EPosixPropertiesOption.SpecialFilesAndProperties().String() // todo: should we default to FullFidelity? It may not always work out well for the customer.
+		} // ignore BlobBlob, these properties persist via metadata & it doesn't make sense to explicitly ignore these bits of metadata.
```
shalinijoshi19, line 648: What is "blobblob" ?

cmd/copy.go
```diff
@@ -638,6 +640,22 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
...
+		// todo: should we consider appending it to a user's custom metadata if it's specified for BlobBlob?
```
shalinijoshi19, line 649: If this scenario is around "side-car'ing" POSIX metadata for uploads to Azure (flat/block) blobs, then yes that was the expectation that we'd leverage the x-ms-meta name:value header here -- One example workflow for eg: 

image

cmd/copy.go
```diff
@@ -907,6 +925,11 @@ func areBothLocationsSMBAware(fromTo common.FromTo) bool {
 	}
 }
 
+func areBothLocationsPOSIXAware(fromTo common.FromTo) bool {
+	// POSIX properties are stored in blob metadata-- They don't need a special persistence strategy for BlobBlob. That said, we silently ignore the BlobBlob case, because customers are assumed to be blind to the inner workings of AzCopy.
```
shalinijoshi19, line 929: Could you elaborate? I am not sure i understand.. What is the diff between localBlob and "BlobBlob" (btw can we rename this a bit to be more clear)?

cmd/copy.go
adreed-msft: BlobBlob is a service-to-service transfer between two Azure Blob Storage endpoints.

cmd/copy.go
adreed-msft: AzCopy uses an enum called "FromTo"-- Effectively, it specifies what direction we're moving files in. The first is the source, the second is the destination. So, LocalBlob would be a upload to Azure Blob Storage, BlobBlob would be a transfer within Azure Blob Storage, and BlobLocal would be a download from Azure Blob Storage.

cmd/copy.go
adreed-msft: Directories, empty or filled, are persisted as an empty blob with the hdi_isfolder metadata.

ste/sourceInfoProvider-Local_linux.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	versions := make([]int64, 3)
+	for k, v := range splits {
+		versions[k], _ = strconv.ParseInt(v, 10, 64)
+	}
+
+	var resp common.UnixStatAdapter
+
+	if versions[0] > 4 || (versions[0] == 4 && versions[1] > 11) { // Can we statx?
```
linuxsmiths, line 41: this isn't a safe way to find statx support especially because many distros backport patches and don't update the version (rather they update the subversion).
Safer/easier way is to make the unix.Statx() call and if it returns ENOSYS error fallback to unix.Stat().

ste/sourceInfoProvider-Local_linux.go
```diff
@@ -0,0 +1,189 @@
...
+		var stat unix.Statx_t
+		// dirfd is a null pointer, because we should only ever be passing relative paths here, and directories will be passed via transferInfo.Source.
+		// AT_SYMLINK_NOFOLLOW is not used, because we automagically resolve symlinks. TODO: Add option to not follow symlinks, and use AT_SYMLINK_NOFOLLOW when resolving is disabled.
```
linuxsmiths, line 44: I think for our usecase (backing up files) AT_SYMLINK_NOFOLLOW should be default. This is because when looking at a symlink file property we will be interested in the properties of the symlink and not of the target. If we need target properties we can always evaluate the symlink and then call statx() on the target.

ste/sourceInfoProvider-Local_linux.go
```diff
@@ -0,0 +1,189 @@
...
+		err = unix.Statx(0, f.transferInfo.Source,
+			unix.AT_STATX_FORCE_SYNC, // We want to sync attributes to ensure correctness.
```
linuxsmiths, line 46: we should not set force sync, else it'll cause lot of getattr calls on the wire for NFS/SMB mounted shares f.e.
Note that in NFS/SMB we have the READDIRPLUS/QUERYDIR calls which get file attributes as part of the dir enumeration response, so that a stat call made within the attribute cache expiry time (which is the common case) will be served from the cache. This hugely saves  on the calls made to the server.

Local filesystems mostly ignore this setting as it doesn't make sense, their attributes are strongly consistent at all times.

common/fe-ste-models.go
```diff
@@ -50,7 +50,8 @@ const (
 
 	//  this is the perm that AzCopy has used throughout its preview.  So, while we considered relaxing it to 0666
 	//  we decided that the best option was to leave it as is, and only relax it if user feedback so requires.
-	DEFAULT_FILE_PERM = 0644
+	DEFAULT_FILE_PERM     = 0644  // the os package will handle base-10 for us.
+	HEX_DEFAULT_FILE_PERM = 0x644 // when making unix syscalls directly, hex form is expected.
```
linuxsmiths, line 54: we should not need to define two and 0x644 is not same as octal 0644.
What we need is just the octal rep i.e. DEFAULT_FILE_PERM and use it in all places.
What the syscall (or any other API) would take is an integer and these are just convenience literal representation in octcal and hex (and decimal).

common/fe-ste-models.go
```diff
@@ -1496,4 +1531,30 @@ func GetClientProvidedKey(options CpkOptions) azblob.ClientProvidedKeyOptions {
...
+
+func (m *LCMMsgType) String() string {
+	return enum.StringInt(m, reflect.TypeOf(m))
+}
+
+type LCMMsg struct {
+	TimeStamp time.Time `json:"TimeStamp"`
+	MsgType   string    `json:"MessageType"`
+	Value     string    `json:"Value"`
+}
```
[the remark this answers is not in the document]
adreed-msft: No-- Weird that it showed up.

ste/sourceInfoProvider-Local_linux.go
linuxsmiths: Any reason why you are not using AT_SYMLINK_NOFOLLOW?
As I said, GetUNIXProperties is a low level call for getting file attributes. It should truthfully get the attributes of the requested file. Whether to follow a symlink or not should be a higher level decision (or if u want pass an option to GetUNIXProperties).
If we don't use AT_SYMLINK_NOFOLLOW  there's no way for the caller to get the symlink properties for cases where it decides to not expand symlinks and copy them as is.

linuxsmiths (approved):
LG mostly. Few comments.

cmd/copyEnumeratorInit.go
```diff
@@ -696,7 +692,7 @@ func newFolderPropertyOption(fromTo common.FromTo, recursive bool, stripTopDir b
 		}
 	}
 
-	bothFolderAware := (fromTo.AreBothFolderAware() || isDfsDfs) && !isDstNull // Copying folders to dev null doesn't make sense.
+	bothFolderAware := (fromTo.AreBothFolderAware() || isDfsDfs || preservePosixProperties) && !isDstNull
```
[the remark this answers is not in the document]
adreed-msft: Blob itself is not inherently folder aware. Thus, it doesn't handle folder uploads by default. If POSIX property persistence is present, we should be carrying over properties on folders, as well.

cmd/zc_enumerator.go
```diff
@@ -305,10 +305,7 @@ type enumerationCounterFunc func(entityType common.EntityType)
 // followSymlinks is only required for local resources (defaults to false)
 // errorOnDirWOutRecursive is used by copy.
 
-func InitResourceTraverser(resource common.ResourceString, location common.Location, ctx *context.Context,
-	credential *common.CredentialInfo, followSymlinks *bool, listOfFilesChannel chan string, recursive, getProperties,
-	includeDirectoryStubs bool, permanentDeleteOption common.PermanentDeleteOption, incrementEnumerationCounter enumerationCounterFunc, listOfVersionIds chan string,
-	s2sPreserveBlobTags bool, logLevel pipeline.LogLevel, cpkOptions common.CpkOptions) (ResourceTraverser, error) {
+func InitResourceTraverser(resource common.ResourceString, location common.Location, ctx *context.Context, credential *common.CredentialInfo, followSymlinks *bool, listOfFilesChannel chan string, recursive, getProperties, includeDirectoryStubs bool, permanentDeleteOption common.PermanentDeleteOption, incrementEnumerationCounter enumerationCounterFunc, listOfVersionIds chan string, s2sPreserveBlobTags bool, logLevel pipeline.LogLevel, cpkOptions common.CpkOptions) (ResourceTraverser, error) {
```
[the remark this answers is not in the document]
adreed-msft: I agree, some effort needs to be put into turning it into an options bag down the line.

cmd/zc_traverser_blob.go
```diff
@@ -261,7 +261,7 @@ func (t *blobTraverser) parallelList(containerURL azblob.ContainerURL, container
 								preprocessor,
 								getObjectNameOnly(strings.TrimSuffix(virtualDir.Name, common.AZCOPY_PATH_SEPARATOR_STRING)),
 								folderRelativePath,
-								common.EEntityType.File(), // folder stubs are treated like files in in the serial lister as well
+								common.EEntityType.Folder(), // folder stubs are treated like files in in the serial lister as well
```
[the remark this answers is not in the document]
adreed-msft: Correct. I need to remove the comment.

common/fe-ste-models.go
```diff
@@ -110,6 +110,40 @@ type PartNumber uint32
...
+		*p = p.FullFidelity()
+		return nil
+	}
+
+	val, err := enum.ParseInt(reflect.TypeOf(p), s, true, true)
+	if err == nil {
+		*p = val.(PosixPropertiesOption)
+	}
+	return err
+}
```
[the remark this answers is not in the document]
adreed-msft: Nope, it can be removed.

ste/sourceInfoProvider-Local_linux.go
adreed-msft: I think once we shift to uploading the symlink path, we should opt for AT_SYMLINK_NOFOLLOW. For now, we still upload the file on the other end of the symlink, so it makes sense to upload the properties of the file on the other end of the symlink.

common/unixStatAdapter.go
```diff
@@ -0,0 +1,381 @@
...
+	MTime() time.Time
+	CTime() time.Time
+}
+
+type UnixStatContainer struct { // Created for downloads
+	statx bool // Does the call contain extended properties (attributes, birthTime)?
+
+	mask       uint32
+	attributes uint64
+	numLinks   uint64
```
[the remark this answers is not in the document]
adreed-msft: I disagree. This struct (and interface) is OS-agnostic, but POSIX specific. If we decide to support some other POSIX-based OS like MacOS or FreeBSD in the future for the POSIX property persistence feature, this could cause problems if we adhere to Linux-specific data typing.

ste/sourceInfoProvider-Local_linux.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (f localFileSourceInfoProvider) GetUNIXProperties() (common.UnixStatAdapter, error) {
+	{ // attempt to call statx, if ENOSYS is returned, statx is unavailable
+		var stat unix.Statx_t
+		// dirfd is a null pointer, because we should only ever be passing relative paths here, and directories will be passed via transferInfo.Source.
+		// AT_SYMLINK_NOFOLLOW is not used, because we automagically resolve symlinks. TODO: Add option to not follow symlinks, and use AT_SYMLINK_NOFOLLOW when resolving is disabled.
+		err := unix.Statx(0, f.transferInfo.Source,
+			unix.AT_STATX_SYNC_AS_STAT, // We want to sync attributes to ensure correctness.
+			unix.STATX_BTIME,           // Let's pull only the special statx properties, and yank the rest from a standard stat_t call.
```
linuxsmiths, line 22: this should be STATX_ALL, now that we are using only statx().
also the " // We want to sync attributes to ensure correctness." comment is not correct, as AT_STATX_SYNC_AS_STAT will usually not sync.

shalinijoshi19:
> The PR remains in draft state, as download is partially broken in some aspects (attributes fail to even be written, etc.), there is no automated testing of the feature yet, and the build fails on other OSes due to certain function signatures..
> 
> This PR implements a new flag, --preserve-posix-properties that allows a user to persist the results of a [statx(2)](https://man7.org/linux/man-pages/man2/statx.2.html) (or [stat(2)](https://man7.org/linux/man-pages/man2/lstat.2.html) if statx is unavailable) syscall.
> 
> To work around cases where the representative device ID may not be related on the destination system, --preserve-posix-properties has two modes: FullFidelity, and SpecialFilesAndProperties, where FullFidelity persists even files representing devices, and SpecialFilesAndProperties skips over these files.

Do update the description wiht this version of the PR pls! thanks

cmd/copy.go
```diff
@@ -638,6 +640,11 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 		cooked.preserveSMBInfo = false
 	}
 
+	cooked.preservePOSIXProperties = raw.preservePOSIXProperties
+	if cooked.preservePOSIXProperties && !areBothLocationsPOSIXAware(cooked.FromTo) {
```
shalinijoshi19, line 644: Nit: interesting that the if condition in go doesnt seem to be enclosed in parenthesis - is this the only syntax in golang for conditions or do they also support paranthesis?

cmd/copy.go
```diff
@@ -638,6 +640,11 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
...
+		return cooked, fmt.Errorf("in order to use --preserve-posix-properties, both the source and destination must be POSIX-aware (Linux->Blob, Blob->Linux, Blob->Blob)")
```
shalinijoshi19, line 645: Nit nit: could just be me but trying to understand what you are trying to convey in the parantheses -- is it the source,destination pairing? Could we a bit more explicit say : "Supported POSIX-aware pairings are .." or even a ("vis" prefix in there)?

cmd/copy.go
```diff
@@ -1869,6 +1886,7 @@ func init() {
...
 	cpCmd.PersistentFlags().BoolVar(&raw.preserveOwner, common.PreserveOwnerFlagName, common.PreserveOwnerDefault, "Only has an effect in downloads, and only when --preserve-smb-permissions is used. If true (the default), the file Owner and Group are preserved in downloads. If set to false, --preserve-smb-permissions will still preserve ACLs but Owner and Group will be based on the user running AzCopy")
 	cpCmd.PersistentFlags().BoolVar(&raw.preserveSMBInfo, "preserve-smb-info", true, "For SMB-aware locations, flag will be set to true by default. Preserves SMB property info (last write time, creation time, attribute bits) between SMB-aware resources (Windows and Azure Files). Only the attribute bits supported by Azure Files will be transferred; any others will be ignored. This flag applies to both files and folders, unless a file-only filter is specified (e.g. include-pattern). The info transferred for folders is the same as that for files, except for Last Write Time which is never preserved for folders.")
+	cpCmd.PersistentFlags().BoolVar(&raw.preservePOSIXProperties, "preserve-posix-properties", false, "On Linux, Preserves property info gleamed from stat and statx.")
```
shalinijoshi19, line 1889: Typo + suggestion around rewording : 'Preserves' property information is gleaned from stat and statx"

cmd/copyEnumeratorInit.go
```diff
@@ -80,10 +81,7 @@ func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 	jobPartOrder.S2SInvalidMetadataHandleOption = cca.s2sInvalidMetadataHandleOption
 	jobPartOrder.S2SPreserveBlobTags = cca.S2sPreserveBlobTags
 
-	traverser, err = InitResourceTraverser(cca.Source, cca.FromTo.From(), &ctx, &srcCredInfo,
-		&cca.FollowSymlinks, cca.ListOfFilesChannel, cca.Recursive, getRemoteProperties,
-		cca.IncludeDirectoryStubs, cca.permanentDeleteOption, func(common.EntityType) {}, cca.ListOfVersionIDs,
-		cca.S2sPreserveBlobTags, cca.LogVerbosity.ToPipelineLogLevel(), cca.CpkOptions)
+	traverser, err = InitResourceTraverser(cca.Source, cca.FromTo.From(), &ctx, &srcCredInfo, &cca.FollowSymlinks, cca.ListOfFilesChannel, cca.Recursive, getRemoteProperties, cca.IncludeDirectoryStubs, cca.permanentDeleteOption, func(common.EntityType) {}, cca.ListOfVersionIDs, cca.S2sPreserveBlobTags, cca.LogVerbosity.ToPipelineLogLevel(), cca.CpkOptions)
```
shalinijoshi19, line 84: In general and this def is not specific to your code but these are a whole lotta parameters :) consider defining a specific parameters class/structure that you can pass in to this initialization function

cmd/sync.go
```diff
@@ -739,6 +746,7 @@ func init() {
 	// smb info/permissions can be persisted in the scenario of File -> File
 	syncCmd.PersistentFlags().BoolVar(&raw.preserveSMBPermissions, "preserve-smb-permissions", false, "False by default. Preserves SMB ACLs between aware resources (Azure Files). This flag applies to both files and folders, unless a file-only filter is specified (e.g. include-pattern).")
 	syncCmd.PersistentFlags().BoolVar(&raw.preserveSMBInfo, "preserve-smb-info", true, "For SMB-aware locations, flag will be set to true by default. Preserves SMB property info (last write time, creation time, attribute bits) between SMB-aware resources (Azure Files). This flag applies to both files and folders, unless a file-only filter is specified (e.g. include-pattern). The info transferred for folders is the same as that for files, except for Last Write Time which is not preserved for folders. ")
+	syncCmd.PersistentFlags().BoolVar(&raw.preservePOSIXProperties, "preserve-posix-properties", false, "On Linux, the flag will be set to true by default. Preserves property info gleamed from stat and statx.")
```
shalinijoshi19, line 749: Nit: very similar string here including similar typo (should be _gleaned_ instead of _gleamed_ ) - consider creating a const string variable you can reference both places

cmd/syncEnumerator.go
```diff
@@ -58,12 +58,11 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 	// TODO: enable symlink support in a future release after evaluating the implications
```
shalinijoshi19, line 58: Nit: Do we need to update this comment in any way now?

shalinijoshi19 (approved):

cmd/copy.go
adreed-msft: They do support parenthesis-- But it's idomatic to not use it around the main conditional.

common/unixStatAdapter.go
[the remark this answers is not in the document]
linuxsmiths: I meant those sizes are POSIX specific not Linux specific ("stat()" is a POSIX call and these sizes are not specific to statx()). So any POSIX OS should have the same sizes.

PR #1782, opened 2022-05-16
Avoid creating .azcopy under HOME if plan/log location is specified elsewhere

tiverma-msft:


tiverma-msft:
There are 2 variables here. Log file loc and plan file loc. If one of the locations is specified by an env variable and other is not, we will still create the .azcopy folder to accommodate the file for which user has not provided a location.

Only in the case of both env variables being provided will we skip creating the .azcopy folder.

One of my concerns... something that I'll have to see is if this folder is used for something other than plan and log files. The location of this folder is being shared by `cmd.Execute(azcopyAppPathFolder, azcopyLogPathFolder, azcopyJobPlanFolder, 0, jobID)` and is then used for `common.NewUserOAuthTokenManagerInstance(...)`
What is that for? (Adding this to the conversation as a personal note to find out.

tiverma-msft:
In such a case (when a token json file is being created), the folder .azcopy will be created, and the token will be stored there. The code -`DPAPIFilePath: azcopyAppPathFolder` in `credentialUtil.go` decides where the json token will go.
Leaving this as is. i.e as long as the user is not creating tokens, this folder is not created.

tiverma-msft:
Changes suggested: the OAuth token should be where the plan files are located. Make changes accordingly

main.go
```diff
@@ -42,12 +42,21 @@ func main() {
 
 	rand.Seed(time.Now().UnixNano()) // make sure our random numbers actually are random (but remember, use crypto/rand for anything where strong/reliable randomness is required
 
+	azcopyLogPathFolder := common.GetLifecycleMgr().GetEnvironmentVariable(common.EEnvironmentVariable.LogLocation())     // user specified location for log files
+	azcopyJobPlanFolder := common.GetLifecycleMgr().GetEnvironmentVariable(common.EEnvironmentVariable.JobPlanLocation()) // user specified location for plan files
+
 	// note: azcopyAppPathFolder is the default location for all AzCopy data (logs, job plans, oauth token on Windows)
 	// but both logs and job plans can be put elsewhere as they can become very large
 	azcopyAppPathFolder := GetAzCopyAppPath()
```
zezha-msft, line 50: azcopyAppPathFolder is obsolete imo, the oauth token should just go with azcopyJobPlanFolder

zezha-msft: perhaps just get rid of it completely, and perform a bit of refactoring to clean it up 😉

tiverma-msft: yes I think so too. If we're going to put OAuth to JobPlanFolder, there isn't a lot of use for the AppPathFolder. I'll make the necessary changes 😎

tiverma-msft: this variable will still be present, but only be local to this function and be used to make azcopyLogPathFolder and azcopyJobPlanFolder if they're empty.

tiverma-msft: changes done. Closing this

tiverma-msft: link to commit - https://github.com/Azure/azure-storage-azcopy/pull/1782/commits/7f67778481dba62323c2636b5ca0e767edf2cf73

nakulkar-msft (approved):

mohsha-msft (approved):

PR #1785, opened 2022-05-17
Introduce xdm-relevant changes to and fix frontend

Strikerzee:
Refer to:
Export azcopyAppPathFolder variable.
https://github.com/Azure/azure-storage-azcopy/commit/7e86c2b4fb20c075b25aba8c8967e680c58d64af

Fix calculation for large file block size.
- This fix is in respect to support 190TB files.
  As of now due to wrong calculation azcopy fails after 24T.
https://github.com/Azure/azure-storage-azcopy/commit/36311199db508cb92b424206ec7b1c7996ca66e8

Handle panic error when plan file is empty.
https://github.com/Azure/azure-storage-azcopy/commit/c7722dec58e912d38a5d140bcedcdfd0c3d5ceaf

A new PR has been created in order to make the changes more modular, and easily reviewable.

### DIFF
--- a/jobsAdmin/JobsAdmin.go
+++ b/jobsAdmin/JobsAdmin.go
@@ -323,12 +323,12 @@ func (ja *jobsAdmin) SuccessfulBytesInActiveFiles() uint64 {
 
 func (ja *jobsAdmin) ResurrectJob(jobId common.JobID, sourceSAS string, destinationSAS string) bool {
 	// Search the existing plan files for the PartPlans for the given jobId
-	// only the files which have JobId has prefix and DataSchemaVersion as Suffix
+	// only the files which are not empty and have JobId has prefix and DataSchemaVersion as Suffix
 	// are include in the result
 	files := func(prefix, ext string) []os.FileInfo {
 		var files []os.FileInfo
 		filepath.Walk(ja.planDir, func(path string, fileInfo os.FileInfo, _ error) error {
-			if !fileInfo.IsDir() && strings.HasPrefix(fileInfo.Name(), prefix) && strings.HasSuffix(fileInfo.Name(), ext) {
+			if !fileInfo.IsDir() && fileInfo.Size() != 0 && strings.HasPrefix(fileInfo.Name(), prefix) && strings.HasSuffix(fileInfo.Name(), ext) {
 				files = append(files, fileInfo)
 			}
 			return nil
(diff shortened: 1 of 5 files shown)


nakulkar-msft (approved):

jobsAdmin/JobsAdmin.go
```diff
@@ -323,12 +323,12 @@ func (ja *jobsAdmin) SuccessfulBytesInActiveFiles() uint64 {
...
 func (ja *jobsAdmin) ResurrectJob(jobId common.JobID, sourceSAS string, destinationSAS string) bool {
 	// Search the existing plan files for the PartPlans for the given jobId
-	// only the files which have JobId has prefix and DataSchemaVersion as Suffix
+	// only the files which are not empty and have JobId has prefix and DataSchemaVersion as Suffix
 	// are include in the result
 	files := func(prefix, ext string) []os.FileInfo {
 		var files []os.FileInfo
 		filepath.Walk(ja.planDir, func(path string, fileInfo os.FileInfo, _ error) error {
-			if !fileInfo.IsDir() && strings.HasPrefix(fileInfo.Name(), prefix) && strings.HasSuffix(fileInfo.Name(), ext) {
+			if !fileInfo.IsDir() && fileInfo.Size() != 0 && strings.HasPrefix(fileInfo.Name(), prefix) && strings.HasSuffix(fileInfo.Name(), ext) {
```
mohsha-msft, line 331: Let's add a check that fileInfo is not null. Otherwise, it'll panic.

ste/mgr-JobPartTransferMgr.go
```diff
@@ -345,7 +345,11 @@ func (jptm *jobPartTransferMgr) Info() TransferInfo {
 				 * we can have 4 blocks in core, waiting for a disk or n/w operation. Any higher block size would *sort of*
 				 * serialize n/w and disk operations, and is better avoided.
 				 */
-				blockSize = sourceSize / common.MaxNumberOfBlocksPerBlob
+				if (sourceSize % common.MaxNumberOfBlocksPerBlob == 0) {
```
mohsha-msft, line 348: Nit: [Ceil function](https://golangbyexample.com/ceil-number-golang/) perhaps?

mohsha-msft (approved):
Please check the comments. Looks good otherwise.

ste/mgr-JobPartTransferMgr.go
Strikerzee: I believe this stems from the comment here: https://github.com/Azure/azure-storage-azcopy/pull/1745/commits/36311199db508cb92b424206ec7b1c7996ca66e8#r858641439

If we are to compare what @nitin-deamon mentioned as a reply there versus this line of code, then in my opinion would be preferable to avoid typecasting of int-float & vice-versa.

PR #1797, opened 2022-05-24
Job Manager changes

Strikerzee:
JobMgr Cleanup.

- Close all the go routines once job is done.

Fix for Report Handler routine cleanup.

Added the support to change the log level.

Memory Leak Fix.

- Set the nil to address allocated memory.

- Fixing some of memory issues.

- JobMgr is freed without setting channel to nil.

### DIFF
--- a/common/chunkStatusLogger.go
+++ b/common/chunkStatusLogger.go
@@ -241,6 +241,7 @@ type ChunkStatusLoggerCloser interface {
 	GetCounts(td TransferDirection) []chunkStatusCount
 	GetPrimaryPerfConstraint(td TransferDirection, rc RetryCounter) PerfConstraint
 	FlushLog() // not close, because we had issues with writes coming in after this // TODO: see if that issue still exists
+	CloseLogger()
 }
 
 type RetryCounter interface {
@@ -315,6 +316,18 @@ func (csl *chunkStatusLogger) FlushLog() {
 	}
 }
 
+// CloseLogger close the chunklogger thread.
+func (csl *chunkStatusLogger) CloseLogger() {
+	// Once logger is closed, we log no more chunks.
+	csl.outputEnabled = false
+
+	/*
+	 * No more chunks will ever be written, let the main logger know about this.
+	 * On closing this channel the main logger will exit from its for-range loop.
+	 */
+	close(csl.unsavedEntries)
+}
+
 func (csl *chunkStatusLogger) main(chunkLogPath string) {
 	f, err := os.Create(chunkLogPath)
 	if err != nil {
@@ -332,18 +345,22 @@ func (csl *chunkStatusLogger) main(chunkLogPath string) {
 	defer doFlush()
 
 	alwaysFlushFromNowOn := false
+
+	// We will exit the following for-range loop after CloseLogger() closes the csl.unsavedEntries channel.
 	for x := range csl.unsavedEntries {
 		if x == nil {
 			alwaysFlushFromNowOn = true
 			doFlush()
 			csl.flushDone <- struct{}{}
 			continue // TODO can become break (or be moved to later if we close unsaved entries, once we figure out how we got stuff written to us after CloseLog was called)
+
 		}
 		_, _ = w.WriteString(fmt.Sprintf("%s,%d,%s,%s\n", x.Name, x.OffsetInFile(), x.reason, x.waitStart))
 		if alwaysFlushFromNowOn {
 			// TODO: remove when we figure out how we got stuff written to us after CloseLog was called. For now, this should handle those cases (if they still exist)
 			doFlush()
 		}
+
 	}
 }
 
--- a/common/logger.go
+++ b/common/logger.go
@@ -47,6 +47,7 @@ type ILoggerCloser interface {
 type ILoggerResetable interface {
 	OpenLog()
 	MinimumLogLevel() pipeline.LogLevel
+	ChangeLogLevel(pipeline.LogLevel)
 	ILoggerCloser
 }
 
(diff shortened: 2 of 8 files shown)


common/chunkStatusLogger.go
```diff
@@ -315,6 +316,18 @@ func (csl *chunkStatusLogger) FlushLog() {
 	}
 }
 
+// CloseLogger close the chunklogger thread.
+func (csl *chunkStatusLogger) CloseLogger() {
+	// Once logger is closed, we log no more chunks.
+	csl.outputEnabled = false
```
mohsha-msft, line 322: This is not thread safe. Some logs will be lost. Is it acceptable?

common/logger.go
```diff
@@ -47,6 +48,7 @@ type ILoggerCloser interface {
 type ILoggerResetable interface {
 	OpenLog()
 	MinimumLogLevel() pipeline.LogLevel
+	ChangeLogLevel(pipeline.LogLevel)
```
mohsha-msft, line 51: `ILoggerResetable` inherits `ILoggerCloser` and `ILoggerCloser` inherits `ILogger`.
I see you are declaring `ChangeLogLevel()` in ILogger. So, by default `ChangeLogLevel()` method should be inherited by `ILoggerResetable`. Why do we need to declare it again?

mohsha-msft: The cleaner solution would have been to create a new logger (at job manager level) to record job manager cleanup. This "cleanup logger" would inherit ILogger and could have an additional method `ChangeLogLevel()`. Doing this would not affect the existing code and keep the changes minimal. Is it feasible to do it based on your code changes?

common/singleChunkReader.go
```diff
@@ -227,14 +227,20 @@ func (cr *singleChunkReader) blockingPrefetch(fileReader io.ReaderAt, isRetry bo
 	// Must use "relaxed" RAM limit IFF this is a retry.  Else, we can, in theory, get deadlock with all active goroutines blocked
 	// here doing retries, but no RAM _will_ become available because its
 	// all used by queued chunkfuncs (that can't be processed because all goroutines are active).
-	cr.chunkLogger.LogChunkStatus(cr.chunkId, EWaitReason.RAMToSchedule())
+	if cr.chunkLogger != nil {
+		cr.chunkLogger.LogChunkStatus(cr.chunkId, EWaitReason.RAMToSchedule())
+	}
+
 	err := cr.cacheLimiter.WaitUntilAdd(cr.ctx, cr.length, func() bool { return isRetry })
```
mohsha-msft, line 234: Just Curious: Can cr.chunkLogger be nil? If yes, can cr.cacheLimiter also be nil? If yes, Why not have nil check for this line cr.cacheLimiter?

jobsAdmin/JobsAdmin.go
```diff
@@ -296,6 +301,41 @@ func (ja *jobsAdmin) JobMgrEnsureExists(jobID common.JobID,
 		})
 }
 
+// JobMgrCleanup cleans up the jobMgr identified by the given jobId. It undoes what NewJobMgr() does, basically it does the following:
+// 1. Stop all go routines started to process this job.
+// 2. Release the memory allocated for this JobMgr instance.
+// Note: this is not thread safe and only one goroutine should call this for a job.
+func (ja *jobsAdmin) JobMgrCleanUp(jobId common.JobID) {
+	// First thing get the jobMgr.
```
mohsha-msft, line 309: Nice. ❤️

ste/jobStatusManager.go
```diff
@@ -106,6 +114,10 @@ func (jm *jobMgr) handleStatusUpdateMessage() {
 			// There is no need to keep sending the same items over and over again
 			js.FailedTransfers = []common.TransferDetail{}
 			js.SkippedTransfers = []common.TransferDetail{}
+
+		case <-jstm.done:
+			fmt.Println("Cleanup JobStatusmgr")
```
mohsha-msft, line 119: Since we're already using loggers, may as well use them instead. We avoid using fmt.Println().
AzCopy has quiet-mode which doesn't print anything on the console and fmt.Print statements can create noise on terminal/

ste/jobStatusManager.go
```diff
@@ -41,6 +43,7 @@ type jobStatusManager struct {
 	listReq     chan bool
 	partCreated chan JobPartCreatedMsg
 	xferDone    chan xferDoneMsg
+	done        chan struct{}
```
mohsha-msft, line 46: I see this channel is getting populatd from `CleanupJobStatusMgr` only once. My question is: why do we need it. Can we not print the "Cleanup done" at the end of `DeferredCleanupJobMgr`. I think you are already doing that. 
`jm.Log(pipeline.LogInfo, "DeferredCleanupJobMgr Exit, Closing the log")`
This way we will not need a channel and the thread performing the clean up can behave as a detached/independent goroutine (faster).

common/logger.go
mohsha-msft: @nakulkar-msft @zezha-msft ?

mohsha-msft:
Please add appropriate unit test cases wherever applicable.

ste/jobStatusManager.go
nitin-deamon: done chan to tell handleStatusUpdateMessage goroutine to close.

common/singleChunkReader.go
nitin-deamon: We are setting cr.chunkLogger to nil, once it done the transfer. So that it release the reference of chunkStatusLogger. 
cr.cacheLimiter never set to nil, so no need to check that.

common/chunkStatusLogger.go
nitin-deamon: This is exposed in ChunkStatusLoggerCloser() interface which is known to JobManager. It's being called from JobManager only exported function DeferredCleanupJobMgr() func. Before calling DeferredCleanupJobMgr we remove jobMgr entry from jobsAdmin.

PR #1803, opened 2022-05-26
Allow calling app to cancel enumeration and handle scanning directory errors in traverser

Strikerzee:
Enable calling app to cancel enumeration
- Used appCtx in parallel-walk so that the app can cancel the enumeration as it requires.
- It doesn't effect azcopy working.

Handle scanning directory errors in traverser.
- Fail visibly when scanning fails for files and directories
- Add in more error paths
- Fix processOneDirectory changes and add more comments

### DIFF
--- a/cmd/zc_enumerator.go
+++ b/cmd/zc_enumerator.go
@@ -392,7 +392,7 @@ func InitResourceTraverser(resource common.ResourceString, location common.Locat
 			output = newListTraverser(baseResource, location, nil, nil, recursive, toFollow, getProperties,
 				globChan, includeDirectoryStubs, incrementEnumerationCounter, s2sPreserveBlobTags, logLevel, cpkOptions)
 		} else {
-			output = newLocalTraverser(resource.ValueLocal(), recursive, toFollow, incrementEnumerationCounter)
+			output = newLocalTraverser(ctx, resource.ValueLocal(), recursive, toFollow, incrementEnumerationCounter)
 		}
 	case common.ELocation.Benchmark():
 		ben, err := newBenchmarkTraverser(resource.Value, incrementEnumerationCounter)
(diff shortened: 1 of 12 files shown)


mohsha-msft:
Please rebase with `dev` branch. The PR is missing unit test cases. Please write test cases to test the fuctionalities.

common/parallel/FileSystemCrawler.go
```diff
@@ -132,18 +145,23 @@ func enumerateOneFileSystemDirectory(dir Directory, enqueueDir func(Directory),
...
 			childEntry := FileSystemEntry{
 				fullPath: filepath.Join(dirString, childInfo.Name()),
 				info:     childInfo,
 			}
+
+			if failable, ok := childInfo.(failableFileInfo); ok && failable.Error() != nil {
+				// while Readdir as a whole did not fail, this particular file info did
+				enqueueOutput(childEntry, failable.Error())
+				continue
+			}
```
mohsha-msft, line 164: QQ: Why did we shift this segment from above?

cmd/zc_enumerator.go
```diff
@@ -392,7 +393,11 @@ func InitResourceTraverser(resource common.ResourceString, location common.Locat
 			output = newListTraverser(baseResource, location, nil, nil, recursive, toFollow, getProperties,
 				globChan, includeDirectoryStubs, incrementEnumerationCounter, s2sPreserveBlobTags, logLevel, cpkOptions)
 		} else {
-			output = newLocalTraverser(resource.ValueLocal(), recursive, toFollow, incrementEnumerationCounter)
+			if ctx != nil {
+				output = newLocalTraverser(*ctx, resource.ValueLocal(), recursive, toFollow, incrementEnumerationCounter, errorChannel)
```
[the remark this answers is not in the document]
Strikerzee: If I do not dereference (or get the value of) it, there will be a mismatch in the type. (context and *context)

# github.com/Azure/azure-storage-azcopy/v10/cmd
cmd/zc_enumerator.go:400:31: cannot use ctx (type *context.Context) as type context.Context in argument to newLocalTraverser:
        *context.Context is pointer to interface, not interface

common/parallel/FileSystemCrawler.go
Strikerzee: We need the childEntry as an argument to the enqueueOutput() method. Hence, the segment has been shifted.

mohsha-msft (approved):

nakulkar-msft (approved):

PR #1804, opened 2022-05-27
Moving log-level to root.go

tiverma-msft:


tiverma-msft:
In 10.15, the user provided log-level was only being given to scanning logger. That's why, the scanning logger was conforming to the log-level, but not job logger. After these changes, both should be able to conform to the needs

cmd/root.go
```diff
@@ -169,8 +184,7 @@ func Execute(azsAppPathFolder, logPathFolder string, jobPlanFolder string, maxFi
 	common.AzcopyJobPlanFolder = jobPlanFolder
 	azcopyMaxFileAndSocketHandles = maxFileAndSocketHandles
 	azcopyCurrentJobID = jobID
-	common.AzcopyCurrentJobLogger = common.NewJobLogger(jobID, common.ELogLevel.Debug(), logPathFolder, "")
```
tiverma-msft, line 172: This line was causing the problem mentioned in the issue. The logger was being created with level Debug irrespective of what the user had flagged.

nakulkar-msft (approved):

mohsha-msft (approved):
Fix the conflicts

PR #1813, opened 2022-06-01
Disabled fast failure on 403s

zezha-msft:


zezha-msft:
No longer needed, service fixed the issue.

PR #1814, opened 2022-06-01
Fix race between Updating JobPart and xferDoneMsg

nakulkar-msft:
XDM folks have seen a race between ReportJobPartDone() and handleStatusUpdateMessage(), which would cause handleStatusUpdateMessage() to report incorrect statistics wrt. JobSummary. This PR will fix the race.

### DIFF
--- a/ste/mgr-JobMgr.go
+++ b/ste/mgr-JobMgr.go
@@ -643,6 +643,8 @@ func (jm *jobMgr) reportJobPartDoneHandler() {
 			isCancelling := jobStatus == common.EJobStatus.Cancelling()
 			shouldComplete := allKnownPartsDone && (haveFinalPart || isCancelling)
 			if shouldComplete {
+				// Inform StatusManager that all parts are done.
+				close(jm.jstm.xferDone)
 				// Wait  for all XferDone messages to be processed by statusManager. Front end
 				// depends on JobStatus to determine if we've to quit job. Setting it here without
 				// draining XferDone will make it report incorrect statistics.


ste/jobStatusManager.go
```diff
@@ -27,34 +27,64 @@ import (
...
-	respChan    chan common.ListJobSummaryResponse
-	listReq     chan bool
-	partCreated chan JobPartCreatedMsg
-	xferDone    chan xferDoneMsg
+	js              common.ListJobSummaryResponse
+	respChan        chan common.ListJobSummaryResponse
+	listReq         chan struct{}
+	partCreated     chan JobPartCreatedMsg
+	xferDone        chan xferDoneMsg
+	xferDoneDrained chan struct{}
```
Strikerzee, line 44: It would greatly help if we could add a comment explaining that these 2 new variables have been added in order to fix the race condition. Otherwise, it does not look very obvious to me

Strikerzee (approved):
LGTM overall. Please rebase with the dev (containing the brand-new app ctx changes) before merge.

ste/jobStatusManager.go
```diff
@@ -67,18 +97,32 @@ func (jm *jobMgr) handleStatusUpdateMessage() {
 	js.JobID = jm.jobID
 	js.CompleteJobOrdered = false
 	js.ErrorMsg = ""
+	allXferDoneHandled := false
 
 	for {
 		select {
-		case msg := <-jstm.partCreated:
+		case msg, ok := <-jstm.partCreated:
```
nitin-deamon, line 104: Once channel closed, this for loop keep running as <-jstm.partCreated always true. That's unnecessary cpu cycles. wasted.  It will delay the processing of other messages.

siminsavani-msft (approved):

PR #1818, opened 2022-06-01
Fix some oopsies in the unix properties PR

adreed-msft:
1. We didn't clone the metadata map when we were going to modify it for individual blobs, meaning we'd encounter data races, and incorrect properties would be written
2. The incorrect seconds value was used for CTime, leading to an invalid value for UnixNano.

### DIFF
--- a/ste/sender-blockBlobFromLocal.go
+++ b/ste/sender-blockBlobFromLocal.go
@@ -47,7 +47,11 @@ func newBlockBlobUploader(jptm IJobPartTransferMgr, destination string, p pipeli
 
 func (s *blockBlobUploader) Prologue(ps common.PrologueState) (destinationModified bool) {
 	if s.jptm.Info().PreservePOSIXProperties {
+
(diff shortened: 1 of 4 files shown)


ste/sender-blockBlobFromLocal.go
```diff
@@ -47,7 +47,11 @@ func newBlockBlobUploader(jptm IJobPartTransferMgr, destination string, p pipeli
 
 func (s *blockBlobUploader) Prologue(ps common.PrologueState) (destinationModified bool) {
 	if s.jptm.Info().PreservePOSIXProperties {
+
 		if unixSIP, ok := s.sip.(IUNIXPropertyBearingSourceInfoProvider); ok {
+			// Clone the metadata before we write to it, we shouldn't be writing to the same metadata as every other blob.
```
shalinijoshi19, line 52: Any unit tests or e2e tests that we could write up for this scenario you think that could have caught this?

nitin-deamon: @adreed-msft we need to do the same for pageBlob too.

zezha-msft (approved):

PR #1823, opened 2022-06-08
Always get properties for file in list file command

tiverma-msft:


cmd/zc_traverser_file.go
```diff
@@ -114,27 +114,26 @@ func (t *fileTraverser) Traverse(preprocessor objectMorpher, processor objectPro
 		var contentProps contentPropsProvider = noContentProps
 		var meta common.Metadata = nil
 
-		// Only get the properties if we're told to
-		if t.getProperties {
```
tiverma-msft, line 118: For files, this means that properties are retrieved only if getProperties arg is true in InitResourceTraverser call from list cmd (which has been hardcoded to false)

There were two ways in which this could have been solved-
1. We make the check if we want to get properties or no and conditionally setting getProperties to true in InitResourceTraverser depending upon if the user has asked for any property
2. Just removing the if statement, so that we always get the properties, which could have meant a tiny bit more time/processing per file, but since the data has already been received from the REST call, we just have to read it, and it wouldn't take a lot of time.

gapra-msft:
getProperties is now set to true for all list traversers. Closing this PR as a result.

PR #1829, opened 2022-06-16
Implement upload/download of symlinks

adreed-msft:
This PR implements a new entity type (Symlink) and adds support for transferring symlinks up to blob as representative files (with the `is_symlink` metadata, as pioneered by BlobFuse), as well as back down from blob. Keep in mind that this uploads the symlink's destination path, not the object on the other end of the symlink.

Creation time of symlinks, nor whether the original file linked to exists or not does not matter, nor will it cause transfer failures.


ste/sender-blobSymlinks.go
```diff
@@ -0,0 +1,131 @@
...
+
+// ditto for s2sCopier
+
+type dummySymlinkS2SCopier struct {
+	blobSymlinkSender
+}
+
+func (d dummySymlinkS2SCopier) GenerateCopyFunc(chunkID common.ChunkID, blockIndex int32, adjustedChunkSize int64, chunkIsWholeFile bool) chunkFunc {
+	// TODO implement me
+	panic("implement me")
```
siminsavani-msft, line 130: Missing implementation?

siminsavani-msft (approved):
LGTM! Just a few minor comments :)

mohsha-msft (changes requested):
Can we add E2E test cases for the same?
We can create a local file (a/file.txt )and it's symlink (symlnk.ink -> a/file.txt) in the local and try to verify the enable/disable flag.

cmd/syncEnumerator.go
```diff
@@ -60,7 +60,7 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 	// TODO: Consider passing an errorChannel so that enumeration errors during sync can be conveyed to the caller.
 	// GetProperties is enabled by default as sync supports both upload and download.
 	// This property only supports Files and S3 at the moment, but provided that Files sync is coming soon, enable to avoid stepping on Files sync work
-	sourceTraverser, err := InitResourceTraverser(cca.source, cca.fromTo.From(), &ctx, &srcCredInfo, nil,
+	sourceTraverser, err := InitResourceTraverser(cca.source, cca.fromTo.From(), &ctx, &srcCredInfo, common.ESymlinkHandlingType.None(),
```
linuxsmiths, line 63: Is this intentional to not use the user supplied symlink behaviour and instead hard code it to None?
Applies to L85 too

cmd/zc_traverser_local.go
```diff
@@ -247,7 +247,28 @@ func WalkWithSymlinks(appCtx context.Context, fullPath string, walkFunc filepath
...
+					skipped, err := getProcessingError(err)
+
+					// If the file was skipped, don't record it.
+					if !skipped {
+						seenPaths.Record(common.ToExtendedPath(result))
+					}
+
+					return err
+				}
+
```
linuxsmiths, line 270: Instead of duplicating the code (and I see that it does not have the HasSeen() handling), why not simply change L249 to

if  (fileInfo.Mode()&os.ModeSymlink != 0) && !symlinkHandling.Follow() {

then the skip check is the only check we need to do

common/fe-ste-models.go
```diff
@@ -1609,3 +1611,33 @@ func (rpt RehydratePriorityType) ToRehydratePriorityType() azblob.RehydratePrior
...
+////////////////////////////////////////////////////////////////////////////////
+type SymlinkHandlingType uint8 // SymlinkHandlingType is only utilized internally to avoid having to carry around two contradictory flags. Thus, it doesn't have a parse method.
+
+// for reviewers: This is different than we usually implement enums, but it's something I've found to be more pleasant in personal projects, especially for bitflags. Should we change the pattern to match this in the future?
+
+type eSymlinkHandlingType uint8
+
+var ESymlinkHandlingType = eSymlinkHandlingType(0)
+
+func (eSymlinkHandlingType) None() SymlinkHandlingType     { return SymlinkHandlingType(0) }
```
linuxsmiths, line 1624: Shall we rename None() to Skip() which is more explicit?

common/fe-ste-models.go
```diff
@@ -1609,3 +1611,33 @@ func (rpt RehydratePriorityType) ToRehydratePriorityType() azblob.RehydratePrior
 		return azblob.RehydratePriorityStandard
 	}
 }
+
+////////////////////////////////////////////////////////////////////////////////
+type SymlinkHandlingType uint8 // SymlinkHandlingType is only utilized internally to avoid having to carry around two contradictory flags. Thus, it doesn't have a parse method.
+
+// for reviewers: This is different than we usually implement enums, but it's something I've found to be more pleasant in personal projects, especially for bitflags. Should we change the pattern to match this in the future?
+
```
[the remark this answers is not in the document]
adreed-msft: We absolutely should during the next AzCopy ship-it.

ste/downloader-blob.go
```diff
@@ -46,6 +47,22 @@ type blobDownloader struct {
...
+	if err != nil {
+		return err
+	}
+	symsip := sip.(ISymlinkBearingSourceInfoProvider) // blob always implements this
+	symlinkInfo, err := symsip.GetSymlinkPath()
+
+	// create the link
+	err = os.Symlink(symlinkInfo, jptm.Info().Destination)
+
+	// todo: write properties
```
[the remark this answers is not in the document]
adreed-msft: methinks this is an artifact from the unix properties upload PR, which this was based on. unix properties download is implemented in another PR.

ste/sender-blobSymlinks.go
adreed-msft: Much like the above dummy sender, most of the functions are panicked out as there is no need nor use for them. They are solely implemented for the sake of copying symlinks.

adreed-msft:
> Can we add E2E test cases for the same? We can create a local file (a/file.txt )and it's symlink (symlnk.ink -> a/file.txt) in the local and try to verify the enable/disable flag.

E2E cases are coming. For now I'm going to resolve the active comments and then implement symlink testing on Linux.

ste/sender-blobSymlinks_linux.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+)
+
+func (s *blobSymlinkSender) getExtraProperties() error {
+	if s.jptm.Info().PreservePOSIXProperties {
+		if unixSIP, ok := s.sip.(IUNIXPropertyBearingSourceInfoProvider); ok {
+			// Clone the metadata before we write to it, we shouldn't be writing to the same metadata as every other blob.
+			s.metadataToApply = common.Metadata(s.metadataToApply).Clone().ToAzBlobMetadata()
+
+			statAdapter, err := unixSIP.GetUNIXProperties()
```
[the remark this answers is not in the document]
adreed-msft: Good catch. You're absolutely correct that we want to grab the properties of the symlink.

ste/sender-blobSymlinks_linux.go
```diff
@@ -0,0 +1,23 @@
+package ste
+
+import (
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+)
+
+func (s *blobSymlinkSender) getExtraProperties() error {
+	if s.jptm.Info().PreservePOSIXProperties {
+		if unixSIP, ok := s.sip.(IUNIXPropertyBearingSourceInfoProvider); ok {
+			// Clone the metadata before we write to it, we shouldn't be writing to the same metadata as every other blob.
```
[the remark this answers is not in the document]
adreed-msft: In Golang, maps are handled as pointers, not as primitive, cloneable types.

If I do
```go
a := map[string]int { "a": 1 }
b := a
```

b is not a deep-copy of a. It's just copied the pointer. if I do `b["b"] = 2` and then say `fmt.Println(a["b"])`, you'll get back 2, because they are the same map.

Now, here's the reason we clone it there: AzCopy never deep-copies the metadata map. It's the same map between every blob. If I modify it on one blob, it modifies it on all the rest of them. This causes data races and incorrect properties being persisted.

ste/sender-blobSymlinks_other.go
```diff
@@ -0,0 +1,7 @@
+// +build !linux
+
+package ste
+
+func (s *blobSymlinkSender) getExtraProperties() error {
+	return nil
```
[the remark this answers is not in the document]
adreed-msft: > why is this not doing what Linux one is doing?
This file is a generic catch-all file for OSes unsupported by the unix properties feature.
> Do you plan to implement it later?
Can you tell me what you think it would do? I don't see anything to implement-- We can't guarantee every other OS has posix properties. Long and short, no I don't plan to implement anything here. It's a no-op designed to satisfy a function call on unsupported platforms.

ste/sender-blobSymlinks.go
```diff
@@ -0,0 +1,131 @@
...
+
+	return out, nil
+}
+
+func (s *blobSymlinkSender) SendSymlink(linkData string) error {
+	err := s.getExtraProperties()
+	if err != nil {
+		return fmt.Errorf("when getting additional folder properties: %w", err)
+	}
+	s.metadataToApply["is_symlink"] = "true" // For BlobFuse
```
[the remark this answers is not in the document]
adreed-msft: I mostly denoted that the header name came from blobfuse

ste/xfer-anyToRemote-symlink.go
```diff
@@ -0,0 +1,65 @@
...
+		jptm.ReportTransferDone()
+		return
+	}
+
+	path, err := symSIP.GetSymlinkPath()
+	if err != nil {
+		jptm.FailActiveSend("getting symlink path", err)
+		commonSenderCompletion(jptm, baseSender, info)
+		return
+	}
```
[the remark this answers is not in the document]
adreed-msft: Made me think about it, and I feel like it's better placed before the prior sender factory block-- No need to initialize a sender or do anything if we can't read the link, right?

ste/mgr-JobPartMgr.go
```diff
@@ -547,8 +547,8 @@ func (jpm *jobPartMgr) createPipelines(ctx context.Context) {
 
 	var statsAccForSip *PipelineNetworkStats = nil // we don't accumulate stats on the source info provider
 
-	// Create source info provider's pipeline for S2S copy.
-	if fromTo == common.EFromTo.BlobBlob() || fromTo == common.EFromTo.BlobFile() {
+	// Create source info provider's pipeline for S2S copy or download (in some cases).
+	if fromTo == common.EFromTo.BlobBlob() || fromTo == common.EFromTo.BlobFile() || fromTo == common.EFromTo.BlobLocal() {
```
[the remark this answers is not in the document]
adreed-msft: Historically we didn't really use the SIP at all for downloads. In fact, it has to manually be created for the downloader if we intend to use it.

This isn't a problem in FileLocal even with SMB properties because files has no OAuth support.

cmd/copy.go
```diff
@@ -85,6 +85,9 @@ type rawCopyCmdArgs struct {
 	legacyExclude         string // used only for warnings
 	listOfVersionIDs      string
 
+	// Indicates the user wants to upload the symlink itself, not the file on the other end
+	preserveSymlinks bool
+
```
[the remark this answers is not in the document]
adreed-msft: Removing --follow-symlinks would be a breaking change. Thus, it must stay.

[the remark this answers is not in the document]
adreed-msft: There is also the case a user wants to do absolutely nothing with symlinks whatsoever.

We _could_ hide follow-symlinks and let it live on in a deprecated state, and return to using a enum instead, but that's not privy to using it as a simple boolean flag (e.g. `--follow-symlinks` would have to be used like `--handle-symlinks=follow`)

cmd/syncEnumerator.go
adreed-msft: Sync has no symlink support, not even follow-symlinks.

cmd/zc_traverser_local.go
adreed-msft: follow symlinks still needs to be handled

common/fe-ste-models.go
adreed-msft: Makes sense to me.

tasherif-msft:
Hey @adreed-msft are we revisiting this? it's marked with 10.17 but it doesn't seem to be a release blocker, given our timeline we should probably move this to 10.18, also it does not seem we've added extensive unit testing here, can we do that as well?

tasherif-msft (changes requested):
Hey @adreed-msft are we revisiting this? it's marked with 10.17 but it doesn't seem to be a release blocker, given our timeline we should probably move this to 10.18, also it does not seem we've added extensive unit testing here, can we do that as well?

nakulkar-msft:
@siminsavani-msft @vibhansa-msft

siminsavani-msft (approved):

cmd/list.go
```diff
@@ -222,7 +222,7 @@ func (cooked cookedListCmdArgs) HandleListContainerCommand() (err error) {
 		}
 	}
 
-	traverser, err := InitResourceTraverser(source, cooked.location, &ctx, &credentialInfo, nil, nil,
+	traverser, err := InitResourceTraverser(source, cooked.location, &ctx, &credentialInfo, common.ESymlinkHandlingType.Skip(), nil,
```
vibhansa-msft, line 225: During traverser we need the actual type as to follow or preserve, why do we pass skip() here?

cmd/zc_traverser_local.go
```diff
@@ -217,7 +217,7 @@ func WalkWithSymlinks(appCtx context.Context, fullPath string, walkFunc filepath
 	// do NOT put fullPath: true into the map at this time, because we want to match the semantics of filepath.Walk, where the walkfunc is called for the root
 	// When following symlinks, our current implementation tracks folders and files.  Which may consume GB's of RAM when there are 10s of millions of files.
 	var seenPaths seenPathsRecorder = &nullSeenPathsRecorder{} // uses no RAM
-	if followSymlinks {
+	if symlinkHandling.Follow() {                              // only if we're following we need to worry about this
```
vibhansa-msft, line 220: In case of broken symlinks do we log when we try to follow it?

ste/sender-blobFolders.go
```diff
@@ -46,9 +46,9 @@ func newBlobFolderSender(jptm IJobPartTransferMgr, destination string, p pipelin
 	}
 	fromTo := jptm.FromTo()
 	if fromTo.IsUpload() {
-		out = &dummyUploader{fsend}
+		out = &dummyFolderUploader{fsend}
 	} else {
-		out = &dummys2sCopier{fsend}
+		out = &dummyFolderS2SCopier{fsend}
```
vibhansa-msft, line 51: These changes are not related to symlink. Do we need them in this PR?

ste/sender-blobSymlinks_linux.go
```diff
@@ -0,0 +1,31 @@
+package ste
+
+import (
+	"errors"
+	"fmt"
+	"github.com/Azure/azure-pipeline-go/pipeline"
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+)
+
+func (s *blobSymlinkSender) getExtraProperties() error {
```
vibhansa-msft, line 10: Have we tried both symlink follow and preserve options in both windows and linux to validate our change is not OS specific at any level.

cmd/list.go
adreed-msft: We need to know how we're handling symlinks. If we're passing in `Follow`, the local traverser has some specific behaviour to follow symlinks-- We could avoid this if we handled following in the STE in the future, but given existing functionality it may be best to delay that into 10.19/20.

For skip-- We could present symlinks as a part of a listing call, but we don't include folder stubs anyway, so, I figured same-same on that general "special blobs" concept.

cmd/zc_traverser_local.go
adreed-msft: For failed symlink traversals, yes.

cmd/zc_traverser_local.go
```diff
@@ -668,8 +691,8 @@ func (t *localTraverser) Traverse(preprocessor objectMorpher, processor objectPr
 				}
 
 				relPath := strings.TrimPrefix(strings.TrimPrefix(cleanLocalPath(filePath), cleanLocalPath(t.fullPath)), common.DeterminePathSeparator(t.fullPath))
-				if !t.followSymlinks && fileInfo.Mode()&os.ModeSymlink != 0 {
-					WarnStdoutAndScanningLog(fmt.Sprintf("Skipping over symlink at %s because --follow-symlinks is false", common.GenerateFullPath(t.fullPath, relPath)))
+				if t.symlinkHandling.None() && fileInfo.Mode()&os.ModeSymlink != 0 {
```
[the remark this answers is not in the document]
adreed-msft: Skip is provided as a general alternative for when following symlinks has too many tradeoffs/doesn't make sense.

For follow, we'll only copy data on the other side of a symlink once. This means data that's symlinked to multiple times will "jump around" depending on whatever's enumerated first. You can test this behavior by having two symlinks in a transfer link to the same file.

For preserve, we won't copy the data at all-- only the fact that there is a symlink, and that it arrives at a specific destination. This is optimal for customers with similar file systems between machines (e.g. if something symlinks to commonly shared libraries, or data within the transfer)

ste/sender-blobFolders.go
adreed-msft: I believe these changes came alongside the merge from dev between AzCopy versions.

ste/sender-blobSymlinks_linux.go
adreed-msft: Preserve symlinks should probably be explicitly limited to Linux-- Symlinks require special permissions on Windows to create, and this is XDM-specific work.

PR #1835, opened 2022-06-22
Reimplement POSIX properties download

adreed-msft:
This PR adds back in previously removed behavior from the POSIX properties upload PR.

In order to achieve persistence of special file types (e.g. pipes, sockets), a special downloader interface was created (creationTimeDownloader). A creationTimeDownloader handles the creation of files, rather than having the STE do that. (Should we just have the STE handle these cases instead to prevent code dupe?)

As for the remainder of the POSIX properties, support for persistence of those properties is handled in a pattern similar to SMB properties.


seanmcc-msft:
Does this need to move to 10.18 or later?

nakulkar-msft:
@nakulkar-msft @tasherif-msft

adreed-msft:
> @nakulkar-msft @tasherif-msft

adding testing, btw.

PR #1839, opened 2022-06-28
"dfs" to "blob" endpoint conversion

mohsha-msft:


zezha-msft (approved):

zezha-msft:
@mohsha-msft could you please also update the help message of set-properties to clarify that the HNS enabled accounts will be processed via blob endpoint.

mohsha-msft:
> @mohsha-msft could you please also update the help message of set-properties to clarify that the HNS enabled accounts will be processed via blob endpoint.

Yeah. Absolutely!

I'm also writing a Wiki for our brand new feature. I'll add there as well. :D

PR #1844, opened 2022-07-08
Implement multi-auth for managed disks (SAS+OAuth) when the managed disk export account requests it

adreed-msft:
This PR implements multi-auth for managed disks as a new credential type.

During credential type discovery, if a SAS token is present, and it is a managed disk account, we'll ping with a get properties call to see if we get back a response requesting additional authorization (targeting `https://disk.azure.com/`). If we do, we opt for the MDOAuthToken CredentialType, signifying we'll need to refresh the OAuth token for the disk target.

Manual testing has been done for download and S2S.


zezha-msft:
Could you please find a place (perhaps in the login help message?) to specify that the OAuth support for MD is in preview?Since the service is also in preview.

cmd/credentialUtil.go
```diff
@@ -135,17 +135,62 @@ func GetOAuthTokenManagerInstance() (*common.UserOAuthTokenManager, error) {
...
 // The implementation logic follows above rule, and adjusts sequence to save web request(for verifying public resource).
-func getBlobCredentialType(ctx context.Context, blobResourceURL string, canBePublic bool, standaloneSAS bool, cpkOptions common.CpkOptions) (common.CredentialType, bool, error) {
+func getBlobCredentialType(ctx context.Context, blobResourceURL string, canBePublic bool, standaloneSAS string, cpkOptions common.CpkOptions) (common.CredentialType, bool, error) {
 	resourceURL, err := url.Parse(blobResourceURL)
 
 	if err != nil {
 		return common.ECredentialType.Unknown(), false, errors.New("provided blob resource string is not in URL format")
 	}
 
+	if standaloneSAS != "" {
```
mohsha-msft, line 145: What is the default of standaloneSAS? If it is empty, what will be the value of RawQuery?
`resourceURL.RawQuery = standaloneSAS`

cmd/credentialUtil.go
```diff
@@ -135,17 +135,62 @@ func GetOAuthTokenManagerInstance() (*common.UserOAuthTokenManager, error) {
...
+		resourceURL.RawQuery = standaloneSAS
+	}
+
 	sas := azblob.NewBlobURLParts(*resourceURL).SAS
+	isMDAccount := strings.HasPrefix(resourceURL.Host, "md-")
```
mohsha-msft, line 150: Is it documented somewhere? 
Nit: Please define "md-" as a constant. And I see it is being reused so maybe write a separate function to check it.

cmd/credentialUtil.go
```diff
@@ -135,17 +135,62 @@ func GetOAuthTokenManagerInstance() (*common.UserOAuthTokenManager, error) {
...
+			if cpkOptions.IsSourceEncrypted {
+				clientProvidedKey = common.GetClientProvidedKey(cpkOptions)
+			}
+
+			bURL := azblob.NewBlobURL(*resourceURL, p)
+			_, err := bURL.GetProperties(ctx, azblob.BlobAccessConditions{}, clientProvidedKey)
+
+			if err != nil {
+				if stgErr, ok := err.(azblob.StorageError); ok {
+					if httpResp := stgErr.Response(); httpResp.StatusCode == 401 {
```
mohsha-msft, line 182: Nit: Please add comments in here to explain what is happening in this segment of code.

cmd/credentialUtil.go
```diff
@@ -135,17 +135,62 @@ func GetOAuthTokenManagerInstance() (*common.UserOAuthTokenManager, error) {
...
+	}
+
 	sas := azblob.NewBlobURLParts(*resourceURL).SAS
+	isMDAccount := strings.HasPrefix(resourceURL.Host, "md-")
+	canBePublic = canBePublic && !isMDAccount // MD accounts cannot be public.
 
 	// If SAS existed, return anonymous credential type.
-	if isSASExisted := sas.Signature() != ""; isSASExisted || standaloneSAS {
+	if isSASExisted := sas.Signature() != ""; isSASExisted {
+		if isMDAccount {
```
mohsha-msft, line 155: Nit: Can we extract this code out in a separate function. Something like `ValidateMDCredential`

mohsha-msft (commented):
Please add E2E test cases to validate these changes.

PR #1849, opened 2022-07-14
Update changelog and version for 10.16.0

siminsavani-msft:


nakulkar-msft (approved):
Add details of #1844 in new features

PR #1859, opened 2022-07-27
Implement MD OAuth testing

adreed-msft:
Implements MD OAuth testing by implementing a new resource manager.

This PR will fail testing until Ze returns as we need OAuth access to grab a Managed Disk SAS token and add the OAuth access requirement to one


e2etest/zt_managed_disks_test.go
```diff
@@ -20,5 +20,56 @@
...
 
+import (
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"testing"
+)
+
 // Purpose: Tests for the special cases that relate to moving managed disks (default local VHD to page blob; special handling for
 //     md- and md-impex URLs.
+
+func TestManagedDisks_NoOAuthRequired(t *testing.T) {
```
vibhansa-msft, line 31: Can we have a negative test case where access is revoked and validate transfer attempt after that fails.

adreed-msft: This feels like it leaves the boundary of testing AzCopy and enters the boundary of testing the service. I don't see the value in adding a test like this.

adreed-msft:
MD OAuth testing is disabled, but this PR should still make it through for the patch release as it contains a fix for one-shot login w/ md oauth

zezha-msft (approved):
lgtm

vibhansa-msft (approved):

PR #1860, opened 2022-07-28
Parallelize listing containers

nakulkar-msft:


cmd/copyEnumeratorInit.go
```diff
@@ -255,6 +234,20 @@ func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 			if cName == "" || (srcLevel == ELocationLevel.Service() && dstLevel > ELocationLevel.Service()) {
 				cName, err = containerResolver.ResolveName(object.ContainerName)
 
+				//attempt creating a container. We'll not create it multiple times.
+				err = cca.createDstContainer(cName, cca.Destination, ctx, existingContainers, azcopyLogVerbosity)
+
+				// if JobsAdmin is nil, we're probably in testing mode.
+				// As a result, container creation failures are expected as we don't give the SAS tokens adequate permissions.
+				// check against seenFailedContainers so we don't spam the job log with initialization failed errors
+				if _, ok := seenFailedContainers[cName]; err != nil && jobsAdmin.JobsAdmin != nil && !ok {
```
adreed-msft, line 243: (Unrelated to this PR)

This functionality often isn't used, and more frequently confuses the heck out of customers and support engineers alike. We should just stop outputting errors from this on the default logging level to help avoid the confusion.

cmd/copyEnumeratorInit.go
```diff
@@ -255,6 +234,20 @@ func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
+					logDstContainerCreateFailureOnce.Do(func() {
+						glcm.Info("Failed to create one or more destination container(s). Your transfers may still succeed if the container already exists.")
+					})
+					jobsAdmin.JobsAdmin.LogToJobLog(fmt.Sprintf("failed to initialize destination container %s; the transfer will continue (but be wary it may fail): %s", bucketName, err), pipeline.LogWarning)
```
adreed-msft, line 247: (Unrelated to this PR)

We have a scanning log these days. Alternative to the log level change, we *could* move the location this is logged to to that, as customers tend not to look at that.

cmd/zc_traverser_blob_account.go
```diff
@@ -51,68 +51,81 @@ func (t *blobAccountTraverser) IsDirectory(_ bool) bool {
...
 	// listContainers will return the cached container list if containers have already been listed by this traverser.
-	cList, err := t.listContainers()
-
-	if err != nil {
-		return err
-	}
+	cList, errChan := t.listContainers()
 
-	for _, v := range cList {
+	for v := range cList {
```
adreed-msft, line 114: This will only run one container listing instance at once. While it will get the job done of separating the process of enumeration of blobs from containers, it will not actually parallelize the listing of blobs.

We could consider starting up multiple goroutines to handle the blob listing on individual containers; it would not increase the price on the listing operation.

adreed-msft: Ditto to the other account level enumerators.

adreed-msft (changes requested):

nakulkar-msft:
Need to rework. This will not fix the original problem of listing large number of containers

PR #1868, opened 2022-08-10
Fix for linux arm build

siminsavani-msft:
Fix for linux arm build error: "ste/sourceInfoProvider-Local_linux.go:126:10: cannot use s.Nlink (type uint32) as type uint64 in return argument" seen in build pipeline


adreed-msft (approved):

zezha-msft:
Included in https://github.com/Azure/azure-storage-azcopy/pull/1917

PR #1872, opened 2022-08-17
Add mitigation for weird NtQuerySecurityObject behavior on NAS sources

adreed-msft:
For whatever odd reason, NtQuerySecurityObject likes to return 0 on the buffer length when it succeeds targeting an object on a NAS source, but still returns a valid security descriptor struct.

This PR adds a check for that, and validates the struct if the edge case conditions were met.

### DIFF
--- a/ste/sourceInfoProvider-Local_windows.go
+++ b/ste/sourceInfoProvider-Local_windows.go
@@ -70,9 +70,11 @@ func (f localFileSourceInfoProvider) GetSDDL() (string, error) {
 			uint32(len(buf)),
 			&bufLen)
 
-		// mitigate weirdness with 0 buflen on success with a NAS source
-		if status == ntdll.STATUS_SUCCESS && bufLen == 0 {
-			bufLen = uint32(len(buf))
+		// get real buffer length, since what's returned by ntquerysecurityobject is questionable for STATUS_SUCCESS
+		if status == ntdll.STATUS_SUCCESS {
+			sd := (*windows.SECURITY_DESCRIPTOR)(unsafe.Pointer(&buf[0])) // ntdll.SecurityDescriptor is equivalent
+
+			bufLen = sd.Length()
 			needValidate = true
 		}
 
--- a/ste/sourceInfoProvider-Local_windows.go
+++ b/ste/sourceInfoProvider-Local_windows.go
@@ -3,7 +3,6 @@
 package ste
 
 import (
-	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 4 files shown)


siminsavani-msft (approved):
LGTM

ste/sourceInfoProvider-Local_windows.go
```diff
@@ -62,20 +61,32 @@ func (f localFileSourceInfoProvider) GetSDDL() (string, error) {
...
+		return status
 	}, &buf, &bufLen)
 
 	if status != ntdll.STATUS_SUCCESS {
-		return "", errors.New(fmt.Sprint("failed to query security object", f.jptm.Info().Source, "ntstatus:", status))
+		return "", fmt.Errorf("failed to query security object %s (ntstatus: %s)", f.jptm.Info().Source, status.String())
 	}
 
 	sd := (*windows.SECURITY_DESCRIPTOR)(unsafe.Pointer(&buf[0])) // ntdll.SecurityDescriptor is equivalent
+	if needValidate && !sd.IsValid() {
```
nakulkar-msft, line 87: This could have been done unconditionally.

ste/sourceInfoProvider-Local_windows.go
```diff
@@ -62,20 +61,32 @@ func (f localFileSourceInfoProvider) GetSDDL() (string, error) {
...
 	status := ntdll.CallWithExpandingBuffer(func() ntdll.NtStatus {
-		return ntdll.NtQuerySecurityObject(
+		status := ntdll.NtQuerySecurityObject(
 			fd,
 			windows.OWNER_SECURITY_INFORMATION|windows.GROUP_SECURITY_INFORMATION|windows.DACL_SECURITY_INFORMATION,
 			(*ntdll.SecurityDescriptor)(unsafe.Pointer(&buf[0])),
 			uint32(len(buf)),
 			&bufLen)
+
+		// mitigate weirdness with 0 buflen on success with a NAS source
```
nakulkar-msft, line 73: Provide a bit more context about the bug in the comments.

nakulkar-msft (approved):

ste/sourceInfoProvider-Local_windows.go
adreed-msft: Technically we probably don't even need to do it.

I discussed with Jason Shay, and from like Vista back, 0 tends to be what gets returned, even if a valid value gets returned.

That said, I felt like best practice was to check it if things didn't seem completely right.

PR #1875, opened 2022-08-19
Fix issues with manipulating URLs for the folder deletion manager

adreed-msft:
Manipulating folder URLs by their full string was causing issues if a user had a % in a folder name, as %25 would get decoded to %, and then fail to re-parse down the line.

This PR fixes this behavior by removing the swap to string, and just manipulating under *url.URL. A clone of the *url.URL is made to prevent any wonkiness with pointers.

### DIFF
--- a/cmd/zt_copy_s2smigration_test.go
+++ b/cmd/zt_copy_s2smigration_test.go
@@ -51,23 +51,22 @@ const (
 var defaultS2SInvalideMetadataHandleOption = common.DefaultInvalidMetadataHandleOption
 
 func (s *cmdIntegrationSuite) SetUpSuite(c *chk.C) {
-	// skip cleaning up the S3 account when not necessary
-	if isS3Disabled() && gcpTestsDisabled() {
-		return
+	if !isS3Disabled() {
+		if s3Client, err := createS3ClientWithMinio(createS3ResOptions{}); err == nil {
+			cleanS3Account(c, s3Client)
+		} else {
+			// If S3 credentials aren't supplied, we're probably only trying to run Azure tests.
+			// As such, gracefully return here instead of cancelling every test because we couldn't clean up S3.
+			c.Log("S3 client could not be successfully initialised")
+		}
 	}
 
-	if s3Client, err := createS3ClientWithMinio(createS3ResOptions{}); err == nil {
-		cleanS3Account(c, s3Client)
-	} else {
-		// If S3 credentials aren't supplied, we're probably only trying to run Azure tests.
-		// As such, gracefully return here instead of cancelling every test because we couldn't clean up S3.
-		c.Log("S3 client could not be successfully initialised")
-	}
-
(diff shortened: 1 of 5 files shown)


cmd/zt_copy_s2smigration_test.go
```diff
@@ -51,23 +51,22 @@ const (
...
-		return
-	}
-
-	if s3Client, err := createS3ClientWithMinio(createS3ResOptions{}); err == nil {
-		cleanS3Account(c, s3Client)
-	} else {
-		// If S3 credentials aren't supplied, we're probably only trying to run Azure tests.
-		// As such, gracefully return here instead of cancelling every test because we couldn't clean up S3.
-		c.Log("S3 client could not be successfully initialised")
+	if !isS3Disabled() {
```
vibhansa-msft, line 54: What is the change here, condition wise it's still the same it feels. It's just earlier we had an early exit in case both S3 and GCP were disabled and now we do the reverse checks and exit. What is the motivation behind this change?

cmd/zt_test.go
```diff
@@ -595,13 +595,18 @@ func deleteGCPBucket(c *chk.C, client *gcpUtils.Client, bucketName string, waitQ
...
 		attrs, err := it.Next()
-		if err == iterator.Done {
-			break
+		if err != nil { // if Next returns an error other than iterator.Done, all subsequent calls will return the same error.
+			if err == iterator.Done {
+				break
+			}
+
+			c.Log("Failed to clear GCS bucket:", err) // todo: maybe this code should be more resilient
+			return
```
vibhansa-msft, line 604: Shouldn't we be returning some sort of failure here. I see function has no return type, feels either we assume it will always succeed or we don't care even if it fails.

common/folderDeletionManager.go
```diff
@@ -155,7 +171,7 @@ func (s *standardFolderDeletionManager) RecordChildDeleted(childFile *url.URL) {
 	}
 
 	s.mu.Lock()
-	folderStatePtr, alreadyKnown := s.contents[folder]
+	folderStatePtr, alreadyKnown := s.contents[s.getMapKey(folder)]
```
vibhansa-msft, line 174: I think better to break this into two statements and log if getMapKey returns unexpected strings like an empty string or something.

cmd/zt_test.go
adreed-msft: I'm in agreement there. I was following the paradigm that was already present (it sounded like cleanup failure was intended to be a non-failing error.

common/folderDeletionManager.go
adreed-msft: I don't see much value in this. "" should be the root directory. A path starting with no slash is weird, but since we're working with parsed URLs here and not constructing them ourselves, I doubt this would occur.

siminsavani-msft (approved):
LGTM!

nakulkar-msft (approved):

vibhansa-msft (approved):

Issue #1887, opened 2022-09-06
Azure AD Workload Identity support


adam-orosz:
I'm trying to use [Azure AD Workload Identity](https://azure.github.io/azure-workload-identity/docs/) with `azcopy` and I'm getting `panic: runtime error: index out of range [1] with length 1`. I'm wondering if this method of authentication is supposed to work.

### Which version of the AzCopy was used? 
AzCopy 10.16.0

### Which platform are you using? (ex: Windows, Mac, Linux)
Linux

### What command did you run?
`azcopy login --identity --identity-client-id < redacted Client ID of a Managed Identity >`

### What problem was encountered?
`panic: runtime error: index out of range [1] with length 1`

### Have you found a mitigation/solution?
no

### Additional notes
The pod was patched by the mutating webhook of the `Azure AD Workload Identity` project.

adreed-msft:
I've reviewed your PR. Thank you for your contribution.

PR #1890, opened 2022-09-08
Fix fixupTokenJson

adam-orosz:
This PR fixes the issue experienced when using [Azure AD Workload Identity](https://github.com/Azure/azure-storage-azcopy/issues/1887)

The token returned by the OIDC token issuer returned an integer (0) for the "not_before" field. This caused a panic as the `fixupTokenJson` function is expecting the field to be a string.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,41 @@
+<!-- BEGIN MICROSOFT SECURITY.MD V0.0.7 BLOCK -->
+
+## Security
+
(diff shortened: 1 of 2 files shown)


adreed-msft (approved):
Rebase from main to dev.

nakulkar-msft (approved):

sashamor:
We need this fix in our deployment workflow 🙏 Very excited for this to merge & be released.

PR #1893, opened 2022-09-13
Do not log request/response for container creation error

nakulkar-msft:


cmd/copyEnumeratorInit.go
```diff
@@ -163,14 +163,15 @@ func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 		// only create the destination container in S2S scenarios
 		if cca.FromTo.From().IsRemote() && dstContainerName != "" { // if the destination has a explicit container name
 			// Attempt to create the container. If we fail, fail silently.
-			err = cca.createDstContainer(dstContainerName, cca.Destination, ctx, existingContainers, azcopyLogVerbosity)
+			err = cca.createDstContainer(dstContainerName, cca.Destination, ctx, existingContainers, common.ELogLevel.None())
```
adreed-msft, line 166: I think Debug, like the other logs, is probably plenty acceptable rather than strictly None.

adreed-msft (approved):

cmd/copyEnumeratorInit.go
nakulkar-msft: We are still capturing the error and will be displaying it if level is debug. It'll be explicitly logged by AzCopy and not sdk

zezha-msft (approved):

PR #1898, opened 2022-09-19
Improve debug-ability of e2e tests by uploading logs of failed jobs

adreed-msft:

PR #1901, opened 2022-09-20
Fix ADLSG2 intermittent failure

adreed-msft:
When a ADLSG2 folder already exists with blobs under it and we attempt to overwrite the properties on it, the transfer will fail. As a result, we opt for a best-effort overwrite.

### DIFF
--- a/ste/sender-blobFolders.go
+++ b/ste/sender-blobFolders.go
@@ -147,37 +147,42 @@ func (b *blobFolderSender) EnsureFolderExists() error {
 		}
 	}
 
-	b.metadataToApply["hdi_isfolder"] = "true" // Set folder metadata flag
-	err = b.getExtraProperties()
-	if err != nil {
-		return fmt.Errorf("when getting additional folder properties: %w", err)
-	}
-
 	err = t.CreateFolder(b.DirUrlToString(), func() error {
-		_, err := b.destination.Upload(b.jptm.Context(),
-			strings.NewReader(""),
-			b.headersToAppply,
-			b.metadataToApply,
-			azblob.BlobAccessConditions{},
-			azblob.DefaultAccessTier, // It doesn't make sense to use a special access tier, the blob will be 0 bytes.
-			b.blobTagsToApply,
-			b.cpkToApply,
-			azblob.ImmutabilityPolicyOptions{})
+		b.metadataToApply["hdi_isfolder"] = "true" // Set folder metadata flag
+		err = b.getExtraProperties()
+		if err != nil {
+			return fmt.Errorf("when getting additional folder properties: %w", err)
+		}
 
-		return err
+		err = t.CreateFolder(b.DirUrlToString(), func() error {
+			_, err := b.destination.Upload(b.jptm.Context(),
+				strings.NewReader(""),
+				b.headersToAppply,
+				b.metadataToApply,
+				azblob.BlobAccessConditions{},
+				azblob.DefaultAccessTier, // It doesn't make sense to use a special access tier, the blob will be 0 bytes.
+				b.blobTagsToApply,
+				b.cpkToApply,
+				azblob.ImmutabilityPolicyOptions{})
+
+			return err
+		})
+
+		if err != nil {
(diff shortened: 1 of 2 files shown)


gapra-msft:
@adreed-msft was this issue reported on Github? If so can we link it here?

adreed-msft:
No-- This was detected through #1898.

gapra-msft:
Can we write a test for this?

tasherif-msft:
Let's add some tests and get prioritize it to get merged in to get our pipelines stable

gapra-msft (changes requested):
TODO: Tests

siminsavani-msft (approved):

ste/sender-blobFolders.go
```diff
@@ -53,6 +54,53 @@ func newBlobFolderSender(jptm IJobPartTransferMgr, destination string, p pipelin
 	return out, nil
 }
 
+func (b *blobFolderSender) overwriteDFSProperties() (string, error) {
+	b.jptm.Log(pipeline.LogWarning, "It is impossible to completely overwrite a folder with existing content under it on a hierarchical namespace storage account. A best-effort attempt will be made, but if CPK does not match the transfer will fail.")
+
+	b.metadataToApply["hdi_isfolder"] = "true" // Set folder metadata flag
```
vibhansa-msft, line 60: Is this function generic as name does not suggest this is only for folders though we are setting hdi_isfolder unconditionally here.

ste/sender-blobFolders.go
```diff
@@ -103,6 +162,26 @@ func (b *blobFolderSender) EnsureFolderExists() error {
 		return fmt.Errorf("when creating folder: %w", err)
 	}
 
+	// Upload ADLS Gen 2 ACLs
+	if b.jptm.FromTo() == common.EFromTo.BlobBlob() && b.jptm.Info().PreserveSMBPermissions.IsTruthy() {
```
vibhansa-msft, line 166: Add test cases for both the changes.

nakulkar-msft (approved):

ste/sender-blobFolders.go
gapra-msft: @vibhansa-msft The type is a blobFolderSender so I think its safe to always add the metadata

ste/sender-blobFolders.go
```diff
@@ -53,6 +54,53 @@ func newBlobFolderSender(jptm IJobPartTransferMgr, destination string, p pipelin
...
+	if err != nil {
+		return "Set Blob Tags", err
+	}
+	_, err = b.destination.SetHTTPHeaders(b.jptm.Context(), b.headersToAppply, azblob.BlobAccessConditions{})
+	if err != nil {
+		return "Set HTTP Headers", err
+	}
+
+	// Upload ADLS Gen 2 ACLs
+	if b.jptm.FromTo() == common.EFromTo.BlobBlob() && b.jptm.Info().PreserveSMBPermissions.IsTruthy() {
```
gapra-msft, line 82: how do we know if the destination is adls?

ste/sender-blobFolders.go
```diff
@@ -103,6 +162,26 @@ func (b *blobFolderSender) EnsureFolderExists() error {
 		return fmt.Errorf("when creating folder: %w", err)
 	}
 
+	// Upload ADLS Gen 2 ACLs
+	if b.jptm.FromTo() == common.EFromTo.BlobBlob() && b.jptm.Info().PreserveSMBPermissions.IsTruthy() {
+		bURLParts := azblob.NewBlobURLParts(b.destination.URL())
```
gapra-msft, line 167: this code seems to be duplicate of above, should we make this a method?

gapra-msft (approved):

vibhansa-msft (approved):

ste/sender-blobFolders.go
adreed-msft: The only place this is called is when we receive an HNS specific error message.

tasherif-msft (approved):

Issue #1631, opened 2021-11-30
AzCopy sync re-copies files even if file already exists and is unchanged (date modified matches).


g7cnp:
##### Note: The version is visible when running AzCopy without any argument
10.13.0
### Which platform are you using? (ex: Windows, Mac, Linux)
windows 11
##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.
azcopy.exe sync "https://scottrichards.file.core.windows.net/sr2/apps/test1?[KEY]" "C:\local copies\apps\test1" 
### What problem was encountered?
On first run all works fine and files copy from remote folder 'test1' to local folder 'test1' 
If the command is then re-run the files get copied again despite neither file being changed. If either source or destination file is modified behaviour is as expected. 
It would appear that the destination file gets over written if it's modified date is less than or equale to the source file rather than just less than? It may be I'm just missing a flag to adjust this behaviour?
### How can we reproduce the problem in the simplest way?
Just run a sync twice and check if files get copied during the second run.
### Have you found a mitigation/solution?
Double check the date/time comparison logic.

g7cnp:
Is there a reason this hasn't been picked up?
All i want to do is copy a folder structure from Azure to the local drive, then keep that local copy UpToDate with any changes i make to the Azure version, but every time i run the command it copies everything not just the changes. I've tried using the sync command and copy with --overwrite=ifSourceNewer

adreed-msft:
I've got a repro on this.

adreed-msft:
> Is there a reason this hasn't been picked up?
I apologize. Unfortunately, sometimes things fall through the cracks. We're a somewhat small team working on AzCopy and we can't keep up with everything, between issues, feature requests, etc. both internal and external.

g7cnp:
I let #1631 drop as it seamed a duplicate of #1666, but either will do!

Regards

Huw Weatherhead
Weatherheads Web
07917890093

adreed-msft:
It seems like this is a bit of funkiness at play with --preserve-smb-info=true (which is the default setting). There are two separate LMTs stored service side-- The actual `Last-Modified` header (which is the last time we've written to the file according to the service) and `x-ms-last-write-time` (which is the last time we've written to the file according to our SMB properties). These don't end up aligning, and this causes issues.

[...]

PR #1930, opened 2022-10-17
Make sync use last write time for Azure Files

adreed-msft:
Read #1631 for details


zezha-msft (approved):
lgtm, please make sure to double check the relevant scenarios to make sure there's no unwanted side effects

siminsavani-msft (approved):
LGTM! Could we add tests for the scenario that was reported?

PR #1932, opened 2022-10-19
UX improvement: avoid crash when copying S2S with user delegation SAS

zezha-msft:
Previously, if a user was copying S2S with a user delegation SAS and had s2s-preserve-tier set to the default value, the tool would crash. We should avoid this problem as the error message is not clear nor actionable.

Instead, we should simply go ahead with the transfer and see if tier is tolerated.


adreed-msft (approved):

nakulkar-msft (approved):

zezha-msft:
Pipeline failed due to storage account configuration change. Rerunning now.

Issue #1956, opened 2022-11-09
Azcopy sync not listing the existing files which are modified but fetching newly added files


shruthiravik:
### Which version of the AzCopy was used? 
10.16.2
### Which platform are you using? (ex: Windows, Mac, Linux)
Linux

### What command did you run?
azcopy sync $env:FILESHARESASURL $env:DESTINATION

### What problem was encountered?
With this new version we were able to see the newly added files but we are not able to see the old files which are modified.

### How can we reproduce the problem in the simplest way?
Modify the existing file and run the above command. you will not see the modified changes 
### Have you found a mitigation/solution?
No

PR #1958, opened 2022-11-10
Fix files sync by determining which LMT to use via smb properties flag

adreed-msft:
https://github.com/Azure/azure-storage-azcopy/issues/1956

It turns out having two completely unreliable LMTs doesn't work well.

### DIFF
--- a/cmd/syncComparator.go
+++ b/cmd/syncComparator.go
@@ -35,11 +35,12 @@ type syncDestinationComparator struct {
 	// storing the source objects
 	sourceIndex *objectIndexer
 
+	preferSMBTime     bool
 	disableComparison bool
 }
 
-func newSyncDestinationComparator(i *objectIndexer, copyScheduler, cleaner objectProcessor, disableComparison bool) *syncDestinationComparator {
-	return &syncDestinationComparator{sourceIndex: i, copyTransferScheduler: copyScheduler, destinationCleaner: cleaner, disableComparison: disableComparison}
+func newSyncDestinationComparator(i *objectIndexer, copyScheduler, cleaner objectProcessor, preferSMBTime, disableComparison bool) *syncDestinationComparator {
+	return &syncDestinationComparator{sourceIndex: i, copyTransferScheduler: copyScheduler, destinationCleaner: cleaner, preferSMBTime: preferSMBTime, disableComparison: disableComparison}
 }
 
 // it will only schedule transfers for destination objects that are present in the indexer but stale compared to the entry in the map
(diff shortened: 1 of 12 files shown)


gapra-msft (changes requested):
are there any tests we can add to validate this change?

vibhansa-msft:
We shall add test case for both the cases where preferSMBTime  : true/false.

tasherif-msft:
why do we prefer smbLMT over service LMT?

adreed-msft:
> why do we prefer smbLMT over service LMT?

Long and dumb story:
1) We have no control of service LMT
2) We do have control of SMB LMT
3) SMB LMT is what's supposedly "real" if the only tool being used to upload is AzCopy
4) SMB LMT is what's persisted back down to local if SMB properties persistence is enabled

There are two different LMTs-- neither are really truly "accurate". The former only updates when the file is updated on the service & cannot be controlled, and the latter only updates when the client updates it.

siminsavani-msft (approved):
LGTM! Adding a test case for this scenario would be good!

cmd/syncComparator.go
```diff
@@ -82,16 +83,18 @@ type syncSourceComparator struct {
 	// storing the destination objects
 	destinationIndex *objectIndexer
 
+	preferSMBTime     bool
 	disableComparison bool
 }
 
-func newSyncSourceComparator(i *objectIndexer, copyScheduler objectProcessor, disableComparison bool) *syncSourceComparator {
-	return &syncSourceComparator{destinationIndex: i, copyTransferScheduler: copyScheduler, disableComparison: disableComparison}
+func newSyncSourceComparator(i *objectIndexer, copyScheduler objectProcessor, preferSMBTime, disableComparison bool) *syncSourceComparator {
```
tasherif-msft, line 90: we really should start creating options bags :'(

adreed-msft: Yes I completely agree.

I would like to take some time in the near future to focus on improving the overall quality of life in the AzCopy codebase, because I generally feel immense pain whenever I have to pass a flag through 8 layers of code and sometimes with 20+ argument functions

cmd/syncComparator.go
```diff
@@ -57,7 +58,7 @@ func (f *syncDestinationComparator) processIfNecessary(destinationObject StoredO
 	// if the destinationObject is present at source and stale, we transfer the up-to-date version from source
 	if present {
 		defer delete(f.sourceIndex.indexMap, destinationObject.relativePath)
-		if f.disableComparison || sourceObjectInMap.isMoreRecentThan(destinationObject) {
+		if f.disableComparison || sourceObjectInMap.isMoreRecentThan(destinationObject, f.preferSMBTime) {
```
vibhansa-msft, line 61: Need test cases for positive and negative scenarios here.

vibhansa-msft (approved):

tasherif-msft (approved):

PR #1961, opened 2022-11-14
[Bug] Correct odd behaviour surrounding folder overwrites

adreed-msft:
# Bug description

Assume a directory structure `C:\like\this`.

On one thread, AzCopy goes to create `this` and set it's properties.
On the other thread, AzCopy goes to create `like` and set it's properties.

`this` creates it's parent directory `like`, but before the folder creation manager can be informed this has occurred, `like` skipped over folder creation, having existed, and locks the mutex on the folder creation manager.

Though we created `like`, we're not sure we did, because it hasn't been inserted into the creation manager yet.

# Solution

Change the folder creation manager such that the act of creating the directory is done under the mutex to prevent getting locked out when trying to inform the manager.

### DIFF
--- a/ste/folderCreationTracker.go
+++ b/ste/folderCreationTracker.go
@@ -37,8 +37,9 @@ func NewFolderCreationTracker(fpo common.FolderPropertyOption, plan *JobPartPlan
 
 type nullFolderTracker struct{}
 
-func (f *nullFolderTracker) RecordCreation(folder string) {
+func (f *nullFolderTracker) CreateFolder(folder string, doCreation func() error) error {
 	// no-op (the null tracker doesn't track anything)
+	return doCreation()
 }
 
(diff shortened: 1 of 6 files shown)


nakulkar-msft (approved):

ste/folderCreationTracker.go
```diff
@@ -76,12 +77,17 @@ func (f *jpptFolderTracker) RegisterPropertiesTransfer(folder string, transferIn
...
+func (f *jpptFolderTracker) CreateFolder(folder string, doCreation func() error) error {
 	f.mu.Lock()
 	defer f.mu.Unlock()
 
 	if folder == common.Dev_Null {
-		return // Never persist to dev-null
+		return nil // Never persist to dev-null
+	}
+
+	err := doCreation()
```
vibhansa-msft, line 88: Kindly add UT and E2E test around this scenario

siminsavani-msft (approved):
LGTM!

PR #1963, opened 2022-11-15
Fixed a bug where preserve permissions would not work with OAuth

gapra-msft:


ste/mgr-JobPartMgr.go
```diff
@@ -581,8 +581,9 @@ func (jpm *jobPartMgr) createPipelines(ctx context.Context) {
 
 		// Consider the ADLSG2->ADLSG2 ACLs case
 		if fromTo == common.EFromTo.BlobBlob() && jpm.Plan().PreservePermissions.IsTruthy() {
+			credential := common.CreateBlobFSCredential(ctx, credInfo, credOption)
 			jpm.secondarySourceProviderPipeline = NewBlobFSPipeline(
-				azbfs.NewAnonymousCredential(),
+				credential,
```
vibhansa-msft, line 586: Is this something that hits only in creating secondarySource. As function is createPipeline sounds like having a wrong credential here should have never worked so just want to understand what condition is failing here.

gapra-msft: this works when using SAS since anonymous credential is for SAS. But if customer uses OAuth we never set that credential (it was anonymous cred)_so we failed for all datalake endpoint operations w/ OAuth

vibhansa-msft (approved):

siminsavani-msft (approved):
LGTM!

PR #1964, opened 2022-11-16
Copy properties from Source

nakulkar-msft:
XDM has a requirement where the posix properties changed on local file system needs to be reflected on the blob container (as metadata) - without explicitly transferring data. To do this,
1. Introduced new Overwrite type
2. Introduced new transfer processor function - anyToRemote-fileproperties, and a propertiesSender Interface
3. This new interface has a chunkFunc which will copy posix attrs. This is common across all blob types.


ste/sender-blockBlob.go
```diff
@@ -273,6 +273,32 @@ func (s *blockBlobSenderBase) Cleanup() {
...
+	return createChunkFunc(true, s.jptm, id, func() {
+		if s.jptm.Info().PreservePOSIXProperties {
+
+			if unixSIP, ok := s.sip.(IUNIXPropertyBearingSourceInfoProvider); ok {
+				// Clone the metadata before we write to it, we shouldn't be writing to the same metadata as every other blob.
+				s.metadataToApply = common.Metadata(s.metadataToApply).Clone().ToAzBlobMetadata()
+	
+				statAdapter, err := unixSIP.GetUNIXProperties()
+				if err != nil {
+					s.jptm.FailActiveSend("GetUNIXProperties", err)
```
vibhansa-msft, line 287: so if we fail to get the properties, we fail entire transfer ? is it what customer would expect ?

nakulkar-msft: Yes, it is a hard failure generally in all our scenarios

ste/xfer-anyToRemote-fileProperties.go
```diff
@@ -0,0 +1,84 @@
...
+		jptm.LogSendError(info.Source, info.Destination, err.Error(), 0)
+		jptm.SetStatus(common.ETransferStatus.Failed())
+		jptm.ReportTransferDone()
+		return
+	}
+
+	//Just one chunk to schedule
+	jptm.ScheduleChunks(s.GenerateCopyMetadata(id))
+
+	commonSenderCompletion(jptm, baseSender, info) // for consistency, always run the standard epilogue
```
nitin-deamon, line 83: @nakulkar-msft  commonSenderCompletion call will mark the ransfer complete, we should mark it complete only once transfer is done. So we need something like this 
jptm.SetNumberOfChunks(1)
jptm.SetActionAfterLastChunk(func() { commonSenderCompletion(jptm, baseSender, info) })

siminsavani-msft (approved):
LGTM

ste/xfer-anyToRemote-fileProperties.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"github.com/Azure/azure-pipeline-go/pipeline"
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+)
+
+// anyToRemote_folder handles all kinds of sender operations for FOLDERS - both uploads from local files, and S2S copies
+func anyToRemote_fileProperties(jptm IJobPartTransferMgr, info TransferInfo, p pipeline.Pipeline, pacer pacer, senderFactory senderFactory, sipf sourceInfoProviderFactory) {
+	// schedule the work as a chunk, so it will run on the main goroutine pool, instead of the
+	// smaller "transfer initiation pool", where this code runs.
+	id := common.NewChunkID(jptm.Info().Source, 0, 0)
```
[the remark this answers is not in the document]
nakulkar-msft: Dummy chunk for bookkeeping.

ste/xfer-anyToRemote-file.go
```diff
@@ -181,6 +181,8 @@ func anyToRemote(jptm IJobPartTransferMgr, p pipeline.Pipeline, pacer pacer, sen
 
 	if info.IsFolderPropertiesTransfer() {
 		anyToRemote_folder(jptm, info, p, pacer, senderFactory, sipf)
+	} else if (jptm.GetOverwriteOption() == common.EOverwriteOption.PosixProperties() && info.EntityType == common.EEntityType.File()) {
```
adreed-msft, line 184: ❓ What if a user specifies overwrite properties, but ends up uploading new files?

ste/xfer-anyToRemote-fileProperties.go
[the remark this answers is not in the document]
adreed-msft: @vibhansa-msft Every file must have at least 1 chunk-- Even 0-byte files have a single chunk in the STE. This is just for status tracking purposes.

ste/xfer-anyToRemote-fileProperties.go
```diff
@@ -0,0 +1,85 @@
...
+	// step 2a. Create sender
+	srcInfoProvider, err := sipf(jptm)
+	if err != nil {
+		jptm.LogSendError(info.Source, info.Destination, err.Error(), 0)
+		jptm.SetStatus(common.ETransferStatus.Failed())
+		jptm.ReportTransferDone()
+		return
+	}
+
+	if (jptm.GetOverwriteOption() == common.EOverwriteOption.PosixProperties() &&
```
adreed-msft, line 52: ❗ It doesn't make sense to be here if overwrite isn't posix properties in the first place, this should be an || and an !=

testSuite/scripts/test_upload_block_blob.py
```diff
@@ -835,3 +835,26 @@ def test_follow_symlinks_upload(self):
 
         self.assertEquals(x.TransfersCompleted, "10")
         self.assertEquals(x.TransfersFailed, "0")
+
+    def test_overwrite_posix_properties(self):
```
adreed-msft, line 839: ❔ I don't really see why this needed the python testing framework-- We could modify the relevant file properties in beforeJobRun & create a differently sized file on the destination to overwrite the properties on; then we could add a hook for beforeValidation to correct our test files to ensure that we did not overwrite the file, but instead just the properties.

ste/sender-blockBlob.go
```diff
@@ -273,6 +273,32 @@ func (s *blockBlobSenderBase) Cleanup() {
 	}
 }
 
+//Currently we've common Metadata Copier across all senders for block blob.
+func (s *blockBlobSenderBase) GenerateCopyMetadata(id common.ChunkID) chunkFunc {
+	return createChunkFunc(true, s.jptm, id, func() {
+		if s.jptm.Info().PreservePOSIXProperties {
```
adreed-msft, line 279: ❓ It feels relevant to implicitly flip PreservePOSIXProperties should a user specify to overwrite POSIX properties.

ste/sender-blockBlob.go
```diff
@@ -273,6 +273,32 @@ func (s *blockBlobSenderBase) Cleanup() {
...
+
+			if unixSIP, ok := s.sip.(IUNIXPropertyBearingSourceInfoProvider); ok {
+				// Clone the metadata before we write to it, we shouldn't be writing to the same metadata as every other blob.
+				s.metadataToApply = common.Metadata(s.metadataToApply).Clone().ToAzBlobMetadata()
```
adreed-msft, line 283: ❗ If we're only going to persist POSIX properties, this would overwrite other unrelated metadata.

adreed-msft (changes requested):

ste/xfer-anyToRemote-file.go
nakulkar-msft: The next line would fail if the file does not exist, it does not upload the file

ste/xfer-anyToRemote-fileProperties.go
nakulkar-msft: Okay. Missed this.

ste/sender-blockBlob.go
nakulkar-msft: The ask was to update all metadata. I think I kind of messed up the pr name.

testSuite/scripts/test_upload_block_blob.py
nakulkar-msft: Put this in the testSuite in cmd

adreed-msft: Not entirely what I meant either. I meant using the `e2etest` suite.

adreed-msft (approved):

Issue #1931, opened 2022-10-18
No error reported downloading non-existent blob


gabe-microsoft:
### Which version of the AzCopy was used? 
10.16.1

### Which platform are you using? (ex: Windows, Mac, Linux)
Linux: Ubuntu 20.04.1 (kernel 5.15.0-1021-azure)

### What command did you run?
Download a non-existent blob (login not required, same behavior is using SAS token):
```bash
azcopy login
azcopy cp https://storageacct.blob.core.windows.net/container/path/to/non/existent/blob .
```

### What problem was encountered?
I would expect AzCopy to report an error and exit with a non-zero code since the requested blob does not exist. However, it returns with a successful (0) exit code with the following output:
```
INFO: Scanning...
INFO: Authenticating to source using Azure AD
INFO: Any empty folders will not be processed, because source and/or destination doesn't have full folder support

Job 0f75537c-c53d-044e-6109-8357cc86191e has started
Log file is located at: /home/azureuser/.azcopy/0f75537c-c53d-044e-6109-8357cc86191e.log

100.0 %, 0 Done, 0 Failed, 0 Pending, 0 Skipped, 0 Total,


Job 0f75537c-c53d-044e-6109-8357cc86191e summary
Elapsed Time (Minutes): 0.0333
Number of File Transfers: 0
Number of Folder Property Transfers: 0
Total Number of Transfers: 0
Number of Transfers Completed: 0
Number of Transfers Failed: 0
Number of Transfers Skipped: 0
TotalBytesTransferred: 0
Final Job Status: Completed
```

gapra-msft:
@tasherif-msft please take a look

tasherif-msft:
Hi @gabe-microsoft I am able to repro this - I will consult with the team regarding this behavior and the next steps as this would be a breaking change. I will keep you posted

marc-hb:
I reproduced something similar in #1975. I think the "to-reproduce" github label should be removed and replaced by "bug".

tasherif-msft:
@marc-hb I have changed the label to improvement. This was something initially done by design - if the file/dir does not exist assume success and show 0 transfers happened. We are now changing the approach to if single file doesn't exist report failure. There's a draft PR atm.

marc-hb:
Do you mean pattern matching / globbing was performed even with a single file argument? I'm just curious.

> There's a draft PR atm.

URL?

tasherif-msft:
[...]

PR #1981, opened 2022-12-06
Fail copy job if single blob does not exist

tasherif-msft:
In this PR we now fail if a targeted blob does not exist for copy

To summarize the behavior:

- if wild card `/dir_doesnt_exist/*` passed on a dir that does not exist - we will pass with 0 transfers **(same behavior)**
- if `/dir_doesnt_exist/ --recursive` - we will pass with 0 transfers **(same behavior)**
- if `/file_or_dir_doesnt_exist --recursive` - we will fail - this is because the customer didn't pass a trailing `/` and therefore the get properties call fails and recursive can indeed be passed when targeting a single file - therefore assume we're targeting a single file and fail **(new behavior)**
- if `/file_doesnt_exist` - fail since file does not exist **(new behavior)**

Resolves https://github.com/Azure/azure-storage-azcopy/issues/1931

### DIFF
--- a/cmd/copyEnumeratorInit_test.go
+++ b/cmd/copyEnumeratorInit_test.go
@@ -41,8 +41,6 @@ func (ce *copyEnumeratorSuite) TestValidateSourceDirThatExists(c *chk.C) {
 	defer deleteContainer(c, containerURL)
 	c.Assert(containerURL, chk.NotNil)
 
-	defer deleteContainer(c, containerURL)
-
 	dirName := "source_dir"
 	createNewDirectoryStub(c, containerURL, dirName)
 	// set up to create blob traverser
@@ -73,8 +71,6 @@ func (ce *copyEnumeratorSuite) TestValidateSourceDirDoesNotExist(c *chk.C) {
 	defer deleteContainer(c, containerURL)
 	c.Assert(containerURL, chk.NotNil)
 
-	defer deleteContainer(c, containerURL)
-
 	dirName := "source_dir/"
 	// set up to create blob traverser
 	ctx := context.WithValue(context.TODO(), ste.ServiceAPIVersionOverride, ste.DefaultServiceApiVersion)
@@ -104,8 +100,6 @@ func (ce *copyEnumeratorSuite) TestValidateSourceFileExists(c *chk.C) {
 	defer deleteContainer(c, containerURL)
 	c.Assert(containerURL, chk.NotNil)
 
-	defer deleteContainer(c, containerURL)
-
 	fileName := "source_file"
 	_, fileName = createNewBlockBlob(c, containerURL, fileName)
 
@@ -130,8 +124,6 @@ func (ce *copyEnumeratorSuite) TestValidateSourceFileDoesNotExist(c *chk.C) {
 	defer deleteContainer(c, containerURL)
 	c.Assert(containerURL, chk.NotNil)
 
-	defer deleteContainer(c, containerURL)
-
 	fileName := "source_file"
 
 	ctx := context.WithValue(context.TODO(), ste.ServiceAPIVersionOverride, ste.DefaultServiceApiVersion)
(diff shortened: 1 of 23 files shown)


cmd/zc_traverser_gcp.go
```diff
@@ -24,20 +24,20 @@ type gcpTraverser struct {
...
 	isDirDirect := !t.gcpURLParts.IsObjectSyntactically() && (t.gcpURLParts.IsDirectorySyntactically() || t.gcpURLParts.IsBucketSyntactically())
 	if !isSource {
-		return isDirDirect
+		return isDirDirect, nil
 	}
 	bkt := t.gcpClient.Bucket(t.gcpURLParts.BucketName)
 	obj := bkt.Object(t.gcpURLParts.ObjectKey)
 	//Directories do not have attributes and hence throw error
 	_, err := obj.Attrs(t.ctx)
 	if err == gcpUtils.ErrObjectNotExist {
```
vibhansa-msft, line 37: For any other error then Not_Exist we shall be returning error here.

cmd/copyEnumeratorInit_test.go
```diff
@@ -0,0 +1,174 @@
...
+	blobTraverser := newBlobTraverser(&rawBlobURLWithSAS, p, ctx, true, true, func(common.EntityType) {}, false, common.CpkOptions{}, false, false, false)
+
+	cca := CookedCopyCmdArgs{StripTopDir: false, Recursive: false}
+	err := cca.validateSourceDir(blobTraverser)
+	c.Assert(err.Error(), chk.Equals, common.FILE_NOT_FOUND)
+	c.Assert(cca.IsSourceDir, chk.Equals, false)
+}
+
+func (ce *copyEnumeratorSuite) TestValidateSourceWithWildCard(c *chk.C) {
+	bsu := getBSU()
```
vibhansa-msft, line 151: Din't get which part is putting WildCard Filter here.

cmd/zc_traverser_blob.go
```diff
@@ -95,15 +97,21 @@ func (t *blobTraverser) IsDirectory(isSource bool) bool {
...
 			azcopyScanningLogger.Log(pipeline.LogError, msg)
 		}
-		return false
+		return false, nil
 	}
 
 	if len(resp.Segment.BlobItems) == 0 {
 		//Not a directory
-		return false
+		if stgErr, ok := blobErr.(azblob.StorageError); ok {
```
gapra-msft, line 105: Is this a valid error for list blobs to throw? I thought it returns no error if file doesnt exist

cmd/zc_traverser_blob.go
```diff
@@ -95,15 +97,21 @@ func (t *blobTraverser) IsDirectory(isSource bool) bool {
 			msg := fmt.Sprintf("Failed to check if the destination is a folder or a file (Azure Files). Assuming the destination is a file: %s", err)
 			azcopyScanningLogger.Log(pipeline.LogError, msg)
 		}
-		return false
+		return false, nil
 	}
 
 	if len(resp.Segment.BlobItems) == 0 {
```
gapra-msft, line 103: Can we add a test for IsDirectory that tests a directory that only has virtual directories in it?

Example: Create a file called mydir/subdir1/file.txt and mydir/subdir2/file.txt and call isDirectory on mydir

cmd/copyEnumeratorInit_test.go
tasherif-msft: StripTopDir is what's set on a wildcard

cmd/zc_traverser_gcp.go
tasherif-msft: We can apply error related changes of other platforms in a separate PR. The IsDirectory changes and error handling in this PR primarily applies to blobs.

cmd/zc_traverser_blob.go
tasherif-msft: that's why we're checking the blobErr, if you scroll up you'll see that this is retrieved on the getProps call. If nothing was listed then we check if blobErr had a not exist - this would confirm that 1. the hdi file doesn't exist AND nothing with this dir prefix exists so throw.

cmd/zc_traverser_blob.go
tasherif-msft: should fall under the NoStub test

siminsavani-msft (approved):

vibhansa-msft (approved):

PR #1983, opened 2022-12-06
darwin arm64

tasherif-msft:


gapra-msft (approved):

siminsavani-msft (approved):

azure-pipelines.yml
```diff
@@ -83,7 +83,12 @@ jobs:
 
       - script: |
           go build -o "$(Build.ArtifactStagingDirectory)/azcopy_darwin_amd64"
-        displayName: 'Generate MacOS Build'
+        displayName: 'Generate MacOS Build with AMD64'
+        condition: eq(variables.type, 'mac-os')
+
+      - script: |
+          GOARCH=arm64 CGO_ENABLED=1 go build -o "$(Build.ArtifactStagingDirectory)/azcopy_darwin_arm64"
```
vibhansa-msft, line 90: We can not enable CGO for cross compilation as that brings a glibc dependency on the package and machine not having similar version will fail to run it.

vibhansa-msft (approved):

azure-pipelines.yml
```diff
@@ -186,7 +191,31 @@ jobs:
...
+        displayName: 'E2E Test MacOs AMD64'
+        condition: eq(variables.type, 'mac-os')
+
+      # Running E2E Tests on Mac
+      - script: |
+          set -e
+          go build -o azcopy_darwin_arm64
+          echo 'starting E2E tests on mac-os'
+          export AZCOPY_E2E_EXECUTABLE_PATH=$(pwd)/azcopy_darwin_arm64
+          go test -timeout 60m -race -cover -v ./e2etest
```
nakulkar-msft, line 203: I assume the devops pipeline does not yet support Mac M1 os, Are we testing M1 build here?

tasherif-msft: yes

tasherif-msft: yea

Issue #1984, opened 2022-12-08
Token refresh failures on large jobs


tasherif-msft:
To repro the issue:
Use device login to transfer between two storage accounts. Create a very large transfer (with many files). Eventually azcopy hit many "client request loop" when refreshing the token. It seems token refreshes are happening too frequently.

Workaround:
Using smaller jobs or SAS on source.

Issue #1957, opened 2022-11-09
azcopy remove with --list-of-files option throws errors


ngorur:
### What command did you run?
azcopy.exe login --aad-endpoint https://login.microsoftonline.us
azcopy remove "container" --recursive=false --list-of-files="FileNameListforDeletion.txt" --delete-snapshots "include"
### What problem was encountered?
Part#=0 failed to refresh token, please check error details and try to log in with azcopy login command again. (Error details: adal: Refresh request failed. Status Code = '400'. Response body: {"error":"invalid_grant","error_description":"AADSTS50196: The server terminated an operation because it encountered a client request loop. Please contact your app vendor.\r\nTrace ID: redacted\r\nCorrelation ID: redacted\r\nTimestamp: 2022-11-09 20:52:49Z","error_codes":[50196],"timestamp":"2022-11-09 20:52:49Z","trace_id":"redacted","correlation_id":"redacted","error_uri":"https://login.microsoftonline.us/error?code=50196"} Endpoint https://login.microsoftonline.us/common/oauth2/token?api-version=1.0)

### How can we reproduce the problem in the simplest way?
run the above command, ensure the FileNameListforDeletion.txt has at least 20 filenames

adreed-msft:
I cannot reproduce this.

Do other commands work over OAuth for you? Or is it *only* delete that fails?

tasherif-msft:
Hi @ngorur we're working on a fix at the moment! will update you once a PR is out

ngorur:
Thanks

Nanda Gorur
Senior Consultant | Azure Cloud and AI | Microsoft Federal
Work: 301 771 8294 | Mobile: 201 213 8364 | ***@***.******@***.***>



________________________________

PR #1986, opened 2022-12-09
OAuth token refreshes bug fix + optimization

tasherif-msft:
Currently any large job using OAuth will eventually hit a service error we're refreshes are happening too frequently, due to the nature of jobs concurrently running in parts, the token is refreshed too many times rather than reusing the already created fresh token. This PR fixes that - it also led to better perf due to the lower number of service calls.

Resolves https://github.com/Azure/azure-storage-azcopy/issues/1984
Resolves https://github.com/Azure/azure-storage-azcopy/issues/1957

### DIFF
--- a/ste/mgr-JobMgr.go
+++ b/ste/mgr-JobMgr.go
@@ -964,10 +964,7 @@ func (jm *jobMgr) scheduleJobParts() {
 				go jm.poolSizer()
 				startedPoolSizer = true
 			}
-			getCredential := true
-			if jm.credential != nil {
-				getCredential = false
-			}
+			getCredential := jm.credential == nil
 			jobPart.ScheduleTransfers(jm.Context(), getCredential, jm.credential)
 			if getCredential {
 				jm.credential = jobPart.SourceCredential()
--- a/ste/mgr-JobMgr.go
+++ b/ste/mgr-JobMgr.go
@@ -964,9 +964,9 @@ func (jm *jobMgr) scheduleJobParts() {
 				go jm.poolSizer()
 				startedPoolSizer = true
 			}
-			getCredential := jm.credential == nil
-			jobPart.ScheduleTransfers(jm.Context(), getCredential, jm.credential)
-			if getCredential {
+			getSourceCredential := jm.credential == nil
+			jobPart.ScheduleTransfers(jm.Context(), getSourceCredential, jm.credential)
+			if getSourceCredential {
 				jm.credential = jobPart.SourceCredential()
 			}
 		}
--- a/ste/mgr-JobMgr.go
+++ b/ste/mgr-JobMgr.go
@@ -964,9 +964,8 @@ func (jm *jobMgr) scheduleJobParts() {
 				go jm.poolSizer()
 				startedPoolSizer = true
 			}
-			getSourceCredential := jm.credential == nil
-			jobPart.ScheduleTransfers(jm.Context(), getSourceCredential, jm.credential)
-			if getSourceCredential {
+			jobPart.ScheduleTransfers(jm.Context(), jm.credential)
+			if jm.credential == nil {
 				jm.credential = jobPart.SourceCredential()
 			}
 		}
--- a/ste/mgr-JobMgr.go
+++ b/ste/mgr-JobMgr.go
@@ -338,7 +338,8 @@ type jobMgr struct {
 	fileCountLimiter    common.CacheLimiter
 	jstm                *jobStatusManager
 
-	isDaemon bool /* is it running as service */
+	isDaemon   bool /* is it running as service */
+	credential pipeline.Factory
 }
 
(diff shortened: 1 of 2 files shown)


ste/mgr-JobMgr.go
```diff
@@ -712,13 +713,15 @@ func (jm *jobMgr) CloseLog() {
 
 // DeferredCleanupJobMgr cleanup all the jobMgr resources.
 // Warning: DeferredCleanupJobMgr should be called from JobMgrCleanup().
-//          As this function neither threadsafe nor idempotient. So if DeferredCleanupJobMgr called
-//          mulitple times, it may stuck as receiving channel already closed. Where as JobMgrCleanup()
-//          safe in that sense it will do the cleanup only once.
+//
```
gapra-msft, line 716: nit Why are there new lines of // added here?

ste/mgr-JobMgr.go
```diff
@@ -961,7 +964,14 @@ func (jm *jobMgr) scheduleJobParts() {
 				go jm.poolSizer()
 				startedPoolSizer = true
 			}
-			jobPart.ScheduleTransfers(jm.Context())
+			getCredential := true
```
gapra-msft, line 967: nit getCredential := jm.credential == nil

ste/mgr-JobMgr.go
```diff
@@ -961,7 +964,14 @@ func (jm *jobMgr) scheduleJobParts() {
...
+			if jm.credential != nil {
+				getCredential = false
+			}
+			jobPart.ScheduleTransfers(jm.Context(), getCredential, jm.credential)
```
gapra-msft, line 971: I'm confused here, shouldnt we get the credential before scheduling transfers?

tasherif-msft: for minimal code changes in the first job part (and its corresponding scheduleTransfers call) we store that credential retrieved when the pipeline is created which is then used for other job parts.

gapra-msft: So basically JobMgr and JobParts will have separate creds?

tasherif-msft: JobMgr will store the cred after the first jobPart does its first scheduleTransfers call and then pass it to the next jobParts

ste/mgr-JobMgr.go
gapra-msft, line 967: also nit: should we call this getSourceCredential? I know in AzCopy they seem to be very explicit about source vs dest names

ste/mgr-JobMgr.go
tasherif-msft: It looks like GO is formatting this automatically

ste/mgr-JobMgr.go
```diff
@@ -961,7 +964,11 @@ func (jm *jobMgr) scheduleJobParts() {
 				go jm.poolSizer()
 				startedPoolSizer = true
 			}
-			jobPart.ScheduleTransfers(jm.Context())
+			getSourceCredential := jm.credential == nil
+			jobPart.ScheduleTransfers(jm.Context(), getSourceCredential, jm.credential)
```
vibhansa-msft, line 968: getSourceCredential is just a flag saying jm.credential is nil or not. Here we are passing down both of parameters here. Flag can be ignored and directly jm.credential can be checked internally?

ste/mgr-JobPartMgr.go
```diff
@@ -561,9 +561,13 @@ func (jpm *jobPartMgr) createPipelines(ctx context.Context) {
 				CallerID: fmt.Sprintf("JobID=%v, Part#=%d", jpm.Plan().JobID, jpm.Plan().PartNum),
 				Cancel:   jpm.jobMgr.Cancel,
 			}
-
-			sourceCred = common.CreateBlobCredential(ctx, jobState.CredentialInfo.WithType(jobState.S2SSourceCredentialType), credOption)
```
vibhansa-msft, line 565: So, actual fix here is to reuse the same credential instead of creating a new one for each pipeline?

vibhansa-msft:
How are we fixing the refreshToken? If multiple goroutine detect token is expired they can still go in parallel to refresh the token,

ste/mgr-JobPartMgr.go
tasherif-msft: yes

tasherif-msft:
> How are we fixing the refreshToken? If multiple goroutine detect token is expired they can still go in parallel to refresh the token,

The issue that causes the job to fail isn't that multiple goroutines are hitting refresh simultaneously, its simply that each job is refreshing the token. When I ran this multiple times on a large job it never hit really hit the case you're referencing. So, I decided to keep changes to a minimum and focus on fixing reusing the token. I'd rather not overengineer here.

ste/mgr-JobMgr.go
tasherif-msft: done

Issue #1553, opened 2021-09-14
azcopy remove fails without read permission for SAS token


DerekHerman-MSFT:
### Which version of the AzCopy was used? 
10.12.1

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows and Storage Explorer
### What command did you run?
`azcopy.exe rm "https://dehermanperms.blob.core.windows.net/testperms/googleleads.JPG?sv=2020-04-08&st=2021-09-14T22%3A35%3A28Z&se=2021-09-15T22%3A35%3A28Z&sr=c&sp=wdl&sig=REDACTED" --recursive --trusted-microsoft-suffixes= --log-level=DEBUG`

### What problem was encountered?
The file was not deleted. See log info:

```
2021/09/14 22:54:59 Job-Command rm https://dehermanperms.blob.core.windows.net/testperms/googleleads.JPG?se=2021-09-15t22%3A35%3A28z&sig=-REDACTED-&sp=wdl&sr=c&st=2021-09-14t22%3A35%3A28z&sv=2020-04-08 --recursive --trusted-microsoft-suffixes= --log-level=DEBUG 
2021/09/14 22:54:59 Number of CPUs: 8
2021/09/14 22:54:59 Max file buffer RAM 4.000 GB
2021/09/14 22:54:59 Max concurrent network operations: 128 (Based on number of CPUs. Set AZCOPY_CONCURRENCY_VALUE environment variable to override)
2021/09/14 22:54:59 Check CPU usage when dynamically tuning concurrency: true (Based on hard-coded default. Set AZCOPY_TUNE_TO_CPU environment variable to true or false override)
2021/09/14 22:54:59 Max concurrent transfer initiation routines: 64 (Based on hard-coded default. Set AZCOPY_CONCURRENT_FILES environment variable to override)
2021/09/14 22:54:59 Max enumeration routines: 16 (Based on hard-coded default. Set AZCOPY_CONCURRENT_SCAN environment variable to override)
2021/09/14 22:54:59 Parallelize getting file properties (file.Stat): false (Based on AZCOPY_PARALLEL_STAT_FILES environment variable)
2021/09/14 22:54:59 Max open files when downloading: 2147483119 (auto-computed)
2021/09/14 22:54:59 ISO 8601 START TIME: to copy files that changed before or after this job started, use the parameter --include-before=2021-09-14T22:54:54Z or --include-after=2021-09-14T22:54:54Z
[...]

PR #1987, opened 2022-12-09
Allow remove to work without read permissions on SAS

gapra-msft:
Should resolve https://github.com/Azure/azure-storage-azcopy/issues/1553

Please take a look. This issue likely also affects the other services, so if this change is ok, I can apply it to others as well.

WIP. Still need to add some more tests.

### DIFF
--- a/cmd/zc_enumerator.go
+++ b/cmd/zc_enumerator.go
@@ -228,6 +228,10 @@ type blobPropsProvider interface {
 	ArchiveStatus() azblob.ArchiveStatusType
 }
 
+func newRemoveStoredObject(morpher objectMorpher, name string, relativePath string, entityType common.EntityType, containerName string) StoredObject {
+	return newStoredObject(morpher, name, relativePath, entityType, time.Now(), 0, noContentProps, noBlobProps, noMetdata, containerName)
+}
+
 // a constructor is used so that in case the StoredObject has to change, the callers would get a compilation error
 // and it forces all necessary properties to be always supplied and not forgotten
 func newStoredObject(morpher objectMorpher, name string, relativePath string, entityType common.EntityType, lmt time.Time, size int64, props contentPropsProvider, blobProps blobPropsProvider, meta common.Metadata, containerName string) StoredObject {
@@ -443,7 +447,7 @@ func InitResourceTraverser(resource common.ResourceString, location common.Locat
 		} else if listOfVersionIds != nil {
 			output = newBlobVersionsTraverser(resourceURL, *p, *ctx, recursive, includeDirectoryStubs, incrementEnumerationCounter, listOfVersionIds, cpkOptions)
 		} else {
-			output = newBlobTraverser(resourceURL, *p, *ctx, recursive, includeDirectoryStubs, incrementEnumerationCounter, s2sPreserveBlobTags, cpkOptions, includeDeleted, includeSnapshot, includeVersion)
+			output = newBlobTraverser(resourceURL, *p, *ctx, recursive, includeDirectoryStubs, incrementEnumerationCounter, s2sPreserveBlobTags, cpkOptions, includeDeleted, includeSnapshot, includeVersion, getProperties)
 		}
 	case common.ELocation.File():
 		resourceURL, err := resource.FullURL()
--- a/cmd/zc_traverser_blob.go
+++ b/cmd/zc_traverser_blob.go
@@ -270,11 +270,6 @@ func (t *blobTraverser) Traverse(preprocessor objectMorpher, processor objectPro
 		searchPrefix += common.AZCOPY_PATH_SEPARATOR_STRING
 	}
 
-	//// if getProperties=false, we need to include the root item in the listing
-	//if !t.getProperties {
-	//	searchPrefix = strings.TrimSuffix(searchPrefix, common.AZCOPY_PATH_SEPARATOR_STRING)
-	//}
-
 	// as a performance optimization, get an extra prefix to do pre-filtering. It's typically the start portion of a blob name.
 	extraSearchPrefix := FilterSet(filters).GetEnumerationPreFilter(t.recursive)
 
--- a/cmd/zc_traverser_blob.go
+++ b/cmd/zc_traverser_blob.go
@@ -61,6 +61,8 @@ type blobTraverser struct {
 	includeSnapshot bool
 
 	includeVersion bool
+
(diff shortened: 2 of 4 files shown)


cmd/zc_traverser_blob.go
```diff
@@ -162,18 +164,33 @@ func (t *blobTraverser) Traverse(preprocessor objectMorpher, processor objectPro
...
 		if stgErr.ServiceCode() == common.CPK_ERROR_SERVICE_CODE {
-			return errors.New("this blob uses customer provided encryption keys (CPK). At the moment, AzCopy does not support CPK-encrypted blobs. " +
-				"If you wish to make use of this blob, we recommend using one of the Azure Storage SDKs")
+			if t.getProperties {
+				return errors.New("this blob uses customer provided encryption keys (CPK). At the moment, AzCopy does not support CPK-encrypted blobs. " +
+					"If you wish to make use of this blob, we recommend using one of the Azure Storage SDKs")
+			} else {
+				blobProperties = &azblob.BlobGetPropertiesResponse{}
+				isBlob = true
+				isDirStub = true
```
gapra-msft, line 177: the only problem here is we have to assume root is a directory to allow further recursive listing.

gapra-msft: We could read the recursive flag and if that's false assume a file though

cmd/zc_traverser_blob.go
```diff
@@ -162,18 +164,33 @@ func (t *blobTraverser) Traverse(preprocessor objectMorpher, processor objectPro
...
+				propErr = nil
+			}
 		}
 
 		if resp := stgErr.Response(); resp == nil {
 			return fmt.Errorf("cannot list files due to reason %s", stgErr)
 		} else {
 			if resp.StatusCode == 403 { // Some nature of auth error-- Whatever the user is pointing at, they don't have access to, regardless of whether it's a file or a dir stub.
-				return fmt.Errorf("cannot list files due to reason %s", stgErr)
+				if t.getProperties {
```
vibhansa-msft, line 186: Here error is not related to CPK and might be some other auth error, do we need to return dummy properties here?

cmd/zc_enumerator.go
```diff
@@ -443,7 +447,7 @@ func InitResourceTraverser(resource common.ResourceString, location common.Locat
 		} else if listOfVersionIds != nil {
 			output = newBlobVersionsTraverser(resourceURL, *p, *ctx, recursive, includeDirectoryStubs, incrementEnumerationCounter, listOfVersionIds, cpkOptions)
 		} else {
-			output = newBlobTraverser(resourceURL, *p, *ctx, recursive, includeDirectoryStubs, incrementEnumerationCounter, s2sPreserveBlobTags, cpkOptions, includeDeleted, includeSnapshot, includeVersion)
+			output = newBlobTraverser(resourceURL, *p, *ctx, recursive, includeDirectoryStubs, incrementEnumerationCounter, s2sPreserveBlobTags, cpkOptions, includeDeleted, includeSnapshot, includeVersion, getProperties)
```
vibhansa-msft, line 450: This method appears to be growing in terms of flags, shall we think to put option bag here instead of passing down so many parameters?

cmd/zc_traverser_blob.go
gapra-msft: We could just skip the getProperty check and blindly add the empty blob properties.

gapra-msft: Also this method is actually Traverse not IsDirectory. I believe IsDirectory is only called for copy and sync

cmd/zc_enumerator.go
gapra-msft: Yes I was thinking the same, not sure if this is a design decision on Azcopy's end or if there hasn't been time to refactor this. I can do it if the team is on board for it.

tasherif-msft: I'm on board with using an options bag as well.

tasherif-msft:
Please mention that this issue is caused by a get properties call which requires read in the SAS permissions - just for future context

cmd/zc_traverser_blob.go
tasherif-msft: I'm still confused on why we don't throw on CPK error immediately in the case of !t.getProperties?

gapra-msft: we don't need CPK to delete a file, so even if we failed to get properties we can still move on and pretend like everything is ok 
https://learn.microsoft.com/en-us/rest/api/storageservices/delete-blob

tasherif-msft: Ah ok I actually thing that would be very useful to add in as an inline comment

cmd/zc_enumerator.go
nakulkar-msft: Option bag would be more clean. We've had the list growing forever.

cmd/zc_traverser_blob.go
```diff
@@ -162,18 +164,33 @@ func (t *blobTraverser) Traverse(preprocessor objectMorpher, processor objectPro
...
 		// If it's a CPK error, we know it's a single blob and that we can't get the properties on it anyway.
 		if stgErr.ServiceCode() == common.CPK_ERROR_SERVICE_CODE {
-			return errors.New("this blob uses customer provided encryption keys (CPK). At the moment, AzCopy does not support CPK-encrypted blobs. " +
-				"If you wish to make use of this blob, we recommend using one of the Azure Storage SDKs")
+			if t.getProperties {
+				return errors.New("this blob uses customer provided encryption keys (CPK). At the moment, AzCopy does not support CPK-encrypted blobs. " +
+					"If you wish to make use of this blob, we recommend using one of the Azure Storage SDKs")
+			} else {
+				blobProperties = &azblob.BlobGetPropertiesResponse{}
+				isBlob = true
```
nakulkar-msft, line 176: Both being true usually conflicts enumerator's decisions. Have we checked the impact? Otherwise, we could pass rename the getProperties flag to isRemoveEnumerator and check only for listing.

gapra-msft: Why would it conflict with the enumerators decisions? What would azcopy do if you had a marker file where the directory exists

cmd/zc_enumerator.go
siminsavani-msft: I like the options bag idea too!

PR #1996, opened 2022-12-17
Store source token and pass to other threads

tasherif-msft:
Resolves https://github.com/Azure/azure-storage-azcopy/issues/1984
Resolves https://github.com/Azure/azure-storage-azcopy/issues/1957

Changes here apply only to copy operation and source OAuth tokens for blobs. Sync will be in a separate PR due to the scale of changes and that most people are hitting this issue on the copy operation.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -26,7 +26,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/Azure/azure-storage-azcopy/v10/jobsAdmin"
 	"io"
 	"math"
 	"net/url"
@@ -35,6 +34,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/Azure/azure-storage-azcopy/v10/jobsAdmin"
 
 	"github.com/Azure/azure-pipeline-go/pipeline"
 
@@ -945,8 +946,8 @@ func areBothLocationsSMBAware(fromTo common.FromTo) bool {
 func areBothLocationsPOSIXAware(fromTo common.FromTo) bool {
 	// POSIX properties are stored in blob metadata-- They don't need a special persistence strategy for BlobBlob.
 	return runtime.GOOS == "linux" && (
-	// fromTo == common.EFromTo.BlobLocal() || TODO
-	fromTo == common.EFromTo.LocalBlob()) ||
+		// fromTo == common.EFromTo.BlobLocal() || TODO
+		fromTo == common.EFromTo.LocalBlob()) ||
 		fromTo == common.EFromTo.BlobBlob()
 }
 
@@ -1113,7 +1114,9 @@ type CookedCopyCmdArgs struct {
 	FollowSymlinks     bool
 	ForceWrite         common.OverwriteOption // says whether we should try to overwrite
 	ForceIfReadOnly    bool                   // says whether we should _force_ any overwrites (triggered by forceWrite) to work on Azure Files objects that are set to read-only
-	autoDecompress     bool
+	IsSourceDir        bool
+
+	autoDecompress bool
 
 	// options from flags
 	blockSize int64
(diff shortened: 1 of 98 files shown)


adreed-msft:
I get deja vu from this PR... I swear I fixed this before, maybe in another PR. Lemme take a look around.

cmd/copy.go
```diff
@@ -1377,6 +1377,51 @@ func (cca *CookedCopyCmdArgs) processRedirectionUpload(blobResource common.Resou
...
+			uotm := GetUserOAuthTokenManagerInstance()
+			// get token from env var or cache
+			if tokenInfo, err := uotm.GetTokenInfo(ctx); err != nil {
+				return srcCredInfo, err
+			} else {
+				cca.credentialInfo.OAuthTokenInfo = *tokenInfo
+				jpo.CredentialInfo.OAuthTokenInfo = *tokenInfo
+			}
+			// if the source is not local then store the credential token if it was OAuth to avoid constant refreshing
+			if cca.FromTo != common.EFromTo.LocalBlob() && cca.FromTo != common.EFromTo.LocalFile() && cca.FromTo != common.EFromTo.LocalBlobFS() {
```
gapra-msft, line 1412: is it possible for it to be S2S and also be LocalBlob/LocalFile/LocalBlobFs?

cmd/copyEnumeratorInit.go
```diff
@@ -28,41 +28,9 @@ type BucketToContainerNameResolver interface {
 	ResolveName(bucketName string) (string, error)
 }
 
-func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrderRequest, ctx context.Context) (*CopyEnumerator, error) {
+func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrderRequest, srcCredInfo common.CredentialInfo, ctx context.Context) (*CopyEnumerator, error) {
```
gapra-msft, line 31: Why is it better to move the credential generation outside?

cmd/copy.go
```diff
@@ -1489,11 +1534,12 @@ func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 		common.EFromTo.BenchmarkFile():
 
 		var e *CopyEnumerator
-		e, err = cca.initEnumerator(jobPartOrder, ctx)
+		srcCredInfo, _ := cca.getSrcCredential(ctx, &jobPartOrder)
```
vibhansa-msft, line 1537: Do we have logs when token refresh is hit so that in future if we see multiple threads doing it in parallel can be debugged from logs itself.

vibhansa-msft (approved):

cmd/copyEnumeratorInit.go
tasherif-msft: makes the code more maintainable to have it as individual method

cmd/copy.go
adreed-msft: No.

cmd/copy.go
```diff
@@ -1377,6 +1377,51 @@ func (cca *CookedCopyCmdArgs) processRedirectionUpload(blobResource common.Resou
...
+			// get token from env var or cache
+			if tokenInfo, err := uotm.GetTokenInfo(ctx); err != nil {
+				return srcCredInfo, err
+			} else {
+				cca.credentialInfo.OAuthTokenInfo = *tokenInfo
+				jpo.CredentialInfo.OAuthTokenInfo = *tokenInfo
+			}
+			// if the source is not local then store the credential token if it was OAuth to avoid constant refreshing
+			if cca.FromTo != common.EFromTo.LocalBlob() && cca.FromTo != common.EFromTo.LocalFile() && cca.FromTo != common.EFromTo.LocalBlobFS() {
+				jpo.CredentialInfo.SourceBlobToken = common.CreateBlobCredential(ctx, srcCredInfo, common.CredentialOpOptions{
```
adreed-msft, line 1413: Initially, I didn't like this, but it's grown on me a lot-- Perhaps we should do this with the destination token(s) as well?

adreed-msft (approved):

cmd/copy.go
tasherif-msft: I'm silly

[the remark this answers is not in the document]
tasherif-msft: that's unnecessary anyway since we're within an S2S check

cmd/copy.go
tasherif-msft: yea I added a TODO in there for dest - should be in a separate PR

PR #1998, opened 2022-12-20
[WIP] Fixed issue where blob enumeration required read permissions for remove

gapra-msft:
This should only be merged after migration to latest service version support for Blob SDK. Swagger needs to support returning properties with virtual directories/directories


cmd/zt_test.go
```diff
@@ -214,8 +214,12 @@ func getFilesystemURL(c *chk.C, bfssu azbfs.ServiceURL) (filesystem azbfs.FileSy
 	return
 }
 
-func getBlockBlobURL(c *chk.C, container azblob.ContainerURL, prefix string) (blob azblob.BlockBlobURL, name string) {
-	name = prefix + generateBlobName()
+func getBlockBlobURL(c *chk.C, container azblob.ContainerURL, prefix string, isPrefix bool) (blob azblob.BlockBlobURL, name string) {
```
gapra-msft, line 217: TODO: Undo this

PR #1999, opened 2022-12-20
Added log indicating a sub-directory is being enqueued

gapra-msft:


siminsavani-msft (approved):

nakulkar-msft (approved):
I might be worthwhile to add this on the CLI as well using glcm.Info. The user would not have to monitor the log file explicitly.

tasherif-msft (approved):

vibhansa-msft (approved):

gapra-msft:
> I might be worthwhile to add this on the CLI as well using glcm.Info. The user would not have to monitor the log file explicitly.

Won't that be a lot of logs to CLI?

PR #2006, opened 2023-01-03
Perf testing for AzCopy

nakulkar-msft:
https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=15328&view=results

### DIFF
--- a/perf-test.yaml
+++ b/perf-test.yaml
@@ -9,7 +9,7 @@ stages:
       strategy:
         matrix:
           Ubuntu-22:
-            imageName: "azcopyPerfUbuntu22"
+            imageName: "azcopyPerfTestUbuntu22.04"
             Description: "AzCopy Perf Test"
 
       pool:
--- a/perf-test.yaml
+++ b/perf-test.yaml
@@ -5,7 +5,7 @@ stages:
 - stage: Smallfiles
   jobs:
     - job: PerformanceTest
-      timeoutInMinutes: 360
+      timeoutInMinutes: 720
       strategy:
         matrix:
           Ubuntu-22:
@@ -19,13 +19,15 @@ stages:
 
       variables:
       - group: AzCopyPerfTestTargets
+      - name: localPath
+        value: "/mnt/storage"
 
       steps:
         - script: |
             echo $(Description)
             hostnamectl
           displayName: 'Print Agent Info'
-
+        
         - task: GoTool@0
           inputs:
             version: '1.19.3'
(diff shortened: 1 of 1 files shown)


adreed-msft:
This is pretty minimal and may produce unexpected results at times-- We should consider utilizing our `bench` command John Rusk lovingly built for us as a better measure of performance, and consider what different conditions could greatly affect performance to try and control the results better. On any given day, the ADO environment may come with an entirely different set of conditions (better CPU, worse CPU, congested/decongested network, etc.) that would give a flawed result.

Effectively, what I'm saying is, we need a controlled environment, we need a wider variety of scenarios to test against, and in theory, we need to prove through repeatability that these results aren't just an anomaly. I don't believe what's in this PR is remotely sufficient to consider as proper perf testing.

adreed-msft (changes requested):
GitHub says I have to leave a comment to finish a review

nakulkar-msft:
> This is pretty minimal and may produce unexpected results at times-- We should consider utilizing our `bench` command John Rusk lovingly built for us as a better measure of performance, and consider what different conditions could greatly affect performance to try and control the results better. On any given day, the ADO environment may come with an entirely different set of conditions (better CPU, worse CPU, congested/decongested network, etc.) that would give a flawed result.
> 
> Effectively, what I'm saying is, we need a controlled environment, we need a wider variety of scenarios to test against, and in theory, we need to prove through repeatability that these results aren't just an anomaly. I don't believe what's in this PR is remotely sufficient to consider as proper perf testing.

This is pretty minimal, for starters. We will build over this. I've tried to reduce the variables in the environment as much as possible. Both the storage accounts are in same region, so we are not impacted by n/w at large. The VM is not an ADO container but Azure vm of fixed SKU. The data is same as well. Service would be unreliable always, and we cannot work around it.

The test results are pretty close, and repeatability is good. Benchmark is here for its own reasons, but not for perf tests. Bench particularly is not reliable in high performance n/w, and is generally for scenarios where we cannot determine config parameters like concurrency for AzCopy. In particular, these tests are to determine what AzCopy is capable of and not test n/w, disk or vm. Azure generally provides good stability for the limits advertised on disk or vm.

perf-test.yaml
```diff
@@ -0,0 +1,71 @@
...
+            azcopy --version
+          displayName: 'Build Azcopy'
+          env:
+            AZCOPY_AUTO_LOGIN_TYPE: $(AZCOPY_AUTO_LOGIN_TYPE)
+            AZCOPY_MSI_CLIENT_ID: $(AZCOPY_MSI_CLIENT_ID)
+            AZCOPY_CONCURRENCY_VALUE: "256"
+            AZCOPY_SHOW_PERF_STATES: "1"
+
+        - script: |
+            time azcopy copy $(Blob2BlobLargeFilesSrc) $(Blob2BlobLargeFilesDst) --recursive --block-size-mb=128 --log-level=ERROR --cap-mbps=40000
```
vibhansa-msft, line 44: Can we parse the output of time and put it in some file for all three test cases and then upload this file in our artifacts.

nakulkar-msft: I'm looking for a better way to do it

gapra-msft, line 44: In Blobfuse2, we also created perf scripts so then we can choose to run them manually also, would that be helpful here?

nakulkar-msft: Can you link an example?

gapra-msft: https://github.com/Azure/azure-storage-fuse/blob/main/test/scripts/fio.sh

adreed-msft (approved):

perf-test.yaml
tasherif-msft: you can refer to blobfuse2 for doing this as well

vibhansa-msft (approved):

PR #2009, opened 2023-01-03
Added CodeQL

gapra-msft:


adreed-msft (approved):

gapra-msft:
Please take a close look at the GCP URL Parts change

common/gcpURLParts.go
```diff
@@ -17,13 +17,13 @@ type GCPURLParts struct {
 	UnparsedParams string
 }
 
-const gcpHostPattern = "^storage.cloud.google.com"
+const gcpHostPattern = "^storage\\.cloud\\.google\\.com"
```
vibhansa-msft, line 20: More then codeQL this appears to be a bug in regex filter. How is this thing working in current code?

vibhansa-msft (approved):

common/gcpURLParts.go
gapra-msft: It passes more than necessary through, so this regex will pass normal google hosts and more.

PR #2015, opened 2023-01-06
Added trivy file

gapra-msft:


vibhansa-msft (approved):

adreed-msft:
Would Github's Dependabot not work as well?
https://docs.github.com/en/code-security/dependabot/working-with-dependabot

gapra-msft:
> Would Github's Dependabot not work as well? https://docs.github.com/en/code-security/dependabot/working-with-dependabot

I can look into this. We just happen to use trivy in Blobfuse so I just integrated that here.

adreed-msft:
> I can look into this. We just happen to use trivy in Blobfuse so I just integrated that here.

Gotcha, gotcha.

gapra-msft:
> > I can look into this. We just happen to use trivy in Blobfuse so I just integrated that here.
> 
> Gotcha, gotcha.

I'll follow this PR up with another one with dependabot since it seems to have some cool features like creating PRs for you.

PR #2020, opened 2023-01-10
Hash based sync

adreed-msft:
This PR provides a set of options for AzCopy sync since LMT is not inherently accurate on any particular service. On Blob, LMT is uncontrollable. On File, there is a non-controllable LMT and a controllable LMT, but both come with their caveats.

Instead, since Sync is used as a unidirectional mirror mechanism, rather than a typical actual bidirectional sync, we can utilize the hashes available from the service, and cache the hashes locally. When a user is syncing with the `--compare-hash=MD5` flag, AzCopy will compare available hashes from the source against the destination; any non-matching hashes are treated as outdated on the destination. In addition, to facilitate the existence of local hashes, `--missing-hash-policy=Generate` will allow AzCopy to hash local files when no cached hash is present. If a cached hash is present, it will be used provided it matches the hash type in use, and the hash is not invalidated by LMT.

Local hash data is cached alongside the target files, on Windows, in an alternate data stream (`<filename>:.azcopysyncmeta`), and on *NIX based systems, a hidden file named `.<filename>.azcopysyncmeta`. It stores the hash, the type of hash, and the LMT at which the hash was taken. The hash must be re-computed, or overwritten at download time if any fail to match.

I am... more than certain this PR will attract a good few comments given the implementation on the local traverser side, and especially the policy applied regarding staleness of files, and I welcome it.


common/hash_data_other.go
```diff
@@ -0,0 +1,70 @@
...
+	// get meta file name
+	dir, fn := filepath.Split(fullpath)
+	var metaFile string
+	if dir == "." { // Hide the file on UNIX-like systems (e.g. Linux, OSX)
+		metaFile = "./." + fn + AzCopyHashDataStream
+	} else {
+		metaFile = dir + "/." + fn + AzCopyHashDataStream
+	}
+	// open file for writing; truncate.
+	f, err := os.OpenFile(metaFile, os.O_CREATE|os.O_TRUNC|os.O_RDWR, 0644)
```
vibhansa-msft, line 51: We are saving the hash data in a separate file. Can we not save it in extended attributes of the source file itself.

cmd/syncComparator.go
```diff
@@ -57,11 +63,23 @@ func (f *syncDestinationComparator) processIfNecessary(destinationObject StoredO
...
+			return f.copyTransferScheduler(sourceObjectInMap)
+		}
+
+		if f.comparisonHashType != common.ESyncHashType.None() {
+			switch f.comparisonHashType {
+			case common.ESyncHashType.MD5():
+				if !reflect.DeepEqual(sourceObjectInMap.md5, destinationObject.md5) {
+					// hash inequality = source "newer" in this model.
+					return f.copyTransferScheduler(sourceObjectInMap)
+				}
```
vibhansa-msft, line 77: For hash type match also we can dump a debug log that we are skipping this file. In future if someone reports and issue with the sync it will be easy to identify what we skipped and what we synced.

seanmcc-msft:
Will we add some tests for this feature?

common/hash_data_other.go
adreed-msft: I opted to write to a separate file as an OS-agnostic solution. I can look into a Linux-specific implementation to use xattr.

adreed-msft:
> Will we add some tests for this feature?

I have been actively adding tests. They are now present.

cmd/syncComparator.go
adreed-msft: Sure. I'll add some debug logging in.

common/hash_data_other.go
adreed-msft: Should we want to use another open-source dependency, https://github.com/pkg/xattr provides an OS-agnostic implementation for most *NIX based operating systems, but more importantly, the two we officially support. The package utilizes the BSD 2-clause "Simplified" license, which, we have pulled other packages in under.

That would be a quick & dirty way to implement that.

adreed-msft: We could have a detection mechanism, but that's a lot of work for minimal gain, and when a transfer can easily span multiple filesystems, we *have* to accept that sometimes, we're just not going to be able to write the xattr, and that HBS is going to come with a tremendous performance caveat every time.

cmd/sync.go
```diff
@@ -792,6 +810,9 @@ func init() {
 	syncCmd.PersistentFlags().BoolVar(&raw.mirrorMode, "mirror-mode", false, "Disable last-modified-time based comparison and overwrites the conflicting files and blobs at the destination if this flag is set to true. Default is false")
 	syncCmd.PersistentFlags().BoolVar(&raw.dryrun, "dry-run", false, "Prints the path of files that would be copied or removed by the sync command. This flag does not copy or remove the actual files.")
 
+	syncCmd.PersistentFlags().StringVar(&raw.compareHash, "compare-hash", "None", "Inform sync to rely on hashes as an alternative to LMT. (None, MD5) Default: None")
```
gapra-msft, line 813: can we add a doc here saying we expect all remote files to have hashes computed on them

common/hash_data_other.go
```diff
@@ -0,0 +1,70 @@
...
+	"encoding/json"
+	"io"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+func TryGetHashData(fullpath string) (SyncHashData, error) {
+	// get meta file name
+	dir, fn := filepath.Split(fullpath)
```
vibhansa-msft, line 16: Using file path we are creating a path to store our hash. This code is duplicate in Get and Set methods. Shall we unify this so that later in case we need to make any change to this logic it's easy to maintain.

common/hash_data_other.go
```diff
@@ -0,0 +1,70 @@
...
+	var metaFile string
+	if dir == "." { // Hide the file on UNIX-like systems (e.g. Linux, OSX)
+		metaFile = "./." + fn + AzCopyHashDataStream
+	} else {
+		metaFile = dir + "/." + fn + AzCopyHashDataStream
+	}
+	// open file for reading
+	f, err := os.OpenFile(metaFile, os.O_RDONLY, 0644)
+	if err != nil {
+		return SyncHashData{}, err
```
vibhansa-msft, line 26: In such cases where Get/Set of hash fails, what will be the impact on sync/copy operation. Do we fail the sync or assume hash is not available and overwrite the content?

vibhansa-msft (approved):

cmd/syncComparator.go
```diff
@@ -57,12 +65,30 @@ func (f *syncDestinationComparator) processIfNecessary(destinationObject StoredO
...
+		if f.comparisonHashType != common.ESyncHashType.None() && sourceObjectInMap.entityType == common.EEntityType.File() {
+			switch f.comparisonHashType {
+			case common.ESyncHashType.MD5():
+				if !reflect.DeepEqual(sourceObjectInMap.md5, destinationObject.md5) {
+					azcopyScanningLogger.Log(pipeline.LogDebug, fmt.Sprintf("File %s overwritten due to differing hash.", destinationObject.relativePath))
+					// hash inequality = source "newer" in this model.
+					return f.copyTransferScheduler(sourceObjectInMap)
+				}
+			default:
+				panic("sanity check: unsupported hash type " + f.comparisonHashType.String())
```
nakulkar-msft, line 82: We should return error, instead of panic. We may specialize the error and quit all at once, but we should avoid panic in general.

cmd/zc_traverser_local.go
```diff
@@ -369,6 +433,194 @@ func (t *localTraverser) Traverse(preprocessor objectMorpher, processor objectPr
...
+			}
+
+			return nil
+		}
+	}
+
+	// wrap the processor, preventing a data race
+	mutexProcessor := func(proc objectProcessor) objectProcessor {
+		return func(object StoredObject) error {
+			commitMutex.Lock()
```
nakulkar-msft, line 469: Nit: defer would be a bit more go-like

ste/xfer-remoteToLocal-file.go
```diff
@@ -417,6 +419,25 @@ func commonDownloaderCompletion(jptm IJobPartTransferMgr, info TransferInfo, ent
...
+			}
+
+			err = common.PutHashData(info.Destination, common.SyncHashData{
+				Mode: common.ESyncHashType.MD5(),
+				LMT:  fi.ModTime(),
+				Data: base64.StdEncoding.EncodeToString(info.SrcHTTPHeaders.ContentMD5),
+			})
+			if err != nil {
+				jptm.FailActiveDownload("saving MD5 data (writing alternate data stream)", err)
+				goto redoCompletion // let fail as expected
```
nakulkar-msft, line 437: Can we get into a loop here?

cmd/syncComparator.go
adreed-msft: Panics are used exclusively for sanity checks, to make sure that values that _should_ for all reasons be correct at the given point in the code are still correct, and that garbage hasn't come in for some reason. They, for all intents and purposes, will never be hit. But if they are hit, we immediately know something is terribly wrong.

cmd/zc_traverser_local.go
adreed-msft: I agree. I forget the semantics of why I didn't use defer here.

ste/xfer-remoteToLocal-file.go
adreed-msft: Nope. At the point whe goto redoCompletion, the jptm is dead due to the failed download. If we wanted to, we could do a recursive call and drop the goto.

common/hash_data_other.go
adreed-msft: For initial hash generation, we let it fail. The user could be syncing from a read-only location.

adreed-msft: For putting the hash at the end of the transfer, we do not let it fail. The destination should be writeable anyway.

common/hash_data_other.go
adreed-msft: Sure.

cmd/sync.go
adreed-msft: Sure.

nakulkar-msft (approved):

cmd/syncComparator.go
```diff
@@ -109,11 +152,47 @@ func (f *syncSourceComparator) processIfNecessary(sourceObject StoredObject) err
...
 			return f.copyTransferScheduler(sourceObject)
 		}
-		// skip if source is more recent
+
+		if f.comparisonHashType != common.ESyncHashType.None() && sourceObject.entityType == common.EEntityType.File() {
+			switch f.comparisonHashType {
+			case common.ESyncHashType.MD5():
+				if sourceObject.md5 == nil {
+					if azcopyScanningLogger != nil {
+						azcopyScanningLogger.Log(pipeline.LogWarning, fmt.Sprintf("File %s skipped, because the source does not have a hash.", sourceObject.relativePath))
```
tasherif-msft, line 165: same as above comment, too much duplication, why don't we move this out in a separate method

cmd/zc_traverser_local.go
```diff
@@ -46,6 +51,10 @@ type localTraverser struct {
 	// a generic function to notify that a new stored object has been enumerated
 	incrementEnumerationCounter enumerationCounterFunc
 	errorChannel                chan ErrorFileInfo
+
+	targetHashType common.SyncHashType
+	// receives fullPath entries and manages hashing of files lacking metadata.
+	hashTargetChannel chan string
```
tasherif-msft, line 57: so this channel is used for sending local paths that need hashing?

cmd/zc_traverser_local.go
```diff
@@ -369,6 +427,194 @@ func (t *localTraverser) Traverse(preprocessor objectMorpher, processor objectPr
...
+		return func(object StoredObject) error {
+			commitMutex.Lock()
+			defer commitMutex.Unlock()
+			err := proc(object)
+
+			return err
+		}
+	}
+
+	for i := 0; i < runtime.NumCPU()/2; i++ {
```
tasherif-msft, line 471: we're duplicating here as well, you have already defined `hashingThreadCount` above, please use that instead

cmd/zc_traverser_local.go
```diff
@@ -369,6 +427,194 @@ func (t *localTraverser) Traverse(preprocessor objectMorpher, processor objectPr
 		return fmt.Errorf("failed to scan path %s due to %s", t.fullPath, err.Error())
 	}
 
+	t.hashTargetChannel = make(chan string, 1_000) // "reasonable" backlog
```
tasherif-msft, line 430: The logic here and below is quite extensive, is there a reason we didn't abstract this out as a helper method? it would make much more readable.

common/hash_data_other.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"encoding/json"
+	"io"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+func getHashPath(fullpath string) string {
```
tasherif-msft, line 14: I think with methods like these, its best to write unit tests as well

cmd/zc_traverser_local.go
adreed-msft: Correct.

tasherif-msft (approved):

PR #2033, opened 2023-01-24
[AzCopyV10]  Add in content-md5 to available list properties

### LINKED ISSUE #2032: azcopy list --properties "ContentMD5" does not return md5sum of the file
azcopy version 10.16.0

MacOS Ventura 13.1

azcopy list https://<stgaccount/container> --properties "Content-MD5"

Would love to just return the md5sum property set for each file 

Choose any file with Content-MD5 and try to use list

No great solution to list all the files

vsmalladi:
Added "ContentMD5" as  new flag to '--properties' in the list command which takes ; separated list of properties and will display these fields in list command output.

Closes #2032

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -180,7 +180,7 @@ func (cooked cookedListCmdArgs) processProperties(object StoredObject) string {
 		case contentEncoding:
 			builder.WriteString(propertyStr + ": " + object.contentEncoding + "; ")
 		case contentMD5:
-			builder.WriteString(propertyStr + ": " + base64.StdEncoding.EncodeString(object.md5) + "; ")
+			builder.WriteString(propertyStr + ": " + base64.StdEncoding.EncodeToString(object.md5) + "; ")
 		case leaseState:
 			builder.WriteString(propertyStr + ": " + string(object.leaseState) + "; ")
 		case leaseStatus:


nakulkar-msft:
Hi @vsmalladi, please submit review to dev branch

cmd/list.go
```diff
@@ -177,6 +178,8 @@ func (cooked cookedListCmdArgs) processProperties(object StoredObject) string {
 			builder.WriteString(propertyStr + ": " + object.contentType + "; ")
 		case contentEncoding:
 			builder.WriteString(propertyStr + ": " + object.contentEncoding + "; ")
+		case contentMD5:
+			builder.WriteString(propertyStr + ": " + object.contentMD5 + "; ")
```
adreed-msft, line 182: Consider base64 encoding or using some standard way to display an MD5 to an end user

adreed-msft (changes requested):

cmd/list.go
vsmalladi: Good idea

vsmalladi:
@adreed-msft any other changes needed?

nakulkar-msft (approved):

cmd/list.go
```diff
@@ -54,6 +55,7 @@ const (
 	blobAccessTier   validProperty = "BlobAccessTier"
 	contentType      validProperty = "ContentType"
 	contentEncoding  validProperty = "ContentEncoding"
+	contentMD5		         validProperty = "md5"
```
adreed-msft, line 58: This name does not align with the helptext. Otherwise, looks good.

adreed-msft (approved):

cmd/list.go
adreed-msft: Also, formatting here.

vsmalladi: Fixed

vsmalladi:
Thanks @adreed-msft and @nakulkar-msft made necessary changes, should be good to go now if you approve.

vibhansa-msft (approved):

gapra-msft:
Hi @vsmalladi, Could you please merge latest dev into your PR? That should resolve some of the failing checks.

vsmalladi:
@gapra-msft Should be good now.

PR #2043, opened 2023-02-01
Blob Checkpoint Infrastructure

nakulkar-msft:
1. Introduce a bitmap to hold remember which chunks have been uploaded
2. Create a checkpoint metafile. This metafile will be at jobManager level. The metafile will be flushed to disk every 10s
3. Plumb metafile to blob Sender


common/bitmap_test.go
```diff
@@ -0,0 +1,72 @@
...
+	chk "gopkg.in/check.v1"
+)
+
+type bitmapTestSuite struct{}
+
+var _ = chk.Suite(&bitmapTestSuite{})
+
+func (b *bitmapTestSuite) TestBitmap(c *chk.C) {
+	numOfBlocks := rand.Int31n(azblob.AppendBlobMaxBlocks)
+	fileMap := NewBitMap(int(numOfBlocks))
```
vibhansa-msft, line 36: We shall also put together tests for different length of bitmask one can create to test boundary conditions and negative cases.

ste/xfer-anyToRemote-file.go
```diff
@@ -589,6 +590,8 @@ func commonSenderCompletion(jptm IJobPartTransferMgr, s sender, info TransferInf
 		// and we know the transfer didn't fail (because just checked its status above and made sure the context was not canceled),
 		// so it must have succeeded. So make sure its not left "in progress" state
 		jptm.SetStatus(common.ETransferStatus.Success())
+		jptm.CheckPoint().TransferDone() // Better place for this line would be Epilogue. But because we've not yet set the
```
vibhansa-msft, line 593: In between if context is cancelled or user aborts the job, we need to clean all this bitmap and other house keeping stuff.

ste/mgr-JobPartTransferMgr.go
```diff
@@ -97,6 +97,14 @@ type IJobPartTransferMgr interface {
...
 	SuccessfulBytesTransferred() int64
 	TransferIndex() (partNum, transferIndex uint32)
+	CheckPoint() ICheckpoint
+}
+
+type ICheckpoint interface {
+	Init(numChunks int)
+	ChunkDone(chunkID int)
+	TransferDone()
+	CompletedChunks() map[int]int
```
vibhansa-msft, line 107: To match the uncommitted list received from server we will take only completed chunks? we can take entire chunk list and save this additional map copy here. Just like 'ChunkDone' we can have a method 'IsChunkDone' where giving an index it just tells whether that chunk was uploaded from our end or not.

adreed-msft (approved):

PR #2050, opened 2023-02-03
Update block ID generation logic

nakulkar-msft:
The block ID names will have a context of the jobID, transfer cookie in the job and block/offset in the source file.
Encoded ID: <5B empty placeholder> <16B GUID of AzCopy re-interpreted as string><5B PartNum><5B Index in the jobPart><5B blockNum>

### DIFF
--- a/common/extensions.go
+++ b/common/extensions.go
@@ -2,7 +2,6 @@ package common
 
 import (
 	"bytes"
-	"crypto/md5"
 	"encoding/base64"
 	"fmt"
 	"net/http"
@@ -11,7 +10,6 @@ import (
 	"strings"
 
 	"github.com/Azure/azure-storage-azcopy/v10/azbfs"
-	"github.com/google/uuid"
 
 	"github.com/Azure/azure-storage-file-go/azfile"
 )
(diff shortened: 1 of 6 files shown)


tasherif-msft (commented):
Please provide a pr description with:
1. why we're doing this change
2. a brief comment on what the change is
Also, can this be tested?

nakulkar-msft:
> Please provide a pr description with:
> 
> 1. why we're doing this change
> 2. a brief comment on what the change is
>    Also, can this be tested?

Shared the reason on email. 
Testing would be check that we are generating same blockID repeatedly for the same input.

nakulkar-msft:
@adreed-msft @tasherif-msft

tasherif-msft (approved):

common/extensions.go
```diff
@@ -178,3 +182,9 @@ func GenerateFullPathWithQuery(rootPath, childPath, extraQuery string) string {
 		return p + "?" + extraQuery
 	}
 }
+
+func GenerateBlockBlobBlockID(blockNamePrefix string, index int32) string {
```
tasherif-msft, line 186: write a small comment above the method mentioning what's encoded in the ID

vibhansa-msft (approved):

PR #2072, opened 2023-02-17
Updating helpMessages.go to further clarify the --compare-hash flag

normesta:
As I update the AzCopy ref files, I'd like to also update the inline help so that these two sources remain in sync. I want to avoid making changes only in the ref article for a couple of reasons - A) consistency and B) so that I don't inadvertently overwrite content in the ref article when taking a major ref refresh from inline help.


gapra-msft:
Hi @normesta Could you please rebase the PR to point to dev instead of main, we generally first merge changes into the dev branch

normesta:
Understood. Will do. My apologies

gapra-msft:
> Understood. Will do. My apologies

No worries! Thank you :)

PR #2079, opened 2023-02-21
Add error when chunkSize is greater than the strict memory limit of the CacheLimiter.

mkrautz:
Scenario:

    export AZCOPY_BUFFER_GB=5
    # Copy a single 5GB file to an Azure Blob Storage Container with a block size of 4000MB
    ./acopy [...] --block-size-mb=4000 [...]

Since the chunk size in that case is greater than the strict memory limit (75% of 5GB) this would end up in a deadlock: CacheLimiter.TryAdd would loop forever, waiting for memory to be freed (which it never will).

This patch tries to remedy that by making it an error if a chunk is greater than the strict memory limit of the associated CacheLimiter.

### DIFF
--- a/common/cacheLimiter.go
+++ b/common/cacheLimiter.go
@@ -27,6 +27,18 @@ import (
 	"time"
 )
 
+// The percentage of a CacheLimiter's Limit that is considered
+// the strict limit.
+var cacheLimiterStrictLimitPercentage = float32(0.75)
+// Rationale for the level of the strict limit: as at Jan 2018, we are using 0.75 of the total as the strict
+// limit, leaving the other 0.25 of the total accessible under the "relaxed" limit.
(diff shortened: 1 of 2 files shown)


adreed-msft (approved):

gapra-msft (approved):

gapra-msft:
Hi @mkrautz Apologies but it looks like this PR was made against main. We generally only accept PRs into the dev branch. I accidentally merged the PR but I had to rollback the change in main. 

If you wouldn't mind, could you please create a new PR against dev? If not, I am also happy to create the PR for you.

mkrautz:
@gapra-msft

Ooops! I created a new PR at https://github.com/Azure/azure-storage-azcopy/pull/2228

gapra-msft:
@mkrautz Thank you! We will work on getting this PR merged in. Thank again for your contribution.

PR #2080, opened 2023-02-21
Redact CPK headers in log output.

mkrautz:
Currently, azcopy will emit the following headers to log files, when using Customer Provided Keys:

x-ms-encryption-key
x-ms-encryption-key-sha256

I believe this is unexpected for some users. This patch redacts those headers from log files, such that encryption keys do not accidentally leak.


mkrautz:
Another PR in the azure-storage-blob-go dependency:

https://github.com/Azure/azure-storage-blob-go/pull/331

PR #2096, opened 2023-03-01
Allow specifying "Cold" tier

adreed-msft:
Supports specifying cold tier for Block Blobs.

Testing is disabled since this isn't available service-side yet, but this PR is nice & simple since we can just pass a new value in and the SDK doesn't validate.

### DIFF
--- a/e2etest/zt_preserve_access_tier_test.go
+++ b/e2etest/zt_preserve_access_tier_test.go
@@ -26,6 +26,19 @@ import (
 	"testing"
 )
 
+//func TestTier_UploadCold(t *testing.T) {
+//	RunScenarios(t, eOperation.Copy(), eTestFromTo.Other(common.EFromTo.LocalBlob()), eValidate.Auto(), anonymousAuthOnly, anonymousAuthOnly, params{
+//		recursive:             true,
+//		accessTier:            common.EBlockBlobTier.Cold().ToAccessTierType(), // this is not valid yet on the service, so this test is disabled.
+//	}, nil, testFiles{
+//		defaultSize: "4M",
+//		shouldTransfer: []interface{}{
+//			folder(""), // root folder
+//			f("filea"),
+//		},
+//	}, EAccountType.Classic(), EAccountType.Standard(), "")
+//}
+
 func TestTier_V2ToClassicAccount(t *testing.T) {
 
 	RunScenarios(t, eOperation.Copy(), eTestFromTo.Other(common.EFromTo.BlobBlob()), eValidate.AutoPlusContent(), anonymousAuthOnly, anonymousAuthOnly, params{
--- a/ste/xfer-anyToRemote-file.go
+++ b/ste/xfer-anyToRemote-file.go
@@ -116,7 +116,7 @@ func BlobTierAllowed(destTier azblob.AccessTierType) bool {
 		// Standard storage account. If it's Hot, Cool, or Archive, we're A-OK.
 		// Page blobs, however, don't have an access tier on Standard accounts.
 		// However, this is also OK, because the pageblob sender code prevents us from using a standard access tier type.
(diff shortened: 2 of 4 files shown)


ste/xfer-anyToRemote-file.go
```diff
@@ -116,7 +116,7 @@ func BlobTierAllowed(destTier azblob.AccessTierType) bool {
 		// Standard storage account. If it's Hot, Cool, or Archive, we're A-OK.
 		// Page blobs, however, don't have an access tier on Standard accounts.
 		// However, this is also OK, because the pageblob sender code prevents us from using a standard access tier type.
-		return destTier == azblob.AccessTierArchive || destTier == azblob.AccessTierCool || destTier == azblob.AccessTierHot
+		return destTier == azblob.AccessTierArchive || destTier == azblob.AccessTierCool || destTier == common.EBlockBlobTier.Cold().ToAccessTierType() || destTier == azblob.AccessTierHot
```
gapra-msft, line 119: Any place where we call SetAccessTier or pass this in, we need to set the context to manually override the service version though right?

e2etest/zt_preserve_access_tier_test.go
```diff
@@ -26,6 +26,19 @@ import (
 	"testing"
 )
 
+//func TestTier_UploadCold(t *testing.T) {
+//	RunScenarios(t, eOperation.Copy(), eTestFromTo.Other(common.EFromTo.LocalBlob()), eValidate.Auto(), anonymousAuthOnly, anonymousAuthOnly, params{
```
vibhansa-msft, line 30: if this is dead code, we can remove it.

ste/xfer-anyToRemote-file.go
vibhansa-msft, line 119: where are we setting the SV to support this new tier.

e2etest/zt_preserve_access_tier_test.go
adreed-msft: This is not dead code. This is a test that can't currently be ran and succeed, because this isn't supported by the service (yet).

ste/xfer-anyToRemote-file.go
adreed-msft: This was to-do, but now is done.

adreed-msft: This is actually implemented as a pipeline policy now.

e2etest/zt_preserve_access_tier_test.go
```diff
@@ -26,6 +26,19 @@ import (
...
+//		recursive:             true,
```
gapra-msft, line 31: Can we add an item in our backlog to remind us to bring this test back once the service is live?

gapra-msft (approved):
Is this safe to merge before the service bump is live?

adreed-msft:
> Is this safe to merge before the service bump is live?

Yes-- It's only kicked in when the user specifies Cold tier, which isn't publicly documented anyway.

e2etest/zt_preserve_access_tier_test.go
adreed-msft: Backlog items tend to get lost in the weeds, but yes, we should do this. I think we should keep it in the current release and just push it back another step each time while we wait so it gets brought up in the sync-up, rather than stick it in "el backlog del backlog backlog" for it to get lost to time.

siminsavani-msft (approved):

seanmcc-msft:
@adreed-msft, Cold Tier should be live in the service - https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-tier

This feature is enabled in central Canada and probably France.  Try the storage account "seanmcccanada" in our XClient Azure subscription.

adreed-msft:
Gotcha-- I'll give it a shot sometime today & merge this if all is happy.

vibhansa-msft (approved):

PR #2106, opened 2023-03-09
Blob Traverser: Include stubs only once

nakulkar-msft:
When we are traversing blob endpoint, and we already have directory stubs, we schedule the directory twice. Once while listing prefixes, and second when listing blobs. This will cause a panic in the folderTrackerManager. 

This can be easily reproduced by copying from a container/virtual-directory in  blobendpoint, with include-directory-stub=true.


adreed-msft:
Needs rebase

PR #2108, opened 2023-03-09
Migrated blob traverser code 

gapra-msft:


common/clientFactory.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
+	blobservice "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
+)
+
+// NewClientCallbacks is a Generic Type to allow client creation error handling to live in a single place (CreateClient)
+// T = Client type
+// U = SharedKeyCredential type
+// Note : Could also make azcore.ClientOptions generic here if one day different storage service clients have additional options. This would also make the callback definitions easier.
+type NewClientCallbacks[T, U any] struct {
```
gapra-msft, line 35: TODO : This type can be non exported.

common/clientFactory.go
```diff
@@ -0,0 +1,106 @@
...
+	TokenCredential        func(string, azcore.TokenCredential, azcore.ClientOptions) (*T, error)
+	NewSharedKeyCredential func(string, string) (*U, error)
+	SharedKeyCredential    func(string, *U, azcore.ClientOptions) (*T, error)
+	NoCredential           func(string, azcore.ClientOptions) (*T, error)
+}
+
+// CreateClient is a generic method to allow client creation error handling to live in a single place
+func CreateClient[T, U any](callbacks NewClientCallbacks[T, U], u string, credInfo CredentialInfo, credOpOptions CredentialOpOptions, options azcore.ClientOptions) (*T, error) {
```
gapra-msft, line 43: TODO : This function can be non exported

cmd/copyEnumeratorInit_test.go
```diff
@@ -127,11 +126,11 @@ func (ce *copyEnumeratorSuite) TestValidateSourceFileDoesNotExist(c *chk.C) {
 	fileName := "source_file"
 
 	ctx := context.WithValue(context.TODO(), ste.ServiceAPIVersionOverride, ste.DefaultServiceApiVersion)
-	p := azblob.NewPipeline(azblob.NewAnonymousCredential(), azblob.PipelineOptions{})
 
 	// List
-	rawBlobURLWithSAS := scenarioHelper{}.getRawBlobURLWithSAS(c, containerName, fileName)
-	blobTraverser := newBlobTraverser(&rawBlobURLWithSAS, p, ctx, true, true, func(common.EntityType) {}, false, common.CpkOptions{}, false, false, false, common.EPreservePermissionsOption.None())
+	rawBlobURLWithSAS := scenarioHelper{}.getBlobClientWithSAS(c, containerName, fileName).URL()
```
vibhansa-msft, line 131: In old code we were working with just raw blob url with sas but with new changes we need both blob and service url with sas ?

cmd/copyUtil.go
```diff
@@ -157,19 +165,15 @@ func (util copyHandlerUtil) urlIsAzureFileDirectory(ctx context.Context, url *ur
...
 // Used to handle special scenarios or conditions.
-func (util copyHandlerUtil) doesBlobRepresentAFolder(metadata azblob.Metadata) bool {
+func (util copyHandlerUtil) doesBlobRepresentAFolder(metadata map[string]*string) bool {
 	// this condition is to handle the WASB V1 directory structure.
 	// HDFS driver creates a blob for the empty directories (let’s call it ‘myfolder’)
 	// and names all the blobs under ‘myfolder’ as such: ‘myfolder/myblob’
 	// The empty directory has meta-data 'hdi_isfolder = true'
-	return metadata["hdi_isfolder"] == "true"
+	// Note: GoLang sometimes sets metadata keys with the first letter capitalized
+	return (metadata["hdi_isfolder"] != nil && strings.ToLower(*metadata["hdi_isfolder"]) == "true") || (metadata["Hdi_isfolder"] != nil && strings.ToLower(*metadata["Hdi_isfolder"]) == "true")
```
vibhansa-msft, line 176: in track-2 while setting the metadata we will always preserve the case user has given but reading is still messy I guess.

gapra-msft: This might be due to intermingling Track 1 and Track 2 with tests. Should we keep this anyway since a customer may have uploaded with Track 1.5?

cmd/copyEnumeratorInit_test.go
gapra-msft: Before we used the raw blob url with sas AND the pipeline, not we pass in the blob URL with the service client. 

Alternatively we could pass the credential here but that seemed less clean to me in this situation

cmd/credentialUtil.go
```diff
@@ -646,6 +648,29 @@ func getCredentialType(ctx context.Context, raw rawFromToInfo, cpkOptions common
 // ==============================================================================================
 // pipeline factory methods
 // ==============================================================================================
+func createClientOptions(logLevel pipeline.LogLevel) azcore.ClientOptions {
+	logOptions := ste.LogOptions{}
+	if azcopyScanningLogger != nil {
+		logOptions.LogOptions = pipeline.LogOptions{
+			Log:       azcopyScanningLogger.Log,
+			ShouldLog: func(level pipeline.LogLevel) bool { return level <= logLevel },
```
vibhansa-msft, line 656: IIRC track-2 sdk does not have this level based logging. We need to register what events we want and then create a callback method that receives those events. How are we taking care of info/error logging in that case based on level we have configured.

ste/xferVersionPolicy.go
```diff
@@ -0,0 +1,52 @@
...
+
+// NewVersionPolicyFactory creates a factory that can override the service version
+// set in the request header.
+// If the context has key overwrite-current-version set to false, then x-ms-version in
+// request is not overwritten else it will set x-ms-version to 207-04-17
+func NewVersionPolicyFactory() pipeline.Factory {
+	return pipeline.FactoryFunc(func(next pipeline.Policy, po *pipeline.PolicyOptions) pipeline.PolicyFunc {
+		return func(ctx context.Context, request pipeline.Request) (pipeline.Response, error) {
+			// get the service api version value using the ServiceAPIVersionOverride set in the context.
+			if value := ctx.Value(ServiceAPIVersionOverride); value != nil {
```
vibhansa-msft, line 45: We have suggested few customers to override the service version using env variable. We shall test this with default version and going backward and forward in terms of version values.

cmd/copyUtil.go
adreed-msft: This is _ok_, but a little ugly, but I still prefer it over the alternative (enumerating over & EqualFolding the keys).

cmd/zc_traverser_blob_account.go
```diff
@@ -116,21 +107,12 @@ func (t *blobAccountTraverser) Traverse(preprocessor objectMorpher, processor ob
...
-func newBlobAccountTraverser(rawURL *url.URL, p pipeline.Pipeline, ctx context.Context, includeDirectoryStubs bool, incrementEnumerationCounter enumerationCounterFunc, s2sPreserveSourceTags bool, cpkOptions common.CpkOptions, preservePermissions common.PreservePermissionsOption) (t *blobAccountTraverser) {
-	bURLParts := azblob.NewBlobURLParts(*rawURL)
-	cPattern := bURLParts.ContainerName
-
-	// Strip the container name away and treat it as a pattern
-	if bURLParts.ContainerName != "" {
-		bURLParts.ContainerName = ""
-	}
-
+func newBlobAccountTraverser(serviceClient *service.Client, container string, ctx context.Context, includeDirectoryStubs bool, incrementEnumerationCounter enumerationCounterFunc, s2sPreserveSourceTags bool, cpkOptions common.CpkOptions, preservePermissions common.PreservePermissionsOption) (t *blobAccountTraverser) {
```
adreed-msft, line 110: Some of these options aren't always required. Should we take the initiative now to options bag this as best we can?

adreed-msft (approved):

cmd/credentialUtil.go
gapra-msft: We are using the AzCopy logger here and choosing to register on none of the Track 2 log events. imo the AzCopy logger is too different from the Track 2 SDK default logger to be able to take advantage of it. We can revisit this after all the APIs are migrated. It will be a fairly simple fix.

cmd/zc_traverser_blob_account.go
gapra-msft: this is a good idea, but I'd still prefer it to be it's own PR (maybe after blob traverser and cmd code are merged)

only reason is changing this results in a lot of test changes so it can get confusing to review.

siminsavani-msft (approved):

common/clientFactory.go
```diff
@@ -0,0 +1,106 @@
...
+	case ECredentialType.OAuthToken():
+		if credInfo.OAuthTokenInfo.IsEmpty() {
+			credOpOptions.panicError(errors.New("invalid state, cannot get valid OAuth token information"))
+		}
+		tc, err := credInfo.OAuthTokenInfo.GetTokenCredential()
+		if err != nil {
+			credOpOptions.panicError(fmt.Errorf("unable to get token credential due to reason (%s)", err.Error()))
+		}
+		return callbacks.TokenCredential(u, tc, options)
+	case ECredentialType.SharedKey():
```
nakulkar-msft, line 54: Shared key should not be required. We do not allow this option.

nakulkar-msft (changes requested):

cmd/copyUtil.go
```diff
@@ -45,23 +45,31 @@ type copyHandlerUtil struct{}
 var gCopyUtil = copyHandlerUtil{}
 
 // checks if a given url points to a container or virtual directory, as opposed to a blob or prefix match
-func (util copyHandlerUtil) urlIsContainerOrVirtualDirectory(url *url.URL) bool {
-	if azblob.NewBlobURLParts(*url).IPEndpointStyleInfo.AccountName == "" {
+func (util copyHandlerUtil) urlIsContainerOrVirtualDirectory(rawURL string) bool {
+	parsedURL, err := url.Parse(rawURL)
```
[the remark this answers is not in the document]
gapra-msft: This url parse is done here mostly to make the Path calls below easier. I could remove it and get the path by appending parts.Container name and parts.Blob name but it might get tricky since we also support IP endpoint style, which means the account name is also part of the path.

common/clientFactory.go
gapra-msft: Why is this in current code to create pipelines? Was it just missed to remove at some point?

Are we sure there are no tests or anything that uses SharedKey?

gapra-msft: Removed SharedKey part

ste/xferVersionPolicy.go
gapra-msft: For the customers we suggest to override service version, is it so that they get access to latest features? or are we asking them to go backwards? @nakulkar-msft @adreed-msft?

nakulkar-msft (approved):

PR #2110, opened 2023-03-10
Migrated more cmd code

gapra-msft:


common/credentialFactory.go
```diff
@@ -166,6 +173,120 @@ func CreateBlobClientFromServiceClient(blobURLParts blob.URLParts, client *blobs
 	return blobClient, nil
 }
 
+func CreateContainerClient(blobURL string, credInfo *CredentialInfo, options *container.ClientOptions, credOpOptions *CredentialOpOptions) (*container.Client, error) {
```
gapra-msft, line 176: Ideally I don't want so much CreateXXXClient code. I will follow up this PR with some code to clean this up a bit and parse the URL, get service client, then get the appropriate container/blob client so we isolate this SAS/OAuth/SharedKey switch statement to one method

cmd/benchmark.go
```diff
@@ -186,12 +186,15 @@ func (raw rawBenchmarkCmdArgs) appendVirtualDir(target, virtualDir string) (stri
...
+		p, err := blob.ParseURL(target)
+		if err != nil {
+			return "", fmt.Errorf("error parsing the url %s. Failed with error %s", target, err.Error())
+		}
 		if p.ContainerName == "" || p.BlobName != "" {
 			return "", errors.New("the blob target must be a container")
 		}
 		p.BlobName = virtualDir
-		result = p.URL()
+		return result.String(), err
```
vibhansa-msft, line 197: 'result' is not used in blob.ParseURL(), is it correct variable to return?

gapra-msft: fixed

vibhansa-msft (approved):

siminsavani-msft (approved):

tasherif-msft (approved):

cmd/copyEnumeratorInit.go
```diff
@@ -460,29 +462,23 @@ func (cca *CookedCopyCmdArgs) createDstContainer(containerName string, dstWithSA
...
+		bsc, err := common.CreateBlobServiceClient(accountRoot, dstCredInfo, nil, options)
 		if err != nil {
 			return err
 		}
+		bcc := bsc.NewContainerClient(containerName)
 
-		bsu := azblob.NewServiceURL(*dstURL, dstPipeline)
-		bcu := bsu.NewContainerURL(containerName)
-		_, err = bcu.GetProperties(ctx, azblob.LeaseAccessConditions{})
+		_, err = bcc.GetProperties(ctx, nil)
```
nakulkar-msft, line 471: We'll save a call to service if we skip this. We already check ignore ContainerAlreadyExists in after the Create call.

adreed-msft (approved):

cmd/copyEnumeratorInit.go
```diff
@@ -460,29 +462,23 @@ func (cca *CookedCopyCmdArgs) createDstContainer(containerName string, dstWithSA
 			return err
 		}
 
-		dstURL, err := url.Parse(accountRoot)
-
+		bsc, err := common.CreateBlobServiceClient(accountRoot, dstCredInfo, nil, options)
```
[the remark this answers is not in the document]
gapra-msft: added to doc as cleanup item

PR #2117, opened 2023-03-15
Move existing HNS delete code into STE to enable cancel

adreed-msft:
Moves a good majority of HNS delete code into the STE so we can utilize existing cancel functionality for stgexp


cmd/copy.go
```diff
@@ -1572,6 +1572,17 @@ func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 		err = e.enumerate()
 
 	case common.EFromTo.BlobFSTrash():
+		if cca.Recursive { // Extra hacky hack: Carrying recursive into STE temporarily to enable cancel of delete job
+			query, err := url.ParseQuery(jobPartOrder.SourceRoot.ExtraQuery)
```
gapra-msft, line 1576: Is this because the delete API doesnt allow you to set recursive as a parameter?

adreed-msft: No, this is because I felt like it did not make sense to tack a whole additional property onto the job plan file format to carry through the recursive option for ADLS Gen 2.

adreed-msft: If we wanted to handle proper enum/delete then this could be completely removed.

cmd/copy.go
```diff
@@ -1542,6 +1542,17 @@ func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 	case cca.FromTo.IsDelete():
 		// Delete gets ran through copy, so handle delete
 		if cca.FromTo.From() == common.ELocation.BlobFS() {
+			if cca.Recursive { // Extra hacky hack: Carrying recursive into STE temporarily to enable cancel of delete job
```
gapra-msft, line 1545: is this so that we don't have to add an extra entry to the plan file?

gapra-msft: Should we just add it to the plan file so it's cleaner? This change seems not friendly to those that may not have historical knowledge in the future.

ste/xfer-deleteBlobFS.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+	"github.com/Azure/azure-pipeline-go/pipeline"
+	"github.com/Azure/azure-storage-azcopy/v10/azbfs"
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"net/url"
+	"strings"
+	"sync"
+)
+
+var logBlobFSDeleteWarnOnce = &sync.Once{}
```
gapra-msft, line 13: can we add an e2e test for dfs delete?

nakulkar-msft (approved):

cmd/copy.go
adreed-msft: > Should we just add it to the plan file so it's cleaner? This change seems not friendly to those that may not have historical knowledge in the future.

Probably.

cmd/removeProcessor.go
```diff
@@ -35,6 +35,7 @@ func newRemoveTransferProcessor(cca *CookedCopyCmdArgs, numOfTransfersPerPart in
 		SourceRoot:          cca.Source.CloneWithConsolidatedSeparators(), // TODO: why do we consolidate here, but not in "copy"? Is it needed in both places or neither? Or is copy just covering the same need differently?
 		CredentialInfo:      cca.credentialInfo,
 		ForceIfReadOnly:     cca.ForceIfReadOnly,
+		BlobFSRecursiveDelete: cca.Recursive,
```
gapra-msft, line 38: nit, do we want to call this BlobFs? What if this becomes relevant to another service?

gapra-msft (approved):
lgtm, just one nit comment

cmd/removeProcessor.go
adreed-msft: We can just rename it later if it does. Not breaking. I don't think it will.

tasherif-msft (approved):

PR #2120, opened 2023-03-16
Do not create parent directory if it is root

nakulkar-msft:


testSuite/scripts/test_blob_download.py
```diff
@@ -41,6 +41,29 @@ def test_download_1kb_blob_to_null(self):
         dst = os.devnull
         result = util.Command("copy").add_arguments(src).add_arguments(dst).add_flags("log-level", "info")
 
+    def test_download_1kb_blob_to_root(self):
```
gapra-msft, line 44: I thought we are trying to avoid adding tests to the testSuite? Shouldn't we add new tests to e2e tests?

nakulkar-msft: The current infra either in e2e or cmd suite does not allow to create custom destination path. There would always be  a temporary directory created for local path.

gapra-msft (approved):

testSuite/scripts/test_blob_download.py
gapra-msft: can you add this as a comment in this test?

adreed-msft (approved):

testSuite/scripts/test_blob_download.py
adreed-msft: Why don't we add this functionality then? It doesn't sound too complicated.

nakulkar-msft: It creates a lot of noise in E2E. The location where we create a temp directory is specific to _resourceLocal_. Which is at least 3 layers down the E2E interface to write tests. Either propagate a variable here - which would mean all the testcases would now take a new variable to say this is a special path to root, or some kind of env or hack so that we dont create a temp directory.

The bug is particular to download path, and does not need to be tested with the all combinations of flags/fromTos in E2E. It is cleaner to do in python suite.

PR #2123, opened 2023-03-20
Change log level to Debug for a few statements

nakulkar-msft:
Reduce spamming by frequent log statements

### DIFF
--- a/common/lifecyleMgr.go
+++ b/common/lifecyleMgr.go
@@ -13,6 +13,7 @@ import (
 	"sync/atomic"
 	"syscall"
 	"time"
+	"unicode"
 
 	"github.com/Azure/azure-pipeline-go/pipeline"
 )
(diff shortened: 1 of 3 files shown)


gapra-msft (approved):

common/lifecyleMgr.go
```diff
@@ -137,6 +138,17 @@ func (lcm *lifecycleMgr) watchInputs() {
 		default:
 		}
 
+		allCharsAreWhiteSpace := true
+		for _, ch := range msg {
+			if !unicode.IsSpace(ch) {
```
vibhansa-msft, line 143: What is this code doing, it waits on a channel until it receives a message with all spaces?
Where do we use this logic.

vibhansa-msft (approved):

common/lifecyleMgr.go
nakulkar-msft: It waits on a channel till it receives a msg with newline at the end. The message is then processed. We changed it to receive special messages (to receive change in throughput etc). But users would hit newlines a bunch of times on their system, and would receive a "Discarding invalid msg" response from AzCopy.

PR #2125, opened 2023-03-23
[AzCopyV10][Bug] Uploading from top directory on linux fails when sub folder has the same name

siminsavani-msft:
Fresh PR for: https://github.com/Azure/azure-storage-azcopy/pull/1762

Context:
- In linux, cx is creating folders under root folder with same name (ex: /event -> root dir, /event/event_stuff/foo.txt)
- Line 19 in cmd/copyEnumeratorHelper.go causes a bug in linux in which the root folder name is removed from path.
i.e. e.SourceRoot.Value = "/event" and transfer.Source = "/event_stuff/foo.txt" --> incorrectly changes transfer.Source to "_stuff/foo.txt"

PR Info:
- The transfer.Source and transfer.Destination is already a relative path and this line of code can be removed because we already pass the relative path for both transfer.Source and transfer.Destination prior to calling addTransfer, see line 265: https://github.com/Azure/azure-storage-azcopy/blob/main/cmd/copyEnumeratorInit.go#L265.

### DIFF
--- a/cmd/copyEnumeratorHelper.go
+++ b/cmd/copyEnumeratorHelper.go
@@ -13,11 +13,7 @@ var EnumerationParallelStatFiles = false
 
 // addTransfer accepts a new transfer, if the threshold is reached, dispatch a job part order.
 func addTransfer(e *common.CopyJobPartOrderRequest, transfer common.CopyTransfer, cca *CookedCopyCmdArgs) error {
-	// Remove the source and destination roots from the path to save space in the plan files
-
-	// TODO: Remove this line because transfer.Source and transfer.Destination are already a relative paths
-	// transfer.Source = strings.TrimPrefix(transfer.Source, e.SourceRoot.Value)
-	// transfer.Destination = strings.TrimPrefix(transfer.Destination, e.DestinationRoot.Value)
+	// Source and destination paths are and should be relative paths. 
 
 	// dispatch the transfers once the number reaches NumOfFilesPerDispatchJobPart
 	// we do this so that in the case of large transfer, the transfer engine can get started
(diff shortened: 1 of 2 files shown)


nakulkar-msft (approved):

cmd/copyEnumeratorHelper.go
```diff
@@ -16,8 +14,10 @@ var EnumerationParallelStatFiles = false
 // addTransfer accepts a new transfer, if the threshold is reached, dispatch a job part order.
 func addTransfer(e *common.CopyJobPartOrderRequest, transfer common.CopyTransfer, cca *CookedCopyCmdArgs) error {
 	// Remove the source and destination roots from the path to save space in the plan files
-	transfer.Source = strings.TrimPrefix(transfer.Source, e.SourceRoot.Value)
-	transfer.Destination = strings.TrimPrefix(transfer.Destination, e.DestinationRoot.Value)
+
+	// TODO: Remove this line because transfer.Source and transfer.Destination are already a relative paths
```
gapra-msft, line 18: Does this need to be a TODO?

cmd/copyEnumeratorHelper_test.go
```diff
@@ -41,7 +41,8 @@ func newRemoteRes(url string) common.ResourceString {
 	return r
 }
 
-func (s *copyEnumeratorHelperTestSuite) TestAddTransferPathRootsTrimmed(c *chk.C) {
+/*func (s *copyEnumeratorHelperTestSuite) TestAddTransferPathRootsTrimmed(c *chk.C) {
```
gapra-msft, line 44: Should we just delete this?

cmd/copyEnumeratorHelper.go
siminsavani-msft: Removed!

tasherif-msft:
I know the PR is already merged but thanks for the clear pr description! @siminsavani-msft

PR #2127, opened 2023-03-27
Redact CPK headers in log output.

nakulkar-msft:
Currently, azcopy will emit the following headers to log files, when using Customer Provided Keys:

x-ms-encryption-key
x-ms-encryption-key-sha256

I believe this is unexpected for some users. This patch redacts those headers from log files, such that encryption keys do not accidentally leak.

Replaces #2080, which is targeted to main


vibhansa-msft (approved):

mkrautz:
Hi @nakulkar-msft
Please note that this also needs https://github.com/Azure/azure-storage-blob-go/pull/331 to be effective...

gapra-msft (approved):

mkrautz:
@nakulkar-msft @gapra-msft @vibhansa-msft 

It seems like https://github.com/Azure/azure-storage-blob-go/pull/331 was never merged?

That part is required, since azure-storage-blob-go also leaks CPK details...

PR #2134, opened 2023-03-29
Migrated identity, get properties and download in STE

gapra-msft:
This PR touches a few things

1. download in STE
2. get properties in STE
3. made CPKInfo/CPKScopeInfo pointers everywhere to make passing it to the methods easier
4. Removed duplicated GetProperties code in XXXsender files. 
5. consolidated Iff and IffNotNil to be generics so we don't have so many methods that basically do the same thing.


common/iff.go
```diff
@@ -20,101 +20,20 @@
 
 package common
 
+// TODO : Remove this?
```
gapra-msft, line 23: this is unused, can we just delete this?

tasherif-msft (approved):

gapra-msft:
Feel free to review after I merge this. Merging this for now to continue forward progress.

Issue #2140, opened 2023-04-01
AzCopy v10.18.0 fails to resolve host when the URI contains a port specification


zigandazag:
### Which version of the AzCopy was used?  
10.18.0

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

### What command did you run?
cp $sourceBlob $localPath --recursive=true

### What problem was encountered?
Fails with  host not found if the blob uri includes a port specification like :443

failed to perform copy command due to error: failed to resolve source: lookup xyz.blob.core.windows.net:443: no such host

### How can we reproduce the problem in the simplest way?
try copying from a blob but include a value for a port in the host name eg.
e.g
azcopy  cp https://xyz.blob.core.windows.net:443/release?sv=.. c:\temp

### Have you found a mitigation/solution?
yes remove the port.

siminsavani-msft:
Hi @zigandazag , I'm curious to know if this was working in v10.17.0, if you can share that information!

zigandazag:
yes it was working in 10.17.0

ljelonek:
I can confirm that it was doing fine on 10.17.0. I am currently facing an issue where it is impossible to sync blobs for local storage (emulated one). As you probably imagine, it is hard to ditch ports when syncing on localhost :)

This is kinda blocker for me, thus I will appreciate fast resolution.
Here is a link to the previous version if someone is interested:
https://azcopyvnext.azureedge.net/release20230123/azcopy_windows_amd64_10.17.0.zip

`Cannot perform sync due to error: failed to resolve destination: lookup 127.0.0.1:10000: no such host`

siminsavani-msft:
Thanks all for confirming! A fix will be made and this thread will be updated accordingly.

siminsavani-msft:
Hi there! The issue has been resolved in the latest version of AzCopy (https://github.com/Azure/azure-storage-azcopy/releases/tag/v10.18.1). Will be closing this issue, but please let us know if you are still experiencing issues!

PR #2143, opened 2023-04-04
Use hostname instead of hostname:port

nakulkar-msft:


vibhansa-msft (approved):

vibhansa-msft:
Add changelog to mark the bug fix

common/util.go
```diff
@@ -15,6 +15,6 @@ func VerifyIsURLResolvable(url_string string) (error) {
 		return err
 	}
 
-	_, err = net.LookupIP(url.Host)
+	_, err = net.LookupIP(url.Hostname())
```
gapra-msft, line 18: Is there a unit test for this method we can add some cases to? If not might be a good idea to add a few tests?

siminsavani-msft:
Fixes: https://github.com/Azure/azure-storage-azcopy/issues/2140

common/util.go
siminsavani-msft: +1

PR #2147, opened 2023-04-04
Use folder deletion manager to manage the deletion of extra folders on the destination in Sync

adreed-msft:
Adjusted the remote & local file deleters to take advantage of a folder deletion manager when relevant.


cmd/syncProcessor.go
```diff
@@ -325,9 +361,33 @@ func (b *remoteResourceDeleter) delete(object StoredObject) error {
...
+
+		objectURL := b.getObjectURL(object)
+		b.folderManager.RecordChildExists(&objectURL)
+
+		b.folderManager.RequestDeletion(&objectURL, func(ctx context.Context, logger common.ILogger) bool {
+			var err error
+			switch b.targetLocation {
+			case common.ELocation.Blob():
+				blobURL := azblob.NewBlobURL(objectURL, b.p)
+				// HNS endpoint doesn't like delete snapshots on a directory
```
gapra-msft, line 376: is this comment relevant here?

gapra-msft (approved):

cmd/syncProcessor.go
adreed-msft: Probably not anymore.

siminsavani-msft (approved):

Issue #2141, opened 2023-04-03
Since 10.18, copy between two file shares create top level dir


Aohzan:
### Which version of the AzCopy was used? 
10.18.0

### Which platform are you using? (ex: Windows, Mac, Linux)
Linux

### What command did you run?
```bash
azcopy copy \
                --overwrite=ifSourceNewer \
                --recursive=true \
                --preserve-smb-info=true \
                --preserve-permissions=true \
                --exclude-path=tmp \
                "https://originstorageaccount.file.core.windows.net/myfileshare/?token" \
                "https://deststorageaccount.file.core.windows.net/myfileshare/?token"
```

I tried with remove trailing slashes but same results

### What problem was encountered?
the origin fileshare has something like
```
root
  |- subdir
     |- file1
     |- file2
```
after 10.18 upgrade, copy leads to
```
root
  |- myfileshare
      |- subdir
         |- file1
         |- file2
```

### How can we reproduce the problem in the simplest way?
run the copy command and see if a top level directory with the share name is created

### Have you found a mitigation/solution?
Downgrading to 10.17

siminsavani-msft:
Hi @Aohzan ! Since you are attempting to preserve permissions with this flag `--preserve-permissions=true`, AzCopy won't strip the top directory, which creates the top level 'myfileshare' directory as you indicated.

Aohzan:
Hello @siminsavani-msft, `--as-subdir=false` fix the problem indeed, thank you

adreed-msft:
> Hello @siminsavani-msft, `--as-subdir=false` fix the problem indeed, thank you

Great to hear it could temporarily mitigate the issue at hand. I'll be reproducing & fixing this for release in 10.19.0 shortly.

adreed-msft:
This has been fixed in AzCopy v10.18.1

PR #2150, opened 2023-04-05
Fix container->container behaviour w/ preserve-permissions

adreed-msft:
In 10.18, we incorrectly adjusted the behavior of container->container w/ preserve-permissions to no longer strip the top directory. The behavior has been adjusted to use as-subdir to resolve the regression.

Linked to #2141

### DIFF
--- a/e2etest/zt_preserve_smb_properties_test.go
+++ b/e2etest/zt_preserve_smb_properties_test.go
@@ -306,3 +306,34 @@ func TestProperties_SMBTimes(t *testing.T) {
 		"",
 	)
 }
+
+func TestProperties_EnsureContainerBehavior(t *testing.T) {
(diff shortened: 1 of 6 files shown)


e2etest/zt_preserve_smb_properties_test.go
```diff
@@ -306,3 +306,34 @@ func TestProperties_SMBTimes(t *testing.T) {
 		"",
 	)
 }
+
+func TestProperties_EnsureContainerBehavior(t *testing.T) {
```
gapra-msft, line 310: is there any test we need to add for datalake? or is this just for smb permissions?

siminsavani-msft (approved):

gapra-msft (approved):

PR #2154, opened 2023-04-12
Added support for trailing dot

gapra-msft:


cmd/removeEnumerator.go
```diff
@@ -75,6 +71,9 @@ func newRemoveEnumerator(cca *CookedCopyCmdArgs) (enumerator *CopyEnumerator, er
 	if cca.IncludeAfter != nil {
 		filters = append(filters, &IncludeAfterDateFilter{Threshold: *cca.IncludeAfter})
 	}
+	if !cca.trailingDot && (cca.FromTo.To() == common.ELocation.File() || cca.FromTo.From() == common.ELocation.File()) {
+		filters = append(filters, &TrailingDotFilter{})
```
gapra-msft, line 75: open question: how to deal with files that have trailing dot if trailing dot is off

cmd/zc_traverser_file.go
```diff
@@ -34,6 +34,8 @@ import (
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 )
 
+const trailingDotErrMsg = "File share contains file/directory %s with a trailing dot but the trailing dot flag was not set to true. AzCopy will automatically filter this file/directory out to prevent data corruption."
```
adreed-msft, line 37: Given we're now raw yes/no and not strip/include/filter, is this message accurate or necessary?

cmd/zc_traverser_file.go
```diff
@@ -67,6 +70,10 @@ func (t *fileTraverser) Traverse(preprocessor objectMorpher, processor objectPro
 
 	// if not pointing to a share, check if we are pointing to a single file
 	if targetURLParts.DirectoryOrFilePath != "" {
+		if t.trailingDot != common.ETrailingDotOption.Enable() && strings.HasSuffix(targetURLParts.DirectoryOrFilePath, ".") {
```
adreed-msft, line 73: Ditto

cmd/zc_traverser_file.go
```diff
@@ -194,14 +201,24 @@ func (t *fileTraverser) Traverse(preprocessor objectMorpher, processor objectPro
 				return fmt.Errorf("cannot list files due to reason %s", err)
 			}
 			for _, fileInfo := range lResp.FileItems {
+				// These conditions are to prevent a GetProperties from returning a 404 when trailing dot is turned off.
```
adreed-msft, line 204: Ditto

e2etest/factory.go
```diff
@@ -60,8 +63,19 @@ func (TestResourceFactory) GetFileServiceURL(accountType AccountType) azfile.Ser
 	if err != nil {
 		panic(err)
 	}
-	pipeline := azfile.NewPipeline(credential, azfile.PipelineOptions{})
-	return azfile.NewServiceURL(*u, pipeline)
+	// Closest to API goes first; closest to the wire goes last
+	f := []pipeline.Factory{
```
adreed-msft, line 67: Can we not re-use the integrated new pipeline functions in AzCopy? Is there a good reason to duplicate this?

e2etest/helpers.go
```diff
@@ -257,7 +258,18 @@ func newNullFolderCreationTracker() ste.FolderCreationTracker {
 func generateParentsForAzureFile(c asserter, fileURL azfile.FileURL) {
 	accountName, accountKey := GlobalInputManager{}.GetAccountAndKey(EAccountType.Standard())
 	credential, _ := azfile.NewSharedKeyCredential(accountName, accountKey)
-	err := ste.AzureFileParentDirCreator{}.CreateParentDirToRoot(ctx, fileURL, azfile.NewPipeline(credential, azfile.PipelineOptions{}), newNullFolderCreationTracker())
+	// Closest to API goes first; closest to the wire goes last
```
adreed-msft, line 261: Ditto

e2etest/factory.go
```diff
@@ -146,7 +160,18 @@ func (TestResourceFactory) GetFileShareULWithSAS(c asserter, accountType Account
 	fullURL, err := url.Parse(rawURL)
 	c.AssertNoErr(err)
 
-	return azfile.NewShareURL(*fullURL, azblob.NewPipeline(azblob.NewAnonymousCredential(), azblob.PipelineOptions{}))
+	// Closest to API goes first; closest to the wire goes last
```
adreed-msft, line 163: Ditto

e2etest/runner.go
```diff
@@ -107,6 +107,9 @@ func (t *TestRunner) SetAllFlags(p params, o Operation) {
 	set("is-object-dir", p.isObjectDir, false)
 	set("debug-skip-files", strings.Join(p.debugSkipFiles, ";"), "")
 	set("check-md5", p.checkMd5.String(), "FailIfDifferent")
+	if p.trailingDot != nil {
```
adreed-msft, line 110: This pattern only really makes sense if the zero-value is not the default; does p.trailingDot need to be a pointer for some reason?

adreed-msft (approved):

e2etest/runner.go
gapra-msft: This is because I need a way to test that if nothing is passed, the trailing dot files still work.

nakulkar-msft (approved):

cmd/zc_traverser_file.go
gapra-msft: You're right, let me fix the verbiage.

e2etest/factory.go
gapra-msft: I can reuse the pipeline creation logic in ste

e2etest/runner.go
adreed-msft: Right; but that's the default behavior. The set() function doesn't apply the parameter if it's default, iirc.

cmd/zc_traverser_file.go
gapra-msft: removed the nil but I will leave the log here as a warning

cmd/zc_traverser_file.go
gapra-msft: this is just a log for warning, so when customers experience weird behavior, they can look at the logs and see why

e2etest/runner.go
gapra-msft: changed this to use set's behavior in the correct way

PR #2157, opened 2023-04-14
Disable lookup before starting traverser

nakulkar-msft:


common/util.go
```diff
@@ -1,20 +1,18 @@
...
-	"net"
-	"net/url"
-)
-
 var AzcopyJobPlanFolder string
 var AzcopyCurrentJobLogger ILoggerResetable
 
 
 func VerifyIsURLResolvable(url_string string) (error) {
+	return nil
```
gapra-msft, line 8: Should we add a link to the issue or comment to why we are "removing" this check?

gapra-msft (approved):

adreed-msft (approved):

PR #2163, opened 2023-04-16
Reuse pipeline in list traverser

nakulkar-msft:
The old code would create http.Client per entry in list, which would exhaust file descriptors in certain environments.


adreed-msft (approved):

cmd/zc_traverser_list.go
```diff
@@ -59,8 +60,12 @@ func (l *listTraverser) Traverse(preprocessor objectMorpher, processor objectPro
 			glcm.Info(fmt.Sprintf("Skipping %s due to error %s", childPath, err))
 			continue
 		}
-
+/*
```
gapra-msft, line 63: what is this comment for?

gapra-msft (approved):

siminsavani-msft (approved):

tasherif-msft (approved):

PR #2165, opened 2023-04-18
Improve --from-to docs

### LINKED ISSUE #2153: --from-to provides insufficent failure information
##### Note: The version is visible when running AzCopy without any argument

AzCopy 10.18.0

Windows

.\azcopy.exe copy $ChildSource $ChildDestSASstring --recursive --overwrite=false  --preserve-smb-permissions  --preserve-smb-info --from-to localfiles

When using the --from-to switch, the failure information provides insufficient information on acceptable options. 

The error provided is "fatal: from-to argument required, PipeBlob (upload) or BlobPipe (download) is acceptable" . However, there are numerous from-to option pairs available. 

From reading the .go file at https://github.com/Azure/azure-storage-azcopy/blob/main/cmd/copy.go the following options appear to be defined:

BenchmarkBlob
BenchmarkBlobFS
BenchmarkFile
BlobBlob
BlobFile
BlobFSLocal
BlobFSNone
BlobFSTrash
BlobLocal
BlobNone
BlobPipe
BlobTrash
FileBlob
FileFile
FileLocal
FileNone
FileTrash
GCPBlob
LocalBlob
LocalBlobFS
LocalFile
PipeBlob
S3Blob

Use AZCopy with the --from-to switch, with any invalid from-to response.

No workaround available as the issue is with an insufficient error, not an application function.

adreed-msft:
Resolves #2153

Per Norm's suggestion, list acceptable values as opposed to just explaining format

### DIFF
--- a/cmd/validators.go
+++ b/cmd/validators.go
@@ -22,7 +22,9 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/JeffreyRichter/enum/enum"
 	"net/url"
+	"reflect"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


cmd/validators.go
```diff
@@ -51,9 +53,36 @@ func ValidateFromTo(src, dst string, userSpecifiedFromTo string) (common.FromTo,
 	return userFromTo, nil
 }
 
-const fromToHelpText = "Valid values are two-word phases of the form BlobLocal, LocalBlob etc.  Use the word 'Blob' for Blob Storage, " +
-	"'Local' for the local file system, 'File' for Azure Files, and 'BlobFS' for ADLS Gen2. " +
-	"If you need a combination that is not supported yet, please log an issue on the AzCopy GitHub issues list."
+const fromToHelpFormat = "Valid FromTo are pairs of Source-Destination words (e.g. BlobLocal, BlobBlob) that specify the source and destination resource types. All valid FromTos are: %s"
```
nakulkar-msft, line 56: We may want to convet that from-to is not required, unless
1. One end is pipe
2. The endpoint does not conform to the normal naming standards.

The issue #2153 is because we parsed source and destination wrong. And even with any valid combo of from-to we would still do that. We could instead make the error clear ( perhaps if we assume user is piping we can throw what the source and destination azcopy saw.)

For case2, I think we should limit to the basic and common options - Blob, File, Local - that fit in for the most customers using a custom domain. These might be confuse the users.

We can instead try improving the Ux  logic for piping and skip modifying from-to help altogether.

cmd/validators.go
```diff
@@ -51,9 +53,36 @@ func ValidateFromTo(src, dst string, userSpecifiedFromTo string) (common.FromTo,
...
+		case common.ELocation.Benchmark(),
+			common.ELocation.None(),
+			common.ELocation.Unknown():
+			return false
+		default:
+			return true
+		}
+	}
+
+	enum.GetSymbols(reflect.TypeOf(common.EFromTo), func(enumSymbolName string, enumSymbolValue interface{}) (stop bool) {
```
gapra-msft, line 72: Does this allow S3 and GCP to be printed as To?

siminsavani-msft (approved):

cmd/validators.go
adreed-msft: The full line reads as follows:

```
Valid FromTo are pairs of Source-Destination words (e.g. BlobLocal, BlobBlob) that specify the source and destination resource types. All valid FromTos are: BlobBlob, BlobFSLocal, BlobFile, BlobLocal, BlobPipe, FileBlob, FileFile, FileLocal, FilePipe, GCPBlob, LocalBlob, LocalBlobFS, LocalFile, PipeBlob, PipeFile, S3Blob
```

adreed-msft: It only iterates over the pre-specified FromTos; it does not create any from-tos itself.

cmd/validators.go
adreed-msft: I think we should still list acceptable values for From-To. I'll indicate that it is optional, and to nudge AzCopy in the right direction.

@normesta thoughts?

gapra-msft (approved):

PR #2190, opened 2023-05-01
[1] Only swap DFS to blob when immediately relevant

adreed-msft:
This PR eliminates the isHNStoHNS flag, instead opting to acknowledge the real location we're working with, and to intelligently swap to the blob endpoint only when necessary.

This eliminates a lot of hackiness going on surrounding the flag and allows us to check without guessing that we're working with an HNS account.

This PR is the base for a second PR involving setting POSIX properties on DFS.


cmd/credentialUtil.go
```diff
@@ -681,7 +673,20 @@ func createBlobPipeline(ctx context.Context, credInfo common.CredentialInfo, log
 		nil,
 		ste.NewAzcopyHTTPClient(frontEndMaxIdleConnectionsPerHost),
 		nil, // we don't gather network stats on the credential pipeline
-	), nil
+	)
+}
+
+func createBlobPipeline(ctx context.Context, credInfo common.CredentialInfo, logLevel pipeline.LogLevel) (pipeline.Pipeline, error) {
```
gapra-msft, line 679: Why did we separate these cases?

cmd/validators.go
```diff
@@ -102,41 +102,22 @@ func inferFromTo(src, dst string) common.FromTo {
...
-		return common.EFromTo.S3Blob()
-	case srcLocation == common.ELocation.Benchmark() && dstLocation == common.ELocation.Blob():
-		return common.EFromTo.BenchmarkBlob()
-	case srcLocation == common.ELocation.Benchmark() && dstLocation == common.ELocation.File():
-		return common.EFromTo.BenchmarkFile()
-	case srcLocation == common.ELocation.Benchmark() && dstLocation == common.ELocation.BlobFS():
-		return common.EFromTo.BenchmarkBlobFS()
-	case srcLocation == common.ELocation.GCP() && dstLocation == common.ELocation.Blob():
-		return common.EFromTo.GCPBlob()
+	out := common.EFromTo.Unknown()
```
gapra-msft, line 105: can we maybe add a comment explaining the logic below since it's a little unintuitive

cmd/zc_enumerator.go
```diff
@@ -486,18 +485,39 @@ func InitResourceTraverser(resource common.ResourceString, location common.Locat
 
 		recommendHttpsIfNecessary(*resourceURL)
 
-		bfsURL := azbfs.NewBfsURLParts(*resourceURL)
+		// Convert BlobFS pipeline to blob-compatible pipeline
+		var credElement azblob.Credential
+		if credential.CredentialType == common.ECredentialType.SharedKey() {
```
gapra-msft, line 490: I thought AzCopy doesnt support shared key?

cmd/zt_copy_s2smigration_test.go
```diff
@@ -1329,7 +1329,7 @@ func (s *cmdIntegrationSuite) TestCopyWithDFSResource(c *chk.C) {
 		c.Assert(err, chk.IsNil)
 
 		// validate that the right number of transfers were scheduled
-		c.Assert(len(mockedRPC.transfers), chk.Equals, 1)
+		c.Assert(len(mockedRPC.transfers), chk.Equals, 2)
```
gapra-msft, line 1332: is this to account for the root as well now?

common/fe-ste-models.go
```diff
@@ -591,15 +598,23 @@ func (ft *FromTo) From() Location {
...
 func (ft *FromTo) IsUpload() bool {
-	return ft.From().IsLocal() && ft.To().IsRemote()
+	return ft.From().IsLocal() && ft.To().IsRemote() && ft.To() != ELocation.None() && ft.To() != ELocation.Unknown()
+}
+
+func (ft *FromTo) IsDelete() bool {
+	return ft.To() == ELocation.Unknown()
+}
+
+func (ft *FromTo) IsSetProperties() bool {
```
gapra-msft, line 616: Should this check that From isnt S3 or GCP? Same with Delete

ste/sender-blobFolders.go
```diff
@@ -214,9 +214,10 @@ func (b *blobFolderSender) SetFolderProperties() error {
 }
 
 func (b *blobFolderSender) DirUrlToString() string {
-	url := b.destination.URL()
-	url.RawQuery = ""
-	return url.String()
+	uri, _ := url.Parse(b.jptm.Info().Destination)
+	uri.RawPath = ""
```
gapra-msft, line 218: Why strip off the path name here?

gapra-msft (commented):
made some comments

cmd/credentialUtil.go
adreed-msft: Technically, we could just add proper support for shared key to the blob endpoint and not hack around it. But I don't think we want to encourage customers to use shared key in the first place; so this was to avoid adding support for that in directly.

cmd/zt_copy_s2smigration_test.go
adreed-msft: Because DFS is no longer treated as blob for S2S, folders are included by default. This change tracks the additional folder.

common/fe-ste-models.go
adreed-msft: We won't permit set properties this way anyway as they're not registered in the FromTo Enum.

ste/sender-blobFolders.go
adreed-msft: Keep in mind that RawPath is only to specify whether characters are encoded; namely for characters that don't need to be encoded.

This just normalizes the path.

cmd/zc_enumerator.go
adreed-msft: It does, actually! Only for DFS though.

gapra-msft (approved):

tasherif-msft (approved):

PR #2192, opened 2023-05-02
Return error from pacer if block cannot be scheduled

nakulkar-msft:


vibhansa-msft (approved):

vibhansa-msft:
Do we have any ways to simulate this in our testing pipelines and validate that it works fine.

nakulkar-msft:
> Do we have any ways to simulate this in our testing pipelines and validate that it works fine.

Pacer is totally untested. I'll write a new test case in a separate review

gapra-msft (approved):

tasherif-msft (approved):

PR #2200, opened 2023-05-04
Gapra/logging

gapra-msft:


common/clientFactory.go
```diff
@@ -53,25 +54,31 @@ func createClient[T, U any](callbacks newClientCallbacks[T, U], u string, credIn
...
 	case ECredentialType.Anonymous():
-		return callbacks.NoCredential(u, options)
+		client, err = callbacks.NoCredential(u, options)
 	default:
-		credOpOptions.panicError(fmt.Errorf("invalid state, credential type %v is not supported", credInfo.CredentialType))
-		return nil, fmt.Errorf("invalid state, credential type %v is not supported", credInfo.CredentialType)
+		err = fmt.Errorf("invalid state, credential type %v is not supported", credInfo.CredentialType)
 	}
+	if err != nil {
+		credOpOptions.panicError(err)
```
adreed-msft, line 73: PanicError will either call `panic` or `o.panic` depending on which one is available.

Typically, we try to avoid `panic` unless it's a sanity check; Realistically, `o.panic` would safely log the error (both in the job log and stdout) and then proceed to cleanly exit AzCopy.

So, I suppose the correct questions here are:

1. Are all of these errors sanity checks?
2. If not, How would we safely exit from here?

adreed-msft, line 73: PanicError will either call `panic` or `o.panic` depending on which one is available.

Typically, we try to avoid `panic` unless it's a sanity check; Realistically, `o.panic` would safely log the error (both in the job log and stdout) and then proceed to cleanly exit AzCopy.

So, I suppose the correct questions here are:

1. Are all of these errors sanity checks?
2. If not, How would we safely exit from here?

adreed-msft (approved):

common/clientFactory.go
gapra-msft: This is a sanity check. The SDK returns an error on client construction but it is always nil in the codebase.

gapra-msft: So basically this will catch either client construction errors (which should always be nil) or wrong credential errors which shouldnt happen anyway

ste/xferLogPolicy.go
```diff
@@ -438,47 +437,20 @@ func newLogPolicy(options LogOptions) policy.Policy {
...
-		"retry-after":                   {},
-		"server":                        {},
-		"traceparent":                   {},
-		"transfer-encoding":             {},
-		"user-agent":                    {},
-		"www-authenticate":              {},
-		"x-ms-request-id":               {},
-		"x-ms-client-request-id":        {},
-		"x-ms-return-client-request-id": {},
+	disallowedHeaders := map[string]struct{}{
```
gapra-msft, line 440: changed this from allow to disallow to make header sanitization easier.

cmd/zc_enumerator.go
```diff
@@ -432,8 +432,8 @@ func InitResourceTraverser(resource common.ResourceString, location common.Locat
 
 		recommendHttpsIfNecessary(*resourceURL)
 
-		if ctx == nil || p == nil {
-			return nil, errors.New("a valid credential and context must be supplied to create a blob traverser")
+		if ctx == nil {
```
nakulkar-msft, line 435: All the case statements, including at line 416, require a context. This check can be advanced to the start of function.

nakulkar-msft (approved):

cmd/zc_enumerator.go
gapra-msft: hmm it looks like benchmark and local don't require a context

gapra-msft: We can consider this after migrating all 3 services. This isn't too big of a deal in terms of changes

PR #2228, opened 2023-05-24
Add error when chunkSize is greater than the strict memory limit of the CacheLimiter.

mkrautz:
Add error when chunkSize is greater than the strict memory limit of the CacheLimiter.

Scenario:

    export AZCOPY_BUFFER_GB=5
    # Copy a single 5GB file to an Azure Blob Storage Container with a block size of 4000MB
    ./acopy [...] --block-size-mb=4000 [...]

Since the chunk size in that case is greater than the strict memory limit (75% of 5GB) this would end up in a deadlock: CacheLimiter.TryAdd would loop forever, waiting for memory to be freed (which it never will).

This patch tries to remedy that by making it an error if a chunk is greater than the strict memory limit of the associated CacheLimiter.

### DIFF
--- a/common/cacheLimiter.go
+++ b/common/cacheLimiter.go
@@ -27,6 +27,18 @@ import (
 	"time"
 )
 
+// The percentage of a CacheLimiter's Limit that is considered
+// the strict limit.
+var cacheLimiterStrictLimitPercentage = float32(0.75)
+// Rationale for the level of the strict limit: as at Jan 2018, we are using 0.75 of the total as the strict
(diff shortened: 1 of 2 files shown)


gapra-msft (approved):

adreed-msft (approved):
I think this is a good change; we should also do this with cap-mbps (though, we could partially move chunks in that scenario with a little refactoring)

siminsavani-msft (approved):

tasherif-msft (approved):
thanks for the contribution

nakulkar-msft (approved):

PR #2238, opened 2023-05-30
Migrated to the new azidentity library

gapra-msft:
azidentity library used for SPN and MSI

custom TokenCredential created for device code and token store


tasherif-msft (approved):

nakulkar-msft (approved):

common/oauthTokenManager.go
```diff
@@ -148,309 +131,92 @@ func (uotm *UserOAuthTokenManager) GetTokenInfo(ctx context.Context) (*OAuthToke
...
+	oAuthTokenInfo := &OAuthTokenInfo{
+		Identity:     true,
+		IdentityInfo: identityInfo,
 	}
 
-	return &oAuthTokenInfo, nil
+	return uotm.validateAndPersistLogin(oAuthTokenInfo, persist)
 }
 
 // SecretLogin is a UOTM shell for secretLoginNoUOTM.
```
adreed-msft, line 177: This comment isn't entirely true anymore.

common/oauthTokenManager.go
```diff
@@ -627,7 +394,7 @@ const TokenRefreshSourceTokenStore = "tokenstore"
 
 // OAuthTokenInfo contains info necessary for refresh OAuth credentials.
 type OAuthTokenInfo struct {
-	azcore.TokenCredential
+	azcore.TokenCredential `json:"-"`
```
adreed-msft, line 397: So the current flow for login is `azcopy login` -> `azcopy whatever` (outside of one-shot login). If we're not persisting the refresh info between runs, then interactive login is completely broken. Are we saving that info some other way?

common/oauthTokenManager.go
```diff
@@ -627,7 +394,7 @@ const TokenRefreshSourceTokenStore = "tokenstore"
...
 	adal.Token
```
adreed-msft, line 398: What is the purpose of maintaining two tokens?

adreed-msft (approved):

common/oauthTokenManager.go
gapra-msft: the adal.Token needs to be used until we can entirely get off of Track 1.5 and it's sole purpose is to hold the refresh token for device code cred

common/oauthTokenManager.go
gapra-msft: Fixed the comment.

common/oauthTokenManager.go
gapra-msft: I took a look at the code again and we should be persisting the token between runs. We can discuss the code changes further if you'd like.

PR #2248, opened 2023-06-08
Resolved a few trailing dot issues

gapra-msft:
1. Fixed an issue where source trailing dot header was passed when source for a S2S copy is not file service
2. Gracefully handle File Share trailing dot paths to Windows/Blob (that do not support trailing dot) by skipping such files
3. Allow trailing dot option to be ignored instead of erroring out in situations it does not apply.


cmd/zc_pipeline_init.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 )
 
-func InitPipeline(ctx context.Context, location common.Location, credential common.CredentialInfo, logLevel pipeline.LogLevel, trailingDot common.TrailingDotOption) (p pipeline.Pipeline, err error) {
+func InitPipeline(ctx context.Context, location common.Location, credential common.CredentialInfo, logLevel pipeline.LogLevel, trailingDot common.TrailingDotOption, from common.Location) (p pipeline.Pipeline, err error) {
```
nakulkar-msft, line 11: Nit: Considering we are already passing 'To' we could just send in entire FromTo() and reduce number of args

nakulkar-msft (approved):

cmd/zc_traverser_file.go
```diff
@@ -66,10 +68,28 @@ func (t *fileTraverser) getPropertiesIfSingleFile() (*azfile.FileGetPropertiesRe
 }
 
 func (t *fileTraverser) Traverse(preprocessor objectMorpher, processor objectProcessor, filters []ObjectFilter) (err error) {
+	invalidBlobOrWindowsName := func(path string) bool {
```
gapra-msft, line 71: TODO: should we only do this check when trailing dot is enabled? If explicitly disabled I think we can skip this error message

cmd/zc_pipeline_init.go
gapra-msft: This causes some issues when the location is Local and such and there is one spot where the location isn't passed as a FromTo. Regardless, this code will be removed with the migration, so shouldnt matter if we keep 1 vs 2 params here.

cmd/zc_traverser_file.go
adreed-msft: Over in copyEnumeratorInit.go there is a function, `pathEncodeRules`. That feels like a decent place to be concerned about this instead of here? It'd be able to apply to more use-cases than just the files enumerator.

cmd/zc_traverser_list.go
```diff
@@ -93,7 +93,7 @@ func (l *listTraverser) Traverse(preprocessor objectMorpher, processor objectPro
 func newListTraverser(parent common.ResourceString, parentType common.Location, credential *common.CredentialInfo,
 	ctx *context.Context, recursive bool, handleSymlinks common.SymlinkHandlingType, getProperties bool, listChan chan string,
 	includeDirectoryStubs bool, incrementEnumerationCounter enumerationCounterFunc, s2sPreserveBlobTags bool,
-	logLevel pipeline.LogLevel, cpkOptions common.CpkOptions, syncHashType common.SyncHashType, preservePermissions common.PreservePermissionsOption, trailingDot common.TrailingDotOption, p pipeline.Pipeline) ResourceTraverser {
+	logLevel pipeline.LogLevel, cpkOptions common.CpkOptions, syncHashType common.SyncHashType, preservePermissions common.PreservePermissionsOption, trailingDot common.TrailingDotOption, p pipeline.Pipeline, destination *common.Location) ResourceTraverser {
```
adreed-msft, line 96: As soon as the track 2 migration is merged, we need to have a discussion as a team about how we want to handle passing around flags in AzCopy in the future. 20+ params is unweildly and often an options bag would at least be more maintainable, but still suffer from the issue of mentally having to calculate all the layers a flag has to get copied through before use.

adreed-msft (approved):

tasherif-msft (approved):

cmd/zc_traverser_file.go
gapra-msft: Would that only apply to copy enumeration then? or also sync?

gapra-msft: Also adding this check in the traverser follows the behavior of S3 sources. PathEncodeRules seems more like a piece of code that encodes if a path is unsafe instead of skipping the path altogether

adreed-msft: NVM then. Trailing dot is something we only really have to deal with from S3 and from Files, anyway. It's fine to implement it as a specific thing for Files.

PR #2250, opened 2023-06-14
Force small blobs to use PutBlob for all sources

nakulkar-msft:


ste/sender-blockBlobFromURL.go
```diff
@@ -69,9 +69,8 @@ func (c *urlToBlockBlobCopier) GenerateCopyFunc(id common.ChunkID, blockIndex in
 		setPutListNeed(&c.atomicPutListIndicator, putListNotNeeded)
 		return c.generateCreateEmptyBlob(id)
 	}
-	// Small blobs from all sources will be copied over to destination using PutBlobFromUrl with the exception of files
```
gapra-msft, line 72: Should we add a test that then checks the block list after validation? should be empty I guess?

murat8603973081:
This issue looks interesting - I'll try to provide some solutions and submit a PR.

siminsavani-msft (approved):

tasherif-msft (approved):

PR #2258, opened 2023-06-19
Support follow-symlinks SMB properties

adreed-msft:
Due to the way symlink following worked (followed in the traverser, then submitted to the plan file as the destination), the flag check may simply be disabled with the consequence being that the actual destination file is checked (in alignment with the suggested behaviour from the files team).

~~I still need to look into if we can run the test suite as an administrator, as it would be required for testing this functionality, since creating Symlinks on Windows requires admin permissions~~

No, we can't test this automatically unless we self-host a Windows agent.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -676,10 +676,6 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 
 	cooked.IncludeDirectoryStubs = raw.includeDirectoryStubs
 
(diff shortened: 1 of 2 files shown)


gapra-msft:
Can we add a comment somewhere indicating how to set up and test this manually? That way if there is a bug or something at some point, we can easily follow those steps and debug

murat8603973081:
I noticed there are some typos in your documentation - if you're okay with it, I can submit a PR to fix them.

gapra-msft (approved):

nakulkar-msft (approved):

PR #2276, opened 2023-06-30
chore: remove refs to deprecated io/ioutil

testwill:


gapra-msft:
Hi, these changes are touching autogenerated files so I don't think we can accept this PR. This issue will be resolved once AzCopy upgrades to the latest Track 2 SDK for datalake when it is ready.

PR #2298, opened 2023-07-14
[DO NOT MERGE]Retry custom request

nakulkar-msft:
Retry http.Conflict on destination pipeline.

### DIFF
--- a/common/version.go
+++ b/common/version.go
@@ -1,6 +1,6 @@
 package common
 
-const AzcopyVersion = "10.19.0"
+const AzcopyVersion = "10.19.0-preview"
 const UserAgent = "AzCopy/" + AzcopyVersion
 const S3ImportUserAgent = "S3Import " + UserAgent
 const GCPImportUserAgent = "GCPImport " + UserAgent
--- a/ste/xferRetrypolicy.go
+++ b/ste/xferRetrypolicy.go
@@ -6,6 +6,7 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
+	"os"
 	"strconv"
 	"time"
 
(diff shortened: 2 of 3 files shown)


ste/xferRetrypolicy.go
```diff
@@ -220,6 +223,12 @@ func NewBFSXferRetryPolicyFactory(o XferRetryOptions) pipeline.Factory {
 				switch {
 				case err == nil:
 					action = "NoRetry: successful HTTP request" // no error
+				case response.Response() != nil && response.Response().StatusCode == http.StatusConflict:
+					if o.RetryHTTPErrorConflict && os.Getenv("AZCOPY_DISABLE_CUSTOM_RETRY") == "" {
```
vibhansa-msft, line 227: Why do we need to check 2 flags here if config is set just retry

nakulkar-msft: We want to retry only on destination pipleine and not source.

vibhansa-msft: o.RetryHTTPErrorConflict :  This represent the flag which is true as harcoded value for now
os.Getenv("AZCOPY_DISABLE_CUSTOM_RETRY") :  What does this signify ? Appears we check this shall be unset to make this retry happen.

nakulkar-msft: If you want to disable retry on 409, you can set this environment var

adreed-msft: I don't think we need the global variable right now. The custom error code retries were intended to be something we'd address in 10.21, right?

If so, then there's not really a need to implement a flag like this right now, especially since it's not really getting used?

adreed-msft (approved):

common/version.go
```diff
@@ -1,6 +1,6 @@
 package common
 
-const AzcopyVersion = "10.19.0"
+const AzcopyVersion = "10.19.0-preview"
```
gapra-msft, line 3: this should be 10.20.0-preview right? or 10.19.1

siminsavani-msft (approved):

common/version.go
nakulkar-msft: This patch is over 10.19 on main.

ste/xferRetrypolicy.go
nakulkar-msft: The flag is not to support a set of error codes. But if the user has to disable retrying on 409 for some reason, it is easier to toggle env than switch builds

PR #2310, opened 2023-07-25
Deprecate login command

nakulkar-msft:
```
PS C:\Users\<>\go\src\github.com\Azure\azure-storage-azcopy> .\azure-storage-azcopy.exe login
INFO: 'azcopy login' command will be deprecated starting release 10.22. Use auto-login instead. Visit https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory#authorize-without-a-secret-store  to know more.
INFO: If you set an environment variable by using the command line, that variable will be readable in your command line history. Consider clearing variables that contain credentials from your command line history.  To keep variables from appearing in your history, you can use a script to prompt the user for their credentials, and to set the environment variable.
```

```
PS C:\Users\<>\go\src\github.com\Azure\azure-storage-azcopy> .\azure-storage-azcopy.exe logout
INFO: 'azcopy logout' command will be deprecated starting release 10.22. Use auto-login instead. Visit https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory#authorize-without-a-secret-store  to know more.
```

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -23,6 +23,7 @@ package cmd
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/Azure/azure-storage-azcopy/v10/common"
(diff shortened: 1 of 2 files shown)


adreed-msft (approved):

weshaggard:
@nakulkar-msft Which section on the doc page is the anchor "authorize-without-a-secret-store" supposed to point at in the documentation? The anchor doesn't seem to exist https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory#authorize-without-a-secret-store

PR #2316, opened 2023-07-27
Support asterisk on list

nakulkar-msft:


cmd/list.go
```diff
@@ -84,14 +85,57 @@ func (raw *rawListCmdArgs) parseProperties(rawProperties string) []validProperty
 	return parsedProperties
 }
 
-func (raw rawListCmdArgs) cook() (cookedListCmdArgs, error) {
+// returns result of stripping and if striptopdir is enabled
+// if nothing happens, the original source is returned
+func (raw *rawListCmdArgs) stripTrailingWildcardOnRemoteSource(location common.Location) (result string, stripTopDir bool, err error) {
```
gapra-msft, line 90: @nakulkar-msft  can we wait until adele's new e2e test framework is merged and add e2e tests for this and then support this feature?

siminsavani-msft: +1

siminsavani-msft (changes requested):

cmd/list.go
siminsavani-msft: +1

PR #2340, opened 2023-08-16
Feature/remove hardcoded lease saver.

afekson:
AzCopy should not be aware of any external files/folder which it needs to preserve.
This code deletes the hard-coded movers lease blob protetion


afekson:
wrong target fork

PR #2343, opened 2023-08-16
Migrated datalake tests in cmd folder

gapra-msft:


siminsavani-msft (approved):

cmd/zt_scenario_helpers_for_test.go
```diff
@@ -510,22 +512,21 @@ func (scenarioHelper) generateShareFilesFromList(a *assert.Assertions, shareClie
 	time.Sleep(time.Second * 3)
 }
 
-func (scenarioHelper) generateBFSPathsFromList(a *assert.Assertions, filesystemURL azbfs.FileSystemURL, fileList []string) {
+func (scenarioHelper) generateBFSPathsFromList(a *assert.Assertions, fsClient *filesystem.Client, fileList []string) {
 	for _, p := range fileList {
-		file := filesystemURL.NewRootDirectoryURL().NewFileURL(p)
+		// TODO : RootDirectoryClient in datalake SDK
```
tasherif-msft, line 517: can't you just create a file client directly under the filesystem? why do you need a root dir client?

tasherif-msft (approved):

cmd/zt_scenario_helpers_for_test.go
gapra-msft: probably not necessary actually
I remember in net and java we added it then removed it due to potential weird behavior

PR #2347, opened 2023-08-22
Fix ADLS Gen 2 ACL tests & copying

adreed-msft:
When updating the DFS handling code to actually record all dfs endpoints as BlobFS and handle the swapping at the time it's needed, some faulty testing caused us to believe that this code was actually working. It was not.

This PR repairs both the ACL copying functionality and the testing for that functionality.

### DIFF
--- a/e2etest/declarativeResourceManagers.go
+++ b/e2etest/declarativeResourceManagers.go
@@ -50,7 +50,7 @@ type downloadContentOptions struct {
 type downloadBlobContentOptions struct {
 	containerClient *container.Client
 	cpkInfo         *blob.CPKInfo
-	cpkScopeInfo *blob.CPKScopeInfo
+	cpkScopeInfo    *blob.CPKScopeInfo
 }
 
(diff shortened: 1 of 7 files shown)


e2etest/declarativeResourceManagers.go
```diff
@@ -326,12 +326,18 @@ func (r *resourceBlobContainer) appendSourcePath(filePath string, useSas bool) {
 }
 
 func (r *resourceBlobContainer) getAllProperties(a asserter) map[string]*objectProperties {
-	objects := scenarioHelper{}.enumerateContainerBlobProperties(a, r.containerClient)
-
+	var fileSystem *azbfs.FileSystemURL
```
gapra-msft, line 329: how hard would it be to just migrate this to the new datalake sdk preemptively?

gapra-msft (approved):
if its possible to write the new code using the new sdk, that would be great. Otherwise looks good.

PR #2348, opened 2023-08-23
sync command scan log fix

souravgupta-msft:


nakulkar-msft (approved):

cmd/syncComparator.go
```diff
@@ -114,10 +114,13 @@ func (f *syncDestinationComparator) processIfNecessary(destinationObject StoredO
 			syncComparatorLog(sourceObjectInMap.relativePath, syncStatusSkipped, syncSkipReasonSameHash, false)
 			return nil
 		} else if sourceObjectInMap.isMoreRecentThan(destinationObject, f.preferSMBTime) {
+			// if destination is stale, schedule source
+			syncComparatorLog(sourceObjectInMap.relativePath, syncStatusOverwritten, syncOverwriteReasonNewerLMT, false)
```
gapra-msft, line 118: We should log this above as well? Line 93

gapra-msft: And below before returning nil, we should log Overwritten Line 130

souravgupta-msft: Added the log at line 93. Did not add it at line 130 because if the destination object is not present, then we should not log this.

gapra-msft (approved):

PR #2356, opened 2023-08-31
Document new behaviour of cache limiter

nakulkar-msft:


cmd/copy.go
```diff
@@ -1973,7 +1973,8 @@ func init() {
...
 		"or the account. Use of this flag is not applicable for copying data from non azure-service to service. More than one blob should be separated by ';'. ")
 	// options change how the transfers are performed
-	cpCmd.PersistentFlags().Float64Var(&raw.blockSizeMB, "block-size-mb", 0, "Use this block size (specified in MiB) when uploading to Azure Storage, and downloading from Azure Storage. The default value is automatically calculated based on file size. Decimal fractions are allowed (For example: 0.25).")
+	cpCmd.PersistentFlags().Float64Var(&raw.blockSizeMB, "block-size-mb", 0, "Use this block size (specified in MiB) when uploading to Azure Storage, and downloading from Azure Storage. The default value is automatically calculated based on file size. Decimal fractions are allowed (For example: 0.25)."+
+	" When uploading or downloading, maximum allowed block size is 0.75 * AZCOPY_BUFFER_GB. Please refer https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-optimize#optimize-memory-use.")
```
normesta, line 1977: Looks great. I'd add a "the" just before "maximum". Also, for the "refer" part, I'd say "To learn more, see the following article: https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-optimize#optimize-memory-use. On the docs side, I'll make sure to update the corresponding ref article so that inline and docs versions are in sync. Thank you for including me in this review!

normesta (changes requested):

gapra-msft (approved):
looks good apart from Norm's comment

PR #2366, opened 2023-09-07
Migrated to track 2 datalake SDK

gapra-msft:


gapra-msft:
These alerts are from files I added to keep supporting logging without pipeline.LogLevel. Should be safe to just dismiss the alerts. Thoughts?

tasherif-msft (approved):
ah thats nice to see :')

nakulkar-msft (approved):
should merge to dev instead

gapra-msft:
> should merge to dev instead

Good catch

siminsavani-msft (approved):

PR #2368, opened 2023-09-07
Added recursive query parameter and fixed some linting errors

gapra-msft:


tasherif-msft (approved):

cmd/syncProcessor.go
```diff
@@ -471,9 +475,11 @@ func (b *remoteResourceDeleter) delete(object StoredObject) error {
 					}
 				}
 			case common.ELocation.BlobFS():
-				directoryClient := common.CreateDatalakeDirectoryClient(objectURL.String(), b.credInfo, nil, b.clientOptions)
-				// TODO : Recursive delete
-				_, err = directoryClient.Delete(ctx, nil)
+				clientOptions := b.clientOptions
```
adreed-msft, line 478: This is a workaround for a lack of support for a continuation token, right? Is there any ongoing work to fix that up?

adreed-msft (commented):
lgtm

adreed-msft (approved):

cmd/syncProcessor.go
gapra-msft: this is a workaround for lack of recursive support. The continuation token I believe will be hidden behind the SDK

gapra-msft: The plan is not to support recursive=false for directories by design in the SDK, so we need to do this to work around that

PR #2373, opened 2023-09-12
Remove optimization for small files in sender-blockBlobFromURL

nakulkar-msft:
Zero byte blobs with and without metadata are already tested by existing tests.

### DIFF
--- a/ste/sender-blockBlobFromURL.go
+++ b/ste/sender-blockBlobFromURL.go
@@ -62,6 +62,15 @@ func newURLToBlockBlobCopier(jptm IJobPartTransferMgr, destination string, pacer
 
 // Returns a chunk-func for blob copies
 func (c *urlToBlockBlobCopier) GenerateCopyFunc(id common.ChunkID, blockIndex int32, adjustedChunkSize int64, chunkIsWholeFile bool) chunkFunc {
+	/*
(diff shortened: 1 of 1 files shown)


ste/sender-blockBlobFromURL.go
```diff
@@ -64,10 +62,6 @@ func newURLToBlockBlobCopier(jptm IJobPartTransferMgr, destination string, pacer
 
 // Returns a chunk-func for blob copies
 func (c *urlToBlockBlobCopier) GenerateCopyFunc(id common.ChunkID, blockIndex int32, adjustedChunkSize int64, chunkIsWholeFile bool) chunkFunc {
```
gapra-msft, line 64: Can you add a comment somewhere here about how this optimization previously existed and we removed it for XXXX reason? That way in the future we can remember and noone will try to add it back again

ste/sender-blockBlobFromURL.go
```diff
@@ -64,10 +62,6 @@ func newURLToBlockBlobCopier(jptm IJobPartTransferMgr, destination string, pacer
...
-	if blockIndex == 0 && adjustedChunkSize == 0 {
```
gapra-msft, line 67: We should also add a test case that matches up with what the customer was trying to do

nakulkar-msft: Customer corrupted the md5 sum. This is another content validation test and is taken care of.

PR #2376, opened 2023-09-14
Enable sync for dfs endpoints

nakulkar-msft:


gapra-msft:
Can we add tests

nakulkar-msft:
> Can we add tests

I've added BlobFS as additional endpoints to be tested in the e2e framework. That will run all the existing sync testcases to be run against DFS.

tasherif-msft (approved):

PR #2377, opened 2023-09-14
Re-architect resource managers & upgrade to track 2

adreed-msft:
The simple stuff: It's an API change to use the track 2 APIs.

The more complex stuff: Objects are now separated out from containers. No more type hinting and `any` containers.

What I really want comments on: How do you guys think we should handle the varied properties between endpoints? I've done my best to keep them separated, but obvious in the ObjectProperties struct. Unfortunately, we can't use generics here, because generics don't exist as "empty" generics, and `Foo[any]` would be just as bad


e2etest/newe2e_resource_managers_blobfs.go
```diff
@@ -0,0 +1,369 @@
...
+func (b *BlobFSServiceResourceManager) Level() cmd.LocationLevel {
+	return cmd.ELocationLevel.Service()
+}
+
+func (b *BlobFSServiceResourceManager) URI() string {
+	return dfsStripSAS(b.internalClient.DFSURL())
+}
+
+func (b *BlobFSServiceResourceManager) ValidAuthTypes() ExplicitCredentialTypes {
+	return EExplicitCredentialType.With(EExplicitCredentialType.OAuth(), EExplicitCredentialType.SASToken(), EExplicitCredentialType.AcctKey())
```
gapra-msft, line 60: why do we support account key in dfs? Is it something we should consider deprecating?

gapra-msft (approved):

e2etest/newe2e_resource_managers_blobfs.go
adreed-msft: We probably should deprecate it in the future yes.

It's supported because back in the day SAS tokens weren't supported

Issue #2380, opened 2023-09-19
File cannot be downloaded when throughput throttled 


mustafakemalsigirci:
### Which version of the AzCopy was used? 

10.16.0 and 10.20.1

### Which platform are you using? (ex: Windows, Mac, Linux)

Windows

### What command did you run?

azcopy copy "https://***.blob.core.windows.net/****.pdf****" "C:\test1.pdf" --cap-mbps 10 --log-level INFO --output-type text 

azcopy copy "https://***.blob.core.windows.net/****.pdf****" "C:\test1.pdf" --cap-mbps 10 --log-level INFO --output-type text --block-size-mb 3

### What problem was encountered?

File is 13MB. If I try to download, it stucks. There is no result like failed or completed. 

### How can we reproduce the problem in the simplest way?

try download files bigger than 10 MB and give --cap-mbps smaller than 20.

### Have you found a mitigation/solution?

If --cap-mbps parameter is 4 times bigger than --block-size-mb parameter, download will be completed.

gapra-msft:
@mustafakemalsigirci are you getting any error in the console when running either command? Could you set the log level to DEBUG and share the latest logs in the HOME/.azcopy directory?

mustafakemalsigirci:
Hi @gapra-msft , than you for your response. There is no ant error in console log. Log files are here.
![console log](https://github.com/Azure/azure-storage-azcopy/assets/108575232/1986ed1c-dde5-4e50-939d-f61d4890394e)
[238addd9-5d08-3346-78c5-4ee2044fb899.log](https://github.com/Azure/azure-storage-azcopy/files/12682215/238addd9-5d08-3346-78c5-4ee2044fb899.log)
[238addd9-5d08-3346-78c5-4ee2044fb899-chunks.log](https://github.com/Azure/azure-storage-azcopy/files/12682217/238addd9-5d08-3346-78c5-4ee2044fb899-chunks.log)
[238addd9-5d08-3346-78c5-4ee2044fb899-scanning.log](https://github.com/Azure/azure-storage-azcopy/files/12682218/238addd9-5d08-3346-78c5-4ee2044fb899-scanning.log)

gapra-msft:
It seems like you are running an old version of AzCopy, is this reproducible in the latest as well?

mustafakemalsigirci:
yes, you can repeat in 10.20.1 version as well

Android için Outlook<https://aka.ms/AAb9ysg> edinin
________________________________

[...]

PR #2385, opened 2023-09-25
Added code to adjust block size based on cap mbps if provided

gapra-msft:
TODO: Tests, I should probably wait for the new e2e test framework for this. 

Resolves https://github.com/Azure/azure-storage-azcopy/issues/2380


gapra-msft:
Note: This is more of a workaround than an actual fix. Ideally, we throttle reading of the actual stream rather than rate limiting by reducing chunk size

Issue #2337, opened 2023-08-14
AzCopy skips the --block-blob-tier arg when uploading to a Premium block blob storage account


JasonYeMSFT:
##### Note: The version is visible when running AzCopy without any argument
10.20.1

### Which platform are you using? (ex: Windows, Mac, Linux)
All

### What command did you run?
`
./azcopy copy "/Users/chunanye/Documents/bad_webpack.js" "https://jasonpremiumblock.blob.core.windows.net/first/bad_webpack.js?sv=2021-10-04&st=st_datetime&se=se_datetime&sr=c&sp=cw&sig=sig_redacted" --overwrite=true --from-to=LocalBlob --blob-type BlockBlob --follow-symlinks --cap-mbps 1000 --check-length=false --put-md5 --follow-symlinks --block-blob-tier=Hot --disable-auto-decoding=false --recursive --log-level=DEBUG;
`

### What problem was encountered?
The --block-blob-tier=Hot is dropped when performing the upload. As a result, the blob always ends up at the default tier. This is not the desirable behavior for a premium block blob storage account. The platform team wants to allow customers to directly upload blobs to a specific tier.

### How can we reproduce the problem in the simplest way?
Create a premium block blob storage account. Use such a command to upload a block blob to a Hot/Cool/Cold/Archive tier.

### Have you found a mitigation/solution?
Not yet.

Related issue:
https://github.com/microsoft/AzureStorageExplorer/issues/7123

seanmcc-msft:
It appears this limitation is by-design:

PUT [https://seanpremiumblockblob.blob.core.windows.net/test-container-c06bca3f-7056-4371-a4d9-09f6a38b5bd1/test-blob-59f67bb1-596b-482b-a6cc-a474f7f92d00 HTTP/1.1](https://seanpremiumblockblob.blob.core.windows.net/test-container-c06bca3f-7056-4371-a4d9-09f6a38b5bd1/test-blob-59f67bb1-596b-482b-a6cc-a474f7f92d00)
Host: seanpremiumblockblob.blob.core.windows.net
x-ms-blob-type: BlockBlob
**x-ms-access-tier: Hot**
x-ms-version: 2023-11-03
Accept: application/xml
x-ms-client-request-id: b1a0ca8c-a226-45c1-af50-4b67be270b97
x-ms-return-client-request-id: true
[...]

PR #2410, opened 2023-10-12
[Storage Explorer] Support hot tier in a Premium block blob storage account

siminsavani-msft:
In certain regions, setting tier on premium block blob is allowed (given the storage account has the feature enabled). Originally, AzCopy prevented setting tier in any premium storage account so in this PR, there is a small change to allow setting tier on premium blockblob accounts. Fixes https://github.com/Azure/azure-storage-azcopy/issues/2337

There are also two e2e tests from Standard to Premium Block Blob with Hot & Cool. These tests are commented out because this feature only works on whitelisted subscriptions and certain regions.


e2etest/zt_preserve_access_tier_test.go
```diff
@@ -99,3 +99,36 @@ func TestTier_V2ToClassicAccountNoPreserveCool(t *testing.T) {
 		},
 	}, EAccountType.Classic(), EAccountType.Standard(), "")
 }
+
+// Set Tier on Premium Block Blob account is a private preview feature that requires a whitelisted subscription
+// and is only available in certain regions. The service version must be 2021-04-10 or above. To test this feature,
+// uncomment TestTier_V2ToPremiumBBAccountCool and TestTier_V2ToPremiumBBAccountHot.
```
gapra-msft, line 105: Please create an item to remind us to reenable this test

gapra-msft (approved):

e2etest/zt_preserve_access_tier_test.go
siminsavani-msft: This test might not be reenabled since it requires a very specific subscription and region enabled.

PR #2426, opened 2023-10-24
Improve release metadata tracking

siminsavani-msft:
Customers see repeated calls to azcopyvnextrelease to determine if there is a new version of AzCopy which can sometimes cause throttling of this account or in some environments, this account cannot be reached. This PR improves the release metadata user experience by:
1. Adding a unit test validate that the release metadata exists, and contents can be downloaded. 
2. Adding a feature where release metadata is cached the metadata for a day to avoid repeated calls. Based on this protoype: https://github.com/Azure/azure-storage-azcopy/pull/1707

The way caching works is that a file (latest_version.txt) is created on the AzCopy log file path (~/.azcopy) and the remote version and expiry time is added to this text file. 

For the time until the expiry time, the remote version is cached (if it is equal or newer than the local version) so everytime AzCopy runs, it checks local version with the cached version in the txt file and notifies the user if their local version is older than the cached/remote version. See scenarios below:
1. Remote version is newer than local version (remote = 10.20.0, local = 10.19.0), remote version is cached for a day
2. Remote version is same as local version (remote = 10.20.0, local = 10.20.0), remote version is cached for a day 
3. Remote version is older than local version (remote = 10.19.0, local = 10.20.0), remote version is not cached, _this is should not happen and typically will not happen_ 
4. With a valid expiry time (1 day), AzCopy will check the cached version instead of calling the release metadata.
5. With an expired expiry time, AzCopy will refresh the cache by calling the release metadata and updating the file. 

Few questions/feedback:
1. Is it okay to create file in the azcopy log level path?
2. Will we need a flag to not cache the release metadata/have normal behavior?


cmd/root.go
```diff
@@ -244,54 +245,69 @@ func beginDetectNewVersion() chan struct{} {
...
+		localVersion, err := NewVersion(common.AzcopyVersion)
+		if err != nil {
+			return
+		}
+
+		// step 1: fetch & validate cached version and if it is updated, return without making API calls
+		filePath := filepath.Join(azcopyLogPathFolder, "latest_version.txt")
+		cachedVersion, err := ValidateCachedVersion(filePath) // same as the remote version
+		if err == nil {
+			if localVersion.OlderThan(*cachedVersion) {
```
gapra-msft, line 263: Is there a way we can restructure to avoid this duplicate code?

gapra-msft (commented):
left a comment

gapra-msft (approved):

PR #2430, opened 2023-10-26
Resolving 500 server timeouts on append blob appends

gapra-msft:
Tested GCP manually since pipeline GCP creds are not working


ste/sender-appendBlobFromLocal.go
```diff
@@ -68,14 +68,18 @@ func (u *appendBlobUploader) GenerateUploadFunc(id common.ChunkID, blockIndex in
 		u.jptm.LogChunkStatus(id, common.EWaitReason.Body())
 		body := newPacedRequestBody(u.jptm.Context(), reader, u.pacer)
 		offset := id.OffsetInFile()
-		_, err := u.destAppendBlobClient.AppendBlock(u.jptm.Context(), body,
+		var timeoutFromCtx bool
+		ctx := withTimeoutNotification(u.jptm.Context(), &timeoutFromCtx)
```
vibhansa-msft, line 72: why do we need context with timeout here, in older code its different right ?

nakulkar-msft (approved):

PR #2441, opened 2023-11-02
Consolidate  client creation

nakulkar-msft:
Because credentials are cached in policy, creating multiple clients will cause too many token refreshes. This PR 
1. Creates clients in frontend itself, these clients are self-contained, and files/blobs will be inherited from these clients
2. Clients in front end and STE are created explicitly, i.e. one client for front end, and one for STE.
3. When interacting with DFS, two clients are created - one for blob and one for dfs endpoint.


cmd/make.go
```diff
@@ -116,7 +138,12 @@ func (cookedArgs cookedMakeCmdArgs) process() (err error) {
 			return err
 		}
 	case common.ELocation.File():
-		shareClient := common.CreateShareClient(cookedArgs.resourceURL.String(), credentialInfo, nil, options, to.Ptr(common.ETrailingDotOption.Disable()), &cookedArgs.resourceLocation)
+		var shareClient *share.Client
+		//nakulkarmerge - create a trailingdot client
+		shareClient, err = share.NewClientWithNoCredential(resourceURL, &share.ClientOptions{ClientOptions: options})
```
gapra-msft, line 143: as long as this is for only share operations, we should be fine without trailing dot options being set

cmd/removeEnumerator.go
```diff
@@ -87,7 +87,27 @@ func newRemoveEnumerator(cca *CookedCopyCmdArgs) (enumerator *CopyEnumerator, er
...
 	}
 
-	transferScheduler := newRemoveTransferProcessor(cca, NumOfFilesPerDispatchJobPart, fpo)
+	targetURL, _ := cca.Source.String()
+	from := cca.FromTo.From()
+	if !from.SupportsTrailingDot() {
+		cca.trailingDot = common.ETrailingDotOption.Disable()
+	}
+	options := createClientOptions(common.AzcopyCurrentJobLogger)
+	var fileClientOptions any
```
gapra-msft, line 96: why is this type any?

nakulkar-msft: Leaves room if blob/dfs get their specific options. For now only AzFiles has, and the var name reflects this.

e2etest/zt_preserve_smb_properties_test.go
```diff
@@ -89,7 +89,7 @@ func TestProperties_SMBPermissionsSDDLPreserved(t *testing.T) {
 //
 //	See https://github.com/Azure/azure-storage-azcopy/issues/113 (which incidentally, I'm not observing in the tests above, for reasons unknown)
 func TestProperties_SMBDates(t *testing.T) {
-	RunScenarios(t, eOperation.CopyAndSync(), eTestFromTo.Other(common.EFromTo.LocalFile(), common.EFromTo.FileLocal()), eValidate.Auto(), anonymousAuthOnly, anonymousAuthOnly, params{
+	RunScenarios(t, eOperation.CopyAndSync(), eTestFromTo.Other(common.EFromTo.LocalFile()), eValidate.Auto(), anonymousAuthOnly, anonymousAuthOnly, params{
```
gapra-msft, line 92: Why did we delete a fromTo here?

cmd/copyEnumeratorInit.go
```diff
@@ -497,13 +500,7 @@ func (cca *CookedCopyCmdArgs) createDstContainer(containerName string, dstWithSA
...
 		return err
 	case common.ELocation.BlobFS():
-		// TODO: Implement blobfs container creation
-		accountRoot, err := GetAccountRoot(dstWithSAS, cca.FromTo.To())
-		if err != nil {
-			return err
-		}
-
-		dsc := common.CreateDatalakeServiceClient(accountRoot, dstCredInfo, nil, options)
+		dsc, _ := sc.DatalakeServiceClient()
```
tasherif-msft, line 503: error handling?

ste/mgr-JobPartMgr.go
```diff
@@ -460,44 +461,49 @@ func (jpm *jobPartMgr) clientInfo() {
 		Cancel:   jpm.jobMgr.Cancel,
 	}
 
-	retryOptions := policy.RetryOptions{
-		MaxRetries:    UploadMaxTries,
-		TryTimeout:    UploadTryTimeout,
-		RetryDelay:    UploadRetryDelay,
-		MaxRetryDelay: UploadMaxRetryDelay,
-	}
+	/*
```
gapra-msft, line 464: let's just go ahead and delete this code, it will be in the github history anyway if we need to go back and reference

gapra-msft (approved):

tasherif-msft (commented):
we need to validate that indeed token refreshes after client consolidation are refreshed correctly, have you validated this?

cmd/copyEnumeratorInit.go
nakulkar-msft: They are available just above. It is just an overhead here.

nakulkar-msft:
> we need to validate that indeed token refreshes after client consolidation are refreshed correctly, have you validated this?

Yes, this has been verified with new login methods, where this bug showed up.

ste/mgr-JobPartMgr.go
nakulkar-msft: There is lot more dead code. I'll clean up either for this release or next. Particularly this file

cmd/copy.go
```diff
@@ -1282,16 +1283,25 @@ func (cca *CookedCopyCmdArgs) processRedirectionDownload(blobResource common.Res
...
 	}
 
+	var blobClient *blockblob.Client
+	if credInfo.CredentialType.IsAzureOAuth() {
+		blobClient, err = blockblob.NewClient(u.String(), credInfo.OAuthTokenInfo.TokenCredential, options)
+	} else {
+		blobClient, err = blockblob.NewClientWithNoCredential(u.String(), options)
+	}
+	if err != nil {
+		return fmt.Errorf("fatal: Could not create client")
```
vibhansa-msft, line 1301: Shall we not log the exact error that we received in err here as well, so that in debugging we get a clear clue on why client creation failed.

vibhansa-msft (approved):

PR #2442, opened 2023-11-03
AzCopy NuGet package (intended for internal use only)

adreed-msft:
Adds a heavily scope-limited NuGet package for running AzCopy from .NET for the time being.

The real meat and potatoes exists in the `Microsoft.AzCopy` package, and the executables live in the subsequent `Microsoft.AzCopy.OS.Arch` packages


Microsoft.AzCopy/Microsoft.AzCopy.Linux.Arm64/Microsoft.AzCopy.Linux.Arm64.csproj
```diff
@@ -0,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <Title>Microsoft.AzCopy.Linux.Arm64</Title>
+    <Version>10.21.0</Version>
```
gapra-msft, line 8: is there any way to pull the version here programmatically? or do we need to change this every release?

gapra-msft (approved):

Microsoft.AzCopy/Microsoft.AzCopy.Linux.Arm64/Microsoft.AzCopy.Linux.Arm64.csproj
adreed-msft: I was looking into this while creating it. I believe you can pull environment variables here, but I didn't wind up experimenting with it enough. We can look into addressing this when we get to automating the creation of this package.

Issue #2213, opened 2023-05-12
Feature Request: Log file splitting\rotating. 


cdiskan:
Feature request: Log file maximum size switch, and\or automatic log splitting.

AZCopy currently creates a single log file for an entire job. When jobs are sufficiently large in file count the log becomes functionally unusable due to the extreme file size of the .log file. 

Example situation: 50TB Azcopy sync job. Current log file with default logging options is 4GB and growing. This is large enough that it cannot be opened with common application. This defeats the purpose of the log file, and prevents diagnosis of failed files. 

Reducing the log verbosity works in some situations, but for copies where a chain of custody is needed for data transfers the full verbosity log may be required. 

Please add automatic log splitting at a reasonable file size (100mb? 500mb?) to the logging function, or a switch that allows setting the maximum log size before log rotation.

adreed-msft:
Hi there!

I'll propose this to the team during our upcoming sync this week. Chances are it will not make it into 10.19.0, but perhaps a later release.

cdiskan:
Thanks. For anyone else who runs into this issue, the Linux App "Less" works well for opening these large logs. There's a version wrapped in GnuWin that works for folks limited to operating on Windows systems.

cdiskan:
@adreed-msft did this ever make it onto a roadmap? Running into this issue again with another large sync (75+TB, with about 25 million files). 

Thanks!

empz:
The logs directory gets out of control pretty quickly in our case. It doesn't have any limit. It will just take up space until the disk is full.

What are the options at the moment to stop this from happening...?

cdiskan:
@adreed-msft Following up again. Was this feature ever added to the roadmap for AZCopy?

gapra-msft:
Hi! This feature request has been resolved as of our latest AzCopy release 10.22.0. Please upgrade! Closing this issue as a result. Thanks again for the feature request!

cdiskan:
Thank you!!!!

PR #2446, opened 2023-11-09
Rotating logger

nakulkar-msft:
Resolves https://github.com/Azure/azure-storage-azcopy/issues/2213

### DIFF
--- a/common/logger.go
+++ b/common/logger.go
@@ -53,6 +53,8 @@ type ILoggerResetable interface {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+const maxLogSize = 500 * 1024 * 1024
+
 type jobLogger struct {
 	// maximum loglevel represents the maximum severity of log messages which can be logged to Job Log file.
 	// any message with severity higher than this will be ignored.
@@ -80,7 +82,7 @@ func (jl *jobLogger) OpenLog() {
 		return
 	}
 
-	file, err := NewRotatingWiter(path.Join(jl.logFileFolder, jl.jobID.String()+jl.logFileNameSuffix))
+	file, err := NewRotatingWiter(path.Join(jl.logFileFolder, jl.jobID.String()+jl.logFileNameSuffix), maxLogSize)
 	PanicIfErr(err)
 
 	jl.file = file
--- a/common/logger.go
+++ b/common/logger.go
@@ -82,7 +82,7 @@ func (jl *jobLogger) OpenLog() {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


common/logger.go
```diff
@@ -80,8 +82,7 @@ func (jl *jobLogger) OpenLog() {
 		return
 	}
 
-	file, err := os.OpenFile(path.Join(jl.logFileFolder, jl.jobID.String()+jl.logFileNameSuffix+".log"),
-		os.O_RDWR|os.O_CREATE|os.O_APPEND, DEFAULT_FILE_PERM)
+	file, err := NewRotatingWiter(path.Join(jl.logFileFolder, jl.jobID.String()+jl.logFileNameSuffix), maxLogSize)
```
gapra-msft, line 85:
```suggestion
	file, err := NewRotatingWriter(path.Join(jl.logFileFolder, jl.jobID.String()+jl.logFileNameSuffix), maxLogSize)
```

common/rotatingWriter.go
```diff
@@ -0,0 +1,94 @@
...
+func (w *rotatingWriter) rotate(suffix int32) error {
+	w.l.RUnlock()
+	defer w.l.RLock()
+
+	w.l.Lock()
+	defer w.l.Unlock()
+
+	if atomic.LoadInt32(&w.currentSuffix) > suffix {
+		// This log has already been rotated.
+		return nil
```
gapra-msft, line 45: is this a valid situation? trying to rotate an already rotated log?

common/rotatingWriter.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (w *rotatingWriter) Close() error {
+	return w.file.Close()
+}
+
+func (w *rotatingWriter) Write(p []byte) (n int, err error) {
+	w.l.RLock()
+	defer w.l.RUnlock()
+
+	currSuffix := atomic.LoadInt32(&w.currentSuffix)
```
gapra-msft, line 79: can we move this line lower to where we actually use currSuffix?

nakulkar-msft: That would leave a race condition. This has to be done here, so that at line 88 we dont use a new value if some other thread has updated log.

gapra-msft (approved):

common/rotatingWriter.go
```diff
@@ -0,0 +1,96 @@
+package common
```
siminsavani-msft, line 1: minor but don't forget to add the commented header here!

siminsavani-msft (approved):

PR #2456, opened 2023-11-15
Resolve 'InvalidBlockList'

siminsavani-msft:
- Maintains the existing behavior by default
- Adds hidden flag '--delete-destination-file'.  When customers specify this flag, the existing destination blob is checked for uncommitted blocks and then is deleted. Then, the upload proceeds as normal.
- This works for Blob to Blob and File to Blob

Manual testing results:
1. Blob to Blob with flag set to **false** = Failed (400 The specified blob or block content is invalid.), original behavior 
2. Blob to Blob with flag set to **true** = Completed
3. File to Blob with flag set to **false** = Failed (400 The specified blob or block content is invalid.), original behavior 
4. File to Blob with flag set to **true** = Completed
5. Container with multiple Blobs (2 with uncommitted blocks + 1 normal block) with flag set to **true** = Completed 
6. Container with multiple Blobs (2 with uncommitted blocks + 1 normal block) with flag set to **false** = Failed (400 The specified blob or block content is invalid.), original behavior 

Added pos & neg unit test that tests DeleteDstBlob to ensure that the function works as intended.


cmd/copy.go
```diff
@@ -180,6 +180,8 @@ type rawCopyCmdArgs struct {
 	rehydratePriority string
 	// The priority setting can be changed from Standard to High by calling Set Blob Tier with this header set to High and setting x-ms-access-tier to the same value as previously set. The priority setting cannot be lowered from High to Standard.
 	trailingDot string
+
+	deleteUncommittedBlocks bool
```
seanmcc-msft, line 184: Please add a comment to describe the purpose of the new flag.

seanmcc-msft (commented):
Please add a test for this case.

ste/xfer-anyToRemote-file.go
```diff
@@ -161,6 +163,22 @@ func ValidateTier(jptm IJobPartTransferMgr, blobTier *blob.AccessTier, client IB
 	}
 }
 
+func DeleteDstBlob(jptm IJobPartTransferMgr, client *blockblob.Client) {
```
gapra-msft, line 166: Is there a reason this code needs to live in xfer-anyToRemote-file even though its specific to block blob senders?

siminsavani-msft:
> Please add a test for this case.

Added results for manual testing in comment. I'm planning to add an item in the backlog and use the new testing framework to test this case as I'm not able to test this case using the current e2e testing.

cmd/copy.go
```diff
@@ -180,6 +180,9 @@ type rawCopyCmdArgs struct {
 	rehydratePriority string
 	// The priority setting can be changed from Standard to High by calling Set Blob Tier with this header set to High and setting x-ms-access-tier to the same value as previously set. The priority setting cannot be lowered from High to Standard.
 	trailingDot string
+
+	// when specified, the dst blob with uncommitted blocks is deleted
+	deleteUncommittedBlocks bool
```
nakulkar-msft, line 185: This is a misnomer - we'll have to make it clear that we're deleting the blob itself and not just un-committted blocks

ste/xfer-anyToRemote-file.go
```diff
@@ -48,6 +49,7 @@ type IBlobClient interface {
 // This sync.Once is present to ensure we output information about a S2S access tier preservation failure to stdout once
 var tierNotAllowedFailure sync.Once
 var checkLengthFailureOnReadOnlyDst sync.Once
+var deleteDstBlobWithUncommitedBlocks sync.Once
```
nakulkar-msft, line 52: Will this make it once per transfer? We're looking do delete every blob which has uncommitted blocks right?

ste/xfer-anyToRemote-file.go
nakulkar-msft: We can move this code to blob-sender Prologue(), so that it is done exactly once per blob.

nakulkar-msft (changes requested):

gapra-msft:
> > Please add a test for this case.
> 
> Added results for manual testing in comment. I'm planning to add an item in the backlog and use the new testing framework to test this case as I'm not able to test this case using the current e2e testing.

Take a look at my PR here. Even though I couldnt test it using the e2e test suite, I added unit tests to test the methods I added and mocked the client 

https://github.com/Azure/azure-storage-azcopy/pull/2430/files#diff-3435557beb691bbe572a99ced6d79bbaf691a0638c9d72f67a48bbeb10f688b0R183

ste/xfer-anyToRemote-file.go
gapra-msft: yes, each blob sender should have a sync.Once but I disagree that it should be in Prologue cause we only want to do it once we get a invalid blob or block

cmd/copy.go
siminsavani-msft: Updated!

gapra-msft (changes requested):
please ensure that at least unit tests are added to this PR

ste/xfer-anyToRemote-file.go
siminsavani-msft: So, when we stage the blocks, the 'chunks' are out of order so if we try to delete the destination blob after getting the error 'InvalidBlockOrBlob', some blocks successfully "stage" and it prematurely deletes the blob causing it to fail. 

Agree with Narasimha. I tested some more and moving the code to Prologue() ensures that if there are uncommitted blocks on a destination blob, they are deleted before we start the chunking process.

ste/xfer-anyToRemote-file.go
siminsavani-msft: Yes, I'll move this to the blockBlobSenderBase so it will delete for every blob.

siminsavani-msft: Done!

ste/xfer-anyToRemote-file.go
siminsavani-msft: Updated!

siminsavani-msft:
Added unit test to this PR.

ste/sender-blockBlob.go
```diff
@@ -192,6 +194,10 @@ func (s *blockBlobSenderBase) Prologue(ps common.PrologueState) (destinationModi
 	if s.jptm.ShouldInferContentType() {
 		s.headersToApply.BlobContentType = ps.GetInferredContentType(s.jptm)
 	}
+	if s.jptm.DeleteUncommittedBlocks() {
+		s.DeleteDstBlob()
```
gapra-msft, line 198: if this is called in the Prologue, is there any way we could call this more than once? If not, the sync.Once is not necessary

ste/sender_blockBlob_test.go
```diff
@@ -53,3 +57,43 @@ func TestGetVerifiedChunkParams(t *testing.T) {
 	_, _, err = getVerifiedChunkParams(transferInfo, memLimit, memLimit)
 	a.Equal(expectedErr, err.Error())
 }
+
+func TestDeleteDstBlob(t *testing.T) {
```
gapra-msft, line 61: can you add a test case testing a situation where we shouldnt delete the blob?

siminsavani-msft: I could, I would need to make a mock Job Part Transfer Manger as seen in your PR: https://github.com/Azure/azure-storage-azcopy/pull/2430/files#diff-3435557beb691bbe572a99ced6d79bbaf691a0638c9d72f67a48bbeb10f688b0 

Then I can use that to run Prologue instead since the actual method does not check for the flag.

siminsavani-msft: Done!

nakulkar-msft (approved):

helge12:
The flag "--delete-destination-file" is supposed to be in latest version 10.23.

But it isn't
C:\progi\azcopy_windows_amd64_10.23.0>azcopy copy "C:\Users\helge\Downloads\iview460_plugins_x64_setup.exe" "https://hexxx.blob.core.windows.net/testblob?si=helge&spr=https&sv=2022-11-02&sr=c&sig=LV%2F4XH%2F62YJnU12XkowB%2XXXXXXXXXXFxF7mB7BEJ6WBU%3D" --overwrite=true --delete-desination-file
**Error: unknown flag: --delete-desination-file**

(the blob link here is invalid)

Either the latest version is not for official download or there is another problem.

Issue #2455, opened 2023-11-15
Add different metadata for each file


stanislaw-czech:
### Which version of the AzCopy was used? 
10.22.0-Preview

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

### What command did you run?
azcopy copy

### What problem was encountered?
I need to add particular metadata for every file being transferred. Currently the same metadata is added for every file based on --metadata parameter. Would there be a way to apply some kind of convention to include information like creation date to every file being transferred? I've seen that some kind of setup to achieve this is already there, are there any reasons why it's not already done this way for local files?

### Have you found a mitigation/solution?
https://github.com/Azure/azure-storage-azcopy/pull/2460
Some convention to add data such as creation date should be established

vibhansa-msft:
Adding a different metadata to each file being uploaded means user needs to provide a file/input which clearly lists what all metadata shall be attached for each file being uploaded. If the data set is large this will cause good processing overhead and slows down entire transfer job. Also, we have not received any such feature request from other customers which means this is a very specific use case for your work flow. 
As this is not going to help majority of our customers and this feature will add processing overhead, I recommend you fork the code base and add the functionality you need for your specific workflow.

PR #2460, opened 2023-11-17
Set metadata on every local file

stanislaw-czech:
Connected with issue https://github.com/Azure/azure-storage-azcopy/issues/2455

Metadata is already present in CopyTransfer so I assume there were some plans to us it
If implementation were to be accepted I will provide proposition for parameter convention


stanislaw-czech:
@microsoft-github-policy-service agree company="Relativity"

PR #2467, opened 2023-11-22
Implement AzCopy runner

adreed-msft:
This is a _very_ basic implementation of things, and _very_ unfinished. It is more intended to ship a semi-functional prototype of the new E2E framework that real-world feedback can be generated against than to bring it to production readiness.

### DIFF
--- a/e2etest/newe2e_resource_manager_azstorage.go
+++ b/e2etest/newe2e_resource_manager_azstorage.go
@@ -2,11 +2,16 @@ package e2etest
 
 import (
 	"fmt"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
+	blobsas "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/sas"
 	blobservice "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
 	blobfscommon "github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake"
+	datalakeSAS "github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake/sas"
 	blobfsservice "github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake/service"
+	filesas "github.com/Azure/azure-sdk-for-go/sdk/storage/azfile/sas"
 	fileservice "github.com/Azure/azure-sdk-for-go/sdk/storage/azfile/service"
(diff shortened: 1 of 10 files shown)


e2etest/newe2e_resource_manager_azstorage.go
```diff
@@ -17,6 +22,112 @@ type AzureAccountResourceManager struct {
...
 }
 
+func (acct *AzureAccountResourceManager) ApplySAS(a Asserter, URI string, loc common.Location) string {
+	a.AssertNow("account must not be nil", Not{IsNil{}}, acct)
+	switch loc {
+	case common.ELocation.Blob():
+		skc, err := blob.NewSharedKeyCredential(acct.accountName, acct.accountKey)
+		a.NoError("Create shared key", err)
+
+		p, err := blobsas.AccountSignatureValues{ // todo: currently this generates a "god SAS", granting all permissions. This is bad! Let's not do this in prod.
```
gapra-msft, line 32: a thought for the future, maybe we make this configurable for the user to pass in a string that represents the sas permissions they want (ideally the user passes the minimal required) and we can parse it as a SignatureValues obj and pass it here

e2etest/newe2e_resource_managers_blob.go
```diff
@@ -83,8 +83,14 @@ func (b *BlobServiceResourceManager) ListContainers(a Asserter) []string {
 	return out
 }
 
-func (b *BlobServiceResourceManager) URI() string {
-	return blobStripSAS(b.internalClient.URL())
+func (b *BlobServiceResourceManager) URI(a Asserter, withSas bool) string {
+	base := blobStripSAS(b.internalClient.URL())
+
+	if withSas {
```
gapra-msft, line 89: why do we need to apply sas below? does the internalClient.URL not contain SAS?

gapra-msft (approved):

e2etest/newe2e_resource_managers_blob.go
adreed-msft: Nope. It's using account key based validation.

e2etest/newe2e_resource_manager_azstorage.go
adreed-msft: Great idea! I was planning on addressing this alongside some stronger auth method selection for RunAzCopy.

tasherif-msft (approved):

PR #2468, opened 2023-11-22
Generic resource management in the name of cleaner syntax

adreed-msft:
As opposed to `CreateResource(...).(ObjectResourceManager)` one now writes `CreateResource[ObjectResourceManager](...)`, and in future versions of Go (>1.21), it will be further reduced to `CreateResource(..., ResourceDefinitionObject{...})`


gapra-msft (approved):
love the generics!

siminsavani-msft (approved):

PR #2475, opened 2023-12-01
Fix tokenstore credential

nakulkar-msft:


common/oauthTokenManager.go
```diff
@@ -731,6 +731,9 @@ func jsonToTokenInfo(b []byte) (*OAuthTokenInfo, error) {
 	if err := json.Unmarshal(b, &OAuthTokenInfo); err != nil {
 		return nil, err
 	}
+	if OAuthTokenInfo.TokenRefreshSource == TokenRefreshSourceTokenStore {
+		_, _ = OAuthTokenInfo.GetTokenStoreCredential()
```
vibhansa-msft, line 735: So there are other values that user can have for refresh source here. I see only in case of token store we are going for refresh and not in other cases, so where do we refresh the token for other cases.

vibhansa-msft (approved):

common/oauthTokenManager.go
```diff
@@ -731,6 +731,9 @@ func jsonToTokenInfo(b []byte) (*OAuthTokenInfo, error) {
 	if err := json.Unmarshal(b, &OAuthTokenInfo); err != nil {
 		return nil, err
 	}
+	if OAuthTokenInfo.TokenRefreshSource == TokenRefreshSourceTokenStore {
```
gapra-msft, line 734: Is there a good way to UT this? We can try to add a UT or add this to the Jan backlog to test the env var

adreed-msft (approved):

siminsavani-msft (approved):

common/oauthTokenManager.go
siminsavani-msft: Would this be possible to test in the python tests we have?

nakulkar-msft: Does not really, we need a way to put oauth token in the very specific path like Stg exp, and better way may be to move to a more away from this to a more testable way than complicate the existing testcases.

PR #2477, opened 2023-12-02
Simplify GetCredType Method

nakulkar-msft:
With oAuth supported on files, we need not have three methods to determine auth type for different targets. This code replaces the old logic with this:
1. For S3/GCP choose type appropriately.
2. If we are targetting a blob as source, check if it is public.
3. If we are targetting a MD account, check if we need oAuth.
4. We are left with blob/file/blobFS. It is anonymous if we've sas, oAuth if not.

### DIFF
--- a/cmd/credentialUtil.go
+++ b/cmd/credentialUtil.go
@@ -35,9 +35,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
-	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/sas"
-	datalakesas "github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake/sas"
-	filesas "github.com/Azure/azure-sdk-for-go/sdk/storage/azfile/sas"
 	"github.com/Azure/azure-storage-azcopy/v10/jobsAdmin"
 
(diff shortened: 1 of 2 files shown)


cmd/credentialUtil.go
```diff
@@ -537,66 +348,168 @@ func logAuthType(ct common.CredentialType, location common.Location, isSource bo
 
 var authMessagesAlreadyLogged = &sync.Map{}
 
+// isPublic reports true if the Blob URL passed can be read without auth.
+func isPublic(ctx context.Context, blobResourceURL string, cpkOptions common.CpkOptions) (isPublicResource bool) {
```
gapra-msft, line 352: should we consider adding unit tests for these standalone methods? 

isPublic, mdAccountNeedsOAuth, doGetCredentialTypeForLocation

cmd/credentialUtil.go
```diff
@@ -537,66 +348,168 @@ func logAuthType(ct common.CredentialType, location common.Location, isSource bo
...
+			challenge := respErr.RawResponse.Header.Get("WWW-Authenticate")
+			if strings.Contains(challenge, common.MDResource) {
+				return true;
+			}
+		}
+	}
+	return false
+}
+
 func getCredentialTypeForLocation(ctx context.Context, location common.Location, resource, resourceSAS string, isSource bool, cpkOptions common.CpkOptions) (credType common.CredentialType, isPublic bool, err error) {
```
gapra-msft, line 423: Why do we need this as a wrapper for doGetCredentialTypeForLocation, can't we just make this one method?

nakulkar-msft: Drafting this. There is one big issue to be resolved.

PR #2480, opened 2023-12-04
[Storage Explorer] Add support for High Throughput Append Blob

siminsavani-msft:
The content length limit for `AppendBlob.AppendBlock()` and `AppendBlob.AppendBlockFromURL()` raised from 4 MB to 100 MB so this PR reflects the needed changes for this upgrade. 

- Hardcoded MaxAppendBlockSize value updated to  100 MB
- In copy.go, where we check the flag `block-size-mb` and blob  type has been updated to support this upgrade
- Included tests testing `block-size-mb` flag and large AppendBlobs (should these be skipped because of large size?)

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -360,9 +360,9 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 	}
 
 	// If the given blobType is AppendBlob, block-size-mb should not be greater than
-	// 4MB.
+	// 100MB.
 	if cookedSize, _ := blockSizeInBytes(raw.blockSizeMB); cooked.blobType == common.EBlobType.AppendBlob() && cookedSize > common.MaxAppendBlobBlockSize {
-		return cooked, fmt.Errorf("block size cannot be greater than 4MB for AppendBlob blob type")
+		return cooked, fmt.Errorf("block size cannot be greater than 100MB for AppendBlob blob type")
 	}
 
 	err = cooked.blockBlobTier.Parse(raw.blockBlobTier)
@@ -1540,7 +1540,7 @@ func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 	options := createClientOptions(common.AzcopyCurrentJobLogger)
 	var azureFileSpecificOptions any
 	if cca.FromTo.From() == common.ELocation.File() {
-		azureFileSpecificOptions = &common.FileClientOptions {
+		azureFileSpecificOptions = &common.FileClientOptions{
 			AllowTrailingDot: cca.trailingDot == common.ETrailingDotOption.Enable(),
 		}
 	}
@@ -1562,8 +1562,8 @@ func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 	}
 
 	if cca.FromTo.To() == common.ELocation.File() {
-		azureFileSpecificOptions = &common.FileClientOptions {
-			AllowTrailingDot: cca.trailingDot == common.ETrailingDotOption.Enable(),
+		azureFileSpecificOptions = &common.FileClientOptions{
+			AllowTrailingDot:       cca.trailingDot == common.ETrailingDotOption.Enable(),
 			AllowSourceTrailingDot: (cca.trailingDot == common.ETrailingDotOption.Enable() && cca.FromTo.From() == common.ELocation.File()),
 		}
 	}
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -362,7 +362,7 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 	// If the given blobType is AppendBlob, block-size-mb should not be greater than
 	// 100MB.
 	if cookedSize, _ := blockSizeInBytes(raw.blockSizeMB); cooked.blobType == common.EBlobType.AppendBlob() && cookedSize > common.MaxAppendBlobBlockSize {
-		return cooked, fmt.Errorf("block size cannot be greater than 100MB for AppendBlob blob type")
+		return cooked, fmt.Errorf("block size cannot be greater than %dMB for AppendBlob blob type", common.MaxAppendBlobBlockSize/1048576)
 	}
 
 	err = cooked.blockBlobTier.Parse(raw.blockBlobTier)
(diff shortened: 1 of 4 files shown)


cmd/copy.go
```diff
@@ -360,9 +360,9 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 	}
 
 	// If the given blobType is AppendBlob, block-size-mb should not be greater than
-	// 4MB.
+	// 100MB.
 	if cookedSize, _ := blockSizeInBytes(raw.blockSizeMB); cooked.blobType == common.EBlobType.AppendBlob() && cookedSize > common.MaxAppendBlobBlockSize {
-		return cooked, fmt.Errorf("block size cannot be greater than 4MB for AppendBlob blob type")
+		return cooked, fmt.Errorf("block size cannot be greater than 100MB for AppendBlob blob type")
```
gapra-msft, line 365: let's try and pull the constant here in the error message so we don't have to change this text every time

common/fe-ste-models.go
```diff
@@ -1073,7 +1073,7 @@ func (i *InvalidMetadataHandleOption) UnmarshalJSON(b []byte) error {
 const (
 	DefaultBlockBlobBlockSize      = 8 * 1024 * 1024
 	MaxBlockBlobBlockSize          = 4000 * 1024 * 1024
-	MaxAppendBlobBlockSize         = 4 * 1024 * 1024
+	MaxAppendBlobBlockSize         = 100 * 1024 * 1024
```
gapra-msft, line 1076: is there a decent way to pull this constant straight from the SDK?

e2etest/zt_basic_copy_sync_remove_test.go
```diff
@@ -77,6 +77,63 @@ func TestBasic_CopyUploadLargeBlob(t *testing.T) {
...
+		afterValidation: func(h hookHelper) {
+			props := h.GetDestination().getAllProperties(h.GetAsserter())
+			bprops, ok := props["test.txt"]
+			h.GetAsserter().Assert(ok, equals(), true)
+			if ok {
+				h.GetAsserter().Assert(bprops.blobType, equals(), dst)
+			}
+		},
+	}, testFiles{
+		defaultSize: "100M",
```
gapra-msft, line 129: should we make the file a little larger than a single block?

cmd/copy.go
vibhansa-msft: If SDK has any constants defined, better to use that so that with upgrade in SDK we always reflect back to the latest limit.

e2etest/zt_basic_copy_sync_remove_test.go
siminsavani-msft: Not sure if I understood your question correctly, but 100 MB is the max size so anything larger will fail and error out.

cmd/copy.go
siminsavani-msft: I completely agree here but the constant values to 104857600 which is unreadable imo. I can still make this change if the rest of the team is aligned with this.

siminsavani-msft: @vibhansa-msft , please see my comment here (https://github.com/Azure/azure-storage-azcopy/pull/2480#discussion_r1420758221). 

I was unable to find the azblob SDK constant that would contain MaxAppendBlobBlock. Here is the Golang SDK PR: https://github.com/Azure/azure-sdk-for-go/pull/21424 and the .NET PR: https://github.com/Azure/azure-sdk-for-net/pull/34548 in regards to high throughput AppendBlob. 

Otherwise I've updated this error message to pull the constant in AzCopy directly.

gapra-msft: if these constants don't already exist, we should add some to the GO SDK. @vibhansa-msft Could you add this item to the Go SDK backlog?

e2etest/zt_basic_copy_sync_remove_test.go
gapra-msft: that parameter is the size of the file, is 100MB the max file size for append blob?

siminsavani-msft: Yes, it is.

gapra-msft: It looks like the max append blob size was several 100GB and now it should be larger due to the increase in block size. https://learn.microsoft.com/en-us/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs#about-append-blobs

cmd/copy.go
siminsavani-msft: I started an email thread in regard to this ("RE: High Throughput Append Blob Constant"). I will resolve this conversation as this is not blocking this PR.

e2etest/zt_basic_copy_sync_remove_test.go
siminsavani-msft: Updated the file size to 101 MB! I didn't make it too big to avoid lagging in tests.

cmd/copy.go
```diff
@@ -360,9 +360,9 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 	}
 
 	// If the given blobType is AppendBlob, block-size-mb should not be greater than
-	// 4MB.
+	// 100MB.
 	if cookedSize, _ := blockSizeInBytes(raw.blockSizeMB); cooked.blobType == common.EBlobType.AppendBlob() && cookedSize > common.MaxAppendBlobBlockSize {
-		return cooked, fmt.Errorf("block size cannot be greater than 4MB for AppendBlob blob type")
+		return cooked, fmt.Errorf("block size cannot be greater than %dMB for AppendBlob blob type", common.MaxAppendBlobBlockSize/1048576)
```
gapra-msft, line 365: Can you use a constant for MB? it just reads cleaner. We might have one already in the codebase somewhere

ste/sender-appendBlob.go
```diff
@@ -63,8 +63,8 @@ func newAppendBlobSenderBase(jptm IJobPartTransferMgr, destination string, pacer
 
 	// compute chunk count
 	chunkSize := transferInfo.BlockSize
-	// If the given chunk Size for the Job is greater than maximum append blob block size i.e 4 MB,
-	// then set chunkSize as 4 MB.
+	// If the given chunk Size for the Job is greater than maximum append blob block size i.e 100 MB,
```
gapra-msft, line 66: here too can you just reference the AzCopy variable? seems unnecessary to call the number out here every time.

nakulkar-msft (approved):

vibhansa-msft (approved):

PR #2493, opened 2023-12-11
Simplify GetCredTypeMethod

nakulkar-msft:
With oAuth supported on files, we need not have three methods to determine auth type for different targets. This code replaces the old logic with this:

For S3/GCP choose type appropriately.
If we are targetting a blob as source, check if it is public.
If we are targetting a MD account, check if we need oAuth.
We are left with blob/file/blobFS. It is anonymous if we've sas, oAuth if not.

This is same as #2477. Accidentally deleted the branch


nakulkar-msft:
Gauri's comments on previous PR.

> [gapra-msft](https://github.com/gapra-msft) [last week](https://github.com/Azure/azure-storage-azcopy/pull/2477#discussion_r1414184274)
should we consider adding unit tests for these standalone methods?

isPublic, mdAccountNeedsOAuth, doGetCredentialTypeForLocation

isPublic now has one. mdAccountNeedsOAuth is a bit messy, I'll check with Adele. doGetCredentialType - I'll add.

nakulkar-msft:
[gapra-msft](https://github.com/gapra-msft) [last week](https://github.com/Azure/azure-storage-azcopy/pull/2477#discussion_r1414184775)
Why do we need this as a wrapper for doGetCredentialTypeForLocation, can't we just make this one method?

This will break the way we test checkAuthSafeForTarget. I'll take this as a separate review after this.

cmd/credentialUtil.go
```diff
@@ -537,66 +348,168 @@ func logAuthType(ct common.CredentialType, location common.Location, isSource bo
...
+
+	blobClient, _ := blob.NewClientWithNoCredential(bURLParts.String(), &blob.ClientOptions{ClientOptions: clientOptions})
+	bURLParts.BlobName = ""
+	bURLParts.Snapshot = ""
+	bURLParts.VersionID = ""
+
+	// Scenario 1: When resourceURL points to a container or a virtual directory
+	// Check if the virtual directory is accessible by doing GetProperties on container.
+	// Virtual directory can be public only when its parent container is public.
+	containerClient, _ := container.NewClientWithNoCredential(bURLParts.String(), &container.ClientOptions{ClientOptions: clientOptions})
```
vibhansa-msft, line 381: Shouldn't we create container client first and then using that create a blob client to share the pipelines?

nakulkar-msft: These are throw away clients, these do not contain any credential information or any special policies.

vibhansa-msft (dismissed):

cmd/credentialUtil.go
```diff
@@ -537,66 +348,168 @@ func logAuthType(ct common.CredentialType, location common.Location, isSource bo
...
+
+	// Scenario 2: When resourceURL points to a blob
+	if _, err := blobClient.GetProperties(ctx, &blob.GetPropertiesOptions{CPKInfo: cpkOptions.GetCPKInfo()}); err == nil {
+		return true
+	}
+
+	return false
+}
+
+func mdAccountNeedsOAuth(ctx context.Context, blobResourceURL string, cpkOptions common.CpkOptions) bool {
```
siminsavani-msft, line 394: min: add comment that describes the purpose of this function

gapra-msft (approved):

siminsavani-msft (dismissed):

vibhansa-msft (approved):

gapra-msft (approved):

Issue #2485, opened 2023-12-07
Invalid URL while running azcopy sync (HTTP 404) in v10.22.0


ljelonek:
### Which version of the AzCopy was used? 
10.22.0
### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

### What command did you run?
```powershell
$path = "http://127.0.0.1:10000/devstoreaccount1/emulated/folder/$sasToken"
Invoke-Expression "$InstallPath\azcopy sync .\localfolder '$path' --from-to LocalBlob --delete-destination=true"
```
### What problem was encountered?
```
Cannot perform sync due to error: cannot list files due to reason GET http://127.0.0.1:10000/emulated
--------------------------------------------------------------------------------
RESPONSE 404: 404 The specified resource does not exist.
ERROR CODE: ResourceNotFound
--------------------------------------------------------------------------------
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Error>
  <Code>ResourceNotFound</Code>
  <Message>The specified resource does not exist.
RequestId:c9b4d898-347e-42a2-9ec9-4aaf8025308c
Time:2023-12-07T11:04:21.095Z</Message>
</Error>
--------------------------------------------------------------------------------
```
### How can we reproduce the problem in the simplest way?

Download 10.22.0 to $InstallPath and run the command above.
### Have you found a mitigation/solution?

No, looks like internal problem with building URL.

vibhansa-msft:
Was this working in older versions? What kind of emulator you are running here?

ljelonek:
I can confirm that 10.17.0 version was working perfectly fine.
Emulator that I am using is `https://hub.docker.com/_/microsoft-azure-storage-azurite` version 3.28.0.

I highly doubt that emulator is the problem here, as changing the url to `http://127.0.0.1:10000/devstoreaccount1/devstoreaccount1/emulated/` returns error `Cannot perform sync due to error: cannot list files due to reason GET http://127.0.0.1:10000/devstoreaccount1`.

This is the second time I need to freeze azcopy version. Previously, I needed to do the same thing to mitigate #2140 . Is it possible to bring more attention to `sync` to avoid similar situations in the future?

gapra-msft:
This should be resolved by this PR https://github.com/Azure/azure-storage-azcopy/pull/2495

gapra-msft:
Hi! This has been fixed in our latest release. Please feel free to open a new issue in case you hit any other issues with AzCopy!

PR #2495, opened 2023-12-11
Fix regression in Azurite support and add Azurite tests

gapra-msft:
Resolves https://github.com/Azure/azure-storage-azcopy/issues/2485

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -117,6 +117,9 @@ func GetServiceClientForLocation(loc Location,
 	switch loc {
 	case ELocation.BlobFS():
 		datalakeURLParts, err := azdatalake.ParseURL(resourceURL)
+		if err != nil {
+			return nil, err
+		}
 		datalakeURLParts.FileSystemName = ""
 		datalakeURLParts.PathName = ""
 		resourceURL = datalakeURLParts.String()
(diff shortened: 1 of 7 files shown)


common/util.go
```diff
@@ -184,6 +190,14 @@ func GetServiceClientForLocation(loc Location,
 		return ret, nil
 
 	case ELocation.File():
+		fileURLParts, err := file.ParseURL(resourceURL)
+		if err != nil {
+			return nil, err
+		}
+		fileURLParts.ShareName = ""
```
vibhansa-msft, line 197: Why we want to explicitly set these variables, if parseURL is not populating them then it shall be empty string already.

e2etest/factory.go
```diff
@@ -48,7 +48,12 @@ type TestResourceFactory struct{}
 
 func (TestResourceFactory) GetBlobServiceURL(accountType AccountType) *blobservice.Client {
 	accountName, accountKey := GlobalInputManager{}.GetAccountAndKey(accountType)
-	resourceURL := fmt.Sprintf("https://%s.blob.core.windows.net/", accountName)
+	var resourceURL string
+	if accountName != "devstoreaccount1" {
+		resourceURL = fmt.Sprintf("https://%s.blob.core.windows.net/", accountName)
+	} else {
+		resourceURL = fmt.Sprintf("http://127.0.0.1:10000/%s/", accountName)
```
vibhansa-msft, line 55: Is this method very specific to Azurite test only? otherwise hardcoding the IP here will not work for other cases.

common/util.go
gapra-msft: We don't know if the URL passed is a service level URL, so we are forcing it here.

e2etest/factory.go
gapra-msft: only azurite has an account name of devstoreaccount1, and we don't have any tests accounts with the same name, so this is safe

siminsavani-msft (approved):

seanmcc-msft (approved):

PR #2502, opened 2023-12-14
Modify backup intent condition

nakulkar-msft:


ste/sender-azureFileFromURL.go
```diff
@@ -72,7 +72,7 @@ func (u *urlToAzureFileCopier) GenerateCopyFunc(id common.ChunkID, blockIndex in
 			return
 		}
 		ctx := u.ctx
-		if u.addFileRequestIntent || (token != nil && u.jptm.FromTo().From() == common.ELocation.File()) {
+		if u.addFileRequestIntent {
```
gapra-msft, line 75: why are we removing this check?

nakulkar-msft: The rest documentation does not require this header when the source is auth'd via a token.

gapra-msft: Can you double check we have a test where source is Authed with Oauth and destination uses SAS for file to file copy? I think that is the scenario I hit when I needed to add it

siminsavani-msft:
@nakulkar-msft Could you please add a description about the change?

PR #2504, opened 2023-12-15
[RCA Repair Item] Add exclude-container flag

siminsavani-msft:
Customers are unable to exclude containers when doing an Account to Account transfer. This PR introduces a new flag, `--exclude-path` that allows the user to add as many container names as they want to be excluded during Account to Account transfer. 

AzCopy sync does not support service level transfers and we spit out this error: `Failed with error service level URLs (https://xxx.blob.core.windows.net/) are not supported in sync:`, therefore, exclude-containers is not added to sync.

AzCopy remove does also does not support deleting at service level. No error message is spit out but containers and blobs are not deleted after a simple `remove https://xxx.blob.core.windows.net/$[SAS]` so exclude-container is not added to remove. Specific containers can be removed by using the container URL.


cmd/copyEnumeratorInit.go
```diff
@@ -65,14 +65,14 @@ func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
...
 	jobPartOrder.S2SSourceChangeValidation = cca.s2sSourceChangeValidation
 	jobPartOrder.DestLengthValidation = cca.CheckLength
 	jobPartOrder.S2SInvalidMetadataHandleOption = cca.s2sInvalidMetadataHandleOption
 	jobPartOrder.S2SPreserveBlobTags = cca.S2sPreserveBlobTags
 
 	dest := cca.FromTo.To()
-	traverser, err = InitResourceTraverser(cca.Source, cca.FromTo.From(), &ctx, &srcCredInfo, cca.SymlinkHandling, cca.ListOfFilesChannel, cca.Recursive, getRemoteProperties, cca.IncludeDirectoryStubs, cca.permanentDeleteOption, func(common.EntityType) {}, cca.ListOfVersionIDs, cca.S2sPreserveBlobTags, common.ESyncHashType.None(), cca.preservePermissions, azcopyLogVerbosity, cca.CpkOptions, nil, cca.StripTopDir, cca.trailingDot, &dest)
+	traverser, err = InitResourceTraverser(cca.Source, cca.FromTo.From(), &ctx, &srcCredInfo, cca.SymlinkHandling, cca.ListOfFilesChannel, cca.Recursive, getRemoteProperties, cca.IncludeDirectoryStubs, cca.permanentDeleteOption, func(common.EntityType) {}, cca.ListOfVersionIDs, cca.S2sPreserveBlobTags, common.ESyncHashType.None(), cca.preservePermissions, azcopyLogVerbosity, cca.CpkOptions, nil, cca.StripTopDir, cca.trailingDot, &dest, cca.excludeContainer)
```
vibhansa-msft, line 75: Not as part of this PR but I see lot of flags being passed down to the function and most of it comes from cca only then why not send cca itself instead of individual flags. Makes life easy to add new flags in future.

cmd/zc_filter.go
```diff
@@ -59,6 +59,39 @@ func (f *excludeBlobTypeFilter) DoesPass(object StoredObject) bool {
...
+	return false // excludeContainerFilter is related to container names, not related to files
+}
+
+func (s *excludeContainerFilter) DoesPass(storedObject StoredObject) bool {
+	if len(s.containerNamesList) == 0 {
+		return true
+	}
+	for _, cName := range s.containerNamesList {
+		// container name matches the excluded container name and return false to exclude
+		if strings.Compare(cName, storedObject.ContainerName) == 0 {
```
vibhansa-msft, line 81: If its an exaxt match that we are doing then it makes more sense to store the configured list in a map and then do a search. Iterating serially and comparing will work fine for smaller list but if someone gives a large number of containers to be excluded this will hit the performance and CPU usage.

cmd/zc_filter.go
```diff
@@ -59,6 +59,39 @@ func (f *excludeBlobTypeFilter) DoesPass(object StoredObject) bool {
...
+		}
+	}
+	return true
+}
+
+func buildExcludeContainerFilter(containerNames []string) []ObjectFilter {
+	excludeContainerSet := make([]string, 0)
+	excludeContainerSet = append(excludeContainerSet, containerNames...)
+
+	return append(make([]ObjectFilter, 0), &excludeContainerFilter{containerNamesList: excludeContainerSet})
```
vibhansa-msft, line 92: For the above comment, here instead of a slice we can just make a map of strings.

cmd/copyEnumeratorInit.go
siminsavani-msft: I was actually thinking about this as well. I feel like it would have to be a separate PR, but we should definitely talk about this in the next AzCopy meeting.

cmd/zc_filter.go
siminsavani-msft: Great idea and now updated!

cmd/zc_filter.go
siminsavani-msft: Updated :)

cmd/copyEnumeratorInit.go
seanmcc-msft: I agree that with both comments, let's hold off this refactor for a different PR.

cmd/zc_traverser_blob_account.go
```diff
@@ -61,27 +63,40 @@ func (t *blobAccountTraverser) listContainers() ([]string, error) {
...
 				}
 
-				cList = append(cList, *v.Name)
+				// get a list of containers that are not excluded
+				if len(t.excludeContainerName) > 0 {
+					so := StoredObject{ContainerName: *v.Name}
+					for _, f := range t.excludeContainerName {
+						if !f.DoesPass(so) {
+							// Ignore the container if the container name should be excluded
+							continue
```
vibhansa-msft, line 86: Should we log this as info that we are excluding this container. User has given the names in the list but the log will help us identify what all we excluded due to filter match.

vibhansa-msft (dismissed):
Minor comment to add logs whenever we skip a container.

cmd/zc_traverser_blob_account.go
siminsavani-msft: This is a great idea! I added it so the run looks like this:

```
INFO: Scanning...
INFO: Any empty folders will not be processed, because source and/or destination doesn't have full folder support
INFO: Skipped container(s): new, stuff, test, test-stuff, test2
```

siminsavani-msft: See added changes in line 112.

nakulkar-msft (approved):

PR #2508, opened 2023-12-19
Reuse clients for sync remove

nakulkar-msft:
This change makes remoteResourceDeleter to take a client instead of URL and client options so that we do not create multiple clients. This is required for #2493 

As a part of this, moved DoWithOverrideReadonly to common, so that it can be used both by STE and remoteDeleteProcessor


common/util.go
```diff
@@ -265,3 +269,73 @@ func (s *ServiceClient) DatalakeServiceClient() (*datalake.Client, error) {
...
+
+type FileClientStub interface {
+	URL() string
+}
+
+// DoWithOverrideReadOnly performs the given action, and forces it to happen even if the target is read only.
+// NOTE that all SMB attributes (and other headers?) on the target will be lost,
+// so only use this if you don't need them any more
+// (e.g. you are about to delete the resource, or you are going to reset the attributes/headers)
+func DoWithOverrideReadOnlyOnAzureFiles(ctx context.Context, action func() (interface{}, error), targetFileOrDir FileClientStub, enableForcing bool) error {
```
gapra-msft, line 292: Can we add some UT for this method?

ste/sender-azureFile.go
```diff
@@ -42,6 +43,8 @@ type FileClientStub interface {
 	URL() string
 }
 
+var doWithOverrideReadOnly = common.DoWithOverrideReadOnlyOnAzureFiles
```
gapra-msft, line 46: why make a variable for the function instead of just calling the function directly?

common/util.go
nakulkar-msft: The function has been unchanged. That should be taken as a diffrent review, does not fit to the purpose of this one

seanmcc-msft: We need to add a new test.  Testing is required to prevent regressions.

adreed-msft: Narasimha is right, it is old functionality, but it should still have a test.

nakulkar-msft: Posted a separate review here https://github.com/Azure/azure-storage-azcopy/pull/2540

gapra-msft (dismissed):

vibhansa-msft (dismissed):

gapra-msft (approved):

vibhansa-msft (approved):

PR #2512, opened 2023-12-28
[BugFix]AzCopy fails when destination directory has ReadOnly attribute set, despite using force-if-read-only flag

pranavmalik-msft:
**Description**
This pull request addresses the reported issue where Azcopy fails to function properly when the destination directory has the 'read-only' attribute set, even when utilizing the 'force-if-read-only' flag. 

**Changes:**
- Implemented a comprehensive fix to rectify the observed behavior, ensuring Azcopy functions correctly even when encountering directories with the 'read-only' attribute.
- Added validation checks.

### DIFF
--- a/ste/sender-azureFile.go
+++ b/ste/sender-azureFile.go
@@ -465,11 +465,7 @@ func (u *azureFileSenderBase) SetFolderProperties() error {
 		u.fileOrDirClient,
 		u.jptm.GetForceIfReadOnly())
 
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 func (u *azureFileSenderBase) DirUrlToString() string {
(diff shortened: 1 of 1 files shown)


ste/sender-azureFile.go
```diff
@@ -450,21 +451,25 @@ func (u *azureFileSenderBase) SetFolderProperties() error {
...
 		},
 		u.fileOrDirClient,
 		u.jptm.GetForceIfReadOnly())
-	return err
+
+	if err != nil {
+		return err
+	}
+
+	return nil
```
souravgupta-msft, line 472: You can directly return err here.
```suggestion
	return err
```

pranavmalik-msft: Done in the latest commit, thanks for the suggestion

nakulkar-msft (approved):

souravgupta-msft (approved):

PR #2513, opened 2023-12-29
Add source credential as a policy

nakulkar-msft:
This PR is to change the source auth token header in Put*FromURL and PutRangeFromURL calls. This is required as the token may expire during retry.

This PR, 
1. Creates an auth policy in STE, this policy is added only to destination clients, and replaces the token only when 'x-ms-copy-source-authorization' header is found.
2. Removes the old GetTokenCredential method to add source token


gapra-msft:
Could a test be added for this PR?

ste/sourceAuthPolicy.go
```diff
@@ -0,0 +1,77 @@
+// Copyright © Microsoft <wastore@microsoft.com>
```
tasherif-msft, line 1: These changes need thorough testing targeting them.

vibhansa-msft:
This appears to be a big change as we are moving source auth tagging to a policy. How are we validating that this is not having any collaterals.

nakulkar-msft:
> Could a test be added for this PR?

Existing tests should work fine. If you're sending in a token and service accepts it without problems, we're fine. I've replaced the token in the APIS to a invalid string for policy to replace this, so if the policy is incorrect, the request would fail.

nakulkar-msft:
> This appears to be a big change as we are moving source auth tagging to a policy. How are we validating that this is not having any collaterals.

The policy is invoked only in S2S transfers where source is Authenticated via oAuth. And adds a token, only when we find a the token header. We can add a test to verify this condition, I do not see this impacting other apis.

gapra-msft (dismissed):

azure-pipelines.yml
```diff
@@ -313,8 +313,11 @@ jobs:
           AWS_SECRET_ACCESS_KEY: $(AWS_SECRET_ACCESS_KEY)
           GOOGLE_CLOUD_PROJECT: $(GOOGLE_CLOUD_PROJECT)
           GOOGLE_APPLICATION_CREDENTIALS: $(ciGCSServiceAccountKey.secureFilePath)
+          AZCOPY_E2E_TENANT_ID: $(OAUTH_TENANT_ID)
```
gapra-msft, line 316: The reason the test sees the env variable as nil is because this set of env variables apply to the smoke tests, not the UT tests. To set for the UT tests, see run-ut.yml in the azurePipelineTemplates directory.

gapra-msft (approved):

siminsavani-msft (approved):

tasherif-msft (approved):

PR #2524, opened 2024-01-04
[StgExp] list-of-versions flag regression

siminsavani-msft:


ste/downloader-blob.go
```diff
@@ -66,9 +66,28 @@ func newBlobDownloader(jptm IJobPartTransferMgr) (downloader, error) {
 		return nil, err
 	}
 
+	blobClient := s.NewContainerClient(jptm.Info().SrcContainer).NewBlobClient(jptm.Info().SrcFilePath)
```
gapra-msft, line 69: it might be better to pass this as a variable in jptm.Info() (like SrcContainer and SrcFilePath) 

See mgr-JobPartTransferMgr line 266 for that method

gapra-msft: I think this is probably a bug in more places than just here too, can you look into where else we may have missed snapshot/version id setting?

ste/downloader-blob.go
```diff
@@ -66,9 +66,23 @@ func newBlobDownloader(jptm IJobPartTransferMgr) (downloader, error) {
...
+
+	if jptm.Info().VersionID != "" {
+		blobClient, err = s.NewContainerClient(jptm.Info().SrcContainer).NewBlobClient(jptm.Info().SrcFilePath).WithVersionID(jptm.Info().VersionID)
```
gapra-msft, line 72: there's no need to re-do the call to NewContainerClient and NewBlobClient here since we do that above. We can just take the blobClient above and call WithVersionId or WithSnapshot

siminsavani-msft:
New PR here: https://github.com/Azure/azure-storage-azcopy/pull/2527

ste/downloader-blob.go
siminsavani-msft: resolved in new pr https://github.com/Azure/azure-storage-azcopy/pull/2527

ste/downloader-blob.go
siminsavani-msft: resolved in new pr https://github.com/Azure/azure-storage-azcopy/pull/2527

PR #2525, opened 2024-01-04
Fix ACL copying with Blob

adreed-msft:
Somewhere along the way, when we moved to generating clients in `cmd`, we opted to generate a dfs and blob client when working with BlobFS, but we didn't do so for working with Blob.

This
1) Creates a DFS and Blob client as long as we're working with either resource, because they're the same resource
2) Enables some feature-parity with sync where specifying include-directory-stubs is used whenever persisting permissions on Blob, because we know that we're working with HNS and will encounter folders
3) Allows our testing suite to *actually* target BlobFS literally, rather than pretending.


gapra-msft (approved):

tasherif-msft (dismissed):

nakulkar-msft (approved):

gapra-msft (dismissed):

e2etest/zt_basic_cli_ps_auth_test.go
```diff
@@ -1,176 +1,176 @@
...
-		defaultSize: "1K",
-		shouldTransfer: []interface{}{
-			"wantedfile",
-			folder("sub/subsub"),
-			"sub/subsub/filea",
-			"sub/subsub/filec",
-		},
-	}, EAccountType.Standard(), EAccountType.Standard(), "")
-}
+//// Purpose: Tests AZCLI and powershell auth tests
```
siminsavani-msft, line 32: Why are these tests removed? /gen

e2etest/zt_basic_copy_sync_remove_test.go
```diff
@@ -41,7 +41,7 @@ import (
 
 // ================================  Copy And Sync: Upload, Download, and S2S  =========================================
 func TestBasic_CopyUploadSingleBlob(t *testing.T) {
-	RunScenarios(t, eOperation.CopyAndSync(), eTestFromTo.AllUploads(), eValidate.AutoPlusContent(), anonymousAuthOnly, allCredentialTypes, params{
+	RunScenarios(t, eOperation.Copy(), eTestFromTo.Other(common.EFromTo.LocalBlobFS()), eValidate.AutoPlusContent(), anonymousAuthOnly, anonymousAuthOnly, params{
```
siminsavani-msft, line 44: Would it be possible to add LocalBlobFS to AllUploads? What about testing LocalFile?

siminsavani-msft (approved):
Had some questions about the e2e tests modifications, otherwise, LGTM!

cmd/syncEnumerator.go
```diff
@@ -55,12 +55,14 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 		}
 	}
 
+	includeDirStubs := cca.fromTo.From().SupportsHnsACLs() && cca.fromTo.To().SupportsHnsACLs() && cca.preservePermissions.IsTruthy()
```
siminsavani-msft, line 58: Would it make sense to replace the isHNSToHNS option cookedSyncCmdArgs to have includeDirStubs?

e2etest/zt_basic_copy_sync_remove_test.go
gapra-msft, line 44: was this changed just for a temporary test?

e2etest/zt_basic_copy_sync_remove_test.go
```diff
@@ -419,8 +419,19 @@ func TestBasic_CopyRemoveFolderHNS(t *testing.T) {
 }
 
 func TestBasic_CopyRemoveContainer(t *testing.T) {
+	bfsRemove := TestFromTo{
```
gapra-msft, line 422:
```suggestion
	allButBfsRemove := TestFromTo{
```

e2etest/zt_basic_cli_ps_auth_test.go
adreed-msft: Accidental, realized later. I was running tests locally, these tests panic if you don't have the right tools.

e2etest/zt_basic_copy_sync_remove_test.go
adreed-msft: Coincidentally, I think this was another accident. I believe I'd just reduced the test conditions to test exactly what I was working on.

adreed-msft: LocalBlobFS was present

cmd/syncEnumerator.go
adreed-msft: isHNStoHNS hasn't existed for an extended period of time.

e2etest/zt_enumeration_filter_test.go
```diff
@@ -162,21 +162,6 @@ func TestFilter_RemoveFolder(t *testing.T) {
 	}, EAccountType.Standard(), EAccountType.Standard(), "")
 }
 
-func TestFilter_RemoveContainer(t *testing.T) {
```
gapra-msft, line 165: why is this test removed? unrelated but this seems like its not testing what the test title claims

e2etest/zt_preserve_properties_test.go
```diff
@@ -44,7 +44,7 @@ func TestProperties_NameValueMetadataIsPreservedS2S(t *testing.T) {
 
 func TestProperties_NameValueMetadataCanBeUploaded(t *testing.T) {
 	expectedMap := map[string]*string{"foo": to.Ptr("abc"), "bar": to.Ptr("def"), "baz": to.Ptr("state=a;b")}
-	RunScenarios(t, eOperation.Copy(), eTestFromTo.AllUploads(), eValidate.Auto(), anonymousAuthOnly, anonymousAuthOnly, params{
+	RunScenarios(t, eOperation.Copy(), eTestFromTo.Other(common.EFromTo.LocalBlobFS()), eValidate.Auto(), anonymousAuthOnly, anonymousAuthOnly, params{
```
gapra-msft, line 47: double check this

e2etest/zt_basic_copy_sync_remove_test.go
```diff
@@ -869,7 +879,7 @@ func TestBasic_OverwriteHNSDirWithChildren(t *testing.T) {
 	RunScenarios(
 		t,
 		eOperation.Copy(),
-		eTestFromTo.Other(common.EFromTo.LocalBlobFS()),
+		eTestFromTo.Other(common.EFromTo.BlobFSBlobFS()),
```
siminsavani-msft, line 882: Why is LocalBlobFS removed?

e2etest/zt_preserve_properties_test.go
```diff
@@ -57,7 +57,7 @@ func TestProperties_NameValueMetadataCanBeUploaded(t *testing.T) {
 }
 
 func TestProperties_HNSACLs(t *testing.T) {
-	RunScenarios(t, eOperation.CopyAndSync(), eTestFromTo.Other(common.EFromTo.BlobBlob()), eValidate.Auto(), anonymousAuthOnly, anonymousAuthOnly, params{
+	RunScenarios(t, eOperation.Copy(), eTestFromTo.Other(common.EFromTo.BlobBlob(), common.EFromTo.BlobFSBlobFS(), common.EFromTo.BlobBlobFS(), common.EFromTo.BlobFSBlob()), eValidate.Auto(), anonymousAuthOnly, anonymousAuthOnly, params{
```
siminsavani-msft, line 60: Should we add a separate test for BlobBlob for sync?

e2etest/zt_enumeration_filter_test.go
adreed-msft: This was a duplicate of another test with a different name that didn't actually test the relevant functionality that was attempted to be tested. I could rewrite this test and add it back in if wanted, but I believe it relies upon account-wide testing functionality, which, makes it probably not something to pursue in this framework.

adreed-msft: (Could probably test it in the new E2E framework though!)

adreed-msft: Actually. I'm not even sure what this is supposed to test. I guess using a wildcard in container name? idk. The name doesn't provide much information, and the test, clearly not testing what it's trying to test, does not make much sense.

gapra-msft (approved):

PR #2527, opened 2024-01-05
list-of-versions flag regression + version id/snapshot id regression

siminsavani-msft:
Adding Version ID and Snapshot ID to Client Creation

### DIFF
--- a/cmd/zt_copy_blob_download_test.go
+++ b/cmd/zt_copy_blob_download_test.go
@@ -962,8 +962,7 @@ func TestDryrunCopyGCPtoBlob(t *testing.T) {
 
 func TestListOfVersions(t *testing.T) {
 	a := assert.New(t)
-	t.Skip("Enable test after allowing versioning on storage account ")
-	bsc := getBlobServiceClient()
+	bsc := getSecondaryBlobServiceClient()
 	// set up the container with single blob with 2 versions
 	containerClient, containerName := createNewContainer(a, bsc)
 	defer deleteContainer(a, containerClient)
@@ -1022,7 +1021,7 @@ func TestListOfVersions(t *testing.T) {
 	mockedRPC.init()
 
 	// construct the raw input to simulate user input
-	rawBlobURLWithSAS := scenarioHelper{}.getRawBlobURLWithSAS(a, containerName, blobName)
+	rawBlobURLWithSAS := scenarioHelper{}.getSecondaryRawBlobURLWithSAS(a, containerName, blobName)
 	raw := getDefaultRemoveRawInput(rawBlobURLWithSAS.String())
 	raw.recursive = true
 	raw.listOfVersionIDs = file.Name()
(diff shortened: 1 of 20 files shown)


ste/xfer-setProperties.go
```diff
@@ -193,7 +193,18 @@ func setPropertiesFile(jptm IJobPartTransferMgr) {
 		transferDone(common.ETransferStatus.Failed(), err)
 		return
 	}
-	srcFileClient := s.NewShareClient(jptm.Info().SrcContainer).NewRootDirectoryClient().NewFileClient(jptm.Info().SrcFilePath)
+
+	sourceShare := s.NewShareClient(jptm.Info().SrcContainer)
```
gapra-msft, line 197: is it possible to set properties or delete on a share snapshot file?

siminsavani-msft: Share snapshot can be deleted but not modified (https://learn.microsoft.com/en-us/azure/storage/files/storage-snapshots-files#capabilities). I will update this accordingly.

gapra-msft: that's a share snapshot, but what about a file in the share snapshot?

cmd/zt_copy_blob_download_test.go
```diff
@@ -955,3 +959,77 @@ func TestDryrunCopyGCPtoBlob(t *testing.T) {
...
+	// construct the raw input to simulate user input
+	rawBlobURLWithSAS := scenarioHelper{}.getRawBlobURLWithSAS(a, containerName, blobName)
+	raw := getDefaultRemoveRawInput(rawBlobURLWithSAS.String())
+	raw.recursive = true
+	raw.listOfVersionIDs = file.Name()
+	runCopyAndVerify(a, raw, func(err error) {
+		a.Nil(err)
+
+		// validate that the right number of transfers were scheduled
+		a.Equal(2, len(mockedRPC.transfers))
```
adreed-msft, line 1033: Let's validate the data piping-- Could you check that the appropriate versions are a part of the registered transfers?

adreed-msft (commented):
looks good to me.

adreed-msft (approved):

ste/xfer-setProperties.go
siminsavani-msft: yeah, I guess that's not possible (https://learn.microsoft.com/en-us/rest/api/storageservices/delete-file2#remarks).

ste/mgr-JobPartTransferMgr.go
```diff
@@ -329,7 +332,6 @@ func (jptm *jobPartTransferMgr) Info() *TransferInfo {
 	}
 
 	if versionID != "" {
-		versionID = "versionId=" + versionID
```
gapra-msft, line 332: line 340 and 342 below, you need to make it 

+= "&versionId=" + versionID
and then the other line
= "versionId=" + versionID


similar change with snapshot

siminsavani-msft: added now!

PR #2531, opened 2024-01-09
Add a symlink to log file

nakulkar-msft:
This change creates a symlink to log file, the symlink is valid while the job is in progress, and will be removed when the job completes.

### DIFF
--- a/common/rotatingWriter.go
+++ b/common/rotatingWriter.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"sync"
 	"sync/atomic"
 )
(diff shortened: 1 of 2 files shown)


tasherif-msft (approved):

gapra-msft (approved):
is this trying to resolve logs sometimes not ending in .log?

nakulkar-msft:
> is this trying to resolve logs sometimes not ending in .log?

Yes, we'll always have a .log file pointing to current log file. So if program crashes (which we ideally dont want to), we still have a log file at the path we display. Kind of similar to syslog

nakulkar-msft:
Closing this, as Vikas suggested a better way.

Issue #2497, opened 2023-12-12
The `--preserve-posix-properties` only works copying to BLOB stroage


edwardsp:
##### Note: The version is visible when running AzCopy without any argument

AzCopy 10.22.0

### Which platform are you using? (ex: Windows, Mac, Linux)
Linux

##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.

```
# copy file to storage (this works - properties show in metadata)
azcopy cp --preserve-posix-properties testfile https://${account}.blob.core.windows.net/${container}

# copy back from storage (properties are not set on the local file)
azcopy cp --preserve-posix-properties https://${account}.blob.core.windows.net/${container}/testfile .
```

### What problem was encountered?

The  `--preserve-posix-properties` flag only works copying to BLOB stroage

### How can we reproduce the problem in the simplest way?

See above

### Have you found a mitigation/solution?

No

gapra-msft:
@edwardsp Could you please share debug level logs for the AzCopy command when you are trying to download?

edwardsp:
@gapra-msft Here is the log:

```
2023/12/13 14:40:33 AzcopyVersion  10.22.0
2023/12/13 14:40:33 OS-Environment  linux
2023/12/13 14:40:33 OS-Architecture  amd64
2023/12/13 14:40:33 Log times are in UTC. Local time is 13 Dec 2023 14:40:33
2023/12/13 14:40:33 ISO 8601 START TIME: to copy files that changed before or after this job started, use the parameter --include-before=2023-12-13T14:40:28Z or --include-after=2023-12-13T14:40:28Z
2023/12/13 14:40:33 Any empty folders will not be processed, because --recursive was not specified
2023/12/13 14:40:33 Job-Command --log-level DEBUG cp --preserve-posix-properties https://ACCOUNTNAME.blob.core.windows.net/test/testfile?SASKEY . 
2023/12/13 14:40:33 Number of CPUs: 8
2023/12/13 14:40:33 Max file buffer RAM 4.000 GB
2023/12/13 14:40:33 Max concurrent network operations: 128 (Based on number of CPUs. Set AZCOPY_CONCURRENCY_VALUE environment variable to override)
2023/12/13 14:40:33 Check CPU usage when dynamically tuning concurrency: true (Based on hard-coded default. Set AZCOPY_TUNE_TO_CPU environment variable to true or false override)
[...]

PR #2533, opened 2024-01-09
Standardize metadata handling to deal with capitalization

gapra-msft:
Resolves https://github.com/Azure/azure-storage-azcopy/issues/2497


tasherif-msft (approved):
great work and thorough tests :)

common/util.go
```diff
@@ -265,3 +265,36 @@ func (s *ServiceClient) DatalakeServiceClient() (*datalake.Client, error) {
...
+
+// Metadata utility functions to work around GoLang's metadata capitalization
+
+func TryAddMetadata(metadata Metadata, key, value string) {
+	if _, ok := metadata[key]; ok {
+		return // Don't overwrite the user's metadata
+	}
+
+	if key != "" {
+		capitalizedKey := strings.ToUpper(string(key[0])) + key[1:]
```
vibhansa-msft, line 277: Generally to make anything case insensitive we just convert everything to lower case and compare. Why we want to just capitalise the first character and keep the remaining as is. Anyway on blob you can not have two meta with same name with just some case change. Code will look more readable if we go that way,

vibhansa-msft (approved):

common/util.go
gapra-msft: Iterating through the metadata is O(n) whereas adjusting our lookups is O(1) so I figured this is slightly better to just search manually.

We don't necessarily care about the full case insensitivity, just what Golang forces the metadata to look like.

PR #2534, opened 2024-01-10
Fix logger name without suffix

nakulkar-msft:
Adds ".log" right when we create a log file, instead of adding it while rotation. KInd of like we used to do earlier.


pranavmalik-msft (approved):
LGTM

gapra-msft (approved):

vibhansa-msft (approved):

PR #2541, opened 2024-01-17
 [RCA Repair Item] Retried deletes should ignore 404 errors

pranavmalik-msft:
**Description:**
This pull request aims to resolve the reported issue related to Azcopy, where it erroneously reports failures in cases of a 404 error, despite a successful background delete operation. The proposed fix involves disregarding 404 errors and treating them as successful operations.

**Changes:**
- Implemented a comprehensive fix to address the observed behavior, ensuring Azcopy operates correctly even when a 404 error is reported.
- Added validation checks to enhance the overall reliability of Azcopy.

### DIFF
--- a/ste/xfer-deleteBlobFS.go
+++ b/ste/xfer-deleteBlobFS.go
@@ -1,8 +1,10 @@
 package ste
 
 import (
+	"errors"
 	"fmt"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake"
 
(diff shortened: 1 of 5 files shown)


vibhansa-msft (dismissed):

ste/xfer-deleteBlobFS.go
```diff
@@ -51,7 +54,16 @@ func doDeleteHNSResource(jptm IJobPartTransferMgr) {
 	transferDone := func(err error) {
 		status := common.ETransferStatus.Success()
 		if err != nil {
-			status = common.ETransferStatus.Failed()
+			var respErr *azcore.ResponseError
```
vibhansa-msft, line 57: We need to add test cases for this and validate that 404 are ignored while others are reported.

seanmcc-msft (changes requested):
Please add tests.

ste/xfer-deleteBlobFS.go
gapra-msft, line 57: Do we only want to apply the ignore 404 to dfs removes? what about file and blob removes?

seanmcc-msft: Agreed with @gapra-msft, we should be consistent.

gapra-msft: It looks like for blob and file, we also check 403, should we do a similar thing here?

pranavmalik-msft: Done in the latest commit

PR #2542, opened 2024-01-17
Preserve the snapshot ID when testing for required auth on MD

adreed-msft:
When supplying a managed disk SAS token, we would historically trim the snapshot (as expected with blob), but this is unnecessary and problematic with managed disk snapshots, where managed disks expect the snapshot as a part of the authorization. When we check for MD auth, we should preserve the snapshot ID, because it is a portion of auth.

This PR includes code changes and tests for both data access authentication mode ("OAuth + SAS") and non-authenticated data access (just a SAS).


e2etest/.azcopy/latest_version.txt
```diff
@@ -0,0 +1 @@
+10.22.2,2024-01-18T12:52:03Z
```
gapra-msft, line 1: what is this file for

gapra-msft (dismissed):
just had a comment about a particular file

e2etest/.azcopy/latest_version.txt
siminsavani-msft: This file was probably accidentally added and is due to the latest version caching

siminsavani-msft (dismissed):

tasherif-msft (approved):

e2etest/.azcopy/latest_version.txt
adreed-msft: Yeah, this is an accident.

gapra-msft (approved):

tasherif-msft (approved):

Issue #2007, opened 2023-01-03
azcopy list --properties "versionid" does not return versions and versionids


Peejay62:
##### Note: The version is visible when running AzCopy without any argument
V10.16.2
### Which platform are you using? (ex: Windows, Mac, Linux)
Windows server 2016
##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.
azcopy,exe list https://<stgaccount/container><sasconnect> --properties "LastModifiedTime;VersionID"

### What problem was encountered?
the above list does only list current versions and also doesn't return a versionid.  I would expect that when I had an object A and older versions of object A (A_v1, A_v2, A_v3 and so on)  azcopy list would return all versions in the listing. 
### How can we reproduce the problem in the simplest way?
Make an object with versioning on, write this object multiple times to create versions and the run azcopy list.
### Have you found a mitigation/solution?
no

grillba:
Why is this labeled a feature request when the functionality is documented as above at https://learn.microsoft.com/en-us/azure/storage/common/storage-ref-azcopy-list? 

it doesnt return any version IDs at all - And the  LastModifiedTime is set to "0001-01-01 00:00:00 +0000 UTC"

siminsavani-msft:
Hi @Peejay62 and @grillba ! Apologies for the confusion here. As an alternative, try using the --list-of-versions flag on azcopy copy command.

Peejay62:
I do not want to copy (versions). I merely want to report on the content of the container (say dir of container). Only the active occurrences of blobs are listed, no "older" versions.  copy --list-of-versions is not suitable in this case

siminsavani-msft:
Hi @Peejay62 and @grillba ! Apologies in taking so long to back to this issue! I am currently working on this issue as seen in the linked PR above and wanted to make sure that the expectations are met. 

With this PR, Blob VersionIds are printed out correctly:
`INFO: hello/stuff/foo.txt; VersionId: 2023-06-05T18:39:41.0295735Z;  Content Length: 3.00 B`

If the blob has more versions, then that will also be printed out:
`INFO: hello/stuff/foo.txt; VersionId: 2023-06-05T18:39:41.0295735Z;  Content Length: 3.00 B`
`INFO: hello/stuff/foo.txt; VersionId: 2023-06-06T10:50:20.0295735Z;  Content Length: 3.00 B`

Please let me know if there are any other questions or concerns!

[...]

PR #2550, opened 2024-01-22
[GitHub] azcopy list with versionids does not show version id

siminsavani-msft:
When doing this `list "https://[accountName].blob.core.windows.net/[containerName]" --properties="VersionID"`, AzCopy does not print out the version IDs of individual blobs (see example below). 

`INFO: hello/stuff/foo.txt; VersionId: ;  Content Length: 3.00 B`

The mentioned list command should print out something list this:
`INFO: hello/stuff/foo.txt; VersionId: 2023-06-05T18:39:41.0295735Z;  Content Length: 3.00 B`

If the blob has more versions, then that will also be printed out:
```
INFO: hello/stuff/foo.txt; VersionId: 2023-06-05T18:39:41.0295735Z;  Content Length: 3.00 B
INFO: hello/stuff/foo.txt; VersionId: 2023-06-06T10:50:20.0295735Z;  Content Length: 2.00 B
```

This PR also handles `--running-tally`:
```
INFO: hello/stuff/foo.txt; VersionId: 2023-06-05T18:39:41.0295735Z;  Content Length: 3.00 B
INFO: hello/stuff/foo.txt; VersionId: 2023-06-06T10:50:20.0295735Z;  Content Length: 2.00 B
INFO: foo.txt; VersionId: 2023-06-07T10:50:20.0295735Z;  Content Length: 10.00 B

INFO: File count: 2
INFO: Total file size 12.00 B
```

**Note that the total file size and count accounts for the latest versions only.**

This fixes the bug in and completes the feature request in https://github.com/Azure/azure-storage-azcopy/issues/2007 . 

I've also added three unit tests in cmd with running rally and version id in property flag.


cmd/zc_traverser_blob.go
```diff
@@ -478,6 +478,8 @@ func (t *blobTraverser) createStoredObjectForBlob(preprocessor objectMorpher, bl
 		object.blobSnapshotID = common.IffNotNil(blobInfo.Snapshot, "")
 	} else if t.includeDeleted && t.includeVersion && blobInfo.VersionID != nil {
 		object.blobVersionID = common.IffNotNil(blobInfo.VersionID, "")
+	} else if blobInfo.VersionID != nil {
+		object.blobVersionID = common.IffNotNil(blobInfo.VersionID, "")
```
gapra-msft, line 482: Does this add version id in cases we might not want it?

siminsavani-msft: That's a good point, I updated the code so that we only add VersionId for list commands if it is mentioned by the user. Let me know what you think!

cmd/zc_traverser_blob.go
```diff
@@ -541,7 +543,7 @@ func (t *blobTraverser) serialList(containerClient *container.Client, containerN
 	return nil
 }
 
-func newBlobTraverser(rawURL string, serviceClient *service.Client, ctx context.Context, recursive, includeDirectoryStubs bool, incrementEnumerationCounter enumerationCounterFunc, s2sPreserveSourceTags bool, cpkOptions common.CpkOptions, includeDeleted, includeSnapshot, includeVersion bool, preservePermissions common.PreservePermissionsOption, isDFS bool) (t *blobTraverser) {
+func newBlobTraverser(rawURL string, serviceClient *service.Client, ctx context.Context, recursive, includeDirectoryStubs bool, incrementEnumerationCounter enumerationCounterFunc, s2sPreserveSourceTags bool, cpkOptions common.CpkOptions, includeDeleted, includeSnapshot, includeVersion bool, preservePermissions common.PreservePermissionsOption, isDFS bool, includeVersionList bool) (t *blobTraverser) {
```
gapra-msft, line 546: Do we need to add a whole new option here to includeVersion? Why not just reuse the existing bool above?

siminsavani-msft: Updated this now!

seanmcc-msft:
We want to list all versions of the base blob.  Watch out for if the customer is using the --running-tally flag, as we probably don't want to sum up all the individual versions of the base blob, since they are incremental.

cmd/zt_list_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"github.com/stretchr/testify/assert"
+	"strings"
+	"testing"
+)
+
+func TestListVersions(t *testing.T) {
```
gapra-msft, line 11: we should also test running tally

cmd/list.go
```diff
@@ -237,23 +247,31 @@ func (cooked cookedListCmdArgs) HandleListContainerCommand() (err error) {
...
 	if err != nil {
 		return fmt.Errorf("failed to initialize traverser: %s", err.Error())
 	}
 
 	var fileCount int64 = 0
 	var sizeCount int64 = 0
+	var objectSummary string
 
 	processor := func(object StoredObject) error {
+		if cooked.RunningTally {
```
gapra-msft, line 261: I think the running tally should still include the size of the latest version of the blob. @adreed-msft thoughts?

cmd/zt_list_test.go
siminsavani-msft: Running tally is on in both tests at line 51 and 123. I can write additional tests covering running tally if needed!

gapra-msft: maybe we can add one with a mixed variety of blobs all together
some with versions, some without. some with 1 version, some with multiple versions, etc

siminsavani-msft: I can update TestListVersionsMultiVersions so that one blob has 2 versions. Right now, TestListVersionsMultiVersions tests: 2 blobs with 1 version and 2 without.

siminsavani-msft: Updated! MultiVersions now tests 1 blob with 2 versions, 1 blob with 1 version, 2 blobs without versions.

cmd/list.go
siminsavani-msft: This makes sense to me. Let me fix this!

siminsavani-msft: Updated, please take a look!

cmd/zc_enumerator.go
```diff
@@ -353,6 +353,10 @@ func InitResourceTraverser(resource common.ResourceString, location common.Locat
 		includeVersion = true
 	}
 
+	// print out version id when using azcopy list
+	if includeVersionsList {
```
vibhansa-msft, line 357: Is this feature behind a CLI flag or it will always print versions if available?

gapra-msft: this is technically a bug fix. it shouldnt go behind a flag

siminsavani-msft: It will only print out versions if specified by the properties flag by doing this: '--properties="VersionId"'

cmd/list.go
```diff
@@ -266,8 +285,33 @@ func (cooked cookedListCmdArgs) HandleListContainerCommand() (err error) {
 		}
 
 		if cooked.RunningTally {
-			fileCount++
-			sizeCount += object.size
+			if containsProperty(cooked.properties, versionId) {
```
gapra-msft, line 288: do we need to make a call and iterate over all the properties so many times? We do it 3 times in this one method

cmd/list.go
```diff
@@ -283,6 +327,15 @@ func (cooked cookedListCmdArgs) HandleListContainerCommand() (err error) {
 	}
 
 	if cooked.RunningTally {
+		if containsProperty(cooked.properties, versionId) {
+			// get file count by getting length of objectVer
+			fileCount = int64(len(objectVer))
+
+			// get size count by incrementing through objectVer and adding all values of versionIdObject.fileSize
+			for _, val := range objectVer {
```
gapra-msft, line 335: this could result in possibly looping over ever blob 2 times. Can we make the code above smarter and update fileCount + sizeCount dynamically instead of looping twice

gapra-msft (approved):

cmd/zt_scenario_helpers_for_test.go
```diff
@@ -345,6 +345,18 @@ func (scenarioHelper) generateBlobsFromList(a *assert.Assertions, containerClien
 	time.Sleep(time.Millisecond * 1050)
 }
 
+func (scenarioHelper) generateVersionsForBlobsFromList(a *assert.Assertions, containerClient *container.Client, blobList []string) {
```
adreed-msft, line 348: This doesn't operate too differently from generateBlobsFromList. Furthermore, we should generate some semblance of actually random data here instead, that way when generating multiple versions it can be proven that each version actually differs(?)

cmd/zt_list_test.go
```diff
@@ -0,0 +1,246 @@
...
+	raw.RunningTally = true
+
+	runListAndVerify(a, raw, func(err error) {
+		a.Nil(err)
+
+		// validate that the no transfers were scheduled
+		a.Nil(mockedRPC.transfers)
+
+		// check if info logs contain the correct version id for each blob
+		msg := mockedLcm.GatherAllLogs(mockedLcm.infoLog)
```
adreed-msft, line 61: nit: is listing in json not applicable here?

adreed-msft (approved):

tasherif-msft (approved):

cmd/zt_scenario_helpers_for_test.go
siminsavani-msft: Updated this method to allow additional of customized strings for data to allow variation of random data which can be set in the test itself.

cmd/zt_list_test.go
siminsavani-msft: It is! The actual contents are pretty much the same, nonetheless, I updated every test to check the json outputs as well!

cmd/zc_enumerator.go
```diff
@@ -334,7 +334,7 @@ type enumerationCounterFunc func(entityType common.EntityType)
...
+func InitResourceTraverser(resource common.ResourceString, location common.Location, ctx *context.Context, credential *common.CredentialInfo, symlinkHandling common.SymlinkHandlingType, listOfFilesChannel chan string, recursive, getProperties, includeDirectoryStubs bool, permanentDeleteOption common.PermanentDeleteOption, incrementEnumerationCounter enumerationCounterFunc, listOfVersionIds chan string, s2sPreserveBlobTags bool, syncHashType common.SyncHashType, preservePermissions common.PreservePermissionsOption, logLevel common.LogLevel, cpkOptions common.CpkOptions, errorChannel chan ErrorFileInfo, stripTopDir bool, trailingDot common.TrailingDotOption, destination *common.Location, excludeContainerNames []string, includeVersionsList bool) (ResourceTraverser, error) {
```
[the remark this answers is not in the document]
siminsavani-msft: I'm in agreement here! I believe this would have to be a separate PR in another release.

gapra-msft (approved):

PR #2557, opened 2024-01-31
Mildly refactor AzCopy runner

adreed-msft:
This mainly targets a few fields:

1) Flags. This is less a refactor, and more of a fleshing-out of the existing work.
2) Environment variables borrows some changes from the flags changes, namely auto-assigning work dirs.
3) Auth modes are a little cleaner, now. Using a RemoteResourceManager (todo: should we just merge RemoteResourceManager into the normal ResourceManager spec, and error out on the relevant fields when calling against local for ease-of-use?)

### DIFF
--- a/e2etest/newe2e_resource_managers_blob.go
+++ b/e2etest/newe2e_resource_managers_blob.go
@@ -26,21 +26,19 @@ TODOs:
 - CPK
 */
 
-// check that everything aligns with interfaces
+// enforce interface compliance at compile time
 func init() {
 	void := func(_ ...any) {} // prevent go from erroring from unused vars
 
-	var sm ServiceResourceManager = &BlobServiceResourceManager{}
-	var cm ContainerResourceManager = &BlobContainerResourceManager{}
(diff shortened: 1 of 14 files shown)


e2etest/newe2e_resource_managers_blob.go
```diff
@@ -86,8 +84,31 @@ func buildCanonForAzureResourceManager(manager ResourceManager) string {
...
-	internalAccount *AzureAccountResourceManager
-	internalClient  *service.Client
+	internalAccount  *AzureAccountResourceManager
+	specificAuthType *ExplicitCredentialTypes
+	internalClient   *service.Client
+}
+
+func (b *BlobServiceResourceManager) ValidAuthTypes() ExplicitCredentialTypes {
+	// Technically AcctKey is valid because of backwards compat integrated for dfs
+	// But we don't want to
```
gapra-msft, line 94: looks like this is incomplete

siminsavani-msft (approved):

PR #2558, opened 2024-01-31
Replace powershell with pwsh in Powershell Login

nakulkar-msft:
This PR replaces the usage of powershell in Powershell login with pwsh. pwsh is supported on linux and mac additionally.

The existing tests have been enabled to run on linux/mac as well. Also, the azure-pipelines is changed to use pwsh instead of powershell.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -154,10 +154,10 @@ jobs:
 
     steps:
       - task: PowerShell@2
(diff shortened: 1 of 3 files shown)


gapra-msft (approved):
Can we link all issues that will be resolved with this PR merging?

vibhansa-msft (approved):

PR #2561, opened 2024-01-31
Specify very large block size to force put blob instead of stage block/put block list

gapra-msft:


e2etest/zt_basic_copy_sync_remove_test.go
```diff
@@ -1149,3 +1149,74 @@ func TestCopySync_DeleteDestinationFileFlag(t *testing.T) {
...
+			folder(""),
+			f("test.txt"),
+		},
+	}, EAccountType.Standard(), EAccountType.Standard(), "")
+}
+
+func TestBasic_PutBlobSizeMultiPart(t *testing.T) {
+	RunScenarios(t, eOperation.CopyAndSync(), eTestFromTo.Other(common.EFromTo.LocalBlob(), common.EFromTo.BlobBlob()), eValidate.Auto(), anonymousAuthOnly, anonymousAuthOnly, params{
+		recursive:     true,
+		putBlobSizeMB: 50, // 256 MB
```
gapra-msft, line 1191:
```suggestion
		putBlobSizeMB: 50, // 50 MB
```

cmd/benchmark.go
```diff
@@ -46,12 +46,13 @@ type rawBenchmarkCmdArgs struct {
...
 
 	// options from flags
-	blockSizeMB float64
-	putMd5      bool
-	checkLength bool
-	blobType    string
-	output      string
-	mode        string
+	blockSizeMB   float64
+	putBlobSizeMB float64
```
vibhansa-msft, line 50: This variable has multiple versions in below code from float64, float32 to int64. Can we check if existing variable of blockSizeMB shall also be uniform across the code and covert this one as well.
Also, why do we take it as float64, Is user allowed to give something like 0.5 here?

ste/mgr-JobPartTransferMgr.go
```diff
@@ -382,6 +383,13 @@ func (jptm *jobPartTransferMgr) Info() *TransferInfo {
 	}
 	blockSize = common.Iff(blockSize > common.MaxBlockBlobBlockSize, common.MaxBlockBlobBlockSize, blockSize)
 
+	// If the putBlobSize is 0, then the user didn't provide any putBlobSize, default to DefaultPutBlobSize
+	putBlobSize := dstBlobData.PutBlobSize
+	if putBlobSize == 0 {
+		putBlobSize = blockSize
+	}
+	putBlobSize = common.Iff(putBlobSize > common.MaxPutBlobSize, common.MaxPutBlobSize, putBlobSize)
```
vibhansa-msft, line 391: Instead of doing this during transfer, we can do this in init phase itself. If user has provided a putBlobSize bigger than MaxPutBlobSize then reset it to MaxPutBlobSize. Here we can have just one comparsion that if its zero then set it to blockSize.

gapra-msft: Its easier to do in ste since this input is taken in in many different commands. This is similar to how we do block size validation as well

cmd/benchmark.go
gapra-msft: Yes, a user can give 0.5, etc to indicate less than 1 MB. Any user facing input is taking in float64, the float32 we see below is just an artifact of how we test, so I wouldn't worry about it too much since we are replacing that anyway

vibhansa-msft: there are other instances taking it in int format also, if user sent a float number do we convert it to number of bytes and then store in int or there will be potential loss of data while converting to int.

gapra-msft: So what we do is transform the param from float (parameter is in MB) to int (these parameters are in bytes since we multiply by MB)

cmd/copy.go
```diff
@@ -354,6 +355,11 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 		return cooked, err
 	}
 
+	cooked.putBlobSize, err = blockSizeInBytes(raw.putBlobSizeMB)
```
adreed-msft, line 358: todo: validate this size is within MaxPutBlobSize

siminsavani-msft (approved):

ste/mgr-JobPartTransferMgr.go
adreed-msft: Didn't see this prior to my last comment. Does it really make sense to override the user's input here instead of indicating it is nonsensical? If we do override the user's input, we should probably log that we did, and make it generally apparent that we are doing something intentionally different, otherwise it may be looked at as a bug.

adreed-msft: Secondarily, DefaultPutBlobSize is unused.

ste/sender-blockBlob.go
```diff
@@ -104,6 +109,25 @@ func getVerifiedChunkParams(transferInfo *TransferInfo, memLimit int64, strictMe
...
+		return
+	}
+
+	if putBlobSize >= strictMemLimit {
+		err = fmt.Errorf("Cannot use a put blob size of %.2fGiB. AzCopy is limited to use only %.2fGiB of memory, and only %.2fGiB of these are available for chunks.",
+			toGiB(putBlobSize), toGiB(memLimit), toGiB(strictMemLimit))
+		return
+	}
+
+	if putBlobSize > common.MaxPutBlobSize {
```
adreed-msft, line 124: We have no error check in the commands, but we have an override in the jptm, but we also have an error, that we probably will never hit, because of the override? If this is intended to be a sanity check, let's make it clear this is a sanity check.

adreed-msft (approved):
Work is mostly fine, I have some comments regarding defaults, overriding, etc.

adreed-msft (approved):

gapra-msft:
> Didn't see this prior to my last comment. Does it really make sense to override the user's input here instead of indicating it is nonsensical? If we do override the user's input, we should probably log that we did, and make it generally apparent that we are doing something intentionally different, otherwise it may be looked at as a bug.

I was mostly copying the pattern of how we deal with BlockBlobBLockSize. We should probably log that we updated it, you're right. I will fix that for both BlockBlobBlockSize and this new variable

gapra-msft:
> Secondarily, DefaultPutBlobSize is unused.

You're right. I think I was planning to use DefaultPutBlobSize then decided it's probably better to use the block size so that customers don't experience any "breaking changes" from this PR. I will get rid of the DefaultPutBLobSize var and update the comment.

cmd/copy.go
gapra-msft: Validation is performed in ste. I prefer it that way since that is the centralized location that all commands call into.

adreed-msft (approved):

Issue #2445, opened 2023-11-08
"500 Operation could not be completed within the specified time" when downloading huge file


schindlerd:
I'm trying to download a big exported disk snapshot (1.6 TB) but azcopy fails upon downloading with error 500. 
### Which version of the AzCopy was used? 
10.21.2
### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

### What command did you run?
azcopy.exe copy "https://md-gdlkcldgsmbx.z1.blob.storage.azure.net/dk4rgsx2pkrj/abcd?sv=2018-03-28&sr=b&si=xxx&sig=xxx" "D:\download\datadisk.vhd"
### What problem was encountered?
```
2023/11/08 13:50:12 PERF: primary performance constraint is Unknown. States: C:  0, R:  0, W:  0, F:  0, H:  0, B:  0, S:  0, P:  0, Q:  0, D:  0, E:  0, T:  0, GRs: 192
2023/11/08 13:50:12 0.0 %, 0 Done, 0 Failed, 1 Pending, 0 Skipped, 1 Total, 
2023/11/08 13:50:12 INFO: [P#0-T#0] Page blob throughput tuner: Target Mbps 4066
2023/11/08 13:50:12 ==> REQUEST/RESPONSE (Try=1/30.1644245s[SLOW >3s], OpTime=30.2698049s) -- RESPONSE STATUS CODE ERROR
   GET https://md-gdlkcldgsmbx.z1.blob.storage.azure.net/dk4rgsx2pkrj/abcd?comp=pagelist&si=xxx&sig=-REDACTED-&sr=b&sv=2018-03-28
   Accept: application/xml
   User-Agent: AzCopy/10.21.2 azsdk-go-azblob/v1.1.0 (go1.19.12; Windows_NT)
   X-Ms-Client-Request-Id: xxx
   x-ms-version: 
   --------------------------------------------------------------------------------
   RESPONSE Status: 500 Operation could not be completed within the specified time.
   Connection: Keep-Alive
   Content-Length: 246
   Content-Type: application/xml
   Date: Wed, 08 Nov 2023 13:50:12 GMT
   Server: Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
   X-Ms-Client-Request-Id: xxx
   X-Ms-Error-Code: OperationTimedOut
   X-Ms-Request-Id: xxx
   X-Ms-Version: 2020-10-02
[...]

PR #2568, opened 2024-02-07
Fixed a bug where page blob download optimizer would behave incorrectly on highly fragemented blobs if the service times out

gapra-msft:
Resolves https://github.com/Azure/azure-storage-azcopy/issues/2445


mock_server/closer.go
```diff
@@ -0,0 +1,34 @@
+//go:build go1.18
```
gapra-msft, line 1: This is all taken from azcore internal mock server implementation that they use for testing. I could not import it in azcopy since its internal. @vibhansa-msft can we ask them if they can make it public for us?

ste/pageRangeOptimizer.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func newPageRangeOptimizer(srcPageBlobClient *pageblob.Client, ctx context.Context) *pageRangeOptimizer {
+	return &pageRangeOptimizer{srcPageBlobClient: srcPageBlobClient, ctx: ctx}
+}
+
+// withNoRetryForBlob returns a context that contains a marker to say we don't want any retries to happen
+// Is only implemented for blob pipelines at present
+func withNoRetryForBlob(ctx context.Context) context.Context {
+	return policy.WithRetryOptions(ctx, policy.RetryOptions{MaxRetries: -1})
```
gapra-msft, line 48: this was one bug, setting MaxRetries to -1 actually forces it to not retry

ste/pageRangeOptimizer.go
```diff
@@ -0,0 +1,117 @@
...
+		pageList, err := pager.NextPage(limitedContext)
+		if err == nil {
+			if p.srcPageList == nil {
+				p.srcPageList = &pageList.PageList
+			} else {
+				p.srcPageList.PageRange = append(p.srcPageList.PageRange, pageList.PageRange...)
+				p.srcPageList.ClearRange = append(p.srcPageList.ClearRange, pageList.ClearRange...)
+				p.srcPageList.NextMarker = pageList.NextMarker
+			}
+		} else {
```
gapra-msft, line 78: adding this else statement is another bug where we need to give up. since we dont retry, our srcpagelist could be wrong if we dont nil it and quit

ste/pageRangeOptimizer.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// according to the REST API documentation:
+	// in a highly fragmented page blob with a large number of writes,
+	// a Get Page Ranges request can fail due to an internal server timeout.
+	// thus, if the page blob is not sparse, it's ok for it to fail
+	// TODO follow up with the service folks to confirm the scale at which the timeouts occur
+	// TODO perhaps we need to add more logic here to optimize for more cases
+	limitedContext := withNoRetryForBlob(p.ctx) // we don't want retries here. If it doesn't work the first time, we don't want to chew up (lots) time retrying
+	pager := p.srcPageBlobClient.NewGetPageRangesPager(&pageblob.GetPageRangesOptions{})
```
gapra-msft, line 66: Changed this from nil to empty options bag due to https://github.com/Azure/azure-sdk-for-go/issues/22356

vibhansa-msft (approved):

pranavmalik-msft (approved):

PR #2577, opened 2024-02-15
Fixed treatment of SAS in new e2e test framework to include container and path names

gapra-msft:


e2etest/newe2e_resource_manager_azstorage.go
```diff
@@ -30,47 +30,65 @@ func (acct *AzureAccountResourceManager) ApplySAS(URI string, loc common.Locatio
...
 	}
 
-	sasVals := opts.AzureOpts.SASValues
-	if sasVals == nil {
-
+	var sasVals GenericServiceSignatureValues
+	if opts.AzureOpts.SASValues == nil {
 		sasVals = GenericServiceSignatureValues{}
+	} else {
+		sasVals = opts.AzureOpts.SASValues.(GenericServiceSignatureValues)
```
adreed-msft, line 37: What if we get Account SAS values?

adreed-msft (changes requested):

e2etest/newe2e_resource_manager_azstorage.go
gapra-msft: Good point, we might need to add a GenericAccountSignatureValues type that keeps track of that

e2etest/newe2e_resource_manager_interface.go
```diff
@@ -24,6 +24,8 @@ type RemoteURIOpts struct {
 type AzureURIOpts struct {
 	// Must be manually specified
 	WithSAS bool
+	// As DirectoryPath in SASValues
+	AsDirectory bool
```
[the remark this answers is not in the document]
gapra-msft: oops, yes I will delete this.

siminsavani-msft (approved):

azure-pipelines.yml
```diff
@@ -127,7 +127,7 @@ jobs:
           display_name: "Linux"
           go_path: '$(go env GOPATH)/bin/'
           suffix: ''
-          run_e2e: 'go test -timeout=1h -v ./e2etest | tee test.txt'
+          run_e2e: 'go test -timeout=2h -v ./e2etest | tee test.txt'
```
adreed-msft, line 130: oh no /s

e2etest/newe2e_resource_manager_azstorage.go
```diff
@@ -30,47 +32,88 @@ func (acct *AzureAccountResourceManager) ApplySAS(URI string, loc common.Locatio
 		return URI
 	}
 
-	sasVals := opts.AzureOpts.SASValues
-	if sasVals == nil {
+	var sasVals GenericSignatureValues
+	if opts.AzureOpts.SASValues == nil {
+		// Default to account level SAS to cover all our bases
```
adreed-msft, line 37: (up to you) we could generate with reasonable defaults on a generic service signature values if we're >container. We can pull the container name and object name, and already do. This would be more "secure" esp. if we delete the container later, which we do.

But, all the resources for the framework disappear at the end anyway, so, it's equally as "secure"

e2etest/newe2e_resource_manager_azstorage.go
```diff
@@ -30,47 +32,88 @@ func (acct *AzureAccountResourceManager) ApplySAS(URI string, loc common.Locatio
...
+		parts, err := datalakesas.ParseURL(URI)
 		common.PanicIfErr(err)
 
-		p, err := sasVals.AsDatalake().SignWithSharedKey(skc)
+		skc, err := blobfscommon.NewSharedKeyCredential(acct.accountName, acct.accountKey)
 		common.PanicIfErr(err)
 
-		parts, err := datalakeSAS.ParseURL(URI)
+		var p datalakesas.QueryParameters
+		if _, ok := sasVals.(GenericServiceSignatureValues); ok {
```
adreed-msft, line 103: (nit) A thought for condensing this: Have an anonymous function above that appends this information? Something like `setupCanon(containerName, objectPath string, isDir bool)`

e2etest/newe2e_resource_manager_interface.go
```diff
@@ -24,6 +24,8 @@ type RemoteURIOpts struct {
 type AzureURIOpts struct {
 	// Must be manually specified
 	WithSAS bool
+	// Location level for SAS token, defaults to the resource's location level.
+	Level cmd.LocationLevel
```
adreed-msft, line 28: Unused?

e2etest/newe2e_resource_manager_azstorage.go
```diff
@@ -20,7 +22,7 @@ type AzureAccountResourceManager struct {
 	armClient *ARMStorageAccount
 }
 
-func (acct *AzureAccountResourceManager) ApplySAS(URI string, loc common.Location, optList ...GetURIOptions) string {
+func (acct *AzureAccountResourceManager) ApplySAS(URI string, loc common.Location, lev cmd.LocationLevel, et common.EntityType, optList ...GetURIOptions) string {
```
adreed-msft, line 25: entity type seems unused? Were these intended to be for a check of some sort? E.g. that using a service sas doesn't make sense on an account?

e2etest/newe2e_sas_generation.go
```diff
@@ -88,6 +91,13 @@ func (vals GenericServiceSignatureValues) withDefaults() GenericServiceSignature
 func (vals GenericServiceSignatureValues) AsBlob() BlobSignatureValues {
 	s := vals.withDefaults()
 
+	blobName := s.ObjectName
+	directoryName := s.DirectoryPath
+	if s.Level != nil && *s.Level != cmd.ELocationLevel.Object() {
```
adreed-msft, line 96: Tying the level to the signature values smells to me. What happens if another container comes along wanting to use the same values, after an object did? The test writer would have to recall this functionality and account for it.

adreed-msft (changes requested):

azure-pipelines.yml
gapra-msft: yeah our e2e tests are getting out of control

e2etest/newe2e_resource_manager_interface.go
gapra-msft: this is intended to be optionally set by devs using the test framework to adjust Level if they need to.

e2etest/newe2e_resource_manager_azstorage.go
gapra-msft: Entity is mostly intended for datalake + file share and only to specify file vs directory SAS. Since its not applicable for container/account levels, we set entity to the empty value.

e2etest/newe2e_resource_manager_azstorage.go
```diff
@@ -30,47 +32,88 @@ func (acct *AzureAccountResourceManager) ApplySAS(URI string, loc common.Locatio
...
+		skc, err := fileservice.NewSharedKeyCredential(acct.accountName, acct.accountKey)
 		common.PanicIfErr(err)
 
-		parts, err := filesas.ParseURL(URI)
+		var p filesas.QueryParameters
+		if _, ok := sasVals.(GenericServiceSignatureValues); ok {
+			v := sasVals.(GenericServiceSignatureValues)
+			v.Level = common.Iff(v.Level == nil, to.Ptr(lev), v.Level)
+			v.ContainerName = parts.ShareName
+			if et == common.EEntityType.Folder() {
```
gapra-msft, line 80: this is where entity type is used

e2etest/newe2e_resource_manager_azstorage.go
gapra-msft: will try this

e2etest/newe2e_resource_manager_azstorage.go
gapra-msft: leave default to account sas

e2etest/newe2e_resource_manager_azstorage.go
gapra-msft: irrelevant now

e2etest/newe2e_resource_manager_interface.go
gapra-msft: irrelevant now

e2etest/newe2e_sas_generation.go
gapra-msft: updated to the new model we agreed on

adreed-msft (approved):

PR #2583, opened 2024-02-21
Added json output to list command

gapra-msft:
This is necessary to facilitate testing of list in new e2e framework.


common/lifecyleMgr.go
```diff
@@ -344,6 +345,18 @@ func (lcm *lifecycleMgr) Dryrun(o OutputBuilder) {
...
 
+func (lcm *lifecycleMgr) Output(o OutputBuilder) {
+	om := ""
+	if o != nil {
+		om = o(lcm.outputFormat)
+	}
+
+	lcm.msgQueue <- outputMessage{
+		msgContent: om,
+		msgType:    eOutputMessageType.Info(),
```
gapra-msft, line 356: instead of the current json implementation, could I just set msgType to ListObject vs ListSummary? Or does this expect only Info vs Debug etc

cmd/list.go
```diff
@@ -333,17 +294,128 @@ func (cooked cookedListCmdArgs) HandleListContainerCommand() (err error) {
...
+	}
 
-		if cooked.MachineReadable {
-			glcm.Info("Total file size: " + strconv.Itoa(int(sizeCount)))
-		} else {
-			glcm.Info("Total file size: " + byteSizeToString(sizeCount))
+	return nil
+}
+
+type AzCopyResponse[T any] struct {
```
adreed-msft, line 313: This is *close* to the original image I had in my head when we chatted about it, but I *think* we may have mis-communicated on this.

I was more referring to the MessageType field of JsonOutputTemplate as a way to specify what type of output we were doing here.

To automate MessageType, we could probably use a custom json marshaller ([documentation](https://pkg.go.dev/encoding/json#Marshaler)) and use the reflect package to pull the type, and/or rely upon an override in an options bag.

common/lifecyleMgr.go
adreed-msft: I believe new enum entries could be used, yeah-- I think processTextOutput will probably have words to say about that unless added to the switch, FYI.

e2etest/newe2e_task_runazcopy_parameters.go
```diff
@@ -181,6 +181,7 @@ type GlobalFlags struct {
 	TrustedSuffixes  []string `flag:"trusted-microsoft-suffixes"`
 	SkipVersionCheck *bool    `flag:"skip-version-check"`
 
+	// TODO : Flags default seems to be broken; WI#26954065
```
adreed-msft, line 184: I think this is fixed in another one of my PRs(?) I think in the local RM? I'll double-check when I get back to E2E work.

adreed-msft (approved):

siminsavani-msft (approved):

adreed-msft (approved):

PR #2585, opened 2024-02-21
Local Resource Manager + Assorted bugfixes

adreed-msft:
A couple of things exist in this PR.

The primary one is that the New E2E framework now has a Local resource manager, a simple upload/download test, and sorts out some of the teething issues with grabbing a root resource in tests (since Local has no concept of an account)

In addition, there are a few bugfixes (for the e2e framework) around. Namely,
- CreateAzCopyTarget can now be called on any sort of ResourceManager, not just a RemoteResourceManager
- Flags structs that were zero-equivalents now have defaults filled in correctly
- Mock resource managers now correctly present their intended location/service type

### DIFF
--- a/e2etest/newe2e_resource_managers_local.go
+++ b/e2etest/newe2e_resource_managers_local.go
@@ -138,6 +138,22 @@ type LocalObjectResourceManager struct {
 	rawPath    string
 }
 
+func (l *LocalObjectResourceManager) ContainerName() string {
+	if l.container != nil {
+		return filepath.Base(l.container.RootPath)
+	}
+
+	return "containerless"
+}
+
+func (l *LocalObjectResourceManager) ObjectName() string {
+	if l.objectPath != "" {
+		return l.objectPath
+	} else {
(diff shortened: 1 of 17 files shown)


e2etest/newe2e_resource_managers_local.go
```diff
@@ -0,0 +1,328 @@
...
+func (l *LocalContainerResourceManager) Parent() ResourceManager {
+	return nil
+}
+
+func (l *LocalContainerResourceManager) Account() AccountResourceManager {
+	return nil
+}
+
+func (l *LocalContainerResourceManager) Canon() string {
+	return fmt.Sprintf("acountless/local/%s", l.ContainerName())
```
gapra-msft, line 70:
```suggestion
	return fmt.Sprintf("accountless/local/%s", l.ContainerName())
```

gapra-msft (approved):

gapra-msft (approved):

siminsavani-msft (approved):

Issue #2036, opened 2023-01-26
azcopy cp fails if the destination path is at the root of the file system (i.e. `/`)


matthiasplappert:
### Which version of the AzCopy was used? 
10.16.2 and 10.17.0

### Which platform are you using? (ex: Windows, Mac, Linux)
Linux

### What command did you run?

```
cd /
azcopy cp <remote-path>/archive.zip archive.zip
```

### What problem was encountered?

`azcopy` crashes with the following error:

```
INFO: Scanning...
INFO: Any empty folders will not be processed, because source and/or destination doesn't have full folder support

Job f16add00-4c91-6540-75ba-a088289c557f has started
Log file is located at: /tmp//f16add00-4c91-6540-75ba-a088289c557f.log

panic: runtime error: slice bounds out of range [:-1]

goroutine 109 [running]:
github.com/Azure/azure-storage-azcopy/v10/common.CreateParentDirectoryIfNotExist({0xc000590800, 0x0}, {0x403886fdf8, 0x17258d8})
	/home/vsts/work/1/s/common/writeThoughFile.go:46 +0x6e
github.com/Azure/azure-storage-azcopy/v10/common.CreateDirectoryIfNotExist({0xc000590800, 0x0}, {0x403886fdf8, 0x17258d8})
	/home/vsts/work/1/s/common/writeThoughFile.go:62 +0x16b
github.com/Azure/azure-storage-azcopy/v10/common.CreateParentDirectoryIfNotExist({0xc000590800, 0x3a}, {0x403886fdf8, 0x17258d8})
	/home/vsts/work/1/s/common/writeThoughFile.go:47 +0x5c
github.com/Azure/azure-storage-azcopy/v10/common.CreateFileOfSizeWithWriteThroughOption({0xc000590800, 0x3a}, 0x3767442, 0x0, {0x403886fdf8, 0x17258d8}, 0xe0)
	/home/vsts/work/1/s/common/writeThoughFile_linux.go:31 +0x45
github.com/Azure/azure-storage-azcopy/v10/ste.createDestinationFile({0x10379b8, 0xc000528070}, {0xc000590800, 0x3a}, 0x3767442, 0xcf)
	/home/vsts/work/1/s/ste/xfer-remoteToLocal-file.go:287 +0xf7
[...]

PR #2588, opened 2024-02-22
[UX][GitHub] azcopy cp fails if the destination path is at the root of the file system (i.e. /)

siminsavani-msft:
While AzCopy is acting as expected (does not transfer to root directory, i.e: "/"), AzCopy crashes with an out of bounds error at line 59 in writeThoughFile.go, especially in Linux. 

When AzCopy gets `lastIndex` by `strings.LastIndex(destinationPath, DeterminePathSeparator(destinationPath))`, and `LastIndex()` will either return a valid index or -1 if last index is not found in the string. In cases that `lastIndex` is not found, we should handle this gracefully by a clear error message instead of letting AzCopy crash as seen in this GitHub issue: https://github.com/Azure/azure-storage-azcopy/issues/2036.

FYI: the workaround is to download the file under a valid directory (i.e. "/tmp", "/home", etc.) and then users can move the downloaded file where they desire as we don't plan on supporting root directories.

### DIFF
--- a/common/writeThoughFile.go
+++ b/common/writeThoughFile.go
@@ -21,6 +21,7 @@
 package common
 
 import (
+	"fmt"
 	"os"
 	"regexp"
 	"strings"
@@ -44,11 +45,10 @@ var RootShareRegex = regexp.MustCompile(`(^\/\/[^\/]*\/?$)`)
 
 func isRootPath(s string) bool {
 	shortParentDir := strings.ReplaceAll(ToShortPath(s), OS_PATH_SEPARATOR, AZCOPY_PATH_SEPARATOR_STRING)
-	return  RootDriveRegex.MatchString(shortParentDir)  ||
-		RootShareRegex.MatchString(shortParentDir)  ||
+	return RootDriveRegex.MatchString(shortParentDir) ||
+		RootShareRegex.MatchString(shortParentDir) ||
 		strings.EqualFold(shortParentDir, "/")
 }
-
 
 func CreateParentDirectoryIfNotExist(destinationPath string, tracker FolderCreationTracker) error {
 	// If we're pointing at the root of a drive, don't try because it won't work.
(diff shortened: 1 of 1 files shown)


common/writeThoughFile.go
```diff
@@ -44,19 +45,26 @@ var RootShareRegex = regexp.MustCompile(`(^\/\/[^\/]*\/?$)`)
...
 	shortParentDir := strings.ReplaceAll(ToShortPath(s), OS_PATH_SEPARATOR, AZCOPY_PATH_SEPARATOR_STRING)
-	return  RootDriveRegex.MatchString(shortParentDir)  ||
-		RootShareRegex.MatchString(shortParentDir)  ||
+	return RootDriveRegex.MatchString(shortParentDir) ||
+		RootShareRegex.MatchString(shortParentDir) ||
 		strings.EqualFold(shortParentDir, "/")
 }
 
-
 func CreateParentDirectoryIfNotExist(destinationPath string, tracker FolderCreationTracker) error {
```
gapra-msft, line 53: any way to add a UT for this

siminsavani-msft: Added a test that mocks the problematic and unproblematic behaviors!

ste/writeThoughFile_test.go
```diff
@@ -0,0 +1,81 @@
...
+package ste
+
+import (
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
+)
+
+func TestCreateParentDirectoryIfNotExist(t *testing.T) {
```
gapra-msft, line 30: I think this test should exist in common since the code its testing lies there

ste/testJobPartTransferManager_test.go
```diff
@@ -370,7 +370,7 @@ func (t *testJobPartTransferManager) GetOverwritePrompter() *overwritePrompter {
 }
 
 func (t *testJobPartTransferManager) GetFolderCreationTracker() FolderCreationTracker {
-	panic("implement me")
+	return t.jobPartMgr.getFolderCreationTracker()
```
gapra-msft, line 373: this change can be undone

common/writeThoughFile.go
```diff
@@ -44,19 +45,26 @@ var RootShareRegex = regexp.MustCompile(`(^\/\/[^\/]*\/?$)`)
...
 	}
 
-	lastIndex := strings.LastIndex(destinationPath, DeterminePathSeparator(destinationPath))
+	pathSeparator := DeterminePathSeparator(destinationPath)
+	lastIndex := strings.LastIndex(destinationPath, pathSeparator)
+
+	// LastIndex() will return -1 if path separator was not found, we should handle this gracefully
+	// instead of allowing AzCopy to crash with an out-of-bounds error.
+	if lastIndex == -1 {
+		return fmt.Errorf("error: Path separator (%s) not found in destination path. On Linux, this may occur if the destination is the root file, such as '/'. If this is the case, please consider changing your destination path.", pathSeparator)
```
vibhansa-msft, line 65: If someone wants to copy at "/" shouldn't we allow it. Anyway if there are access issues, creation of file will fail with EPERM.

gapra-msft: This is a good point, but should we consider this a feature request? We can add it to our backlog for future consideration.

Currently we panic on copying to / so this improves the experience in the meantime.

pranavmalik-msft (approved):

vibhansa-msft (approved):

pranavmalik-msft:
Please sync with main before merging, else looks good to me

PR #2591, opened 2024-02-23
[BugFix] Minimise mumber of calls to folder create in case of Blob to File Copy

pranavmalik-msft:
**Description:**
This pull request addresses a regression stemming from the existing logic for Blob to File Copy. Previously, the default case returned a nullFolderTracker, lacking the mechanism to track already created folders. Considering this a corner case, the proposed solution introduces the AllFolderCopy approach. With this enhancement, the foldertracker logic now effectively monitors previously created folders, thereby reducing redundant folder creation during Blob to File operations.

**Changes:**
- Introduced the AllFolderCopy approach to address the regression in Blob to File Copy, effectively resolving the identified issue.


siminsavani-msft (commented):
Is there any way to test this?

cmd/copyEnumeratorInit.go
```diff
@@ -674,6 +674,12 @@ func NewFolderPropertyOption(fromTo common.FromTo, recursive, stripTopDir bool,
 		}
 	}
 
+	// In case of a Blob to File Copy, we will go for all folder copy
+	if fromTo == common.EFromTo.BlobFile() {
```
gapra-msft, line 678: I thought we decided to use the mitigation (customer sets include directory flag) and get to this issue in a later release after we have a solid plan?

@seanmcc-msft  @vibhansa-msft thoughts?

adreed-msft (changes requested):
This is neither a proper fix to the issue (If the source has no directory stubs, the issue resurfaces), nor is it what was discussed (temp mitigation by manual flag setting, spike for next release)

cmd/copyEnumeratorInit.go
vibhansa-msft: We can consider this fix for the next release (not for 10.23). Once we have more clarity on the flow, may be we can improvise the fix.

vibhansa-msft: We also need to consider the part where putting such check may take away the capabilities to skip empty folders while copying.

PR #2592, opened 2024-02-26
[Test] Resume_Job Fix E2E Tests

pranavmalik-msft:
**Description**
- This pull request addresses the reported issue where job resumptions in the Azure Storage AzCopy tool resulted in the creation of multiple log files.
  
**Changes**
- Introduced end-to-end tests for scenarios where resuming a copy job generates consistent log files.

### DIFF
--- a/e2etest/zt_resume_test.go
+++ b/e2etest/zt_resume_test.go
@@ -1,9 +1,12 @@
 package e2etest
 
 import (
+	"os"
+	"strings"
+	"testing"
+
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
-	"testing"
 
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


e2etest/zt_resume_test.go
```diff
@@ -125,3 +128,52 @@ func TestResume_PublicSource_ContainerTarget(t *testing.T) {
...
+		},
+		&hooks{
+			afterValidation: func(h hookHelper) {
+				a := h.GetAsserter()
+				logDir := os.Getenv("AZCOPY_E2E_LOG_OUTPUT")
+				dir, err := os.Open(logDir)
+				a.Assert(err, equals(), nil)
+				defer dir.Close()
+
+				entries, err := dir.Readdir(-1)
```
vibhansa-msft, line 156: @adreed-msft : is there a way to validate the logs and check if the copy failed in first attempt and was success due to resume.

e2etest/zt_resume_test.go
```diff
@@ -125,3 +128,52 @@ func TestResume_PublicSource_ContainerTarget(t *testing.T) {
 		EAccountType.Standard(), EAccountType.Standard(), "",
 	)
 }
+
+// TestResume_ConsistentLogs_PublicSource_BlobTarget validates log file consistency for a job initially initiated as a copy operation and later resumed.
+// During the copy operation, a jobId is created and associated with its log file. If a resume operation follows, the same jobId from the copy is used for the log file, and the resumed logs are appended to it.
+// The test ensures that at most two log files are generated: one for scanning logs and one for normal logs.
+
+func TestResume_ConsistentLogs_PublicSource_BlobTarget(t *testing.T) {
```
siminsavani-msft, line 136: It would be good to validate the contents of the log. Is it possible for you to programmatically open the log file and do some validation?

e2etest/zt_resume_test.go
pranavmalik-msft: Parsing the logs may seem like an unnecessary complication. The flag debugSkipFiles already ensures a scenario where the copy operation fails, allowing the resume operation to proceed smoothly. Introducing log parsing would only add unnecessary complexity to the process.

siminsavani-msft (approved):

e2etest/zt_resume_test.go
```diff
@@ -125,3 +128,52 @@ func TestResume_PublicSource_ContainerTarget(t *testing.T) {
...
+
+				entries, err := dir.Readdir(-1)
+				a.Assert(err, equals(), nil)
+				var logFiles int
+				for _, entry := range entries {
+					if !entry.IsDir() && strings.HasSuffix(entry.Name(), ".log") {
+						logFiles++
+					}
+				}
+				a.Assert(logFiles <= 2, equals(), true)
```
vibhansa-msft, line 164: If its resume job, shouldn't there be only one log file. with <= 2 we are accepting 2 log files as well?

pranavmalik-msft: In totality - there would be a scanning log-file and a normal log file pre-pended by the JobID hence we are checking for two log files

gapra-msft (approved):

PR #2598, opened 2024-02-29
Fix AzCLI in tests/Make Autologin mode ephemeral in tests

adreed-msft:
This is a cherrypicked change from #2542 to get folks unblocked.

It serves two purposes
1) Specifying tenant ID to AzCopy (why did this suddenly become required? todo)
2) Make it so that AzCLI/PSCred isn't permanent across the test suite, because prior the env var never got cleared.


gapra-msft (approved):
lgtm, thanks for also adding some of those code cleanup changes

pranavmalik-msft (approved):

siminsavani-msft (approved):
LGTM!

PR #2604, opened 2024-03-08
Upgrade Go version to 1.21

gapra-msft:
Merge after 10.24 release


go.mod
```diff
@@ -76,4 +75,4 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-go 1.19
+go 1.21
```
adreed-msft, line 78: I'm curious: Any particular reason to *not* use Go 1.22?

main.go
```diff
@@ -36,7 +36,7 @@ import (
 var glcm = common.GetLifecycleMgr()
 
 func main() {
-	rand.Seed(time.Now().UnixNano()) // make sure our random numbers actually are random (but remember, use crypto/rand for anything where strong/reliable randomness is required
+	rand.New(rand.NewSource(time.Now().UnixNano())) // make sure our random numbers actually are random (but remember, use crypto/rand for anything where strong/reliable randomness is required
```
adreed-msft, line 39: This is incorrect. `rand.New` returns a new `*rand.Rand`, and does not set the seed globally. Please track down all uses of `rand` and ensure they use this `*rand.Rand`

adreed-msft (changes requested):

main.go
adreed-msft: Actually lemme double correct this-- According to the docs, as of Go 1.20 this is no longer necessary at all. https://pkg.go.dev/math/rand#Seed

gapra-msft: will remove this line entirely

go.mod
gapra-msft: linting doesnt support 1.22 yet so I just set this to 121 for now. we can upgrade to 1.22 if we want to.

gapra-msft: actually it wasnt linting, it was the trivy vulnerability scanner that doesnt support 1.22 yet

pranavmalik-msft (approved):

main.go
```diff
@@ -21,23 +21,18 @@
...
-
-	"github.com/Azure/azure-storage-azcopy/v10/cmd"
-	"github.com/Azure/azure-storage-azcopy/v10/common"
 )
 
 // get the lifecycle manager to print messages
 var glcm = common.GetLifecycleMgr()
 
 func main() {
-	rand.Seed(time.Now().UnixNano()) // make sure our random numbers actually are random (but remember, use crypto/rand for anything where strong/reliable randomness is required
```
siminsavani-msft, line 39: What is the reasoning behind removing this randomization? /gen

siminsavani-msft (approved):

main.go
gapra-msft: This is no longer necessary with the latest Go version. Now rand is by default generating a random seed.

siminsavani-msft (approved):

Issue #2584, opened 2024-02-21
AzCopy sync failes to write hash data when using a readonly VSS shadow copy as the source


ktrace2:
### Which version of the AzCopy was used? 
10.23.0

### Which platform are you using?
Windows

### What command did you run?
sync #SourcePath# #BlobStoragePath# --recursive=true --exclude-pattern=*humbs.db --exclude-attributes=S;H --put-md5 --delete-destination=true --compare-hash=md5 --cap-mbps=900

### What problem was encountered?
Seeing this in the output:
INFO: One or more hash storage operations (read/write) have failed. Check the scanning log for details.

Seeing this in the logs:
failed to write hash data for #RelativePath#/#Filename#: failed to open data stream: open \\?\C:\Backup\SymbolicLink\#RelativePath#\#Filename#:.azcopysyncmeta: The media is write protected.

### How can we reproduce the problem in the simplest way?
- Create a VSS snapshot of a volume using diskshadow.exe
- Create a Symbolic Link to the shadow copy
   cmd /c mklink /d $SymbolicLinkPath "$ShadowDevicePath\"
- Run AzCopy sync using the Symbolic Link as the source

### Have you found a mitigation/solution?
No

The sync works but I am not clear if this will cause issues long term. It seems AzCopy saves the hash of every file as an Alternate Data Stream on the source files however this is now failing because the shadow copy is read only.

I ran the sync command directly on the source files before I started using VSS snapshots. Will this now cause problems because the read only snapshot copy has an out of date hash that can no longer be updated? I am seeing a lot of these messages:

File #FilePath# was skipped because the source has the same hash

adreed-msft:
(also, please re-open this issue if this doesn't entirely solve your issue" ^)

ktrace2:
I am trying this but it is failing with the following error (for every file):

failed to write hash data for Folder1/Filename.doc: failed to open hash meta file: open T:\AzCopyMetdata\ContainerA\Folder1\.Filename.doc.azcopysyncmeta: The system cannot find the path specified.

The arguments I passed were the following:

--compare-hash=md5 --local-hash-storage-mode=HiddenFiles --hash-meta-dir=T:\AzCopyMetdata\ContainerA

To confirm, T:\AzCopyMetdata\ContainerA is an empty directory. Should AzCopy create all the directories and files in here automatically?

adreed-msft:
[...]

PR #2611, opened 2024-03-14
Automatically create the hash data directory when using an alternative

adreed-msft:
Under current behavior, AzCopy appears to expect a similar file structure when specifying an alternate data directory for hashes. To remediate this issue (#2584), automatically create the parent directories when saving hash data to the disk.


common/hash_data_adapter_hidden_files.go
```diff
@@ -6,15 +6,18 @@ import (
...
 	"os"
 	"path/filepath"
+	"sync"
 )
 
 type HiddenFileDataAdapter struct {
 	hashBasePath string // "" == dataBasePath
 	dataBasePath string
+
+	createHashDirOnce *sync.Once
```
gapra-msft, line 16: Should we use this in the lines added below to create the hash dir? I am missing where this is used currently.

gapra-msft (commented):
Test changes look good. I have an open question about the sync.Once

common/hash_data_adapter_hidden_files.go
adreed-msft: It was a remnant of prior code changes in this PR-- I forgot to remove this, but, I'll trim it now.

Long story short-- We can't know the entire directory structure up front here, so we have to create it as we go.

gapra-msft (approved):

siminsavani-msft (approved):

e2etest/newe2e_sas_generation.go
```diff
@@ -205,6 +206,8 @@ func (vals GenericAccountSignatureValues) AsBlob() BlobSignatureValues {
 func (vals GenericAccountSignatureValues) AsFile() FileSignatureValues {
 	s := vals.withDefaults()
 
+	s.Permissions = strings.ReplaceAll(s.Permissions, "a", "") // remove 'a', because it's invalid and causes panics.
```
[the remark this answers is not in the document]
adreed-msft: Yes, it is not valid on a files signature.

gapra-msft (approved):

PR #2595, opened 2024-02-28
Gapra/list argument

gapra-msft:

PR #2615, opened 2024-03-18
[Docs] Clarify syntax for ambiguous flags (e.g. list-of-files, list-of-versions, metadata, tags)

siminsavani-msft:
Updating description in helpMessages.go and other <command>.go files to clarify syntax.


cmd/helpMessages.go
```diff
@@ -129,7 +129,7 @@ Download a subset of containers within a storage account by using a wildcard sym
 
   - azcopy cp "https://[srcaccount].blob.core.windows.net/[container*name]" "/path/to/dir" --recursive
 
-Download all the versions of a blob from Azure Storage to local directory. Ensure that source is a valid blob, destination is a local folder and versionidsFile which takes in a path to the file where each version is written on a separate line. All the specified versions will get downloaded in the destination folder specified.
+Download all the versions of a blob from Azure Storage to local directory listed in a text file (i.e, versionidsFile). Ensure that source is a valid blob, destination is a local folder and versionidsFile is a text file where each version is written on a separate line. All the specified versions will get downloaded in the destination folder specified.
```
schoag-msft, line 132: From https://github.com/Azure/azure-storage-azcopy/wiki/Listing-specific-files-to-transfer:

>> The reason that it's undocumented is that, at high file counts, it has slow performance for downloads and service-to-service copies. We want to fix that before we document it more officially

We should think about whether we want to peel this text back as the param --list-of-files is marked hidden for copy commands

cmd/helpMessages.go
```diff
@@ -365,7 +365,7 @@ Remove specified version ids of a blob from Azure Storage. Ensure that source is
 
   - azcopy rm "https://[srcaccount].blob.core.windows.net/[containername]/[blobname]" "/path/to/dir" --list-of-versions="/path/to/dir/[versionidsfile]"
 
-Remove specific blobs and virtual directories by putting their relative paths (NOT URL-encoded) in a file:
+Remove specific blobs and virtual directories by putting their relative paths (NOT URL-encoded) in a text file (i.e., list.txt) using the --list-of-files flag. In the text file, each blob and virtual directory is written on a separate line, see file contents below.
```
schoag-msft, line 368: From https://github.com/Azure/azure-storage-azcopy/wiki/Listing-specific-files-to-transfer:

>> The reason that it's undocumented is that, at high file counts, it has slow performance for downloads and service-to-service copies. We want to fix that before we document it more officially

We should think about whether we want to peel this text back as the param --list-of-files is marked hidden for copy commands

schoag-msft:
- list-of-files is hidden in copy but available in remove. There is no example to use this flag in helpMessages.go for copy (remove does have example).
- list-of-versions is available in copy and remove and there are examples for both commands in helpMessages.go

cmd/helpMessages.go
siminsavani-msft: list-of-files is marked hidden for copy and this particular text is for `remove`

cmd/helpMessages.go
```diff
@@ -365,7 +373,7 @@ Remove specified version ids of a blob from Azure Storage. Ensure that source is
...
```
gapra-msft, line 376: Why does this say see file contents below?

gapra-msft (approved):

cmd/copy.go
```diff
@@ -2035,23 +2035,23 @@ func init() {
...
-	cpCmd.PersistentFlags().StringVar(&raw.forceWrite, "overwrite", "true", "Overwrite the conflicting files and blobs at the destination if this flag is set to true. (default 'true') Possible values include 'true', 'false', 'prompt', and 'ifSourceNewer'. For destinations that support folders, conflicting folder-level properties will be overwritten this flag is 'true' or if a positive response is provided to the prompt.")
-	cpCmd.PersistentFlags().BoolVar(&raw.autoDecompress, "decompress", false, "Automatically decompress files when downloading, if their content-encoding indicates that they are compressed. The supported content-encoding values are 'gzip' and 'deflate'. File extensions of '.gz'/'.gzip' or '.zz' aren't necessary, but will be removed if present.")
-	cpCmd.PersistentFlags().BoolVar(&raw.recursive, "recursive", false, "Look into sub-directories recursively when uploading from local file system.")
+	cpCmd.PersistentFlags().StringVar(&raw.listOfFilesToCopy, "list-of-files", "", "Defines the location of text file which has the list of files to be copied. The text file should contain of each file name or directory written on a separate line.")
```
adreed-msft, line 2050: This should be noted as paths from the root, rather than just file names.

cmd/helpMessages.go
adreed-msft, line 132: move listed in a text file prior to local directory, currently reads like the directory should be in the text file

cmd/helpMessages.go
```diff
@@ -206,6 +206,14 @@ Copy all buckets to Blob Storage from Google Cloud Storage (GCS) by using a serv
 Copy a subset of buckets by using a wildcard symbol (*) in the bucket name from Google Cloud Storage (GCS) by using a service account key and a SAS token for destination. First, set the environment variables GOOGLE_APPLICATION_CREDENTIALS and GOOGLE_CLOUD_PROJECT=<project-id> for GCS source
  
   - azcopy cp "https://storage.cloud.google.com/[bucket*name]/" "https://[destaccount].blob.core.windows.net/?[SAS]" --recursive=true
+
+Copy a subset of files modified on or after the given date/time (in ISO8601 format) in a container by using the include-after flag.
```
adreed-msft, line 210: Note we output these in job logs

cmd/helpMessages.go
adreed-msft: list-of-files incurs a performance and monetary cost, even in remove. It should be re-implemented at some point to blindly perform the operations, imo. The theory behind list-of-files is that the enumeration work is done for us, so, that would make sense to do that way.

Either way-- I'd note that here (and everywhere list-of-files is visible), if we're revealing it. I'm not sure if this same thing applies to list-of-versions, too.

cmd/list.go
```diff
@@ -171,10 +171,12 @@ func init() {
...
 	}
 
-	listContainerCmd.PersistentFlags().BoolVar(&raw.MachineReadable, "machine-readable", false, "Lists file sizes in bytes.")
-	listContainerCmd.PersistentFlags().BoolVar(&raw.RunningTally, "running-tally", false, "Counts the total number of files and their sizes.")
-	listContainerCmd.PersistentFlags().BoolVar(&raw.MegaUnits, "mega-units", false, "Displays units in orders of 1000, not 1024.")
-	listContainerCmd.PersistentFlags().StringVar(&raw.Properties, "properties", "", "delimiter (;) separated values of properties required in list output.")
+	listContainerCmd.PersistentFlags().BoolVar(&raw.MachineReadable, "machine-readable", false, "False by default. Lists file sizes in bytes.")
+	listContainerCmd.PersistentFlags().BoolVar(&raw.RunningTally, "running-tally", false, "False by default. Counts the total number of files and their sizes.")
+	listContainerCmd.PersistentFlags().BoolVar(&raw.MegaUnits, "mega-units", false, "False by default. Displays units in orders of 1000, not 1024.")
+	listContainerCmd.PersistentFlags().StringVar(&raw.Properties, "properties", "", "Properties to be displayed in list output."+
```
adreed-msft, line 177: Can we implement some auto-updating solution like in #2595 here? This will keep the docs up to date for us.

cmd/remove.go
```diff
@@ -101,29 +101,29 @@ func init() {
...
+	deleteCmd.PersistentFlags().BoolVar(&raw.dryrun, "dry-run", false, "False by default. Prints the path files that would be removed by the command. This flag does not trigger the removal of the files.")
 	deleteCmd.PersistentFlags().StringVar(&raw.fromTo, "from-to", "", "Optionally specifies the source destination combination. For Example: BlobTrash, FileTrash, BlobFSTrash")
-	deleteCmd.PersistentFlags().StringVar(&raw.permanentDeleteOption, "permanent-delete", "none", "This is a preview feature that PERMANENTLY deletes soft-deleted snapshots/versions. Possible values include 'snapshots', 'versions', 'snapshotsandversions', 'none'.")
+	deleteCmd.PersistentFlags().StringVar(&raw.permanentDeleteOption, "permanent-delete", "none", "This is a preview feature that PERMANENTLY deletes soft-deleted snapshots/versions. Possible values include 'snapshots', 'versions', 'snapshotsandversions', 'none'. Default is 'none'.")
```
adreed-msft, line 117: ditto

cmd/remove.go
```diff
@@ -101,29 +101,29 @@ func init() {
...
-	deleteCmd.PersistentFlags().StringVar(&raw.trailingDot, "trailing-dot", "", "'Enable' by default to treat file share related operations in a safe manner. Available options: Enable, Disable. " +
-		"Choose 'Disable' to go back to legacy (potentially unsafe) treatment of trailing dot files where the file service will trim any trailing dots in paths. This can result in potential data corruption if the transfer contains two paths that differ only by a trailing dot (ex: mypath and mypath.). If this flag is set to 'Disable' and AzCopy encounters a trailing dot file, it will warn customers in the scanning log but will not attempt to abort the operation." +
+	deleteCmd.PersistentFlags().StringVar(&raw.trailingDot, "trailing-dot", "", "'Enable' by default to treat file share related operations in a safe manner. Available options: Enable, Disable. "+
```
adreed-msft, line 120: ditto

adreed-msft (approved):
Minor comments and maintainability options recommended, but otherwise changes look good

adreed-msft (approved):

PR #2616, opened 2024-03-18
Fail on File to Blob OAuth early since service does not support it yet

gapra-msft:


e2etest/zt_newe2e_error_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+)
+
+func init() {
+	suiteManager.RegisterSuite(&ErrorTestSuite{})
+}
+
+type ErrorTestSuite struct{}
+
+func (s *ErrorTestSuite) Scenario_FileBlobOAuthError(svm *ScenarioVariationManager) {
```
adreed-msft, line 13: Can we validate the particular reason for failure in this test?

adreed-msft (approved):

PR #2619, opened 2024-03-21
Added support for workload identity

gapra-msft:


common/oauthTokenManager.go
```diff
@@ -765,6 +796,11 @@ func (credInfo *OAuthTokenInfo) GetTokenCredential() (azcore.TokenCredential, er
 	if credInfo.PSCred {
 		return credInfo.GetPSContextCredential()
 	}
+
+	if credInfo.WorkloadIdentity {
```
adreed-msft, line 800: Q: is there a reason we continue to use several booleans here instead of a simple enum/switch like you've changed it to in login.go?

common/environment.go
```diff
@@ -164,6 +165,20 @@ func (EnvironmentVariable) ManagedIdentityResourceString() EnvironmentVariable {
...
+	}
+}
+
+func (EnvironmentVariable) WorkloadIdentityTokenFilePath() EnvironmentVariable {
+	return EnvironmentVariable{
+		Name:        "AZCOPY_WORKLOAD_TOKEN_FILE_PATH",
+		Description: "Token file path for Workload identity. This variable is only used for auto login, please use the command line flag instead when invoking the login command.",
+	}
+}
+
```
sugibuchi, line 181: Client IDs and paths of service account token files for Azure Workload Identity are supposed not to be manually set by users.

Instead, those values are supposed to be automatically set by the runtime environment (the Azure Workload Identity Webhook in AKS environment, for example) as the following environment variables.

| Environment variable | Description |
| -- | -- |
| AZURE_AUTHORITY_HOST | The Azure Active Directory (AAD) endpoint. |
| AZURE_CLIENT_ID | The client ID of the AAD application or user-assigned managed identity. |
| AZURE_TENANT_ID | The tenant ID of the registered AAD application or user-assigned managed identity. |
| AZURE_FEDERATED_TOKEN_FILE | The path of the projected service account token file. |

https://azure.github.io/azure-workload-identity/docs/quick-start.html#7-deploy-workload -> See "Output"

Therefore, we should retrieve a client ID and the path of the service account token file from the env variables `AZURE_CLIENT_ID` and `AZURE_FEDERATED_TOKEN_FILE`, not from `AZCOPY_WORKLOAD_CLIENT_ID` and `AZCOPY_WORKLOAD_TOKEN_FILE_PATH`.

In addition, we should also retrieve a tenant ID from `AZURE_TENANT_ID`.

This is the default behavior implemented by Azure Identity SDK. Example:

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/workload_identity.go#L42-L53

common/oauthTokenManager.go
gapra-msft: oh whoops this was an oversight from before I changed to the string version. I'll clean this up.

common/oauthTokenManager.go
```diff
@@ -476,7 +490,7 @@ func (credInfo *OAuthTokenInfo) Refresh(ctx context.Context) (*adal.Token, error
 	if err != nil {
 		return nil, err
 	}
-	if credInfo.TokenRefreshSource == "tokenstore" || credInfo.Identity || credInfo.ServicePrincipalName {
+	if credInfo.TokenRefreshSource == "tokenstore" || credInfo.Identity || credInfo.ServicePrincipalName || credInfo.AzCLICred || credInfo.PSCred || credInfo.WorkloadIdentity {
```
gapra-msft, line 493: TODO: clean this up

common/oauthTokenManager.go
```diff
@@ -425,7 +432,9 @@ type OAuthTokenInfo struct {
 	ServicePrincipalName    bool `json:"_spn"`
 	SPNInfo                 SPNInfo
 	AzCLICred               bool
-	PSCred					bool
+	PSCred                  bool
```
gapra-msft, line 435: remove these booleans

common/environment.go
gapra-msft: All these environment variables will work by default as we use the Identity SDK under the hood.

gapra-msft: @adreed-msft This is what I was mentioning in the description of the PR. I think it might be cleaner just to always use environment variables and let the identity SDK handle the reading and parsing of them. Then we can get rid of our environment variables for this and our CLI params. The only thing the user would have to do is either do azcopy login login-type=workload or set the auto login env var to workload

cmd/login.go
```diff
@@ -110,45 +122,19 @@ type loginCmdArgs struct {
 }
 
 func (lca loginCmdArgs) validate() error {
```
gapra-msft, line 124: TODO : This function def needs to change to func (lca *loginCmdArgs) validate() error {

common/environment.go
```diff
@@ -83,17 +83,18 @@ func (EnvironmentVariable) UserDir() EnvironmentVariable {
...
+	AutologinTypeAzCLI    = "azcli"
+	AutologinTypePsCred   = "pscred"
+	AutologinTypeWorkload = "workload"
 )
 
 func (EnvironmentVariable) AutoLoginType() EnvironmentVariable {
 	return EnvironmentVariable{
 		Name:        "AZCOPY_AUTO_LOGIN_TYPE",
-		Description: "Specify the credential type to access Azure Resource without invoking the login command and using the OS secret store, available values SPN, MSI, DEVICE, AZCLI, and PSCRED  - sequentially for Service Principal, Managed Service Identity, Device workflow, Azure CLI, or Azure PowerShell.",
+		Description: "Specify the credential type to access Azure Resource without invoking the login command and using the OS secret store, available values SPN, MSI, DEVICE, AZCLI, and PSCRED, WORKLOAD  - sequentially for Service Principal, Managed Service Identity, Device workflow, Azure CLI, Azure PowerShell, or Workload Identity.",
```
adreed-msft, line 97: Nit: We should maybe list all of these out for maintainability reasons, that way, if somebody adds a credential type, it automatically appears here.

common/oauthTokenManager.go
```diff
@@ -420,18 +423,9 @@ type OAuthTokenInfo struct {
 	ActiveDirectoryEndpoint string `json:"_ad_endpoint"`
 	TokenRefreshSource      string `json:"_token_refresh_source"`
 	ApplicationID           string `json:"_application_id"`
-	Identity                bool   `json:"_identity"`
+	LoginType               string `json:"_login_type"`
```
adreed-msft, line 426: This might be better as an enum, and could pair into the previous comment.

common/environment.go
gapra-msft: this is a good idea

e2etest/zt_newe2e_workload_test.go
```diff
@@ -0,0 +1,70 @@
...
+	suiteManager.RegisterSuite(&WorkloadIdentitySuite{})
+}
+
+type WorkloadIdentitySuite struct{}
+
+// Run only in environments that support and are set up for Workload Identity (ex: Azure Pipeline, Azure Kubernetes Service)
+func (s *WorkloadIdentitySuite) Scenario_SingleFileUploadDownloadWorkloadIdentity(svm *ScenarioVariationManager) {
+	// Run only in environments that support and are set up for Workload Identity (ex: Azure Pipeline, Azure Kubernetes Service)
+	if os.Getenv("NEW_E2E_ENVIRONMENT") != "AzurePipeline" {
+		svm.InvalidateScenario()
```
adreed-msft, line 19: Nit: logging skips may be a good idea-- But there's no proper mechanism for this at the moment.

InvalidateScenario was intended to basically be "in no context does this scenario ever make sense, or should it ever exist", like local to local.

common/oauthTokenManager.go
```diff
@@ -476,7 +470,7 @@ func (credInfo *OAuthTokenInfo) Refresh(ctx context.Context) (*adal.Token, error
 	if err != nil {
 		return nil, err
 	}
-	if credInfo.TokenRefreshSource == "tokenstore" || credInfo.Identity || credInfo.ServicePrincipalName {
+	if credInfo.TokenRefreshSource == "tokenstore" || credInfo.LoginType != AutologinTypeDevice {
```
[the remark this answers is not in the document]
gapra-msft: will change the var name to LoginType and use the json rep _token_refresh_source to prevent breaking storage explorer.

adreed-msft (approved):

e2etest/zt_newe2e_workload_test.go
gapra-msft: I can add a svm.Skip thing and enhance the e2e framework then.

pranavmalik-msft (approved):
LGTM

PR #2620, opened 2024-03-23
Fix path joining in and test --delete-destination

adreed-msft:
This PR is presently based on create-hash-data-folder, because much of the work is re-used.

Currently, `--delete-directory` is completely broken, because the usage of `path.Join` in `syncProcessor.go` was incorrect. This PR adds testing in for the feature in the E2E testing framework, and re-uses the work from it's base PR for objects creating their parents in the E2E framework.


gapra-msft (approved):
lgtm

siminsavani-msft (approved):

PR #2630, opened 2024-04-01
Alter stdout handling/parsing to be performed mid-job rather than after the fact

adreed-msft:
Based off Gauri's list json PR-- Alters log scanning to be performed as the job is running, rather than after the fact. This should help support any interactive functions of AzCopy in the E2E framework.


gapra-msft:
Can we resolve the conflicts on this PR and prioritize getting this merged? This will be very helpful in debugging issues with New E2E tests.

e2etest/newe2e_runazcopy_stdout.go
```diff
@@ -0,0 +1,198 @@
...
+				var object cmd.AzCopyListObject
+				err = json.Unmarshal([]byte(line.MessageContent), &object)
+				if err != nil {
+					return
+				}
+
+				a.InsertObject(object)
+
+			case "ListSummary":
+				_ = json.Unmarshal([]byte(line.MessageContent), &a.Summary)
```
gapra-msft, line 144: should we catch an error here?

adreed-msft: Probably. Actually, there's room to suggest that "probably not", because if writing fails, the exec.Cmd stops running.

e2etest/newe2e_assertions.go
```diff
@@ -240,11 +240,16 @@ func (m MapContains[K, V]) Assert(items ...any) bool {
 		kvPair, ok := v.(KVPair[K, V])
 
 		if !ok {
-			val, ok := v.(V)
-			if !ok {
-				panic("MapContains only accepts KVPair[K,V] or V as items")
+			if val, ok := v.(V); ok {
+				kvPair = m.ValueToKVPair(val)
```
gapra-msft, line 244: Should we remove the condition of v.(V) since we are now using the condition below?

gapra-msft (approved):

siminsavani-msft (approved):

e2etest/newe2e_assertions.go
adreed-msft: I don't think it's bad functionality to have, honestly, even as a fallback.

gapra-msft (approved):

siminsavani-msft (approved):

Issue #2606, opened 2024-03-13
azcopy copy memory error on 10.23.0 MacOS via brew using SAS


bfox-sugarshot:
### What problem was encountered? `INFO: Failed to create one or more destination container(s). Your transfers may still succeed if the container already exists.
INFO: Any empty folders will not be processed, because source and/or destination doesn't have full folder support
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x10162b3a0]

goroutine 1 [running]:
github.com/Azure/azure-storage-azcopy/v10/cmd.(*blobTraverser).Traverse(0x1400021a230, 0x0?, 0x0?, {0x10219dae0, 0x0, 0x0})
	github.com/Azure/azure-storage-azcopy/v10/cmd/zc_traverser_blob.go:231 +0x570
github.com/Azure/azure-storage-azcopy/v10/cmd.(*CopyEnumerator).enumerate(0x140000fa140)
	github.com/Azure/azure-storage-azcopy/v10/cmd/zc_enumerator.go:766 +0x48
github.com/Azure/azure-storage-azcopy/v10/cmd.(*CookedCopyCmdArgs).processCopyJobPartOrders(0x1400010a900)
	github.com/Azure/azure-storage-azcopy/v10/cmd/copy.go:1610 +0xb5c
github.com/Azure/azure-storage-azcopy/v10/cmd.(*CookedCopyCmdArgs).process(0x14000252a20?)
	github.com/Azure/azure-storage-azcopy/v10/cmd/copy.go:1262 +0x74
github.com/Azure/azure-storage-azcopy/v10/cmd.init.2.func2(0x14000423b80?, {0x140004e80e0?, 0x2?, 0x1016416f6?})
	github.com/Azure/azure-storage-azcopy/v10/cmd/copy.go:2014 +0x1b0
github.com/spf13/cobra.(*Command).execute(0x14000423b80, {0x140004e80a0, 0x2, 0x2})
	github.com/spf13/cobra@v1.4.0/command.go:860 +0x550
github.com/spf13/cobra.(*Command).ExecuteC(0x102107e80)
	github.com/spf13/cobra@v1.4.0/command.go:974 +0x318
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v1.4.0/command.go:902
[...]

PR #2635, opened 2024-04-03
AzCopy panics when uploading 16.5 TB disk snapshot to a container in blob storage

siminsavani-msft:
- Retrieve LMT through a helper method that handles if LMT is nil or not
- Retrieve ContentLength through a helper method that handles if ContentLength is nil or not
- E2E tests written to test locally and avoid lengthy testing time in CI due to large size of managed disk snapshot (7.9 TB). Must be set up locally.
- Unit tests that mocks service GetProperties response
- Resolves https://github.com/Azure/azure-storage-azcopy/issues/2606


e2etest/zt_managed_disks_test.go
```diff
@@ -147,3 +147,61 @@ func TestManagedDisks_OAuthRequired(t *testing.T) {
 		"",
 	)
 }
+
+// Test a managed disk of size 7.9 TB
+func TestManagedDisks_LargeManagedDisk(t *testing.T) {
```
gapra-msft, line 152: Instead of adding to the legacy e2e tests I'd like to start seeing people use the new e2e test framework.

cmd/zc_traverser_blob.go
```diff
@@ -223,16 +223,17 @@ func (t *blobTraverser) Traverse(preprocessor objectMorpher, processor objectPro
...
 			"",
-			getEntityType(blobProperties.Metadata),
-			*blobProperties.LastModified,
-			*blobProperties.ContentLength,
-			blobPropertiesResponseAdapter{blobProperties},
-			blobPropertiesResponseAdapter{blobProperties},
-			blobProperties.Metadata,
+			getEntityType(blobPropsAdapter.Metadata),
+			blobPropsAdapter.LastModified(),
+			*blobPropsAdapter.ContentLength,
```
gapra-msft, line 233: Should we apply this fix for ContentLength too?

e2etest/zt_managed_disks_test.go
```diff
@@ -147,3 +147,61 @@ func TestManagedDisks_OAuthRequired(t *testing.T) {
...
+	if runtime.GOOS != "linux" {
+		t.Skip("Limit runs to Linux so no simultaneous runs occur")
+		return
+	}
+
+	RunScenarios(
+		t,
+		eOperation.Copy(),
+		eTestFromTo.Other(common.EFromTo.BlobLocal(), common.EFromTo.BlobLocal()),
```
gapra-msft, line 161: this is repeating BlobLocal twice

cmd/zc_traverser_blob.go
siminsavani-msft: I thought about this as well but decided against it as it's not entirely related to the context of the PR. I can totally do this in a separate PR if you are cool with that!

adreed-msft (changes requested):

e2etest/zt_managed_disks_test.go
gapra-msft, line 152: Are we trying to download an 8TB md in the pipeline? will that even work? this will probably also eat up waaaay too much time if we run this live.

siminsavani-msft: That's correct and that is a valid concern. I could potentially skip the test entirely and/or pursue testing in unit tests or new e2e.

cmd/zt_traverser_managed_disk_test.go
```diff
@@ -0,0 +1,52 @@
...
+	//a.Nil(blobProps.LastModified)
+	//
+	//blobTraverser := newBlobTraverser(largeMDSnapshot, serviceClientWithSAS, ctx, true, false, func(common.EntityType) {}, false, common.CpkOptions{}, false, false, false, common.EPreservePermissionsOption.None(), false)
+	//blobDummyProcessor := dummyProcessor{}
+	//err = blobTraverser.Traverse(noPreProccessor, blobDummyProcessor.process, nil)
+	//a.NoError(err)
+}
+
+func TestLargeManagedDisk(t *testing.T) {
+	//a := assert.New(t)
```
gapra-msft, line 31: I think I prefer these tests existing in the E2E test framework instead of as a unit test. Instead of commenting you should be able to write code to skip the test and enable it based on a flag.

seanmcc-msft:
1. Create end-to-end test to test locally
2. Run this test locally
3. Mark test as skip, follow up with Gauri on how to do this.
4. Create unit test with mocks

cmd/zt_traverser_managed_disk_test.go
```diff
@@ -0,0 +1,52 @@
+package cmd
+
+import "testing"
+
+func TestLargeManagedDiskSnapshot(t *testing.T) {
```
adreed-msft, line 5: technically the issue is "covered" by the other test, but I wonder if we can't mock just the single request or two that'd get made by the HTTP client under the hood here.

adreed-msft (approved):

cmd/zt_traverser_managed_disk_test.go
siminsavani-msft: GetProperties request/responses are mocked here cmd/zt_traverser_blob_test.go!

adreed-msft (approved):

cmd/zt_traverser_managed_disk_test.go
gapra-msft: Why do we want to keep these tests in comments?

cmd/zt_traverser_blob_test.go
```diff
@@ -289,3 +295,116 @@ func TestGetEntityType(t *testing.T) {
 	a.Equal(common.EEntityType.Symlink(), entityType)
 
 }
+
+func TestBlobPropertiesAdapter_LMTAndContentLength(t *testing.T) {
```
gapra-msft, line 299: can we delete the first 2 tests here since the last should cover this right?

gapra-msft (approved):

gapra-msft (approved):

PR #2643, opened 2024-04-08
Switch to using track 2 identity token caching for Device Login

pranavmalik-msft:
### Description
This pull request aims to facilitate the migration process from the existing ADAL library to the MSAL framework, specifically for device login functionalities.

### Changes
- The existing flow for the ADAL framework has been removed regarding Device Login with the MSAL framework.
- All references to ADAL library has been removed -for token caching, refreshing and persistence.
- Adding new token persistence logic


cmd/credentialUtil.go
```diff
@@ -138,7 +129,7 @@ func GetOAuthTokenManagerInstance() (*common.UserOAuthTokenManager, error) {
 			return
 		}
 
-		lca.persistToken = false
+		lca.persistToken = true
```
gapra-msft, line 132: Why are we changing behavior to start persisting tokens on auto login?

cmd/copy.go
```diff
@@ -1449,7 +1449,7 @@ func (cca *CookedCopyCmdArgs) getSrcCredential(ctx context.Context, jpo *common.
 			// get token from env var or cache
 			if tokenInfo, err := uotm.GetTokenInfo(ctx); err != nil {
 				return srcCredInfo, err
-			} else if _, err := tokenInfo.GetTokenCredential(); err != nil {
+			} else if _, err := tokenInfo.GetTokenCredential(true); err != nil {
```
gapra-msft, line 1452: Same here, why are we hardcoding the persistence behavior?

cmd/jobsResume.go
```diff
@@ -288,7 +288,7 @@ func (rca resumeCmdArgs) getSourceAndDestinationServiceClients(
 			return nil, nil, err
 		}
 
-		tc, err = tokenInfo.GetTokenCredential()
+		tc, err = tokenInfo.GetTokenCredential(true)
```
gapra-msft, line 291: ditto

cmd/loginStatus.go
```diff
@@ -51,7 +52,7 @@ func init() {
 			uotm := GetUserOAuthTokenManagerInstance()
 			tokenInfo, err := uotm.GetTokenInfo(ctx)
 
-			if err == nil && !tokenInfo.IsExpired() {
+			if err == nil && tokenInfo == nil {
```
gapra-msft, line 55: I'm a little lost here. How does tokenInfo being nil mean the token refreshed successfully?

cmd/sync.go
```diff
@@ -700,7 +700,7 @@ func (cca *cookedSyncCmdArgs) process() (err error) {
 		// Get token from env var or cache.
 		if tokenInfo, err := uotm.GetTokenInfo(ctx); err != nil {
 			return err
-		} else if _, err := tokenInfo.GetTokenCredential(); err != nil {
+		} else if _, err := tokenInfo.GetTokenCredential(true); err != nil {
```
gapra-msft, line 703: same concern about persistence

common/credCacheInternal_linux.go
```diff
@@ -54,22 +53,6 @@ func (c *CredCacheInternalIntegration) HasCachedToken() (bool, error) {
...
-// SaveToken saves an oauth token.
-func (c *CredCacheInternalIntegration) SaveToken(token OAuthTokenInfo) error {
-	c.lock.Lock()
-	err := c.saveTokenInternal(token)
-	c.lock.Unlock()
-	return err
-}
-
 // LoadToken gets the cached oauth token.
 func (c *CredCacheInternalIntegration) LoadToken() (*OAuthTokenInfo, error) {
```
gapra-msft, line 57: Do we even need any credCache code in AzCopy anymore? Could we just remove all of these files?

common/oauthTokenManager.go
```diff
@@ -305,33 +291,7 @@ func (uotm *UserOAuthTokenManager) UserLogin(tenantID, activeDirectoryEndpoint s
 // If refresh token is expired, the method will fail and return failure reason.
 // Fresh token is persisted if access token or refresh token is changed.
 func (uotm *UserOAuthTokenManager) getCachedTokenInfo(ctx context.Context) (*OAuthTokenInfo, error) {
```
gapra-msft, line 293: Can we just delete this if its always returning nil, nil?

common/oauthTokenManager.go
```diff
@@ -340,12 +300,68 @@ func (uotm *UserOAuthTokenManager) HasCachedToken() (bool, error) {
 		return true, nil
 	}
 
-	return uotm.credCache.HasCachedToken()
+	// TODO: Revisit
```
gapra-msft, line 303: Does this still need to be actioned on?

common/oauthTokenManager.go
```diff
@@ -396,11 +412,7 @@ func (uotm *UserOAuthTokenManager) getTokenInfoFromEnvVar(ctx context.Context) (
 	}
 
 	if tokenInfo.TokenRefreshSource != TokenRefreshSourceTokenStore {
-		refreshedToken, err := tokenInfo.Refresh(ctx)
-		if err != nil {
-			return nil, fmt.Errorf("get token from environment variable failed to ensure token fresh, %v", err)
-		}
-		tokenInfo.Token = *refreshedToken
+		panic("Invalid Token Refresh Source")
```
gapra-msft, line 415: Why are we panicking here? Refreshing should mean re checking the environment variable.

common/oauthTokenManager.go
```diff
@@ -434,6 +448,17 @@ type OAuthTokenInfo struct {
 	ClientID string `json:"_client_id"`
 }
 
+func (t *OAuthTokenInfo) Expires() time.Time {
+	s, err := t.ExpiresOn.Float64()
+	if err != nil {
```
gapra-msft, line 453: why are we overwriting the expiry time if there is no error?

common/oauthTokenManager.go
```diff
@@ -523,61 +500,35 @@ func getAuthorityURL(tenantID, activeDirectoryEndpoint string) (*url.URL, error)
...
-	if err != nil {
-		return azcore.AccessToken{}, fmt.Errorf("get cached token failed in Token Store Mode(SE), %v", err)
-	}
-
-	tsc.token = &azcore.AccessToken{
-		Token:     tokenInfo.AccessToken,
-		ExpiresOn: tokenInfo.Expires(),
-	}
-
+	tsc.token = &azcore.AccessToken{}
```
gapra-msft, line 524: Why are we returning an empty token?

common/oauthTokenManager.go
```diff
@@ -617,7 +568,7 @@ func (credInfo *OAuthTokenInfo) GetManagedIdentityCredential() (azcore.TokenCred
 	return tc, nil
 }
 
-func (credInfo *OAuthTokenInfo) GetClientCertificateCredential() (azcore.TokenCredential, error) {
+func (credInfo *OAuthTokenInfo) GetClientCertificateCredential(persist bool) (azcore.TokenCredential, error) {
```
gapra-msft, line 571: Instead of passing persist to each method, we should store it as part of OAuthTokenInfo

gapra-msft (changes requested):

cmd/loginStatus.go
pranavmalik-msft: Token refresh is now seamlessly handled by the Az-identity codebase, alleviating concerns about token expiration.

common/oauthTokenManager.go
pranavmalik-msft: We relocated this function to a new location since its previous file was removed due to non-usage.

common/credCacheInternal_linux.go
pranavmalik-msft: Removed in the latest commit

cmd/credentialUtil.go
pranavmalik-msft: Done in the latest commit.

cmd/copy.go
pranavmalik-msft: Done in the latest commit.

cmd/jobsResume.go
pranavmalik-msft: Done in the latest commit.

cmd/sync.go
pranavmalik-msft: Done in the latest commit.

common/oauthTokenManager.go
pranavmalik-msft: Done in the latest commit.

common/oauthTokenManager.go
pranavmalik-msft: Done in the latest commit.

common/oauthTokenManager.go
pranavmalik-msft: I need some context here - lets discuss offline on this.

common/oauthTokenManager.go
pranavmalik-msft: I need some context here - lets discuss offline on this.

common/oauthTokenManager.go
pranavmalik-msft: Done in the latest commit.

cmd/loginStatus.go
gapra-msft: This doesnt quite answer the question. If the token is refreshed, why would it be nil?

common/credCacheInternal_linux.go
gapra-msft: Why are we unable to? What piece of code in AzCopy still relies on this (other than cred cache)?

cmd/loginStatus.go
vibhansa-msft: If AzIdentity is handling everything about the token, do we really need to even check token is there or not ?

gapra-msft: I think we might still want to check the expiry of the token since this is checking status of the login.

common/oauthTokenManager.go
```diff
@@ -118,13 +112,9 @@ func (uotm *UserOAuthTokenManager) GetTokenInfo(ctx context.Context) (*OAuthToke
 		if err != nil { // this is the case when env var exists while get token info failed
 			return nil, err
 		}
-	} else { // Scenario: session mode which get token from cache
```
gapra-msft, line 121: What scenarios is this removing?

common/oauthTokenManager.go
```diff
@@ -290,62 +282,75 @@ func (uotm *UserOAuthTokenManager) UserLogin(tenantID, activeDirectoryEndpoint s
...
-func (uotm *UserOAuthTokenManager) getCachedTokenInfo(ctx context.Context) (*OAuthTokenInfo, error) {
-	hasToken, err := uotm.credCache.HasCachedToken()
-	if err != nil {
-		return nil, fmt.Errorf("no cached token found, please log in with azcopy's login command, %v", err)
-	}
-	if !hasToken {
-		return nil, errors.New("no cached token found, please log in with azcopy's login command")
+// HasCachedToken returns if there is cached token in token manager.
+func (uotm *UserOAuthTokenManager) HasCachedToken() (bool, error) {
+	if uotm.stashedInfo != nil {
```
gapra-msft, line 290: this should technically check the record right?

common/oauthTokenManager.go
```diff
@@ -290,62 +282,75 @@ func (uotm *UserOAuthTokenManager) UserLogin(tenantID, activeDirectoryEndpoint s
...
-	return uotm.credCache.RemoveCachedToken()
+	filePath := filepath.Join(AzcopyJobPlanFolder, TokenCred)
+
+	// Check if the file exists
+	if _, err := os.Stat(filePath); err == nil {
+		// File exists, so delete it
+		if err := os.Remove(filePath); err != nil {
+			return fmt.Errorf("error deleting file: %v", err)
+		}
+		fmt.Println("File deleted successfully.")
```
gapra-msft, line 347: Is this normal for logging in the OAuth token manager?

common/oauthTokenManager.go
adreed-msft: When we try to figure out what type of credential to use, AzCopy makes no attempt to poll the cache on this branch, and falls back to "Unknown", which should always generate an error on Blob.

adreed-msft (changes requested):

common/oauthTokenManager.go
pranavmalik-msft: Hey @adreed-msft ,
Could you please clarify the specific scenarios you're encountering? Both logging in and copying data seem to be functioning correctly on my end. Initially, I log in via DEVICE-Login, then proceed to copy data between two blobs, which operates smoothly. Even after logging out successfully, I repeat the process by logging in again, and the scenario continues to work seamlessly.

gapra-msft: Have you tried performing the login status command?

PR #2644, opened 2024-04-08
[Supportability] Improve error message when performing uploads with a missing SAS token

siminsavani-msft:
INFO messages are updated to indicate preferred authentication methods, given that credential type is Unknown and resource is Azure.

```
INFO: Scanning...
INFO: Autologin not specified.
INFO: Authenticating to destination using Unknown, Please authenticate using Microsoft Entra ID (https://aka.ms/AzCopy/AuthZ), use AzCopy login, or append a SAS token to your Azure URL.
INFO: Authenticating to source using Unknown, Please authenticate using Microsoft Entra ID (https://aka.ms/AzCopy/AuthZ), use AzCopy login, or append a SAS token to your Azure URL.
```


cmd/credentialUtil.go
```diff
@@ -345,6 +345,9 @@ func logAuthType(ct common.CredentialType, location common.Location, isSource bo
 		name = "Azure AD (Managed Disk)"
 	}
 	message := fmt.Sprintf("Authenticating to %s using %s", resource, name)
+	if ct == common.ECredentialType.Unknown() && location.IsAzure() {
+		message += ", Please authenticate using Microsoft Entra ID (https://aka.ms/AzCopy/AuthZ), use AzCopy login, or append SAS to your Azure URL."
```
adreed-msft, line 349: "append *a* SAS token"

Second note-- Noting whether this is the source or destination can help customers determine the issue.

e2etest/zt_newe2e_basic_functionality_test.go
```diff
@@ -64,3 +65,55 @@ func (s *BasicFunctionalitySuite) Scenario_SingleFileUploadDownload(svm *Scenari
...
 	}, true)
 }
+
+func (s *BasicFunctionalitySuite) Scenario_SingleFileUploadDownload_EmptySAS(svm *ScenarioVariationManager) {
+	azCopyVerb := ResolveVariation(svm, []AzCopyVerb{AzCopyVerbCopy, AzCopyVerbSync})
+
+	dstObj := CreateResource[ContainerResourceManager](svm, GetRootResource(svm, ResolveVariation(svm, []common.Location{common.ELocation.Local(), common.ELocation.Blob(), common.ELocation.File(), common.ELocation.BlobFS()})), ResourceDefinitionContainer{}).GetObject(svm, "test", common.EEntityType.File())
+	// The object must exist already if we're syncing.
+	if azCopyVerb == AzCopyVerbSync {
+		dstObj.Create(svm, NewZeroObjectContentContainer(0), ObjectProperties{})
```
adreed-msft, line 75: There is no need to make resources if we're just going to fail intentionally in a case where we'll never truly hit the service

adreed-msft (approved):

cmd/credentialUtil.go
siminsavani-msft: The info message starts with listing if the source or destination is being authenticated, I don't think it's necessary to repeat the information - `INFO: Authenticating to destination using Unknown`.

gapra-msft (approved):

PR #2652, opened 2024-04-19
Non S3 URI as source resulting on local filesystem scanning

pranavmalik-msft:
### Description
An issue surfaced in Azcopy where users encountered a bug during copy operations with a source URI of "https://isd-storage.obs.ae-ad-1.g42cloud.com". This triggered a scan of the local file system, resulting in failures. Upon investigation, it was discovered that unsupported cloud platforms, such as **g42cloud**, were incorrectly identified as local sources by the backend.
As can be observed below-
f47954df-28a8-46ac-b3bb-fc5f388404d4

This PR addresses this issue by categorizing all unsupported cloud platforms as `Unknown `instead of `Local`, thus preventing repetitive scans of local files and ensuring early failure in processing.
According to the new flow it would be as follows:

### Implementation:
- Implemented a check to categorize unsupported cloud platforms as `Unknown`.

### Tests:
- Added Unit Tests to verify this flow and enhance code coverage.


cmd/validators.go
```diff
@@ -164,6 +165,9 @@ func InferArgumentLocation(arg string) common.Location {
 			if common.IsGCPURL(*u) {
 				return common.ELocation.GCP()
 			}
+
+			// If none of the above conditions match, return Unknown
+			return common.ELocation.Unknown()
```
siminsavani-msft, line 170: How does this impact non standard URLs?

siminsavani-msft (commented):
Let's add some tests here!

cmd/validators.go
pranavmalik-msft: non standard URL wont be processed , they would follow the default case.

adreed-msft:
Needs tests, otherwise lgtm

adreed-msft (changes requested):
Needs tests

pranavmalik-msft:
> Needs tests

Done

siminsavani-msft (approved):

adreed-msft (approved):

PR #2658, opened 2024-04-23
Switch to using track 2 identity token caching for Device Login

pranavmalik-msft:
### Description
- This pull request aims to facilitate the migration process from the existing ADAL library to the MSAL framework, specifically for device login functionalities.

### Design Doc
[Azcopy-Track2 Migration](https://microsoftapc-my.sharepoint.com/:w:/g/personal/pranavmalik_microsoft_com/EU-pvoYU7TFKnwe8qIz334EBnnNYpmJlhro6DfgGVarSww?e=P28yq1)

### Changes
- The existing flow for the ADAL framework has been removed regarding Device Login with the MSAL framework.
- All references to ADAL library has been removed -for token caching, refreshing and persistence.
- Adding new token persistence logic

### Tests
-  added the reintegration with CredCache tests.


adreed-msft:
Hi, Pranav-- Could you more clearly label your PRs in the future? "Donot merge" isn't very concise, and Github has a feature for draft PRs to add a barrier between merging.

common/credCacheInternal_linux.go
```diff
@@ -105,7 +105,14 @@ func (c *CredCacheInternalIntegration) removeCachedTokenInternal() error {
 	return errors.New("Not implemented")
 }
 
+// saveTokenInternal persists an oauth token on disk.
```
vibhansa-msft, line 108: If not useful, why do we want to implement this function? Where is it used currently?

pranavmalik-msft: So this is existing in our original workflow , right now its not used anywhere hence not implemented .

common/credCache_darwin.go
```diff
@@ -25,7 +25,7 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/wastore/keychain" // forked and customized from github.com/keybase/go-keychain, todo: make a release to ensure stability
+	"github.com/keybase/go-keychain" // forked and customized from github.com/keybase/go-keychain, todo: make a release to ensure stability
```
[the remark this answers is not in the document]
pranavmalik-msft: yes, Az identity-cache module  brings github.com/keybase/go-keychain dependency with itself , we had removed github.com/wastore/keychain dependency because it was causing build failures and co-existing issues , hence we had to fall back to this dependency in order to avoid conflicts.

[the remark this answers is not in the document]
gapra-msft: it says wastore is forked and customized from keybase, what are the customizations that were made and do they need to exist to continue to work with storage explorer?

common/credCacheInternal_linux.go
pranavmalik-msft: changes for this module has been un-done.

common/credCache_darwin.go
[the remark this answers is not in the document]
pranavmalik-msft: github.com/wastore/keychain- is a slim library for accessing the Keychain for macOS in Go its build ontop of github.com/keybase/go-keychain it contains wrappers around the functionalities present in wastore/keychain . I dont believe it should cause a problem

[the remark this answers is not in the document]
vibhansa-msft: Does this code compile and work on Mac ?

common/oauthTokenManager.go
```diff
@@ -300,38 +326,122 @@ func (uotm *UserOAuthTokenManager) UserLogin(tenantID, activeDirectoryEndpoint s
...
+	// Create oAuthTokenInfo from the existing AuthenticationRecord
+	oAuthTokenInfo := &OAuthTokenInfo{
+		Tenant:   record.TenantID,
+		ClientID: record.ClientID,
+	}
+
+	tc, err := oAuthTokenInfo.GetTokenCredential()
 	if err != nil {
-		return nil, fmt.Errorf("get cached token failed, %v", err)
+		return nil, err
```
vibhansa-msft, line 380: Does these changes mess with the error message also as we have removed the older message and now just returning the error object itself.

cmd/loginStatus.go
```diff
@@ -51,7 +53,7 @@ func init() {
 			uotm := GetUserOAuthTokenManagerInstance()
 			tokenInfo, err := uotm.GetTokenInfo(ctx)
 
-			if err == nil && !tokenInfo.IsExpired() {
+			if err == nil && tokenInfo.AccessToken.ExpiresOn.After(time.Now()) {
```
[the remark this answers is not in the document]
pranavmalik-msft: the IsExpired method is part of the adal library that we had removed , I have kept the implementation intact here for the same.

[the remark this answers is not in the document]
gapra-msft: What about this method here https://github.com/Azure/azure-storage-azcopy/pull/2658/files#diff-9b3756871a6cee5f762bf39ab9400fea295daedbb2c56808546af29b06a48977R561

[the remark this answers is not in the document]
pranavmalik-msft: I think the link is redirecting here only.

seanmcc-msft (changes requested):
Tests are required.

cmd/loginStatus.go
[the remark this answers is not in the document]
gapra-msft: oauthTokenManager line 561

common/credCacheInternal_linux.go
```diff
@@ -106,6 +106,7 @@ func (c *CredCacheInternalIntegration) removeCachedTokenInternal() error {
 }
 
 // loadTokenInternal restores a Token object from file cache.
+//
```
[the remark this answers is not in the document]
pranavmalik-msft: Done in the latest commit

common/credCache_darwin.go
gapra-msft, line 28:
```suggestion
	"github.com/keybase/go-keychain"
```

common/oauthTokenManager.go
pranavmalik-msft: That section of code has been moved up, its implementation has be kept intact with the same error messages getting returned.

cmd/loginStatus.go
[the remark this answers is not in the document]
vibhansa-msft: So earlier isExpired() shall also be checking the same only right ? AzIdentity does not offer any method to check this part ?

common/oauthTokenManager.go
```diff
@@ -300,38 +326,122 @@ func (uotm *UserOAuthTokenManager) UserLogin(tenantID, activeDirectoryEndpoint s
...
-		return nil, fmt.Errorf("get cached token failed to ensure token fresh, please log in with azcopy's login command again, %v", err)
+		return nil, err
 	}
+	oAuthTokenInfo.AccessToken = token
 
-	// Update token cache, if token is updated.
-	if freshToken.AccessToken != tokenInfo.AccessToken || freshToken.RefreshToken != tokenInfo.RefreshToken {
-		tokenInfo.Token = *freshToken
-		if err := uotm.credCache.SaveToken(*tokenInfo); err != nil {
+	// Update the credential cache with the new token
```
vibhansa-msft, line 393: Token retreived from AzIdentity is getting stored in credCache below?

common/oauthTokenManager.go
```diff
@@ -189,6 +190,7 @@ func (uotm *UserOAuthTokenManager) MSILogin(identityInfo IdentityInfo, persist b
 	oAuthTokenInfo := &OAuthTokenInfo{
 		Identity:     true,
 		IdentityInfo: identityInfo,
+		Persist:      persist,
 	}
 
 	return uotm.validateAndPersistLogin(oAuthTokenInfo, persist)
```
vibhansa-msft, line 196: - 'persist' is part of oAuthTokenInfo already. 'validateAndPersistLogin' is taking both inputs which is redundant and can be removed.
- 'validateAndPersistLogin' is persisting in credCache. We discussed to always use AzIdentity cache first and if not available to use credCache. Do not see those changes in these flow.

vibhansa-msft: Only in device login we want to use AzIdentity cache and no other login like SPN/MSI/SecretaBased login etc.

common/oauthTokenManager.go
```diff
@@ -244,42 +248,64 @@ func (uotm *UserOAuthTokenManager) UserLogin(tenantID, activeDirectoryEndpoint s
...
+		}
+		dc, err = azidentity.NewDeviceCodeCredential(&azidentity.DeviceCodeCredentialOptions{
+			AuthenticationRecord: record,
+			TenantID:             tenantID,
+			TokenCachePersistenceOptions: &azidentity.TokenCachePersistenceOptions{
+				AllowUnencryptedStorage: true,
+				Name:                    TokenCache,
+			}})
+		// If record is empty
+		if record == (azidentity.AuthenticationRecord{}) {
```
vibhansa-msft, line 271: 'record' is created at line 2589, then we try to get a token. Post that we check whether record is empty or not. We shall check record is empty or not first as in case of empty record. NewDeviceCodeCredential is just taking it as input to fetch token. If we call that before the record is populate how is it working?
- fetch record
- create create wth record (which is empty in the first place)
- if record is empty, populate the record
- so in this flow DeviceCredential is alredy created 'dc' with empty reocrd. How does populating the record now make any difference to 'dc'.

common/oauthTokenManager.go
```diff
@@ -244,42 +248,64 @@ func (uotm *UserOAuthTokenManager) UserLogin(tenantID, activeDirectoryEndpoint s
...
+		TenantID: tenantID,
+		Scopes:   scopes,
+	})
 	if err != nil {
-		return fmt.Errorf("failed to login with tenantID %q, Azure directory endpoint %q, %v",
-			tenantID, activeDirectoryEndpoint, err)
+		return err
 	}
-
+	/*
```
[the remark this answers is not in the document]
pranavmalik-msft: Done in the latest commit.

common/oauthTokenManager.go
pranavmalik-msft: resolved in the latest commit .

common/oauthTokenManager.go
pranavmalik-msft: we have removed the redundant code in validateAndPersistLogin, and are now using the persist information in oAuthtokenInfo.
azidentity does not support Authentication Record in SPN/MSI/Secret based login , hence we have not used that flow here.

common/oauthTokenManager.go
pranavmalik-msft: resolved in the latest commit.

common/credCache_darwin.go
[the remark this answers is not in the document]
pranavmalik-msft: Yes, we have made test runs over a macOS-vm to verify the correctness and stability.
Test runs are attached over in the Design Doc for reference.

common/oauthTokenManager.go
```diff
@@ -163,23 +163,27 @@ func (uotm *UserOAuthTokenManager) validateAndPersistLogin(oAuthTokenInfo *OAuth
 }
 
 func (uotm *UserOAuthTokenManager) AzCliLogin(tenantID string) error {
+
+	// CLI creds will not be persisted. AzCLI would have already persistd that
 	oAuthTokenInfo := &OAuthTokenInfo{
```
vibhansa-msft, line 168: In case of CLI login we want to persist the token then why flag is false here.

common/oauthTokenManager.go
```diff
@@ -300,38 +326,111 @@ func (uotm *UserOAuthTokenManager) UserLogin(tenantID, activeDirectoryEndpoint s
...
+			tokenInfo.AccessToken = *freshToken
+			if err := uotm.credCache.SaveToken(*tokenInfo); err != nil {
+				return nil, err
+			}
+		}
+		return tokenInfo, nil
+	}
+
+	// Create oAuthTokenInfo from the existing AuthenticationRecord
+	oAuthTokenInfo := &OAuthTokenInfo{
```
vibhansa-msft, line 373: Don't we need to pass record as well in this case to getToken from identity library.

common/oauthTokenManager.go
```diff
@@ -300,38 +326,111 @@ func (uotm *UserOAuthTokenManager) UserLogin(tenantID, activeDirectoryEndpoint s
...
+// If the access token has expired, it is refreshed.
+// If the refresh token has expired, the method fails and returns the reason for failure.
+// The fresh token is persisted if either the access token or refresh token has changed.
 func (uotm *UserOAuthTokenManager) getCachedTokenInfo(ctx context.Context) (*OAuthTokenInfo, error) {
-	hasToken, err := uotm.credCache.HasCachedToken()
+
+	// retrieveRecord gets an AuthenticationRecord from the local disk
+	record, err := retrieveRecord()
 	if err != nil {
-		return nil, fmt.Errorf("no cached token found, please log in with azcopy's login command, %v", err)
```
vibhansa-msft, line 310: If retrieve record fails then we still need to try out credCache as that might be the legit source of token here.

common/oauthTokenManager.go
```diff
@@ -343,8 +442,23 @@ func (uotm *UserOAuthTokenManager) HasCachedToken() (bool, error) {
 	return uotm.credCache.HasCachedToken()
 }
 
-// RemoveCachedToken delete all the cached token.
+// RemoveCachedToken deletes the token and the AuthenticationRecord as part of the logout command execution.
+// Returns an error if encountered during file deletion or checking.
 func (uotm *UserOAuthTokenManager) RemoveCachedToken() error {
+	filePath := filepath.Join(AzcopyJobPlanFolder, defaultAuthFileName)
```
vibhansa-msft, line 448: We need to remove the record here and if that fails still we shall try to delete the credCache.

common/oauthTokenManager.go
pranavmalik-msft: the comment above it is explaining the behavior observed here in this case , in our original flow also persist flag for AzCliLogin was kept as false therefore to maintain consistency we have passed Persist as false here.

common/oauthTokenManager.go
pranavmalik-msft: Retrieve record's implementation is such that if  Authentication-record does not exist it will return an empty Authentication-record otherwise it fetches the existing record and tries to create the Authentication-record by un-marshaling[parsing the json record to create Authentication-record] this error is returned in the case where un-marshaling was at failure otherwise we are looking in the credcache basis of the empty record found.

common/oauthTokenManager.go
```diff
@@ -300,38 +326,111 @@ func (uotm *UserOAuthTokenManager) UserLogin(tenantID, activeDirectoryEndpoint s
...
-	tokenInfo, err := uotm.credCache.LoadToken()
+	tc, err := oAuthTokenInfo.GetTokenCredential()
 	if err != nil {
-		return nil, fmt.Errorf("get cached token failed, %v", err)
+		return nil, err
 	}
+	scopes := []string{StorageScope}
 
-	freshToken, err := tokenInfo.Refresh(ctx)
+	token, err := tc.GetToken(context.TODO(), policy.TokenRequestOptions{
```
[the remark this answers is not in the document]
pranavmalik-msft: no, Authentication-record is passed whenever we are creating credential, if you observe at line no 373 we are creating Oauthtokeninfo based on the  Authentication-record , and then passing it over the GetTokenCredential() method where in case of Device Code credential its passing in the auth-record for tokenCredential creation.

common/oauthTokenManager.go
pranavmalik-msft: RemoveCachedToken() function is invoked only in the case of Logout command execution and this is by default in our original code , the design of this function has been kept such that if we call the logout command :

- we will delete the Authentication record 
-  if we face some problems in deleting report that problem
- in case the Authentication record file does not exist we simply display it over the console
- Complete the original flow of deleting the token in the credcache.

common/oauthTokenManager.go
```diff
@@ -343,8 +442,23 @@ func (uotm *UserOAuthTokenManager) HasCachedToken() (bool, error) {
...
 func (uotm *UserOAuthTokenManager) RemoveCachedToken() error {
+	filePath := filepath.Join(AzcopyJobPlanFolder, defaultAuthFileName)
+
+	// Check if the file exists
+	if _, err := os.Stat(filePath); err == nil {
+		// File exists, so delete it
+		if err := os.Remove(filePath); err != nil {
+			return fmt.Errorf("error deleting file: %v", err)
+		}
+		fmt.Println("File deleted successfully.")
```
vibhansa-msft, line 456: These logs do not give any indication on what is happening or what has failed and on top of that these are going to user console. Imagine user getting an error message "file does not exist.". What is user suppose to imply from this message?

common/oauthTokenManager.go
```diff
@@ -152,7 +152,7 @@ func (uotm *UserOAuthTokenManager) validateAndPersistLogin(oAuthTokenInfo *OAuth
 	}
 	uotm.stashedInfo = oAuthTokenInfo
 
-	if persist && err == nil {
+	if oAuthTokenInfo.Persist && err == nil {
```
vibhansa-msft, line 155: There is no case where err is not nil here, then why this check?

pranavmalik-msft: That check exist of err is not nil exist just above when we are getting the token.

common/oauthTokenManager.go
pranavmalik-msft: Done in the latest commit.

common/oauthTokenManager.go
```diff
@@ -343,8 +440,23 @@ func (uotm *UserOAuthTokenManager) HasCachedToken() (bool, error) {
...
+	filePath := filepath.Join(AzcopyJobPlanFolder, defaultAuthFileName)
+
+	// Check if the file exists
+	if _, err := os.Stat(filePath); err == nil {
+		// File exists, so delete it
+		if err := os.Remove(filePath); err != nil {
+			return fmt.Errorf("error deleting AuthenticationRecord: %v", err)
+		}
+		fmt.Println("AuthenticationRecord deleted successfully.")
+	} else if os.IsNotExist(err) {
```
vibhansa-msft, line 455: both these else cases can be removed as we need to clear credcache after deleting record file.

pranavmalik-msft: Done in the latest commit.

common/zt_credCache_test.go
```diff
@@ -21,22 +21,18 @@
...
-		ExpiresIn:    "3599",
-		ExpiresOn:    "1528108397",
-		NotBefore:    "1528104497",
-		Resource:     "https://storage.azure.com",
-		Type:         "Bearer",
+	AccessToken: azcore.AccessToken{
```
adreed-msft, line 32: What is the nature of this token? Is it something hand-crafted, is it a well-known fake token? Let's document the where here, so that if security comes up and says "Wow, that's an OAuth token in your repo!", we can point and say why it's safe.

adreed-msft (changes requested):
This code still does not work. This is a build I produced with the current revision of this branch.

```
PS C:\Users\adreed\go\src\github.com\azure\azure-storage-azcopy> azcopy login
To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the code AV6XY26HE to authenticate.
INFO: Login succeeded.
PS C:\Users\adreed\go\src\github.com\azure\azure-storage-azcopy> azcopy copy "https://adeledevtest1.blob.core.windows.net/dest" "C:\Users\adreed\Downloads\oauthdltest" --recursive
INFO: Scanning...
INFO: Autologin not specified.
INFO: Authenticating to source using Unknown, Please authenticate using Microsoft Entra ID (https://aka.ms/AzCopy/AuthZ), use AzCopy login, or append a SAS token to your Azure URL.
INFO: Any empty folders will not be processed, because source and/or destination doesn't have full folder support

failed to perform copy command due to error: cannot start job due to error: cannot list files due to reason GET https://adeledevtest1.blob.core.windows.net/dest
RESPONSE 401: 401 Server failed to authenticate the request. Please refer to the information in the www-authenticate header.
ERROR CODE: NoAuthenticationInformation
﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>NoAuthenticationInformation</Code><Message>Server failed to authenticate the request. Please refer to the information in the www-authenticate header.
RequestId:1d348fad-201e-003c-353d-9b7ec5000000
Time:2024-04-30T20:34:26.4069956Z</Message></Error>
.

PS C:\Users\adreed\go\src\github.com\azure\azure-storage-azcopy> 
```

common/zt_credCache_test.go
pranavmalik-msft: This token is the same to the one found in our credcache_test.go file from earlier. Unfortunately, there's no historical context for this token since there are no comments or other documentation indicating its origin. Therefore, it's unclear whether this token was manually created or generated by some other means.

PR #2660, opened 2024-04-25
Updates to Test for Empty SAS Supportability 

siminsavani-msft:


gapra-msft (approved):

adreed-msft (approved):
Ran locally, appears to be working as expected. Thanks for the quick fix!

PR #2668, opened 2024-05-02
Preserve blob tags should not fail  the transfer if SAS does not have enough permissions

pranavmalik-msft:
### Description
This pull request addresses the need for clearer error messages during transfer failures, specifically when the preserve-blob-tags option is enabled. Previously, the error response incorrectly indicated a 403 permission issue, as evidenced below:
image (6)

 This improvement will enable Storage Explorer to accurately inform users of transfer failures related to tag preservation, such as "Failed to set tags due to permissions error."
The new flow would return a better error message as can be observed below:
image


cmd/copy.go
```diff
@@ -1652,6 +1652,9 @@ func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 	if err != nil {
 		if err == NothingToRemoveError || err == NothingScheduledError {
 			return err // don't wrap it with anything that uses the word "error"
+		} else if cca.S2sPreserveBlobTags {
```
siminsavani-msft, line 1655: Shouldn't we check if the preserve blob tags flag is enabled **and** the error message is the AuthorizationPermissionMismatch?

siminsavani-msft:
Personally, I would keep the service error along with the newly added message: "Failed to set blob tags due to permission error"

cmd/copy.go
pranavmalik-msft: We are doing exactly the same thing here in line no 1655

siminsavani-msft (approved):

adreed-msft (approved):

cmd/copy.go
```diff
@@ -1652,6 +1652,9 @@ func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 	if err != nil {
 		if err == NothingToRemoveError || err == NothingScheduledError {
 			return err // don't wrap it with anything that uses the word "error"
+		} else if cca.S2sPreserveBlobTags && strings.Contains(err.Error(), "AuthorizationPermissionMismatch") {
+			// If S2sPreserveBlobTags is true, return a more descriptive error message indicating permission issue.
+			return fmt.Errorf("Failed to set blob tags due to permission error: %s.\n", err)
```
adreed-msft, line 1657: Is this the only way we can get this back? Could this not occur later in a job, after enumeration?

e2etest/zt_newe2e_basic_functionality_test.go
```diff
@@ -160,4 +162,48 @@ func (s *BasicFunctionalitySuite) Scenario_Copy_EmptySASErrorCodes(svm *Scenario
...
+					},
+				}),
+			},
+			Flags: CopyFlags{
+				CopySyncCommonFlags: CopySyncCommonFlags{
+					Recursive:           pointerTo(true),
+					S2SPreserveBlobTags: pointerTo(true),
+				},
+			},
+			ShouldFail: false,
```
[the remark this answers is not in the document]
pranavmalik-msft: yes, transfer should fail as done in the latest commit.

cmd/copy.go
pranavmalik-msft: this is the only place we can  get this back as of now , even after job enumeration we are coming here.

gapra-msft (changes requested):
Please also make sure that this error message is returned when the source/destination is pointing to the blob level URL

gapra-msft:
@pranavmalik-msft I don't see the addition of a test that checks that this behavior works when pointing to the blob level URL, which was my previous comment. 

The existing test checks this behavior when pointed to a container level URL.

adreed-msft (changes requested):

PR #2679, opened 2024-05-10
Bump AzCopy Go version to 1.22.3

siminsavani-msft:


.github/workflows/build_m1.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: '1.21'
+          go-version: '1.22.3'
```
gapra-msft, line 24: Can you manually test this workflow? This won't run automatically

siminsavani-msft: Workflow found here: https://github.com/Azure/azure-storage-azcopy/actions/runs/9100202508

I manually tested the build from the workflow on the Mac Mini as well!

gapra-msft (approved):

adreed-msft (approved):

PR #2680, opened 2024-05-10
Fixed teardown and setup suite flow to actually register methods and run them when expected

gapra-msft:


e2etest/newe2e_suite_manager.go
```diff
@@ -66,13 +66,11 @@ func (sm *SuiteManager) RunSuites(t *testing.T) {
 
 			if setupIdx != -1 {
 				// todo: call setup with suite manager
-				t.Run("Setup", func(t *testing.T) {
-					defer func() {
-						NewFrameworkAsserter(t).AssertNow("Scenario setup panicked (recovered)", NoError{stackTrace: true}, recover())
-					}()
+				defer func() {
```
gapra-msft, line 69: Choosing not to run this within t.Run to keep balance with how teardown happens (we cant call t.Cleanup within t.Run so we have to use t.Cleanup alone)

Let me know if you feel strongly about keeping Setup in t.Run.

siminsavani-msft (approved):

e2etest/newe2e_suite_manager.go
adreed-msft: No particular qualms here.

adreed-msft (approved):

adreed-msft (approved):

PR #2698, opened 2024-05-30
Update M1 build to use OAuth

adreed-msft:
Updates the M1 build pipeline to use OAuth instead of SAS tokens


.github/workflows/build_m1.yml
```diff
@@ -26,5 +30,12 @@ jobs:
       - name: Build
         run: CGO_ENABLED=1 go build -o azcopy_darwin_arm64
 
+      - name: Azure Login
+        uses: Azure/login@v2.1.0
+        with:
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
```
gapra-msft, line 36: is this using the workload identity?

gapra-msft (approved):

.github/workflows/build_m1.yml
adreed-msft: This is using `azcopy-nextgen-msi`

pranavmalik-msft (commented):
can you give some description to the PR , and sync with main

PR #2705, opened 2024-06-06
Adding unknown CVE to scanner

vibhansa-msft:
CSI Driver scans also check for UNKNOWN type of CVE issues and this is what our trivy pipeline is missing.
This results into CSI team firing CVE issues post our release. This changeset will help identify those before we publish our release.


gapra-msft (approved):

adreed-msft (approved):

vibhansa-msft:
> > CSI Driver scans also check for UNKNOWN type of CVE issues and this is what our trivy pipeline is missing. This results into CSI team firing CVE issues post our release. This changeset will help identify those before we publish our release.
> 
> Shouldn't this result in trivy step to fail right now? Why is it still not catching the Go vuln?

I checked the item where CVE was reported recently. It reports issue with "stdlib". There is no direct go dependency on that. When we compile with CGO this might be coming in. I tried on both Ubn20 and 22 but on local scanning it's not reporting any issue. My guess here is it depends on what system we are compiling the binary (stdlib version coming in from there).

gapra-msft:
@vibhansa-msft Can this get updated so we can merge this soon?

vibhansa-msft:
I created a container image based on ubuntu-22.04 and did a trivy analysis on that but that shows exactly the same list for the container image with or without azcopy binary being packaged.

PR #2710, opened 2024-06-10
[Storage Explorer] _token_refresh_source type change

dphulkar-msft:
Issue: https://msazure.visualstudio.com/One/_workitems/edit/28320197

Error while unmashalling  OAuthTokenInfo._token_refresh_source as the type has been changed from string to uint8 and the field is received as string type from user.

Bug-fix: Implemented two interfaces **json.Marshaler** and **json.Unmarshaler** on **AutoLoginType** to marshal and unmarshal it as a string.

[PR](https://github.com/Azure/azure-storage-azcopy/pull/2619) reference where _token_refresh_source has been changes from string to uint8


common/oauthTOkenManager_test.go
```diff
@@ -0,0 +1,110 @@
...
+					"access_token": "dummy_access_token",
+					"refresh_token": "dummy_refresh_token",
+					"expires_in": 0,
+					"expires_on": 0,
+					"not_before": 0,
+					"resource": "dummy_resource",
+					"token_type": "dummy_token_type",
+					"_tenant": "dummy_tenant",
+					"_ad_endpoint": "dummy_ad_endpoint",
+					"_token_refresh_source": "255",
```
vibhansa-msft, line 59: Add one more case where this is int and parses through fine. Also, we can add negative case where this is not an integer and it fails to parse the token.

dphulkar-msft: handled

common/oauthTokenManager.go
```diff
@@ -771,13 +770,41 @@ func (credInfo *OAuthTokenInfo) GetTokenCredential() (azcore.TokenCredential, er
 	}
 }
 
+// MarshalJSON customizes the JSON encoding for AutoLoginTypeJSON
+func (alt AutoLoginTypeJSON) MarshalJSON() ([]byte, error) {
+	return json.Marshal(alt.AutoLoginType)
```
adreed-msft, line 775: Let's do the .string call here instead, given we'd prefer to support it as strings

common/oauthTokenManager.go
```diff
@@ -771,13 +770,41 @@ func (credInfo *OAuthTokenInfo) GetTokenCredential() (azcore.TokenCredential, er
...
+	var v interface{}
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+
+	switch value := v.(type) {
+	case float64:
+		alt.AutoLoginType = AutoLoginType(value)
+	case string:
+		uintValue, err := strconv.ParseUint(value, 10, 8)
```
adreed-msft, line 789: I think we may have miscommunicated a bit. This isn't the intended behavior-- `AutoLoginType` is an enum, if you check the type's `String` and `Parse` functions, this is what we're looking for, regarding a string value here.

common/environment.go
```diff
@@ -87,8 +88,14 @@ func (EnvironmentVariable) UserDir() EnvironmentVariable {
 
 var EAutoLoginType = AutoLoginType(0)
 
+// Wrapper type for AutoLoginType that handles custom JSON marshaling and unmarshaling
 type AutoLoginType uint8
 
+// Wrapper type for AutoLoginType that handles custom JSON marshaling and unmarshaling
+type AutoLoginTypeJSON struct {
```
adreed-msft, line 95: I think a wrapper type is unnecessary here, any defined type can have functions attached. I *believe* `json.Unmarshal` and `json.Marshal` always check even if the underlying type is a primitive(?), but lmk if I'm wrong.

adreed-msft (changes requested):
It seems we miscommunicated on the mode of failure we're trying to fix here, but overall, looks good.

common/environment.go
```diff
@@ -87,6 +88,7 @@ func (EnvironmentVariable) UserDir() EnvironmentVariable {
 
 var EAutoLoginType = AutoLoginType(0)
 
+// Wrapper type for AutoLoginType that handles custom JSON marshaling and unmarshaling
```
gapra-msft, line 91: nit: this comment is no longer relevant

common/oauthTokenManager.go
```diff
@@ -771,6 +770,29 @@ func (credInfo *OAuthTokenInfo) GetTokenCredential() (azcore.TokenCredential, er
 	}
 }
 
+// MarshalJSON customizes the JSON encoding for AutoLoginType
+func (alt AutoLoginType) MarshalJSON() ([]byte, error) {
```
gapra-msft, line 774: This should ideally be defined alongside the AutoLoginType definition

common/oauthTokenManager.go
dphulkar-msft: done

common/environment.go
dphulkar-msft: Updated

common/oauthTokenManager.go
dphulkar-msft: moved MarshalJSON and UnmarshalJSON to environment.go file

common/oauthTokenManager.go
dphulkar-msft: updated the code accordingly

common/oauthTokenManager.go
```diff
@@ -84,7 +84,7 @@ func newAzcopyHTTPClient() *http.Client {
 				Timeout:   10 * time.Second,
 				KeepAlive: 10 * time.Second,
 				DualStack: true,
-			}).Dial,                   /*Context*/
```
vibhansa-msft, line 87: This file does not have a code change so may be you can sync this up with main and keep the file as is.

common/environment.go
```diff
@@ -115,6 +117,29 @@ func (d *AutoLoginType) Parse(s string) error {
...
+func (d *AutoLoginType) UnmarshalJSON(data []byte) error {
+	var v interface{}
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+	if strValue, ok := v.(string); ok {
+		return d.Parse(strValue)
+	}
+	// Handle numeric values
+	if numValue, ok := v.(float64); ok {
```
vibhansa-msft, line 135: Here we are checking for string and float, this field is considered at float64 by the Json definition ?

common/oauthTokenManager.go
dphulkar-msft: done

common/environment.go
dphulkar-msft: removed

common/environment.go
gapra-msft, line 135: why parse the number as a float64 if only uint8s are allowed for AutoLoginType?

seanmcc-msft:
@dphulkar-msft, let's prioritize this PR, this is the last item we need to complete before the patch release.

common/oauthTokenManager_test.go
```diff
@@ -0,0 +1,202 @@
...
+		wantErr bool
+	}{
+		{
+			name: "This UT tests if AutoLoginType filled is parsed properly from string to uint8 data type",
+			uotm: &UserOAuthTokenManager{},
+			args: args{
+				ctx: context.Background(),
+			},
+			setup: func(t *testing.T) {
+				tokenInfoJSON := `{
```
gapra-msft, line 49: this might look a little cleaner if we had a string format definition and just specify the things we are changing in the test using fmt.Sprintf

Similar for OAuthTokenInfo, we can create a variable with the default variables and set the auto login type as needed for the test

dphulkar-msft: done

common/environment.go
dphulkar-msft: In Go, JSON numbers are decoded into float64 by default. Therefore, during JSON unmarshalling, we cannot directly check for uint8 type. Instead, we need to handle float64 and then ensure that the value fits within the range of uint8.

vibhansa-msft (approved):

gapra-msft (approved):

common/environment.go
```diff
@@ -115,6 +118,32 @@ func (d *AutoLoginType) Parse(s string) error {
 	return err
 }
 
+// MarshalJSON customizes the JSON encoding for AutoLoginType
+func (d AutoLoginType) MarshalJSON() ([]byte, error) {
+	return json.Marshal(d.String())
+}
+
+// UnmarshalJSON customizes the JSON decoding for AutoLoginType
+func (d *AutoLoginType) UnmarshalJSON(data []byte) error {
```
adreed-msft, line 127: Looking good!

adreed-msft (approved):
LGTM

PR #2732, opened 2024-06-21
TraverserTests

mawhite-msft:
Added tests with mocked responses for zc_traverser_blob


cmd/zt_traverser_blob_test.go
```diff
@@ -61,6 +65,989 @@ func TestIsSourceDirWithStub(t *testing.T) {
...
+
+	client, err := blobservice.NewClientWithSharedKeyCredential(rawURL, credential,
+		&blobservice.ClientOptions{
+			ClientOptions: azcore.ClientOptions{
+				Transport: srv, //passing in mock server
+			}})
+	a.Nil(err)
+
+	//the mock server responds with a status code error
+	srv.AppendResponse(mock_server.WithStatusCode(403), mock_server.WithHeader("x-ms-error-code", "BlobNotFound"), mock_server.WithBody([]byte(MockErrorBody("BlobNotFound"))))
```
gapra-msft, line 307: nit: BlobNotFound is usually a 404 in http world. If we want to use 403, we should take a look at common error codes for Blob and choose an appropriate one.

cmd/zt_traverser_blob_test.go
```diff
@@ -61,6 +65,989 @@ func TestIsSourceDirWithStub(t *testing.T) {
...
+
+	//test method and validate
+	seenFiles := make(map[string]StoredObject)
+
+	err = blobTraverser.Traverse(nil, func(storedObject StoredObject) error {
+		seenFiles[storedObject.relativePath] = storedObject
+		return nil
+	}, []ObjectFilter{})
+	a.Equal(map[string]StoredObject{}, seenFiles)
+	a.Equal("this blob uses customer provided encryption keys (CPK). At the moment, AzCopy does not support CPK-encrypted blobs. If you wish to make use of this blob, we recommend using one of the Azure Storage SDKs", err.Error())
```
adreed-msft, line 284: This might be out of scope, but for these more specific errors, perhaps we should just have a `zt_traverser_errors.go` file where we store all of these strings in one place, that way, if it gets used in multiple places, or changed in the future, it's a zero-brainpower operation to find all the places to change it in tests and such.

adreed-msft (approved):
Looking good, nice work!

PR #2747, opened 2024-07-10
Device Code Migration to AzIdentity Track 2

### LINKED ISSUE #2361: Auto-login not persisting token when AZCOPY_AUTO_LOGIN_TYPE=DEVICE
10.20.1

Linux

### What command did you run?
export AZCOPY_AUTO_LOGIN_TYPE=DEVICE
export AZCOPY_TENANT_ID=_tenant id_
... multiple azcopy commands

Have to enter code into web browser for _every_ azcopy command encountered a script.  Not a usable solution when there many commands.

Should be easy to reproduce.  It appears from the code that auto-login disables token persistence altogether.

Don't use auto-login, use `azcopy login` instead, but it's getting removed in 10.22.

Azure CLI persists tokens in ~/.azure folder; could azcopy do the same?

gapra-msft:
Resolves #2361 

Note: We need to test with personal accounts right now.


e2etest/zt_newe2e_oauth_device_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// no local->local
+	if srcObj.Location().IsLocal() && dstObj.Location().IsLocal() {
+		svm.InvalidateScenario()
+		return
+	}
+
+	sasOpts := GenericAccountSignatureValues{}
+
+	stdout, _ := RunAzCopy(
```
adreed-msft, line 46: Can we validate that stdout is noting that we're connecting via AAD?

adreed-msft (approved):

common/oauthTokenManager.go
```diff
@@ -682,59 +641,48 @@ func (credInfo *OAuthTokenInfo) GetWorkloadIdentityCredential() (azcore.TokenCre
...
 
-	if err := spt.RefreshWithContext(ctx); err != nil {
-		return nil, err
+	if record == (azidentity.AuthenticationRecord{}) {
+		// No stored record; call Authenticate to acquire one
+		record, err = tc.Authenticate(context.TODO(), &policy.TokenRequestOptions{Scopes: []string{StorageScope}})
+		if err != nil {
+			return nil, err
+		}
+		if credInfo.Tenant == DefaultTenantID {
```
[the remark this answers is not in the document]
dphulkar-msft: Could you please help me explain a scenario why this needs to be sync.Once? I can't see tenant as a shared resource in the login workflow. Please correct me if I am wrong.

e2etest/zt_newe2e_oauth_device_test.go
dphulkar-msft: I am not sure if we can get this information from stdout for validating if we are connected via AAD. Will check.

common/oauthTokenManager.go
```diff
@@ -42,6 +42,9 @@ import (
 	"github.com/Azure/go-autorest/autorest/date"
 
 	"github.com/Azure/go-autorest/autorest/adal"
```
vibhansa-msft, line 44: We still have reference to adal.Token in the code which shall be removed as adal package itself is deprecated now and as part of our Track-2 migration we have moved to msal. Refresh() method in this file still relies on adal.Token so we might need wider changes to remove adal dependency completely here.

common/oauthTokenManager.go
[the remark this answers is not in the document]
adreed-msft: On the off-chance we call GetDeviceCodeCredential multiple times for whatever reason, stdout might look like...

```
INFO: Logging in under the...
INFO: If you plan to use...
INFO: Logging in under the...
INFO: If you plan to use...
INFO: Logging in under the...
INFO: If you plan to use...
```

This is quite spammy. Granted, it would be a decent indicator a mistake has been made, though.

e2etest/zt_newe2e_oauth_device_test.go
adreed-msft: We most definitely do log this information. Please check `credentialUtil.go`, it is logged by `logAuthType`. If it is not, and we are authorizing something, that is a bug.

common/oauthTokenManager.go
gapra-msft: Unfortunately we need to keep this to maintain compatibility with storage explorer. We can take up an item to work with them to change the format of the token they send but that may take another release cycle or 2. 

For now to remove the dep I should be able to manually copy the adal.Token type into our codebase, so I will do that.

e2etest/zt_newe2e_oauth_device_test.go
gapra-msft: Done

e2etest/zt_newe2e_oauth_device_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"github.com/Azure/azure-storage-azcopy/v10/testSuite/cmd"
+	"path"
+	"strings"
+	"time"
+)
+
+var runDeviceCodeTest = flag.Bool("device-code", false, "Whether or not to run device code tests. These must be run manually due to interactive nature.")
+
+var manualEnv AzCopyEnvironment
```
adreed-msft, line 14: Unused; code smell as documented in next comment

e2etest/zt_newe2e_oauth_device_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+func init() {
+	if runDeviceCodeTest != nil && *runDeviceCodeTest {
+		suiteManager.RegisterSuite(&DeviceLoginManualSuite{})
+	}
+}
+
+func Scenario_CopySync(svm *ScenarioVariationManager, env *AzCopyEnvironment) {
```
adreed-msft, line 22: env also is unused and this has a code smell about it

If the goal were to set up a consistent environment for the suite, the best idea would be to add it as a field under the suite struct, but between a global function and a global variable, this feels like an odd implementation.

That said, neither of these wind up getting used, so, it should get removed.

e2etest/zt_newe2e_oauth_device_test.go
```diff
@@ -0,0 +1,118 @@
...
+		})
+
+	ValidateResource[ObjectResourceManager](svm, dstObj, ResourceDefinitionObject{
+		Body: body,
+	}, true)
+
+	ValidateAADAuth(svm, stdout, srcObj.Location(), dstObj.Location())
+}
+
+func ValidateAADAuth(a Asserter, stdout AzCopyStdout, src, dst common.Location) {
```
adreed-msft, line 77: These also have a code smell about them-- I wanted to say we have a way to validate that stdout contains certain lines already, but, this is probably fine.

e2etest/zt_newe2e_oauth_device_test.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	a.AssertNow("login should be successful", Equal{}, loggedIn, true)
+}
+
+func (s *DeviceLoginManualSuite) TeardownSuite(a Asserter) {
+	RunAzCopyLoginLogout(a, AzCopyVerbLogout)
+}
+
+func (s *DeviceLoginManualSuite) Scenario_CopySync(svm *ScenarioVariationManager) {
+	Scenario_CopySync(svm, nil)
```
adreed-msft, line 117: ditto, this is odd.

e2etest/newe2e_task_runazcopy_login_logout.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	return
+}
+
+func (a *AzCopyInteractiveStdout) String() string {
+	return strings.Join(a.RawStdout(), "\n")
+}
+
+func RunAzCopyLoginLogout(a Asserter, verb AzCopyVerb) AzCopyStdout {
```
adreed-msft, line 45: This is also interesting to me-- Is there a reason to prefer a separate function over integrating with RunAzCopy?

e2etest/newe2e_task_runazcopy.go
```diff
@@ -337,8 +342,10 @@ func RunAzCopy(a ScenarioAsserter, commandSpec AzCopyCommand) (AzCopyStdout, *Az
 		0, command.ProcessState.ExitCode())
 
 	a.Cleanup(func(a ScenarioAsserter) {
-		UploadLogs(a, out, stderr, DerefOrZero(commandSpec.Environment.LogLocation))
-		_ = os.RemoveAll(DerefOrZero(commandSpec.Environment.LogLocation))
+		if !commandSpec.Environment.ManualLogin {
```
adreed-msft, line 345: This also interests me-- I suspect this is to prevent capturing the login token, but, could we not filter that out?

adreed-msft (approved):
Tests have some funk about them, but otherwise lgtm. If we don't do anything about it now, I'll go in and do some housekeeping on the testing framework when I get time in December.

e2etest/zt_newe2e_oauth_device_test.go
gapra-msft: this is accidental leftover from some of my debugging. will remove.

e2etest/zt_newe2e_oauth_device_test.go
gapra-msft: this is accidental leftover from some of my debugging. will remove.

e2etest/zt_newe2e_oauth_device_test.go
gapra-msft: I did this so that we can eventually have more suites for the other OAuth methods, then we can just reference the one test and call it like this. But this might be overthinking/overplanning things

e2etest/newe2e_task_runazcopy_login_logout.go
gapra-msft: I did this mostly because login needs some interactive stdout, whereas RunAzCopy is just reading the stdout. There is probably a clean way to integrate it with the other method. this can be cleaned this up after the PR is merged

e2etest/zt_newe2e_oauth_device_test.go
gapra-msft: Forgot about that, will fix to use that instead.

e2etest/newe2e_task_runazcopy.go
gapra-msft: Currently the new e2e test framework is designed such that each of the scenarios creates a new temp log location. When you log in it creates a file in the log location and we want that log in to be the same across all e2e test framework logs, so we avoid removing the log folder. There is probably  a cleaner way to do this that we can look into during a cleanup PR.

e2etest/zt_newe2e_oauth_device_test.go
gapra-msft: This can easily be refactored during the cleanup PR, so I will leave this for now.

vibhansa-msft (approved):

PR #2762, opened 2024-07-26
Add prefix to the xattr

tsvietOK:


tsvietOK:
Seems like the test (from main branch, not from mine) on Linux is reporting not supported result:
image

tsvietOK:
Tested my change:
image

tsvietOK:
Maybe also there is needed some back compatibility mode (if user used old version of azcopy on Mac OS and file has xattr without "user" prefix) for Mac OS where xattr works fine (according to test execution).
image

adreed-msft (changes requested):
This is good, but it comes within a catch. This changes the metadata name, and somebody could have come to rely upon that. Two things need to happen to allow this through

1) When reading, let's check for both.
2) When writing, let's check that the "oldschool" version doesn't exist; if it does, delete it and replace it with the new attribute.

tsvietOK:
Seems like the requested changes are way above my knowledge of the Go. So, feel free to create a new PR or push to this one. By the way, wouldn't it be better to use this module https://github.com/pkg/xattr which covers support/error handling for some specific OSs like FreeBSD, NetBSD and MacOS?

adreed-msft:
> Seems like the requested changes are way above my knowledge of the Go. So, feel free to create a new PR or push to this one. By the way, wouldn't it be better to use this module https://github.com/pkg/xattr which covers support/error handling for some specific OSs like FreeBSD, NetBSD and MacOS?

Sounds like I'll be picking this up if I get some spare time.

PR #2763, opened 2024-07-29
Fixed jobs list broken command

dphulkar-msft:


ste/JobPartPlan.go
```diff
@@ -142,8 +143,12 @@ func (jpph *JobPartPlanHeader) Transfer(transferIndex uint32) *JobPartPlanTransf
 
 // CommandString returns the command string given by user when job was created
 func (jpph *JobPartPlanHeader) CommandString() string {
-	data := unsafe.Pointer(uintptr(unsafe.Pointer(jpph)) + unsafe.Sizeof(*jpph)) // Address of Job Part Plan + Command String Length
-	return unsafe.String((*byte)(data), int(jpph.CommandStringLength))
+	// Calculate the start address of the command string
```
adreed-msft, line 146: Sorry if I'm missing something obvious, how is this functionally different?

vibhansa-msft (approved):

ste/JobPartPlan.go
dphulkar-msft: **Existing code:** Directly converts the pointer to a string, which could lead to issues if the underlying memory is not correctly managed or if there are alignment issues. unsafe.String expects that the memory pointed to is valid and correctly aligned.
**New Code:** Converts the pointer to a byte slice first, which is more robust. unsafe.Slice creates a slice that provides better handling and validation of the byte array, and converting this slice to a string is less prone to errors.

Additionally, New Code is written in a more understandable and readable manner, making it easier to follow and maintain.

adreed-msft (approved):

seanmcc-msft:
@dphulkar-msft please add a testcase

vibhansa-msft (approved):

PR #2778, opened 2024-08-09
Allow AzCopy to receive OAuth tokens via GRPC for internal integrations

adreed-msft:
adds functionality to receive OAuth tokens over a HTTP/1.1 or HTTP/2 grpc message.

All functionality is behind a compile tag, `grpc`.

The token is shot in at about once per minute in the currently used integration. Requests trying to get an OAuth token pull a read lock on an RWMutex, and if the token is expired when the read is pulled, they wait up to 3x the validity of the last OAuth token for a fresh one. If no OAuth token is ever received within that timespan, it gives up, and signals to other requests to stop trying.

Injected tokens pull a write lock on the same RWMutex, and signals the `sync.Cond` to inform the requests waiting on a token a fresh one may be available.


common/oauthTokenManager.go
```diff
@@ -785,6 +803,124 @@ func jsonToTokenInfo(b []byte) (*OAuthTokenInfo, error) {
 
 // ====================================================================================
 
+var globalGRPCOAuthTokenLock = &sync.RWMutex{}
+
+// Pass in the RLocker to the sync.Cond so we're only requesting read locks, not write locks.
+var globalRPCOAuthTokenState = &GRPCOAuthToken{Mutex: sync.NewCond(globalGRPCOAuthTokenLock.RLocker())}
+
+func init() {
```
vibhansa-msft, line 811: Do we need this in init() or we can move it to a function which can be called on general init of azcopy.

release-pipeline.yml
```diff
@@ -110,6 +110,10 @@ stages:
               GOARCH=amd64 GOOS=linux go build -tags "netgo,se_integration" -o "$(root_dir)/azcopy_linux_se_amd64"
             displayName: 'Generate Linux AMD64 SE Integration'
 
+          - script: |
```
vibhansa-msft, line 113: GRPC feature can be triggered by CLI, do we need to build a seperate binary for that.

dphulkar-msft, line 113: Could you please run the pipeline on this branch once to verify if the changes are working as expected?

release-pipeline.yml
```diff
@@ -860,6 +962,13 @@ stages:
               archiveFile: '$(archives)/azcopy_windows_386_$(azcopy_version).zip'
             continueOnError: true
 
+          - task: ArchiveFiles@2
+            displayName: 'Archive Windows 32 bit Build'
```
dphulkar-msft, line 966: Shouldnt it be "Archive Windows 32 bit GRPC Build"

release-pipeline.yml
```diff
@@ -860,6 +962,13 @@ stages:
...
+            inputs:
+              rootFolderOrFile: '$(windows_386)'
```
dphulkar-msft, line 968: This will override the existing zip file created in previous step. You will have to use $(windows_grpc_386)

release-pipeline.yml
```diff
@@ -860,6 +962,13 @@ stages:
...
+              archiveFile: '$(archives)/azcopy_windows_386_$(azcopy_version).zip'
```
dphulkar-msft, line 969: Rename the zip file to azcopy_windows_grpc_386_$(azcopy_version).zip' as azcopy_windows_386_$(azcopy_version).zip already exists

common/oauthTokenManager.go
adreed-msft: Is there somewhere better you'd suggest it be?

release-pipeline.yml
adreed-msft: It's more intended to be a separate build so a customer has to go out of their way, fully understanding we won't support it, in order to use it. The other point is so that we don't introduce new dependencies in the normal build. If they choose to use it, they should expect things to break and change.

By not including it in the normal build, we release nothing anybody could accidentally come to depend on without intentionally expecting a footgun.

release-pipeline.yml
adreed-msft: Nice catch! Thank you.

release-pipeline.yml
adreed-msft: Done. https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=23352&view=results

bhansalivikas (approved):

grpcctl/events.go
```diff
@@ -0,0 +1,26 @@
...
+
+// 0 = none, -1 = infinite
+var validSubscriptionCounts = map[string]int64{
+	fetchSubscriptionName(OAuthTokenUpdate{}): 1,
+}
+
+// Inputs
+
+// EventSubscriber = func(OAuthTokenUpdate) EmptyResponse
+type OAuthTokenUpdate struct {
```
gapra-msft, line 17: Why not just reference the one in common?

ste/metadata-hack.go
```diff
@@ -0,0 +1,24 @@
+package ste
+
+import (
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"regexp"
+)
+
+var ormetadataregex = regexp.MustCompile("((([A-F]|[a-f]|[0-9])+-?)+_?){2}")
```
gapra-msft, line 8: This isn't necessary anymore right?

gapra-msft: Also all the related changes in ste

gapra-msft (approved):

common/oauthTokenManager.go
```diff
@@ -792,6 +811,137 @@ func jsonToTokenInfo(b []byte) (*OAuthTokenInfo, error) {
...
+				AzcopyCurrentJobLogger.Log(LogInfo, "Released.")
+			}
+
+			// send wait unblock
+			waitch <- true
+		}()
+
+		// Time out eventually, so that AzCopy can exit.
+		select {
+		case _, _ = <-GrpcServerFailed:
```
wonwuakpa-msft, line 903: there's a linting error with using _

wonwuakpa-msft (approved):

common/grpc_setup.go
```diff
@@ -0,0 +1,72 @@
+//go:build grpc
```
dphulkar-msft, line 1: nsft license header missing

common/grpc_shim.go
```diff
@@ -0,0 +1,28 @@
+package common
```
dphulkar-msft, line 1: msft license header missing

grpcctl/events.go
```diff
@@ -0,0 +1,26 @@
+package grpcctl
```
dphulkar-msft, line 1: same comment as above

release-pipeline.yml
```diff
@@ -120,6 +120,10 @@ stages:
               GOARCH=amd64 GOOS=linux go build -tags "netgo,se_integration" -o "$(root_dir)/azcopy_linux_se_amd64"
             displayName: 'Generate Linux AMD64 SE Integration'
 
+          - script: |
```
dphulkar-msft: This file has been removed from the main branch as we are now using build-1es-pipeline.yaml.
Please make these changes in the build-1es-pipeline.yaml file instead.

PR #2781, opened 2024-08-16
Correcting panic & formatting issues with sync

adreed-msft:
Fixes up some issues regarding --dry-run with sync.

When some paths have a % in them, but don't parse correctly from local, AzCopy will panic. In addition, paths are displayed incorrectly, with an extra path separator between the root path and the child path.


vibhansa-msft (approved):

dphulkar-msft (approved):

vibhansa-msft (approved):

e2etest/newe2e_resource_manager_getter.go
```diff
@@ -24,8 +24,13 @@ func GetRootResource(a Asserter, location common.Location, varOpts ...GetResourc
 
 		return NewLocalContainer(a)
 	case common.ELocation.Blob(), common.ELocation.BlobFS(), common.ELocation.File():
-		// acct handles the dryrun case for us
-		acct := GetAccount(a, DerefOrDefault(opts.PreferredAccount, PrimaryStandardAcct))
+		// do we have a hns acct attached, if so, and we're requesting blobfs, let's use it
```
gapra-msft, line 27: something about this code seems like the problem with the tests. currently we are using a hns account when we shouldnt be. there are some HNS account doesnt support this yet type errors.

PR #2782, opened 2024-08-19
Fixed golangci-lint errors

dphulkar-msft:
image


ste/sender-appendBlob.go
```diff
@@ -228,7 +229,7 @@ func (s *appendBlobSenderBase) GetMD5(offset, count int64) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	if response.ContentMD5 != nil && len(response.ContentMD5) > 0 {
+	if len(response.ContentMD5) > 0 {
```
vibhansa-msft, line 232: If its nil does len() method takes care of it, othewise this will result in panic

dphulkar-msft: yes it takes care of it as suggested by gosimple in the screenshot added at the top.

adreed-msft (approved):

ste/sender-appendBlob.go
adreed-msft: Nil maps and slices return 0 when `len()`'d over.

https://go.dev/play/p/lZ55EjV-xXQ

vibhansa-msft (approved):

PR #2785, opened 2024-08-21
AzCopy sync with --hash-meta-dir causes original files to get hidden

dphulkar-msft:


dphulkar-msft:
refer this PR: https://github.com/Azure/azure-storage-azcopy/pull/2786

Issue #2758, opened 2024-07-24
AzCopy sync with hash-meta-dir hides original files


tsvietOK:
### Which version of the AzCopy was used? 
10.25.1
### Which platform are you using? (ex: Windows, Mac, Linux)
Windows
### What command did you run?
azcopy sync ".\testData\\" "https://\<storage>.blob.core.windows.net/\<container>/\<folderName>/?sp=\<sig>" --delete-destination true --exclude-path "backup;import;system/caches" --compare-hash "MD5" --put-md5 --cap-mbps 200 --local-hash-storage-mode HiddenFiles --hash-meta-dir "E:\hash-meta"
### What problem was encountered?
Command succeeded, but it created a new folder E:\hash-meta with metadata. After that, I noticed that all the files in the original folder were hidden.
### How can we reproduce the problem in the simplest way?
1. Create folder `testData` with test files
2. Run powershell in the testData folder
3. Run command (change hash meta dir to your place) `azcopy sync ".\testData\" "https://<storage>.blob.core.windows.net/<container>/<folderName>/?sp=<sig>" --delete-destination true --exclude-path "backup;import;system/caches" --compare-hash "MD5" --put-md5 --cap-mbps 200 --local-hash-storage-mode HiddenFiles --hash-meta-dir "E:\hash-meta"`
4. Check that the data files (in folder testData) are hidden and meta data (folder hash-meta) files are not hidden.
### Have you found a mitigation/solution?
No

tsvietOK:
![image](https://github.com/user-attachments/assets/74a6772f-8145-4a1f-b6ba-050d1fac1314)
![image](https://github.com/user-attachments/assets/44117d6d-8f57-4793-a32b-7a605760721c)

tsvietOK:
I guess here should be `metaFile`:
https://github.com/Azure/azure-storage-azcopy/blob/main/common/hash_data_adapter_hidden_files.go#L88

ktrace2:
azcopy kept crashing on me when using the hash-meta-dir option. I assumed it was related to this bug.

ashruti-msft:
> azcopy kept crashing on me when using the hash-meta-dir option. I assumed it was related to this bug.

@ktrace2 which version of azcopy are you using when it crashes? Also can you share your debug logs for that particular run?

adreed-msft:
This looks very much like a bug, yes... I'll add this to our backlog, hopefully we can get to it this release.

[...]

PR #2786, opened 2024-08-21
AzCopy sync with --hash-meta-dir causes original files to get hidden

dphulkar-msft:
**Command:**
.\azure-storage-azcopy.exe sync "C:\testdata" "https://dpazcopy2.blob.core.windows.net/samplecont?<sas-token>" --delete-destination true --exclude-path "backup;import;system/caches" --compare-hash "MD5" --put-md5 --cap-mbps 200 --local-hash-storage-mode HiddenFiles --hash-meta-dir "C:\hash-meta"

**Issue:** https://github.com/Azure/azure-storage-azcopy/issues/2758

**BugFix:**
       According to the command the metadata file should be hidden and not the original file.
SS2

image
image


adreed-msft (approved):

adreed-msft:
We should add an automated test to this, but the change itself looks good.

seanmcc-msft:
> We should add an automated test to this, but the change itself looks good.

Yes, please add a test case.

adreed-msft (changes requested):
rc: test case needed

dphulkar-msft:
> rc: test case needed

dphulkar-msft:
@adreed-msft  @vibhansa-msft Added the testcase. Please review

e2etest/zt_basic_copy_sync_remove_test.go
```diff
@@ -860,6 +862,9 @@ func TestBasic_HashBasedSync_HashDir(t *testing.T) {
 				hashFile := filepath.Join(hashStorageDir, ".asdf.txt"+common.AzCopyHashDataStream)
 				_, err = os.Stat(hashFile)
 				a.AssertNoErr(err)
+
+				isHidden := osScenarioHelper{}.IsFileHidden(a, hashFile)
```
adreed-msft, line 866: Quick check: make sure the data file is also not hidden. Otherwise, lgtm.

vibhansa-msft: Yes we shall check both ways, meta shall be hidden and data shall not be hidden (which is the exact issue we are trying to solve here)

common/hash_data_adapter_hidden_files.go
```diff
@@ -66,9 +67,18 @@ func (a *HiddenFileDataAdapter) SetHashData(relativePath string, data *SyncHashD
...
+	_, err := os.Stat(metaFile)
+	if os.IsNotExist(err) {
+		f, err = os.OpenFile(metaFile, os.O_CREATE|os.O_TRUNC|os.O_RDWR, 0644)
+		if err != nil {
+			return fmt.Errorf("failed to create hash meta file: %w", err)
+		}
+	} else {
+		f, err = os.OpenFile(metaFile, os.O_TRUNC|os.O_RDWR, 0644)
+		if err != nil {
+			return fmt.Errorf("failed to open hash meta file: %w", err)
```
vibhansa-msft, line 80: error log going to console, rather shall go to log file.

common/hash_data_adapter_hidden_files.go
```diff
@@ -66,9 +67,18 @@ func (a *HiddenFileDataAdapter) SetHashData(relativePath string, data *SyncHashD
 
 	metaFile := a.getHashPath(relativePath)
 
-	f, err := os.OpenFile(metaFile, os.O_CREATE|os.O_TRUNC|os.O_RDWR, 0644)
-	if err != nil {
-		return fmt.Errorf("failed to open hash meta file: %w", err)
+	var f *os.File
+	_, err := os.Stat(metaFile)
+	if os.IsNotExist(err) {
```
vibhansa-msft, line 72: Both if and else are doing almost the same thing, we shall rather unify the case. Only the log differs it seems here.

common/hash_data_adapter_hidden_files.go
```diff
@@ -83,9 +93,8 @@ func (a *HiddenFileDataAdapter) SetHashData(relativePath string, data *SyncHashD
 
 	// Push types around to check for OS-specific hide file method
 	if adapter, canHide := any(a).(interface{ HideFile(string) error }); canHide {
-		dataFile := a.getDataPath(relativePath)
-
-		err := adapter.HideFile(dataFile)
+		metaFile := a.getHashPath(relativePath)
```
vibhansa-msft, line 96: We shall put some comments here on why we are changing this code and why meta shall be hidden

common/hash_data_adapter_hidden_files.go
dphulkar-msft: There is one use case where it fails if we use a single OpenFile() API with os.O_CREATE|os.O_TRUNC|os.O_RDWR. 
Once the metafile is hidden and if we try to run the same azcopy command it fails to open the meta file with these flags being used combined os.O_CREATE|os.O_TRUNC|os.O_RDWR for which I have seperated the 2 calls.

common/hash_data_adapter_hidden_files.go
dphulkar-msft: added

e2etest/zt_basic_copy_sync_remove_test.go
dphulkar-msft: added

vibhansa-msft (approved):

common/hash_data_adapter_hidden_files.go
dphulkar-msft: Added the comment in the code as well.

PR #2790, opened 2024-08-26
Add Scenario for BlobToPipe download and ORP copy operation

dphulkar-msft:
TC 1: Added test Scenario for older version of AzCopy appends version check error message to piped download
TC 2: Added test scenario for "AzCopy cannot copy blob from a storage account with Object Replication policy"


azurePipelineTemplates/run_scenarios.yml
```diff
@@ -0,0 +1,109 @@
...
+
+        # Test copying from Object Replication Policy source container to destination container
+        # Assign an object replication policy to a storage account. 
+        # Perform azcopy copy operation to copy the blob to another container.
+
+        echo "Scenario 2: Test copying from Object Replication Policy source container to destination container"
+
+        echo "Create test file myfile.txt and copy the file to the source container srcorpcontainer under storage account azcopyorpsrcaccount having the ORP policy"
+        echo "Hello, Azure!" > myfile.txt
+        & "./$build_name" "cp" "myfile.txt" "https://azcopyorpsrcaccount.blob.core.windows.net/srcorpcontainer"
```
souravgupta-msft, line 93: The account name and container should not be exposed and be taken as pipeline variables

dphulkar-msft: done

azurePipelineTemplates/run_scenarios.yml
```diff
@@ -0,0 +1,117 @@
...
+      inlineScript: |
+        # Print "Building executable"
+        Write-Output "Building executable"
+        
+        # Set platform-specific environment variables and tags
+        $tags = ""
+        $suffix = ""
+        $build_name = ""
+        $display_name = ""
+        if ($IsWindows) {
```
vibhansa-msft, line 32: Based on the env you are setting env variables, not sure if this if-else is the best way to do this. We shall explore if we can do something better here.

azure-pipelines.yml
```diff
@@ -409,4 +408,59 @@ jobs:
           python ./tool_distributed_mutex.py unlock "$(MUTEX_URL)"
         name: 'Release_the_distributed_mutex'
         # this runs even if the job was canceled (only if the mutex was acquired by this job)
-        condition: and(always(), eq(variables['isMutexSet'], 'true'))
\ No newline at end of file
+        condition: and(always(), eq(variables['isMutexSet'], 'true'))
+
+  - job: Test_Scenarios
```
vibhansa-msft, line 413: This can be renamed to Additional E2E Tests

azure-pipelines.yml
```diff
@@ -409,4 +408,59 @@ jobs:
...
+          imageName: 'macos-latest'
+          type: 'macos'
+          build_name: 'azcopy_darwin_amd64'
+          container_name: 'testcontainer3'
+        
+    pool:
+      vmImage: $(imageName)
+
+    steps:
+      - task: PowerShell@2
```
vibhansa-msft, line 437: Does this step work on linux and mac?

azure-pipelines.yml
```diff
@@ -409,4 +408,59 @@ jobs:
...
+          container_name: 'testcontainer1'
+        Windows:
+          imageName: 'windows-latest'
+          type: 'windows'
+          build_name: 'azcopy_windows_amd64.exe'
+          container_name: 'testcontainer2'
+        MacOS:
+          imageName: 'macos-latest'
+          type: 'macos'
+          build_name: 'azcopy_darwin_amd64'
```
vibhansa-msft, line 430: build_name is not used in this pipeline. template files refines it based on the underlying environment,

azure-pipelines.yml
dphulkar-msft: yes it works

azure-pipelines.yml
dphulkar-msft: will update it

azurePipelineTemplates/run_scenarios.yml
```diff
@@ -0,0 +1,117 @@
...
+
+        ######################################################## Scenario 2 ########################################################
+
+        # Test copying from Object Replication Policy source container to destination container
+        # Assign an object replication policy to a storage account. 
+        # Perform azcopy copy operation to copy the blob to another container.
+
+        echo "Scenario 2: Test copying from Object Replication Policy source container to destination container"
+
+        echo "Create test file myfile.txt and copy the file to the source container srcorpcontainer under storage account azcopyorpsrcaccount having the ORP policy"
```
souravgupta-msft, line 99: Rename `srcorpcontainer` and `azcopyorpsrcaccount`

azurePipelineTemplates/run_scenarios.yml
```diff
@@ -0,0 +1,117 @@
...
+        if ($LASTEXITCODE -eq 0) {
+            Write-Output "Successfully executed the scenario."
+        } else {
+            Write-Error "AzCopy command failed with exit code $LASTEXITCODE."
+        }
+
+        # Perform cleanup
+        echo "Perform cleanup"
+        echo "Remove the test file from the destination container"
+        & "./$build_name" "rm" "https://azcopyorpdestaccount.blob.core.windows.net/testcontainer/myfile.txt"
```
souravgupta-msft, line 115:
```suggestion
        & "./$build_name" "rm" "https://${{ parameters.orp_destination_account_name }}.blob.core.windows.net/${{parameters.destination_container_name}}/myfile.txt"
```

azurePipelineTemplates/run_scenarios.yml
```diff
@@ -0,0 +1,117 @@
...
+
+        if ($LASTEXITCODE -eq 0) {
+            Write-Output "Successfully executed the scenario."
+        } else {
+            Write-Error "AzCopy command failed with exit code $LASTEXITCODE."
+        }
+
+        # Perform cleanup
+        echo "Perform cleanup"
+        echo "Remove the test file from the destination container"
```
souravgupta-msft, line 114: Should we also do cleanup for source ORP account?

dphulkar-msft: No we have to create this src ORP account manually. Just cleaning up the destination container would work.

souravgupta-msft: I meant deleting the test file from the source container

azurePipelineTemplates/run_scenarios.yml
dphulkar-msft: ok will change it

azure-pipelines.yml
dphulkar-msft: will fix it

azurePipelineTemplates/run_scenarios.yml
dphulkar-msft: will update

souravgupta-msft (approved):

vibhansa-msft (approved):

azurePipelineTemplates/run_scenarios.yml
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        # Perform cleanup
+        echo "Perform cleanup"
+        echo "Remove the test file from the source and destination container"
+        & "./$build_name" "rm" "https://${{ parameters.orp_destination_account_name }}.blob.core.windows.net/${{parameters.destination_container_name}}/myfile.txt"
+        & "./$build_name" "rm" "https://${{ parameters.orp_source_account_name }}.blob.core.windows.net/${{ parameters.orp_source_container_name }}/myfile.txt"
+      displayName: "Run Scenarios"
+
+  # - task: AzureCLI@2
```
adreed-msft, line 108: This commented out block-- Do you think we'll need it in the future, or should it be removed?

azurePipelineTemplates/run_scenarios.yml
```diff
@@ -0,0 +1,189 @@
...
+  - name: goarch
+    type: string
+  - name: cgo_enabled
+    type: string
+  - name: build_name
+    type: string
+  - name: azcopy_msi_app_id
+    type: string
+
+steps:
```
adreed-msft, line 25: Is there some reason why this can't be done within the E2E framework? If so, this'd be a really valuable bit of feedback on it.

adreed-msft (changes requested):

azurePipelineTemplates/run_scenarios.yml
adreed-msft: More specifically, understanding what makes using that difficult is the aim. I have marked this PR as requested changes, in case it is feasible. If it is not, I'll unblock.

dphulkar-msft: There are two test cases added as part of this Additional E2E 
1. Test end-to-end functionality for copying data between ORP source and destination accounts, noting that ORP accounts are not currently supported in the E2E framework.

2. Verify that older versions of AzCopy do not append version messages to piped downloads by comparing output files from older and newer versions to ensure consistency fow which we are downloading the older version and executing the command.

azurePipelineTemplates/run_scenarios.yml
dphulkar-msft: removed

azurePipelineTemplates/run_scenarios.yml
adreed-msft: These are both absolutely possible within the E2E test framework. Especially number 2 would not require any additional static accounts. I'll let this go since the work is done, but please prefer to implement things in the E2E framework in the future.

Just because it doesn't support something now, doesn't mean it's outside of your range to try to implement, nor does it mean we should work around it. We own 100% of that code. If something doesn't work yet in it, we can implement it.

adreed-msft (approved):

PR #2793, opened 2024-08-27
Use a different message when writing tags fails due to permissions issues, add test

adreed-msft:
If the failure point includes data about tags, we now treat that as a failure writing tags, and log about needing tags permissions (per stgexp's request).

A single test case covers Upload & S2S cases for writing tags, as well as different blob types, hitting all code paths.

### DIFF
--- a/ste/sender-blobFolders.go
+++ b/ste/sender-blobFolders.go
@@ -199,17 +199,33 @@ func (b *blobFolderSender) EnsureFolderExists() error {
 	}
 
 	err = t.CreateFolder(b.DirUrlToString(), func() error {
+		blobTags := b.blobTagsToApply
+		setTags := separateSetTagsRequired(blobTags)
+		if setTags || len(blobTags) == 0 {
+			blobTags = nil
+		}
+
 		// It doesn't make sense to use a special access tier for a blob folder, the blob will be 0 bytes.
 		_, err := b.destinationClient.Upload(b.jptm.Context(), streaming.NopCloser(bytes.NewReader(nil)),
 			&blockblob.UploadOptions{
 				HTTPHeaders:  &b.headersToApply,
 				Metadata:     b.metadataToApply,
-				Tags:         b.blobTagsToApply,
+				Tags:         blobTags,
 				CPKInfo:      b.jptm.CpkInfo(),
 				CPKScopeInfo: b.jptm.CpkScopeInfo(),
 			})
+		if err != nil {
+			b.jptm.FailActiveSend(common.Iff(len(blobTags) > 0, "Upload symlink (with tags)", "Upload symlink"), err)
+		}
 
-		return err
+		if setTags {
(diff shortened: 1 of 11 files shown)


tanyasethi-msft (approved):
Looks good to me, the pipelines seem to fail because of the MaxStorageAccountsCountPerSubscriptionExceeded, waiting for the cleanup, along with a string linter error which might need attention.

seanmcc-msft:
Related https://github.com/Azure/azure-storage-azcopy/issues/2798

vibhansa-msft (approved):

ste/sender-blobSymlinks.go
```diff
@@ -68,16 +68,33 @@ func (s *blobSymlinkSender) SendSymlink(linkData string) error {
...
 			HTTPHeaders:  &s.headersToApply,
 			Metadata:     s.metadataToApply,
 			Tier:         s.destBlobTier,
-			Tags:         s.blobTagsToApply,
+			Tags:         blobTags,
 			CPKInfo:      s.jptm.CpkInfo(),
 			CPKScopeInfo: s.jptm.CpkScopeInfo(),
 		})
-	return err
+	if err != nil {
```
gapra-msft, line 86: was this a bug - us not doing a separate set tags in certain situations for symlinks? Should we add an explicit symlink test?

ste/sender-blobSymlinks.go
```diff
@@ -68,16 +68,33 @@ func (s *blobSymlinkSender) SendSymlink(linkData string) error {
...
+		s.jptm.FailActiveSend(common.Iff(len(blobTags) > 0, "Upload symlink (with tags)", "Upload symlink"), err)
```
gapra-msft, line 87: I think we need to also make this change with sender-blobFolders

ste/sender-blobSymlinks.go
adreed-msft: Yes, it appears so. I'll port this to folders and modify the tests.

ste/sender-blobFolders.go
```diff
@@ -199,17 +199,33 @@ func (b *blobFolderSender) EnsureFolderExists() error {
...
 			&blockblob.UploadOptions{
 				HTTPHeaders:  &b.headersToApply,
 				Metadata:     b.metadataToApply,
-				Tags:         b.blobTagsToApply,
+				Tags:         blobTags,
 				CPKInfo:      b.jptm.CpkInfo(),
 				CPKScopeInfo: b.jptm.CpkScopeInfo(),
 			})
+		if err != nil {
+			b.jptm.FailActiveSend(common.Iff(len(blobTags) > 0, "Upload symlink (with tags)", "Upload symlink"), err)
```
gapra-msft, line 218:
```suggestion
			b.jptm.FailActiveSend(common.Iff(len(blobTags) > 0, "Upload folder (with tags)", "Upload folder"), err)
```

gapra-msft (approved):

PR #2795, opened 2024-08-27
Added support for OAuth with List Operation + e2e test

wonwuakpa-msft:
Removed the blocker preventing users from authentication with OAuth for list operations. 
Extended `Scenario_ListBasic` e2e test to support File and OAuth variations


e2etest/zt_newe2e_list_test.go
```diff
@@ -38,12 +39,15 @@ func (s *ListSuite) Scenario_ListBasic(svm *ScenarioVariationManager) {
...
 			Verb: AzCopyVerbList,
 			Targets: []ResourceManager{
-				srcObj.Parent().(RemoteResourceManager).WithSpecificAuthType(EExplicitCredentialType.SASToken(), svm, CreateAzCopyTargetOptions{
-					SASTokenOptions: GenericServiceSignatureValues{
-						ContainerName: srcObj.ContainerName(),
-						Permissions:   (&blobsas.ContainerPermissions{Read: true, List: true}).String(),
-					},
-				}),
+				srcObj.Parent().(RemoteResourceManager).WithSpecificAuthType(ResolveVariation(svm,
+					[]ExplicitCredentialTypes{EExplicitCredentialType.OAuth(),
```
vibhansa-msft, line 43: Does this mean the test case will now use OAuth for source and SAS for destination while it was using SAS for both sides earlier. If that's the case we shall add this as a new test case instead of removing the existing one.

wonwuakpa-msft: For this test, it extends the previous test implementation that was only testing a basic list operation with SAS tokens with Blobs. The updated test now incorporates variations in both authentication (OAuth and SAS) and location (File and Blob). Since this is a list operation and not a data transfer, not sure how the concept of a 'destination' is applicable here. The intent was to cover more variations within a scenario of a single test.

adreed-msft (approved):
Nice work extending the existing test!

vibhansa-msft (approved):

dphulkar-msft (approved):

Issue #2703, opened 2024-06-04
AzCopy v10 "panic: send on closed channel" on 403 Response


tommywilkinson:
### Which version of the AzCopy was used? 
10.25

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

### What command did you run?
"copy <local path> <destination> --log-level=INFO --recursive"
(redacted exact details, since it's not terribly relevant to the repro)

### What problem was encountered?
User account had owner permissions on storage account but not Storage Blob Data Contributor. Full error message from running azcopy locally:

```
INFO: Authentication failed, it is either not correct, or expired, or does not have the correct permission PUT https://<storageaccount>.blob.core.windows.net/1js/pr/3045226/wdjs/wac/we-c/debug/suiteux-shell/js/suiteux.shell.npm_mecontrol.rtl.js
--------------------------------------------------------------------------------
RESPONSE 403: 403 This request is not authorized to perform this operation using this permission.
ERROR CODE: AuthorizationPermissionMismatch
--------------------------------------------------------------------------------
﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>AuthorizationPermissionMismatch</Code><Message>This request is not authorized to perform this operation using this permission.
RequestId:3cd177f3-c01e-0050-18a5-b6a35a000000
Time:2024-06-04T17:34:48.4836791Z</Message></Error>
--------------------------------------------------------------------------------

panic: send on closed channel

goroutine 140 [running]:
github.com/Azure/azure-storage-azcopy/v10/ste.(*jobMgr).SendXferDoneMsg(0xc037b4fe08?, {{0xc000ef3b00, 0x7a}, {0xc000ef3b80, 0x80}, 0x0, 0xfffffffa, 0x12385, 0x0})
        D:/a/1/s/ste/jobStatusManager.go:72 +0x4e
github.com/Azure/azure-storage-azcopy/v10/ste.(*jobPartMgr).SendXferDoneMsg(0x180a2c3?, {{0xc000ef3b00, 0x7a}, {0xc000ef3b80, 0x80}, 0x0, 0xfffffffa, 0x12385, 0x0})
        D:/a/1/s/ste/mgr-JobPartMgr.go:694 +0x6e
github.com/Azure/azure-storage-azcopy/v10/ste.(*jobPartTransferMgr).ReportTransferDone(0xc000f08230)
        D:/a/1/s/ste/mgr-JobPartTransferMgr.go:992 +0x16c
[...]

PR #2804, opened 2024-09-10
Fix to the panic due to the send on a closed channel

wonwuakpa-msft:
PR for GitHub issue https://github.com/Azure/azure-storage-azcopy/issues/2703

Implemented a RWMutex to ensure exclusive access to critical parts of the codebase i.e channel closures and sends. The issue linked above is related specifically to the `xferDone `channel. Similar implementation was done for other channels in `jobStatusManager` with the mutually exclusive read and write locks.


ste/jobStatusManager.go
```diff
@@ -61,15 +65,36 @@ func (jm *jobMgr) statusMgrClosed() bool {
...
+		}
 	}
 }
 
 func (jm *jobMgr) SendXferDoneMsg(msg xferDoneMsg) {
-	jm.jstm.xferDone <- msg
+	jm.jstm.flagMutex.RLock()
+	defer jm.jstm.flagMutex.RUnlock()
+	if jm.jstm.isXferDoneClosed {
+		fmt.Println("Cannot send message on closed channel")
```
vibhansa-msft, line 87: This message will pop up on user console. I guess it will not make any sense to user/customer. Either we can just dump it to logs or put a meaningful message for user. If channel is closed for a valid reason then we can skip showing something to user and just add a log message.

ste/jobStatusManager.go
```diff
@@ -61,15 +65,36 @@ func (jm *jobMgr) statusMgrClosed() bool {
...
+		return
+	}
+	// channel is open, can send message
+	select {
+	case jm.jstm.xferDone <- msg:
+		fmt.Println("Message sent successfully!")
```
vibhansa-msft, line 93: Again, this is going to user console and will not make any sense to end user.

ste/jobStatusManager.go
```diff
@@ -61,15 +65,36 @@ func (jm *jobMgr) statusMgrClosed() bool {
 
 /* These functions should not fail */
 func (jm *jobMgr) SendJobPartCreatedMsg(msg JobPartCreatedMsg) {
-	jm.jstm.partCreated <- msg
-	if msg.IsFinalPart {
-		// Inform statusManager that this is all parts we've
-		close(jm.jstm.partCreated)
+	if jm.jstm.partCreated != nil { // Sends not allowed if channel is closed
+		jm.jstm.flagMutex.RLock()
```
vibhansa-msft, line 69: Channel push and pop are locked internally. Are we just trying to safeguard against parallel closure of the channel here ?

wonwuakpa-msft: We're trying to safeguard against panics from sending a message on a closed channel.  That might have been closed by another concurrent operation or go routine. Wrapping channel sends and closures with mutexes allows exclusive access to those parts of the code. What do you mean by locked internally?

wonwuakpa-msft: Not a problem! I'll replace the Mutex locking and unlocking mechanisms with this combo of `defer` and `recover` for channel sends

ste/jobStatusManager.go
```diff
@@ -155,19 +181,23 @@ func (jm *jobMgr) handleStatusUpdateMessage() {
 		case <-jstm.listReq:
 			/* Display stats */
 			js.Timestamp = time.Now().UTC()
+			jstm.flagMutex.RLock() // Read-Lock
 			jstm.respChan <- *js
+			jstm.flagMutex.RUnlock()
```
dphulkar-msft, line 186: jstm.flagMutex.RLock() 
defer jstm.flagMutex.RUnlock() 
jstm.respChan <- *js

Using defer ensures that the lock is always released, even if the send operation panics or encounters an error.

ste/jobStatusManager.go
```diff
@@ -155,19 +181,23 @@ func (jm *jobMgr) handleStatusUpdateMessage() {
...
 			js.FailedTransfers = []common.TransferDetail{}
 			js.SkippedTransfers = []common.TransferDetail{}
 
 			if allXferDoneHandled {
+				jstm.flagMutex.Lock() // Write Lock
 				close(jstm.statusMgrDone)
 				close(jstm.respChan)
 				close(jstm.listReq)
 				jstm.listReq = nil
 				jstm.respChan = nil
```
dphulkar-msft, line 199: jstm.statusMgrDone=nil
Should we also set the statusMgrDone channel to nil?

ste/mgr-JobMgr.go
```diff
@@ -715,7 +715,14 @@ func (jm *jobMgr) reportJobPartDoneHandler() {
 				(isCancelling && !haveFinalPart) // If we're cancelling, it's OK to try to exit early; the user already accepted this job cannot be resumed. Outgoing requests will fail anyway, so nothing can properly clean up.
 			if shouldComplete {
 				// Inform StatusManager that all parts are done.
-				close(jm.jstm.xferDone)
+
+				jm.jstm.flagMutex.Lock()
+				if !jm.jstm.isXferDoneClosed { //channel is open
```
dphulkar-msft, line 720: Did we introduce **isXferDoneClosed** variable just to keep status of xferDone channel as closed or open?
If so we can eliminate this variable and add a function to incorporate the same.

func (jm *jobMgr) xferDoneCLosed() bool {
    select {
    case <-jm.jstm.xferDone:
        return true
    default:
        return false
    }
}

dphulkar-msft:
Just wondering if we could add a testcase around it.
@wonwuakpa-msft

ste/mgr-JobMgr.go
wonwuakpa-msft: I modified the implementation to recover gracefully when a panic while sending on a closed channel occurs. So, there is no longer a need for the variable or function to check if the channel is closed.

wonwuakpa-msft:
@dphulkar-msft This bug fix, to me, seems tricky to test because of its non-deterministic nature. I was able to repro the error the customer had without having a panic on my end. In some cases, it might panic, in others it might not. It boils down to the timing of closing a channel before sending a message down it.

adreed-msft (approved):

vibhansa-msft (approved):

dphulkar-msft (approved):

dphulkar-msft:
@wonwuakpa-msft Please close this PR if not required. 
I can see one more PR with similar fix. https://github.com/Azure/azure-storage-azcopy/pull/2824

PR #2805, opened 2024-09-12
Build azcopy go version with 10.23.1

dphulkar-msft:


adreed-msft (approved):
Quick check: Have you checked the changelogs to validate nothing between 10.22.5 and 10.23.1 will affect us?

tanyasethi-msft (approved):

PR #2809, opened 2024-09-17
Check for both `folder` and `folder/` as hdi_isfolder blobs

adreed-msft:
The old implementation was designed against an HNS account, not a FNS account, and was almost always tested against HNS.

On HNS, if you were using the `blob` endpoint, and you listed, you wouldn't get returned folder stub blobs. You'd have to manually ping them for properties. Two things happened because of this.

1) We manually checked for `folder`.
2) We could schedule `folder` twice working against blob on FNS, because it'd now appear inside the listed items. This would cause a panic, because we'd try to deschedule a successful folder twice.

A third problem arose with certain customers, where they were following this format, but misunderstood it, and actually named their folders `folder/`. In this case, we'd find `folder/` normally, and it'd function like expected. But the original code wasn't explicitly checking for it.

So the solution is, always ignore the folder blobs when we're sent them as list items. Always check both `folder` and `folder/`. This means we'll always catch it, even on HNS.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -31,6 +31,7 @@ import (
 	"net/url"
 	"os"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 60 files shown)


cmd/copy.go
```diff
@@ -1347,6 +1348,27 @@ func (cca *CookedCopyCmdArgs) processRedirectionDownload(blobResource common.Res
 func (cca *CookedCopyCmdArgs) processRedirectionUpload(blobResource common.ResourceString, blockSize int64) error {
 	ctx := context.WithValue(context.TODO(), ste.ServiceAPIVersionOverride, ste.DefaultServiceApiVersion)
 
+	// Use the concurrency environment value
+	concurrencyEnvVar := glcm.GetEnvironmentVariable(common.EEnvironmentVariable.ConcurrencyValue())
+
+	pipingUploadParallelism := pipingUploadParallelism
```
vibhansa-msft, line 1354: Is this PR trying to fix multiple things. These changes does not appear to be related to folder/ thingy.

adreed-msft: Uhhh... This was unintentionally included actually... At the time, I was helping Wendi out with a PR in similar vein. I'll remove this.

adreed-msft: Not there anymore! Oh well.

vibhansa-msft (approved):

gapra-msft (approved):

Issue #2753, opened 2024-07-22
azcopy url-decodes slashes from directory names


frank-siebert-tracetronic:
### Which version of the AzCopy was used? 
10.24.0
### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

### What command did you run?
 azcopy copy "C:/Users/XXX/testproject/test_folder_%5C_something_wrong" "https://xxx.blob.core.windows.net/zzz?sp=racw&st=2024-07-22T10:00:55Z&se=2024-07-22T18:00:55Z&spr=https&sv=2022-11-02&sr=c&sig=abc" --recursive
### What problem was encountered?
I have some directories that have backward slashes in their names (they originally come from unix systems), so they get url encoded on my disk (therefore the %5C in the directory name).
What I am trying to do is upload them in the exact same manner, so no decoding taking place, just the plain "%5C" within the directory name.
However what is happening instead is the %5C gets replaced with a slash and then gets interpreted as a seperator, so in the storage I will get a subdirectory named "_something_wrong"

Expected: storageSource/test_folder_%5C_something_wrong
Actual: storageSource/test_folder_/_something_wrong

I have tried to do a double url encoding by renaming the directory to %255C but that didn't work either and will not be decoded at all, so it stays that way and therefore is also not a solution.

### How can we reproduce the problem in the simplest way?
Create a simple folder locally with a %5C in the directory name and transfer it to an arbitrary storage container via azcopy copy.

### Have you found a mitigation/solution?
No

ashruti-msft:
Okay I can reproduce this on windows not on linux. I'll get back to you regarding this.

frank-siebert-tracetronic:
> Okay I can reproduce this on windows not on linux. I'll get back to you regarding this.

Hi, are there any updates on this issue?

ashruti-msft:
Hi, can you please share your debug logs for this run. Thanks!

frank-siebert-tracetronic:
Well, those are corporate specific logs, so this is unfortunately not possible. I thought you could reproduce this. What part of the logs do you need?

Data-Engineer-Joe:
Hello @ashruti-msft, you said you can reproduce this issue. What's the next step here, this behaviour hurts pretty much because it rips our hive partitioning into pieces. Please follow up on this. I hope there will be a fix. Thanks.

jschelter-co:
[...]

PR #2811, opened 2024-09-19
Azcopy copy url-decodes slashes from directory names for windows

dphulkar-msft:
Issue: https://github.com/Azure/azure-storage-azcopy/issues/2753

When we do a copy operation from local to remote the path Encode Rules should not be applied on the source path.


seanmcc-msft:
We need to put this behind a flag to not break existing customers.  Default behavior should not change.

PR #2818, opened 2024-10-01
Correct behavior when syncing with delete-destination-files

adreed-msft:
The original code for delete-destination-files accidentally passed through all files when delete-destination-files was enabled.

This PR removes the accidental code and implements a test to validate that the override does not occur, further validating the two use cases this flag was intended to cover--

1) A blob previously written to via the dfs endpoint
2) A blob previously written to with another tool that uses different block ID lengths.

### DIFF
--- a/cmd/syncComparator.go
+++ b/cmd/syncComparator.go
@@ -65,13 +65,12 @@ type syncDestinationComparator struct {
 
 	comparisonHashType common.SyncHashType
 
-	preferSMBTime             bool
-	disableComparison         bool
-	deleteDestinationFileSync bool
+	preferSMBTime     bool
+	disableComparison bool
 }
 
-func newSyncDestinationComparator(i *objectIndexer, copyScheduler, cleaner objectProcessor, comparisonHashType common.SyncHashType, preferSMBTime, disableComparison bool, deleteDestinationFile bool) *syncDestinationComparator {
-	return &syncDestinationComparator{sourceIndex: i, copyTransferScheduler: copyScheduler, destinationCleaner: cleaner, preferSMBTime: preferSMBTime, disableComparison: disableComparison, comparisonHashType: comparisonHashType, deleteDestinationFileSync: deleteDestinationFile}
+func newSyncDestinationComparator(i *objectIndexer, copyScheduler, cleaner objectProcessor, comparisonHashType common.SyncHashType, preferSMBTime, disableComparison bool) *syncDestinationComparator {
+	return &syncDestinationComparator{sourceIndex: i, copyTransferScheduler: copyScheduler, destinationCleaner: cleaner, preferSMBTime: preferSMBTime, disableComparison: disableComparison, comparisonHashType: comparisonHashType}
 }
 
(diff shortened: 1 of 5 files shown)


cmd/syncComparator.go
```diff
@@ -65,13 +65,12 @@ type syncDestinationComparator struct {
...
-	preferSMBTime             bool
-	disableComparison         bool
-	deleteDestinationFileSync bool
+	preferSMBTime     bool
+	disableComparison bool
 }
 
-func newSyncDestinationComparator(i *objectIndexer, copyScheduler, cleaner objectProcessor, comparisonHashType common.SyncHashType, preferSMBTime, disableComparison bool, deleteDestinationFile bool) *syncDestinationComparator {
-	return &syncDestinationComparator{sourceIndex: i, copyTransferScheduler: copyScheduler, destinationCleaner: cleaner, preferSMBTime: preferSMBTime, disableComparison: disableComparison, comparisonHashType: comparisonHashType, deleteDestinationFileSync: deleteDestinationFile}
+func newSyncDestinationComparator(i *objectIndexer, copyScheduler, cleaner objectProcessor, comparisonHashType common.SyncHashType, preferSMBTime, disableComparison bool) *syncDestinationComparator {
```
vibhansa-msft, line 72: During sync we do delete file if its not present on the other side, are we revoking this feature now ?

adreed-msft: This is a confusion of two very similarly named flags that have significantly different functionality.

There are two flags vaguely named "delete destination". One of them cleans up files that are not present at the destination, the other deletes an object with uncommitted blocks before writing to the destination.

The former is used to basically mirror the source to the destination.

The latter is used to handle cases like writing to a blob with other tools, or writing to a blob prior written by the dfs endpoint.

adreed-msft: In this case, I've removed the latter, not the former. The former still happens in the destinationCleaner-- which is handled by the sync deleter.

e2etest/zt_newe2e_sync_test.go
```diff
@@ -170,3 +172,92 @@ func (s *SyncTestSuite) Scenario_TestSyncRemoveDestination(svm *ScenarioVariatio
...
+			overwriteName: ResourceDefinitionObject{Body: srcData},
+			ignoreName:    ResourceDefinitionObject{Body: srcData},
+		},
+	})
+
+	dstData := NewRandomObjectContentContainer(svm, 1024)
+	if !svm.Dryrun() {
+		time.Sleep(time.Second * 10) // Make sure this file is newer
+
+		CreateResource[ObjectResourceManager](svm, dstRes, ResourceDefinitionObject{
```
gapra-msft, line 228: won't this overwrite anything we did above for destination setup?

gapra-msft: oh nvm we are setting up two different files

gapra-msft (approved):

vibhansa-msft (approved):

PR #2819, opened 2024-10-04
Create msi for windows

dphulkar-msft:


release-pipeline.yml
```diff
@@ -1,2160 +1,2178 @@
...
-              sudo ls -lRt $(root_dir)/
-              cp $(root_dir)/azcopy_linux* $(Build.ArtifactStagingDirectory)
-            displayName: 'List the binaries and add to staging directory'
-
-          - task: PublishBuildArtifacts@1
-            inputs:
-              artifactName: 'azCopy-binaries'
-            displayName: 'Publish binaries'
-            condition: succeededOrFailed()
+# variables:
```
gapra-msft, line 1: Should we undo this commentting here?

azcopy.wxs
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
+  <Product Id="*" Name="azcopy" Language="1033" Version="10.26.0" Manufacturer="Microsoft" UpgradeCode="abeee33a-0cca-4060-89d7-0687f7374f7a">
```
gapra-msft, line 3: Will this version need to change every release? If so, we should add a section to the release process to double check this.

setup/create_windows_msi.yml
```diff
@@ -0,0 +1,72 @@
+parameters:
+  working_dir: ''
```
gapra-msft, line 2:
```suggestion
    - name: working_dir
      type: string
```

nit. Similar changes for the other params,

gapra-msft (commented):
made a few comments

azcopy.wxs
dphulkar-msft: yes this is the PR in progress not ready for review. Yes will not be using the Version as static value. Will make the changes for it

PR #2779, opened 2024-08-15
Fix to slow perf of PipeBlob uploads 

wonwuakpa-msft:
Assigned the pipingUploadParallelism to the user defined environment variable AZCOPY_CONCURRENCY_VALUE if present. If not, fell back on the default concurrency value and errored out for AUTO and non-integer values.

PR #2821, opened 2024-10-07
Fix to slow performance of PipeBlob uploads

wonwuakpa-msft:
Assigned the `pipingUploadParallelism `to the user defined environment variable `AZCOPY_CONCURRENCY_VALUE` if present. If not, fell back on the default concurrency value. Cleanly errored out for `AUTO` and non-integer values.


wonwuakpa-msft:
Same as this PR https://github.com/Azure/azure-storage-azcopy/pull/2779

vibhansa-msft (approved):

gapra-msft (approved):

PR #2824, opened 2024-10-09
Fix to the panic due to the send on a closed channel

wonwuakpa-msft:


ste/jobStatusManager.go
```diff
@@ -61,14 +61,28 @@ func (jm *jobMgr) statusMgrClosed() bool {
 
 /* These functions should not fail */
 func (jm *jobMgr) SendJobPartCreatedMsg(msg JobPartCreatedMsg) {
-	jm.jstm.partCreated <- msg
-	if msg.IsFinalPart {
-		// Inform statusManager that this is all parts we've
-		close(jm.jstm.partCreated)
+	defer func() {
```
gapra-msft, line 64: these fixes will technically resolve the issue, but have we looked into if this function is being called when it shouldnt be?

gapra-msft, line 64: is there an easy test that we can add to our repo that reproduces the bug

wonwuakpa-msft: The main issue with this bug is that it's a bit arbitrary and non-deterministic to reproduce. It is all about the timing of when the send was made on a channel that has already been closed. This doesn't always happen. But in the case, it does in the future, we have some safeguards around that.

wonwuakpa-msft: Yes, I looked into that and this function is called only one other time. When we update the jobPart Status with the status Manager -- which seems to be fine to me!

gapra-msft (approved):

vibhansa-msft (approved):

dphulkar-msft (approved):

ste/jobStatusManager.go
```diff
@@ -61,15 +61,34 @@ func (jm *jobMgr) statusMgrClosed() bool {
...
+		if recErr := recover(); recErr != nil {
+			jm.Log(common.LogError, "Cannot send message on closed channel")
+		}
+	}()
+	if jm.jstm.partCreated != nil { // Sends not allowed if channel is closed
```
gapra-msft, line 69: Do we need to do this check?

wonwuakpa-msft: Yes, because sending on a nil channel will cause a deadlock and block the routine indefinitely.

gapra-msft: If it is nil should an error be returned? Also should we add this nil check to the xferDoneMsg channel?

wonwuakpa-msft: The channel is set to nil only after it's been closed. And, we already have support for cleanly erroring out for sends on a closed channel.

ste/jobStatusManager.go
```diff
@@ -155,8 +176,19 @@ func (jm *jobMgr) handleStatusUpdateMessage() {
...
-			jstm.respChan <- *js
-
+			if jstm.respChan != nil {
+				select {
+				case jstm.respChan <- *js:
+					// Send on the channel
+				case <-jstm.statusMgrDone:
+					// If we time out, no biggie. This isn't world-ending, nor is it essential info. The other side stopped listening by now.
+				}
+				defer func() { // Exit gracefully if panic
```
gapra-msft, line 186: Should this code come before reading from respChan?

ste/jobStatusManager.go
```diff
@@ -61,15 +61,36 @@ func (jm *jobMgr) statusMgrClosed() bool {
...
-	jm.jstm.partCreated <- msg
-	if msg.IsFinalPart {
-		// Inform statusManager that this is all parts we've
-		close(jm.jstm.partCreated)
+	defer func() {
+		if recErr := recover(); recErr != nil {
+			jm.Log(common.LogError, "Cannot send message on closed channel")
+		}
+	}()
+	if jm.jstm.partCreated != nil { // Sends not allowed if channel is closed
```
adreed-msft, line 69: Since we're not nilling things out anymore, let's remove the nil checks, they're unnecessary now.

ste/jobStatusManager.go
adreed-msft: A read won't panic; this is probably unnecessary.

adreed-msft (approved):

ste/jobStatusManager.go
wonwuakpa-msft: `partCreated` is still being nil, should I remove that and others?

gapra-msft: Where is that being niled? I had to remove the nil setting below to fix the sync test timeout

wonwuakpa-msft: Yes just double checked it's removed

gapra-msft (approved):

PR #2827, opened 2024-10-10
Added support for trailing dot transfers to unsafe locations

wonwuakpa-msft:
PR made to address [this customer issue](https://msazure.visualstudio.com/One/_backlogs/backlog/XStore-DevExp/Epics?showParents=true&workitem=27349680) concerning transferring trailing dot files. Windows natively does not support trailing dots in filenames, so a (initially just `Enable` and `Disable`) third option `AllowToUnsafeDestination` was added to AzCopy to enable this functionality. 

- Updated the documentation to include the feature
- Tests using the E2E framework were added for this additional trailing dot `AllowToUnsafeDestination` option
- Persisted the `trailingDotOption` information when the job is resumed
- Update the code to prevent nil dereference error from one of the tests


cmd/jobsResume.go
wonwuakpa-msft, line 1: Implements persistence of the  `trailingDot` flag option when a job is resumed after pausing

cmd/jobsResume.go
```diff
@@ -295,8 +295,28 @@ func (rca resumeCmdArgs) getSourceAndDestinationServiceClients(
...
+	}
 
-	srcServiceClient, err := common.GetServiceClientForLocation(fromTo.From(), source, srcCredType, tc, &options, nil)
+	var fileSrcClientOptions common.FileClientOptions
+	if fromTo.From() == common.ELocation.File() {
+		fileSrcClientOptions = common.FileClientOptions{
+			AllowTrailingDot: getJobDetailsResponse.TrailingDot.IsEnabled(), //Access the trailingDot option of the job
+		}
+	}
+	srcServiceClient, err := common.GetServiceClientForLocation(fromTo.From(), source, srcCredType, tc, &options, fileSrcClientOptions)
```
adreed-msft, line 319: I don't *expect* this to cause issues, but, ideally our client options should probably be an `any` with either `nil` or the target client options we want as opposed to just always passing in file client options.

cmd/jobsResume.go
```diff
@@ -306,11 +326,17 @@ func (rca resumeCmdArgs) getSourceAndDestinationServiceClients(
 		srcCred = common.NewScopedCredential(tc, srcCredType)
 	}
 	options = createClientOptions(common.AzcopyCurrentJobLogger, srcCred)
-	dstServiceClient, err := common.GetServiceClientForLocation(fromTo.To(), destination, dstCredType, tc, &options, nil)
+	var fileClientOptions any
```
adreed-msft, line 329: like this yes

cmd/zc_traverser_file.go
```diff
@@ -145,7 +149,7 @@ func (t *fileTraverser) Traverse(preprocessor objectMorpher, processor objectPro
 			WarnStdoutAndScanningLog(fmt.Sprintf(invalidNameErrorMsg, targetURLParts.DirectoryOrFilePath))
 			return common.EAzError.InvalidBlobOrWindowsName()
 		}
-		if t.trailingDot != common.ETrailingDotOption.Enable() && strings.HasSuffix(targetURLParts.DirectoryOrFilePath, ".") {
+		if !t.trailingDot.IsEnabled() && strings.HasSuffix(targetURLParts.DirectoryOrFilePath, ".") {
```
adreed-msft, line 152: Unrelated thought, but something we should consider.  This is definitely a @gapra-msft type question as well... Should we be trimming the trailing dot if we're letting cx shoot themselves in the foot?

It's possible they actually want to properly fetch the data anyway, but that's always possible as `azcopy copy "sourcefile." "destination"`

adreed-msft (approved):

cmd/zc_traverser_file.go
wonwuakpa-msft: @adreed-msft if you're talking about the `isEnabled()` method, it lets users to use `Enable` and `AllowToUnsafeDestination`

adreed-msft: Discussed offline; this is in reference to functionality. Currently, if we copy a directory containing a file `file.` with the `AllowToUnsafeDestination` option enabled, `file.` will fail to transfer as it is not a valid filename on the destination.

If a customer was trying to copy one of these files, which they would be, knowingly enabling AllowToUnsafeDestination, they would likely want `file.` to become `file` when travelling to the unsafe destination.

gapra-msft: oh I see, so you mean "should we strip the trailing dot for customers on AllowToUnsafeDEstinations for unsafe destinations".

gapra-msft: I think for this particular bug, what has been implemented is sufficient - we just allow the destination filesystem to do what it will with the trailing dot

cmd/copy.go
```diff
@@ -2134,9 +2134,11 @@ func init() {
...
+		"'Enable'(Default) treats trailing dot file operations in a safe manner between systems that support these files. On Windows, the transfers will not occur to stop risk of data corruption. See 'AllowToUnsafeDestination' to bypass this."+
+		"'Disable' reverts to the legacy functionality, where trailing dot files are ignored. This can result in potential data corruption if the transfer contains two paths that differ only by a trailing dot (E.g 'path/foo' and 'path/foo.'). If this flag is set to 'Disable' and AzCopy encounters a trailing dot file, it will warn customers in the scanning log but will not attempt to abort the operation."+
+		"If the destination does not support trailing dot files (Windows or Blob Storage), AzCopy will fail if the trailing dot file is the root of the transfer and skip any trailing dot paths encountered during enumeration."+
+		"'AllowToUnsafeDestination' supports transferring trailing dot files to systems that do not support them e.g Windows or Blob Storage. Use with caution acknowledginge risk of data corruption, when two files with different contents 'path/bar' and 'path/bar.' (differ only by a trailing dot) are seen as identical.")
```
gapra-msft, line 2141: Should we fix this comment to only include Windows? Cause Blob wont be supported here right?

cmd/zc_traverser_file.go
```diff
@@ -123,6 +123,10 @@ func (t *fileTraverser) getPropertiesIfSingleFile() (*file.GetPropertiesResponse
 func (t *fileTraverser) Traverse(preprocessor objectMorpher, processor objectProcessor, filters []ObjectFilter) (err error) {
 	invalidBlobOrWindowsName := func(path string) bool {
 		if t.destination != nil {
+			if t.trailingDot == common.ETrailingDotOption.AllowToUnsafeDestination() {
```
gapra-msft, line 126: should we only return false for Local here? Since Blob will fail on the customer.

ste/jobStatusManager.go
```diff
@@ -67,10 +67,7 @@ func (jm *jobMgr) SendJobPartCreatedMsg(msg JobPartCreatedMsg) {
 		}
 	}()
 	if jm.jstm.partCreated != nil { // Sends not allowed if channel is closed
-		select {
-		case jm.jstm.partCreated <- msg:
-		case <-jm.jstm.statusMgrDone: // Nobody is listening anymore, let's back off.
-		}
+		jm.jstm.partCreated <- msg
```
gapra-msft, line 70: Is this change part of this PR?

wonwuakpa-msft: It shouldn't be, I think I had some issues with the bases of my branch

e2etest/zt_newe2e_basic_functionality_test.go
```diff
@@ -412,6 +414,73 @@ func (s *BasicFunctionalitySuite) Scenario_Copy_EmptySASErrorCodes(svm *Scenario
 	ValidateContainsError(svm, stdout, []string{"https://aka.ms/AzCopyError/NoAuthenticationInformation", "https://aka.ms/AzCopyError/ResourceNotFound"})
 }
 
+// Test of Copy and Sync commands to UnsafeDestinations (Windows Local and Blob dest)
```
gapra-msft, line 417:
```suggestion
// Test of Copy and Sync commands to UnsafeDestinations (Windows Local dest)
```

gapra-msft (approved):

PR #2828, opened 2024-10-11
Fix making too many create directory call

dphulkar-msft:
**Issue:**
In the folderCreationTracker.go file, we addressed an issue where multiple createDirectory REST calls were being sent for the same folder during the copy operation, particularly with deeply nested directory structures. This was caused by the StopTracking() function prematurely removing folder entries from the contents map, leading to redundant createDirectory requests.

**Fix:**
To resolve this issue while maintaining memory optimization (which was the original reason for removing entries), we have replaced the contents map with a trie data structure. The trie allows us to efficiently track folder creation and ensures that redundant requests are avoided, while also optimizing memory usage.

### DIFF
--- a/common/trieForDirPath.go
+++ b/common/trieForDirPath.go
@@ -5,9 +5,10 @@ import (
 )
 
 type TrieNode struct {
-	Children map[string]*TrieNode
-	Value    uint32
-	isEnd    bool
+	dirPathSeg             map[string]*TrieNode
+	status                 uint32
+	unregisteredButCreated bool
+	isEnd                  bool
 }
 
 type Trie struct {
@@ -16,68 +17,73 @@ type Trie struct {
 
 func NewTrie() *Trie {
 	return &Trie{
-		Root: &TrieNode{Children: make(map[string]*TrieNode)},
+		Root: &TrieNode{dirPathSeg: make(map[string]*TrieNode)},
 	}
 }
 
-func (t *Trie) Insert(key string, value uint32) {
+// @brief InsertStatus inserts the status of the directory path in the trie
+func (t *Trie) InsertStatus(dirPath string, dirCreationStatus uint32) {
 	node := t.Root
-	segments := strings.Split(key, "/")
+	segments := strings.Split(dirPath, "/")
 	for _, segment := range segments {
-		child, exists := node.Children[segment]
+		child, exists := node.dirPathSeg[segment]
 		if !exists {
-			child = &TrieNode{Children: make(map[string]*TrieNode)}
-			node.Children[segment] = child
+			child = &TrieNode{dirPathSeg: make(map[string]*TrieNode)}
+			node.dirPathSeg[segment] = child
 		}
 		node = child
 	}
-	node.Value = value
+	node.status = dirCreationStatus
 	node.isEnd = true
 }
 
-func (t *Trie) Get(key string) (uint32, bool) {
+// @brief InsertUnregisteredStatus inserts the unregistered status of the directory path in the trie
+func (t *Trie) InsertUnregisteredStatus(dirPath string, unregisteredButCreated bool) {
 	node := t.Root
-	segments := strings.Split(key, "/")
+	segments := strings.Split(dirPath, "/")
 	for _, segment := range segments {
-		child, exists := node.Children[segment]
+		child, exists := node.dirPathSeg[segment]
+		if !exists {
+			child = &TrieNode{dirPathSeg: make(map[string]*TrieNode)}
+			node.dirPathSeg[segment] = child
+		}
+		node = child
+	}
+
+	node.unregisteredButCreated = unregisteredButCreated
+	node.isEnd = true
+}
+
+// @brief GetStatus returns the status of the directory path in the trie
+func (t *Trie) GetStatus(dirPath string) (uint32, bool) {
+	node := t.Root
+	segments := strings.Split(dirPath, "/")
+	for _, segment := range segments {
+		child, exists := node.dirPathSeg[segment]
 		if !exists {
 			return 0, false
 		}
 		node = child
 	}
 	if node.isEnd {
-		return node.Value, true
+		return node.status, true
 	}
 	return 0, false
 }
 
-func (t *Trie) Delete(key string) bool {
-	segments := strings.Split(key, "/")
-	return t.deleteHelper(t.Root, segments, 0)
-}
-
-func (t *Trie) deleteHelper(node *TrieNode, segments []string, depth int) bool {
-	if node == nil {
-		return false
-	}
-
-	// If we have reached the end of the key
-	if depth == len(segments) {
-		if !node.isEnd {
-			return false // Key does not exist
+// @brief CheckIfUnregisteredButCreated checks if the directory path is unregistered but created in the trie
+func (t *Trie) CheckIfUnregisteredButCreated(dirPath string) (bool, bool) {
+	node := t.Root
+	segments := strings.Split(dirPath, "/")
+	for _, segment := range segments {
+		child, exists := node.dirPathSeg[segment]
+		if !exists {
+			return false, false
 		}
-		node.isEnd = false // Unmark the end of the key
-		node.Value = 0
-
-		// If the node has no Children, it can be deleted
-		return len(node.Children) == 0
+		node = child
 	}
-
-	segment := segments[depth]
-	if t.deleteHelper(node.Children[segment], segments, depth+1) {
-		delete(node.Children, segment)
-		return !node.isEnd && len(node.Children) == 0
+	if node.isEnd {
+		return node.unregisteredButCreated, true
 	}
-
-	return false
+	return false, false
 }
--- a/common/trieForDirPath.go
+++ b/common/trieForDirPath.go
@@ -38,7 +38,7 @@ func (t *Trie) InsertStatus(dirPath string, dirCreationStatus uint32) {
 }
 
 // @brief InsertUnregisteredStatus inserts the unregistered status of the directory path in the trie
-func (t *Trie) InsertUnregisteredStatus(dirPath string, unregisteredButCreated bool) {
+func (t *Trie) SetUnregisteredStatus(dirPath string, unregisteredButCreated bool) {
 	node := t.Root
 	segments := strings.Split(dirPath, "/")
 	for _, segment := range segments {
--- a/common/trieForDirPath.go
+++ b/common/trieForDirPath.go
@@ -6,7 +6,7 @@ import (
 
 type TrieNode struct {
 	Children map[string]*TrieNode
-	Value    *uint32
+	Value    uint32
 	isEnd    bool
 }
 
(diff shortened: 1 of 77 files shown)


dphulkar-msft:
@gapra-msft @adreed-msft
Feel free to suggest alternative approaches. 
This is just the reference MR showcasing a simple fix for the issue.

ste/xfer-anyToRemote-folder.go
```diff
@@ -82,7 +82,7 @@ func anyToRemote_folder(jptm IJobPartTransferMgr, info *TransferInfo, pacer pace
 	} else {
 
 		t := jptm.GetFolderCreationTracker()
-		defer t.StopTracking(s.DirUrlToString()) // don't need it after this routine
+		//defer t.StopTracking(s.DirUrlToString()) // don't need it after this routine
```
vibhansa-msft, line 85: @adreed-msft We see this map deletion was introduced to save some memory footprints, but this is causing duplicate directory creation calls as the directory creation relies on entry being present in the map. Kindly verify.

vibhansa-msft (approved):

ste/xfer-anyToRemote-folder.go
```diff
@@ -82,7 +82,7 @@ func anyToRemote_folder(jptm IJobPartTransferMgr, info *TransferInfo, pacer pace
 	} else {
 
 		t := jptm.GetFolderCreationTracker()
```
gapra-msft, line 84: is there any way we can test this by mocking the storage clients?

dphulkar-msft: Not sure. Will need to check

adreed-msft (changes requested):
A couple concerns I thought of...

1) This only affects folders we're directly tracking to transfer properties. What about those that we're not?
2) The mention of the use of a Trie instead of a hashmap on Vikas' comment
3) What about when we're not trying to persist directories at all? e.g. blob->file. We need to carefully track the existing directory structure for a valid implementation of this.

ste/xfer-anyToRemote-folder.go
gapra-msft: I like the idea of a trie for mapping

ste/folderCreationTracker.go
```diff
@@ -127,8 +203,8 @@ func (f *jpptFolderTracker) ShouldSetProperties(folder string, overwrite common.
 		defer f.mu.Unlock()
 
 		var created bool
-		if idx, ok := f.contents[folder]; ok {
-			created = f.plan.Transfer(idx).TransferStatus() == common.ETransferStatus.FolderCreated()
+		if idx, ok := f.contents.Get(folder); ok {
+			created = f.plan.Transfer(*idx).TransferStatus() == common.ETransferStatus.FolderCreated()
```
vibhansa-msft, line 207: We shall check for status created or success like the other check. Once creation is complete the status is set to success from created.

ste/folderCreationTracker.go
```diff
@@ -9,6 +9,78 @@ import (
...
+	}
+}
+
+func (t *Trie) Insert(key string, value uint32) {
+	node := t.root
+	for _, char := range key {
+		if _, exists := node.children[char]; !exists {
+			node.children[char] = &TrieNode{children: make(map[rune]*TrieNode)}
+		}
+		node = node.children[char]
```
vibhansa-msft, line 34: We are searchig the map twice here once in if condition and then to get the node back. I think we can do this in one shot only during if check

ste/folderCreationTracker.go
```diff
@@ -9,6 +9,78 @@ import (
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 )
 
+type TrieNode struct {
```
vibhansa-msft, line 12: May be better to put this implementation of Trie in a seperate file and have some UT around it.

ste/folderCreationTracker.go
```diff
@@ -9,6 +9,78 @@ import (
...
+	children map[rune]*TrieNode
+	value    *uint32
```
jainakanksha-msft, line 14: Do we require to store value address or can we simply store the value of that?

ste/folderCreationTracker.go
```diff
@@ -167,17 +243,22 @@ func (f *jpptFolderTracker) StopTracking(folder string) {
 		return // Not possible to track this
 	}
 
+	//add a log mentioning we are in stotracking and folder is being deleted
```
jainakanksha-msft, line 246: Typo stotracking

ashruti-msft: Added for debugging purposes. Removed now.

ste/folderCreationTracker.go
ashruti-msft: you are correct we should store the value directly.

ste/folderCreationTracker.go
```diff
@@ -55,7 +55,7 @@ func (f *nullFolderTracker) StopTracking(folder string) {
 type jpptFolderTracker struct {
 	plan                   IJobPartPlanHeader
 	mu                     *sync.Mutex
-	contents               map[string]uint32
+	contents               *common.Trie
 	unregisteredButCreated map[string]struct{}
```
gapra-msft, line 59: I think we should be able to also contain the unregistered but created info in the same Trie right?

dphulkar-msft: Addressed. Please review

common/trieForDirPath.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		node = child
+	}
+	if node.isEnd {
+		return node.Value, true
+	}
+	return 0, false
+}
+
+func (t *Trie) Delete(key string) bool {
```
gapra-msft, line 54: this can be addressed after the PR is merged but Delete is no longer used right? So it can be deleted

ste/folderCreationTracker.go
```diff
@@ -23,8 +22,8 @@ func NewFolderCreationTracker(fpo common.FolderPropertyOption, plan *JobPartPlan
 		return &jpptFolderTracker{ // This prevents a dependency cycle. Reviewers: Are we OK with this? Can you think of a better way to do it?
 			plan:                   plan,
 			mu:                     &sync.Mutex{},
-			contents:               make(map[string]uint32),
-			unregisteredButCreated: make(map[string]struct{}),
+			contents:               common.NewTrie(),
+			unregisteredButCreated: common.NewTrie(),
```
gapra-msft, line 26: I'm also fine if we address this after the PR is merged but shouldnt both contents and unregisteredButCreated be able to be tracked in the same Trie? We can just add a boolean for unregisteredButCreated.

dphulkar-msft: I initially considered combining the data to optimize memory usage, but after further reflection, I realized that in terms of code readability, it's better to keep the data separate. By keeping the data distinct, the logic remains clearer, making it easier for developers to understand and maintain the code. 
Let me know your thoughts on this.

common/trieForDirPath.go
dphulkar-msft: we are using this delete function for removing the entries from unregisteredButCreated

ste/folderCreationTracker.go
dphulkar-msft: Something of this sort right?

type TrieNode struct {
	Children               map[string]*TrieNode
	Value                  uint32
	unregisteredButCreated bool
	isEnd                  bool
}

gapra-msft: yep

common/trieForDirPath.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+type TrieNode struct {
+	dirPathSeg             map[string]*TrieNode
+	status                 uint32
+	unregisteredButCreated bool
+	isEnd                  bool
+}
+
+type Trie struct {
```
adreed-msft, line 14: This is fine, but could be improved. Golang supports generics now, and if we implement this in a generic fashion, we can re-use this work to improve the scale of Sync. It's fine if this merges as is now, it's just tech debt.

common/trieForDirPath.go
```diff
@@ -0,0 +1,72 @@
...
+	return &Trie{
+		Root: &TrieNode{dirPathSeg: make(map[string]*TrieNode)},
+	}
+}
+
+// @brief InsertStatus inserts the status of the directory path in the trie
+func (t *Trie) InsertStatus(dirPath string, dirCreationStatus uint32) {
+	node := t.Root
+	segments := strings.Split(dirPath, "/")
+	for _, segment := range segments {
```
adreed-msft, line 28: Tree-walking is duplicate code. This could be DRY.

common/trieForDirPath.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func NewTrie() *Trie {
+	return &Trie{
+		Root: &TrieNode{dirPathSeg: make(map[string]*TrieNode)},
+	}
+}
+
+// @brief InsertStatus inserts the status of the directory path in the trie
+func (t *Trie) InsertStatus(dirPath string, dirCreationStatus uint32) {
```
adreed-msft, line 25: It is also important to note that because this is a map of pointers, we don't have to re-walk the entire tree every time. Any call in this situation, even if we already have the path, is equally expensive time-wise.

We could re-implement the portions of these functions to work on pre-walked nodes, if say, we called GetDirDetails followed by another one of these calls.

common/trieForDirPath.go
```diff
@@ -0,0 +1,72 @@
...
+	segments := strings.Split(dirPath, "/")
+	for _, segment := range segments {
+		child, exists := node.dirPathSeg[segment]
+		if !exists {
+			child = &TrieNode{dirPathSeg: make(map[string]*TrieNode)}
+			node.dirPathSeg[segment] = child
+		}
+		node = child
+	}
+	node.status = dirCreationStatus
```
adreed-msft, line 36: This smells slightly. Is unregistered but created not in itself a status? Is it important to note that we created something, but that it is not a transfer we "own"?

common/trieForDirPath.go
```diff
@@ -0,0 +1,72 @@
+package common
+
+import (
+	"strings"
+)
+
+type TrieNode struct {
+	dirPathSeg             map[string]*TrieNode
+	status                 uint32
```
adreed-msft, line 9: Upon further thinking, the same problem returns to existence prior to the job plan tracker. When AzCopy gets restarted, e.g. `jobs resume`, we don't persist these brand new folders we created into the plan file.

adreed-msft: Separate comment, because unrelated to this PR, but future work, wrt scaling, part of the equation is going to be reducing plan file size after we finish sets of transfers. We'll want some efficient way to keep track of the folder structure we've created down the line. This is more of a TODO for me than you.

adreed-msft: Back to the comment re: unregisteredButCreated. If we're adding these folders into the plan file, we can just set status to the index in the plan file. Since we're never dropping things from the trie to save memory now, we should also worry about adding folders we do register in the plan files into this trie, even if the tx is already completed.

adreed-msft (changes requested):
Apologies for the wall of text, I was reading the code and just more things kept popping into my head.

adreed-msft (approved):

PR #2833, opened 2024-10-15
Run workload identity tests first to avoid timeout

adreed-msft:


vibhansa-msft (approved):

azure-pipelines.yml
```diff
@@ -215,7 +215,189 @@ jobs:
...
+          testRunTitle: 'Go on $(display_name)'
+
+      - task: PublishCodeCoverageResults@1
+        condition: succeededOrFailed()
+        inputs:
+          codeCoverageTool: Cobertura
+          summaryFileLocation: $(System.DefaultWorkingDirectory)/**/$(display_name)_coverage.xml
+          additionalCodeCoverageFiles: $(System.DefaultWorkingDirectory)/**/$(display_name)_coverage.html
+
+  - job: New_E2E_Framework
```
gapra-msft, line 294: if these 2 jobs are the same barring the line that runs the e2e tests, we could create a template and pass that line as a variable

gapra-msft (approved):

PR #2838, opened 2024-10-21
Adding OAuth as supported source account AuthN method using blob/file server-to-server copies 

derdanu:
OAuth is supported beside SAS and public on the source account as AuthN method using blob and file server-to-server copies.


gapra-msft:
@adreed-msft could you take a quick look when you get a chance?

adreed-msft (approved):

gapra-msft (approved):

gapra-msft:
https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=23759&view=results

Pipeline run

PR #2845, opened 2024-10-28
Improved some logging messages and --log-level NONE behavior

dphulkar-msft:
1. If --log-level is set to NONE it should not display the log file location message on the cmd prompt. (Applicable for copy, sync, jobs resume)
2. We do support --log-level as DEBUG but the details of it are missing from the help command that we display.
3. Corrected some logging message level.


cmd/copy.go
```diff
@@ -279,6 +279,11 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 		azcopyScanningLogger.CloseLog()
 	})
 
+	// if no logging, set this empty so that we don't display the log location
+	if azcopyLogVerbosity == common.LogNone {
```
gapra-msft, line 283: We should be able to test this behavior in the e2e test framework.

cmd/sync.go
```diff
@@ -140,6 +140,11 @@ func (raw *rawSyncCmdArgs) cook() (cookedSyncCmdArgs, error) {
 		azcopyScanningLogger.CloseLog()
 	})
 
+	// if no logging, set this empty so that we don't display the log location
+	if azcopyLogVerbosity == common.LogNone {
```
gapra-msft, line 144: When log level is none, do we also not produce a log file at all?

dphulkar-msft: yeah correct. We dont create the log file at all.

cmd/copy.go
```diff
@@ -573,7 +578,7 @@ func (raw rawCopyCmdArgs) cook() (CookedCopyCmdArgs, error) {
 		return cooked, errors.New("blob tags can only be set when transferring to blob storage")
 	}
 	if cooked.FromTo.To() == common.ELocation.None() && strings.EqualFold(raw.blobTags, common.MetadataAndBlobTagsClearFlag) { // in case of Blob and BlobFS
-		glcm.Info("*** WARNING *** BlobTags will be cleared because of input --blob-tags=clear ")
+		glcm.Warn("*** WARNING *** BlobTags will be cleared because of input --blob-tags=clear ")
```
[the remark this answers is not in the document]
dphulkar-msft: Having separate methods for Warn and Info is generally a good practice, even if their implementations are currently identical. This separation enhances readability, flexibility, and semantic clarity in your codebase.

wonwuakpa-msft (commented):
NA

wonwuakpa-msft (approved):

seanmcc-msft:
@dphulkar-msft we should be good to merge this PR after you update the branch with main.

wonwuakpa-msft (approved):

PR #2847, opened 2024-10-29
Catch an object named after the root with a trailing `/` & carefully handle trailing `/` files on sync

adreed-msft:
Two issues still existed following the folder naming PR.

1) Sync with `--delete-destination-files`; files ending with `/` were not caught properly.
2) Root directories ending with `/`, when not containing the `hdi_isfolder` meta were caught, but not properly put into the plan file.

### DIFF
--- a/cmd/zc_traverser_blob.go
+++ b/cmd/zc_traverser_blob.go
@@ -559,6 +559,11 @@ func (t *blobTraverser) serialList(containerClient *container.Client, containerN
 
 			storedObject := t.createStoredObjectForBlob(preprocessor, blobInfo, relativePath, containerName)
 
+			// edge case, blob name happens to be the same as root and ends in /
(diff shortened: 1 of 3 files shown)


cmd/zc_traverser_blob.go
```diff
@@ -419,6 +419,11 @@ func (t *blobTraverser) parallelList(containerClient *container.Client, containe
 
 				storedObject := t.createStoredObjectForBlob(preprocessor, blobInfo, strings.TrimPrefix(*blobInfo.Name, searchPrefix), containerName)
 
+				// edge case, blob name happens to be the same as root and ends in /
+				if storedObject.relativePath == "" && strings.HasSuffix(storedObject.name, "/") {
```
gapra-msft, line 423: does this code change need to apply for serial listing too?

gapra-msft (approved):

gapra-msft (approved):

PR #2883, opened 2024-11-28
Added proper error message for system to system container copy

dphulkar-msft:
## Description

Customers attempt to copy $blobchangefeed in a source to $blobchangefeed in a destination. 

We should allow copy of system containers to non-system containers, but need better error messaging for system container to system container copies when customers try to execute
sysContainerBug

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Manual Testing
sysContainerCopy

sysContainerSync


common/util.go
```diff
@@ -382,3 +382,15 @@ func DoWithOverrideReadOnlyOnAzureFiles(ctx context.Context, action func() (inte
 	_, err = action()
 	return err
 }
+
+// @brief Thia API check if the container name provided is a system container or not
+func IsSystemContainer(containerName string) bool {
+	// define the system variables for the system containers
+	systemContainers := []string{"$blobchangefeed", "$logs"}
```
vibhansa-msft, line 389: These containers as destination of copy shall be errored out but as source it shall still work. Add test cases around such combinations in E2E

common/util.go
```diff
@@ -382,3 +382,15 @@ func DoWithOverrideReadOnlyOnAzureFiles(ctx context.Context, action func() (inte
 	_, err = action()
 	return err
 }
+
+// @brief Thia API check if the container name provided is a system container or not
+func IsSystemContainer(containerName string) bool {
```
gapra-msft, line 387: is it possible that the input $ is encoded when it gets here? Would be good to check that case as well

wonwuakpa-msft (commented):
left comment

wonwuakpa-msft (approved):

cmd/copy.go
```diff
@@ -1634,6 +1634,17 @@ func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 		return fmt.Errorf("S2S copy from Azure File authenticated with Azure AD to Blob/BlobFS is not supported")
 	}
 
+	// Check if destination is system container
+	if cca.FromTo.IsS2S() {
+		dstContainerName, err := GetContainerName(cca.Destination.Value, cca.FromTo.To())
+		if err != nil {
+			return fmt.Errorf("failed to get container name from destination (is it formatted correctly?)")
```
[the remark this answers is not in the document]
dphulkar-msft: done

e2etest/zt_newe2e_s2s_test.go
```diff
@@ -619,3 +621,52 @@ func (s *S2STestSuite) Scenario_S2SDirectoryMultipleFilesStripTopDirNonRecursive
 		Objects: dstObjs,
 	}, true)
 }
+
+func (s *S2STestSuite) Scenario_SystemContainerCopy(svm *ScenarioVariationManager) {
```
gapra-msft, line 625: This test is failing since we are trying to create a file in the blobchangefeed container, which is not allowed. I think there is a way to just Get the resource instead of trying to create it, which is what we should do here.

common/util.go
dphulkar-msft: Added

e2etest/zt_newe2e_s2s_test.go
dphulkar-msft: Updated the test.

e2etest/zt_newe2e_s2s_test.go
```diff
@@ -619,3 +620,39 @@ func (s *S2STestSuite) Scenario_S2SDirectoryMultipleFilesStripTopDirNonRecursive
...
+	azCopyVerb := ResolveVariation(svm, []AzCopyVerb{AzCopyVerbCopy, AzCopyVerbSync}) // Calculate verb early to create the destination object early
+
+	dstObj := GetAccount(svm, PrimaryStandardAcct).GetService(svm, common.ELocation.Blob()).GetContainer("$logs")
+
+	body := NewRandomObjectContentContainer(SizeFromString("10K"))
```
gapra-msft, line 629: no need to actually make data for this test since its going to fail anyway. ok if we do this after the release.

gapra-msft (approved):

vibhansa-msft (approved):

PR #2887, opened 2024-12-06
Implement a policy to reauthorize the credential when invalid

adreed-msft:
## Description

- **Feature / Bug Fix**: A bit of both actually. With long enough jobs under device code auth, the underlying token could totally time out and cause a job to abruptly fail. This PR adds in a pipeline policy that catches failures of this nature, and provides a prompt to reauthorize on the fly.

- **Related Links**:
- Email Subject: "AzCopy always fails trying to download a large file due to token expiration"
- ADO Work Item # 29466291

## Type of Change

- [x] Bug fix
- [x] New feature

## How Has This Been Tested?

1) Manually tested by trying to perform a transfer against a stg account I do not have access to, reauth-ing to an identity that does have access
2) Logic test in `TestDestReauthPolicy` ensures the pipeline policy fundamentally works.


ste/destReauthPolicy.go
```diff
@@ -0,0 +1,114 @@
...
+		if reauth { // If it is, pull the lock if we can, reauth
+			m := reauthLock.L.(*sync.Mutex)
+
+			if m.TryLock() { // Fetch the lock and try until we get auth.
+				for {
+					if ctx.Value("noPrompt") == nil {
+						_ = common.GetLifecycleMgr().Prompt("Authentication is required to continue the job. Reauthorize and continue?", common.PromptDetails{
+							PromptType: common.EPromptType.Reauth(),
+							ResponseOptions: []common.ResponseOption{
+								common.EResponseOption.Yes(),
```
gapra-msft, line 72: Why is Yes the only option here? Is it cause yes is really the only way to make sure the job completes (and presumably the customer wants that)?

ste/destReauthPolicy.go
```diff
@@ -0,0 +1,114 @@
...
+				debugCtx = context.WithValue(debugCtx, "reauthSrc", "InvalidAuthenticationInfo")
+			}
+		}
+
+		if reauth { // If it is, pull the lock if we can, reauth
+			m := reauthLock.L.(*sync.Mutex)
+
+			if m.TryLock() { // Fetch the lock and try until we get auth.
+				for {
+					if ctx.Value("noPrompt") == nil {
```
gapra-msft, line 68: Is this purely for test?

ste/destReauthPolicy.go
adreed-msft: Yes, precisely. "No, I don't want to reauth" does make sense as a potential option but given that we have other ways to cancel a job... meh?

ste/destReauthPolicy.go
adreed-msft: Yep. We want to test the policy, and not get caught up waiting for a response that can never come.

ste/destReauthPolicy.go
```diff
@@ -0,0 +1,114 @@
...
+var reauthLock *sync.Cond = sync.NewCond(&sync.Mutex{})
+
+func NewDestReauthPolicy(cred *common.ScopedAuthenticator) policy.Policy {
+	return &destReauthPolicy{cred}
+}
+
+func (d *destReauthPolicy) Do(req *policy.Request) (*http.Response, error) {
+retry:
+	ctx := req.Raw().Context()
+	debugCtx := context.WithValue(ctx, "destAuthDebug", true)
```
gapra-msft, line 38: make these constants

gapra-msft (approved):

dphulkar-msft:
There are some linting errors that need to be addressed. Could you please take a look and resolve them?

dphulkar-msft (approved):

vibhansa-msft (approved):

gapra-msft (approved):

PR #2894, opened 2024-12-17
Create missing resources in sync, allow blob to overlap

adreed-msft:
Creates parity with accidental functionality of blob, extends functionality on FNS

## Description

As it stands, when using a Blob destination, AzCopy is capable of creating a file if it is missing, as well as a directory if it is missing. However, if a directory stub exists, sync does not allow it to be overwritten, or overlapped, if using trailing slash directory stubs (e.g. `foo/`).

In addition, this PR extends the resource creation functionality to containers, shares, and filesystems.

- **Feature / Bug Fix**: Fleshing out the result of undefined behavior with Sync, creating parity with other destination resources.

- **Related Links**:
- Email Subject `Re: SciClone/AzCopy issues summary and resolution plan`
- ADO `#30057272`

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Both sides of the feature, blob overlapping and resource creation, are exhaustively tested with two new E2E tests, covering their relevant scenarios.


cmd/zc_traverser_blob.go
```diff
@@ -71,14 +71,25 @@ type blobTraverser struct {
...
+func (t *blobTraverser) IsDirectory(isSource bool) (isDirectory bool, err error) {
 	isDirDirect := copyHandlerUtil{}.urlIsContainerOrVirtualDirectory(t.rawURL)
 
 	// Skip the single blob check if we're checking a destination.
 	// This is an individual exception for blob because blob supports virtual directories and blobs sharing the same name.
 	// On HNS accounts, we would still perform this test. The user may have provided directory name without path-separator
 	if isDirDirect { // a container or a path ending in '/' is always directory
-		return true, nil
+		// If it's a container, let's ensure that container exists. Listing is a safe assumption to be valid, because how else would we enumerate?
+		blobURLParts, err := blob.ParseURL(t.rawURL)
```
gapra-msft, line 84: Do we also want this to hit if the URL is a virtual directory? I think isDirDirect could also be true for virtual directories.

adreed-msft: Good catch, this was lazy 4 am coding; I thought about this directly and then proceeded to not check.

cmd/syncEnumerator.go
```diff
@@ -93,11 +97,51 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
...
-		return nil, errors.New("trying to sync between different resource types (either file <-> directory or directory <-> file) which is not allowed." +
-			"sync must happen between source and destination of the same type, e.g. either file <-> file or directory <-> directory." +
-			"To make sure target is handled as a directory, add a trailing '/' to the target.")
+	destIsDir, err := destinationTraverser.IsDirectory(true)
+
+	var resourceMismatchError = errors.New("trying to sync between different resource types (either file <-> directory or directory <-> file) which is not allowed." +
+		"sync must happen between source and destination of the same type, e.g. either file <-> file or directory <-> directory." +
+		"To make sure target is handled as a directory, add a trailing '/' to the target.")
+
+	if cca.fromTo.To() == common.ELocation.Blob() || cca.fromTo.To() == common.ELocation.BlobFS() {
```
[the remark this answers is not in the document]
adreed-msft: Nope! We want all blob based targets to hit this case, bypassing the check for equivalent resource types.

cmd/zc_traverser_blob.go
adreed-msft: No, this was not. If it is a vdir, we should still validate the container exists.

cmd/sync.go
```diff
@@ -849,6 +852,7 @@ func init() {
 	syncCmd.PersistentFlags().StringVar(&raw.trailingDot, "trailing-dot", "", "'Enable' by default to treat file share related operations in a safe manner. Available options: "+strings.Join(common.ValidTrailingDotOptions(), ", ")+". "+
 		"Choose 'Disable' to go back to legacy (potentially unsafe) treatment of trailing dot files where the file service will trim any trailing dots in paths. This can result in potential data corruption if the transfer contains two paths that differ only by a trailing dot (ex: mypath and mypath.). If this flag is set to 'Disable' and AzCopy encounters a trailing dot file, it will warn customers in the scanning log but will not attempt to abort the operation."+
 		"If the destination does not support trailing dot files (Windows or Blob Storage), AzCopy will fail if the trailing dot file is the root of the transfer and skip any trailing dot paths encountered during enumeration.")
+	syncCmd.PersistentFlags().BoolVar(&raw.includeRoot, "include-root", false, "Disabled by default. Enable to include the root directory's properties when ")
```
gapra-msft, line 855: This sentence is incomplete.

cmd/syncEnumerator.go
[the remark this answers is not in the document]
gapra-msft: This is a nit comment so not a big deal either way. I think my original comment was misunderstood - Should we make the if condition 
_**err != nil &&**_ cca.fromTo.To() == common.ELocation.Blob() || cca.fromTo.To() == common.ELocation.BlobFS()

Because in the code below we are checking HasCode ContainerNotFound - which will only be true if err != nil anyway. Just seeing if we can make the code somewhat similar to the Azure Files handling below it.

cmd/zc_traverser_blob.go
```diff
@@ -71,14 +71,28 @@ type blobTraverser struct {
...
+		return false, err
+	}
+
 	// Skip the single blob check if we're checking a destination.
 	// This is an individual exception for blob because blob supports virtual directories and blobs sharing the same name.
 	// On HNS accounts, we would still perform this test. The user may have provided directory name without path-separator
 	if isDirDirect { // a container or a path ending in '/' is always directory
-		return true, nil
+		if blobURLParts.ContainerName != "" && blobURLParts.BlobName == "" {
+			// If it's a container, let's ensure that container exists. Listing is a safe assumption to be valid, because how else would we enumerate?
```
gapra-msft, line 89: Might be good to also note that GetProperties is not being called because there is a possibility of not having those specific permissions.

e2etest/zt_newe2e_fns_dir_test.go
```diff
@@ -202,3 +202,168 @@ func (*FNSSuite) Scenario_SyncTrailingSlashDeletion(a *ScenarioVariationManager)
...
+
+	ValidateResource(a, dst, ResourceDefinitionContainer{
+		Objects: ObjectResourceMappingFlat(JoinMap(dstMap, srcMap)),
+	}, false)
+}
+
+/*
+This tests that sync is capable of creating the destination resource if it does not exist, including the root container.
+*/
+//func (*FNSSuite) Scenario_OverlapWithTrailingSlash(a *ScenarioVariationManager) {
```
gapra-msft, line 278: Should this be uncommented?

gapra-msft (commented):
looks good. just had one question about the commented test

cmd/sync.go
adreed-msft: oooooooooauuuugghhhhhhhh

e2etest/zt_newe2e_fns_dir_test.go
adreed-msft: It should be deleted, it was leftover from a prior iteration of testing on this PR.

cmd/zc_traverser_blob.go
adreed-msft: Hmmmm. Permissions are a good thought there, but that wasn't actually a part of the original thought.

cmd/syncEnumerator.go
```diff
@@ -93,11 +97,51 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
...
+			- We are a tool, tools can be misused. It is up to the customer to validate everything they intend to do.
+		*/
+
+		if bloberror.HasCode(err, bloberror.ContainerNotFound) { // We can resolve a missing container. Let's create it.
+			bt := destinationTraverser.(*blobTraverser)
+			sc := bt.serviceClient                                                   // it being a blob traverser is a relatively safe assumption, because
+			bUrlParts, _ := blob.ParseURL(bt.rawURL)                                 // it should totally have succeeded by now anyway
+			_, err = sc.NewContainerClient(bUrlParts.ContainerName).Create(ctx, nil) // If it doesn't work out, this will surely bubble up later anyway. It won't be long.
+			if err != nil {
+				glcm.Info(fmt.Sprintf("Failed to create the missing destination container: %v", err))
```
dphulkar-msft, line 129: Should we treat it as a warning message? Just a thought

cmd/syncEnumerator.go
```diff
@@ -93,11 +97,51 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
...
 	// verify that the traversers are targeting the same type of resources
 	sourceIsDir, _ := sourceTraverser.IsDirectory(true)
-	destIsDir, _ := destinationTraverser.IsDirectory(true)
-	if sourceIsDir != destIsDir {
-		return nil, errors.New("trying to sync between different resource types (either file <-> directory or directory <-> file) which is not allowed." +
-			"sync must happen between source and destination of the same type, e.g. either file <-> file or directory <-> directory." +
-			"To make sure target is handled as a directory, add a trailing '/' to the target.")
+	destIsDir, err := destinationTraverser.IsDirectory(true)
+
+	var resourceMismatchError = errors.New("trying to sync between different resource types (either file <-> directory or directory <-> file) which is not allowed." +
```
dphulkar-msft, line 102: Should we define this error statement as constant?

dphulkar-msft (commented):
"Looks good to me! However, it seems that some tests are failing. Could you please take a look?"

cmd/syncEnumerator.go
adreed-msft: Probably, yes.

cmd/syncEnumerator.go
adreed-msft: Also probably yes

adreed-msft: actually idk. It's not relied upon anywhere, and it's not really used outside of this function.

cmd/syncEnumerator.go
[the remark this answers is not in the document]
adreed-msft: ig, that makes sense.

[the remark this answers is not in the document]
adreed-msft: wait actually no-- this is to prevent the check for a resource mismatch

vibhansa-msft (approved):

gapra-msft (approved):

Issue #2685, opened 2024-05-17
Unable to delete the files not completed automaticly when fileshare is full.


v-youxie:
### Which version of the AzCopy was used? 
10.24.0
### Which platform are you using? (ex: Windows, Mac, Linux)
Windows

##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.
.\azcopy.exe copy "C:\Users\v-youxie\Downloads\123" "https://xie********.file.core.chinacloudapi.cn/testing?sv**********************************************" --preserve-smb-permissions=true --preserve-smb-info=true --recursive

### What problem was encountered?
I have an Azure fileshare with 1GB quota. While the whole size of the source directory is 1.2GB（11 files）
When uploading source directory to Azure fileshare. The process ended witwithin 10s（obviously no file transferred completly）. The file transfers completed is 0. But in Azure fileshare ,some files generated. They consume capacity，but ccould not be used cause not completed.

![image](https://github.com/Azure/azure-storage-azcopy/assets/170086424/558718c3-548f-4e90-a50d-38b1a8f9f39f)
![image](https://github.com/Azure/azure-storage-azcopy/assets/170086424/9b2de5b8-a42a-4dc1-84d1-08e3bef4fef2)

In more tests, the file  transfers completed changes randomly. Still a lot of uncompleted files generated in Azure fileshare.
![image](https://github.com/Azure/azure-storage-azcopy/assets/170086424/2e5ff82a-e4b8-40dc-8fd5-c42115359ba1)
![image](https://github.com/Azure/azure-storage-azcopy/assets/170086424/4593b437-0749-4986-9a64-7c69827f6498)

First of all ,  uncompleted files need to be deleted automaticly to keep the consistency.
Secondly, when the source size is larger than fileshare quota， need to have a certain policy，no matter fullfilling the fileshare or uploading nothing，rather than uploading some of the files randomly.

### How can we reproduce the problem in the simplest way?
Create a Azuzre fileshare，change the quota to a small number.
Upload a directory larger than fileshare with azcopy.

### Have you found a mitigation/solution?
no

adreed-msft:
Normally, we would delete failed objects. In this case though, it seems AzCopy got back a 403... Typically, we interpret that as the literal meaning of the status code (forbidden, access denied), and exit the transfer early. In this case, this isn't the correct behavior. This is a bug. Thanks for reporting this!

[...]

PR #2895, opened 2024-12-17
Fix for File Share Limit Reached Error

wonwuakpa-msft:
## Description

- Changes:
   - Added nil and closure checks to ensure there's no `panic` from closing channels. 
   - Added a description for the `ShareSizeLimitReached` error to guide users on how to handle the error
   - Added test to catch `ShareSizeLimitReached` error

User reported this [AzCopy behavior](https://github.com/Azure/azure-storage-azcopy/issues/2685) when uploading data to a file share that exceeds the quota (i.e uploading 4GB to a 1GB file share). When the quota is hit, the ongoing job is stopped and the already uploaded files are not deleted. [AzCopy team](https://msazure.visualstudio.com/One/_backlogs/backlog/XStore-DevExp/Epics?showParents=true&workitem=30212338) concluded this is not a bug. There is a high chance of unnecessary failed API calls after the `ShareSizeLimitReached` error is hit. Thus, the current behavior of stopping the job is valid.

- **Related Links**:
- [Issues](https://github.com/Azure/azure-storage-azcopy/issues/2685)
- Email: Expected AzCopy Behavior When Exceeding Quotas

## Type of Change

- [X] Bug fix
- [ ] New feature
- [X] Documentation update required
- [ ] Other (describe):

## How Has This Been Tested?

Existing e2e tests and new `Scenario_UploadFileQuota` test
AzCopy randomly uploads files to the file share until the quota is hit. Wrote a script to validate that these uploaded files are non-empty. If empty files are uploaded, they are correctly deleted in the cleanup handler in the codebase. 


cmd/make.go
```diff
@@ -208,6 +208,6 @@ func init() {
 		},
 	}
 
-	makeCmd.PersistentFlags().Uint32Var(&rawArgs.quota, "quota-gb", 0, "Specifies the maximum size of the share in gigabytes (GiB), 0 means you accept the file service's default quota.")
+	makeCmd.PersistentFlags().Uint32Var(&rawArgs.quota, "quota-gb", 0, "Specifies the maximum size of the share in gigabytes (GiB), 0 means you accept the file service's default quota. If `ShareSizeLimitReached` error is hit, increase quota and call resume command.")
```
gapra-msft, line 211: This doc will only be seen by customers if they happen to be creating the share with AzCopy and they happen to be reading the help docs. 

Would it make more sense to catch this error in ste and print a helpful log message?

wonwuakpa-msft: It seems like this specific error is not included in the AzCopy or Azure SDK for Go. I opened an [ issue ](https://github.com/Azure/azure-sdk-for-go/issues/23833) to address this, it has not been merged in yet

gapra-msft: I dont think you need the specific error in the SDK, temporarily you can just use a string as an input to the HasErrorCode function.

ste/mgr-JobPartTransferMgr.go
```diff
@@ -864,6 +864,10 @@ func (jptm *jobPartTransferMgr) failActiveTransfer(typ transferErrorCode, descri
 				common.GetLifecycleMgr().Info(fmt.Sprintf("Authentication failed, it is either not correct, or expired, or does not have the correct permission %s", err.Error()))
 			}
 
+			if serviceCode == "ShareSizeLimitReached" {
```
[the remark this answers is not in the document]
gapra-msft: Also in the test you should be able to set a small file share quota I think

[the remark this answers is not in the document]
wonwuakpa-msft: Yep, I wrote `Scenario_UploadFilesWithQuota` test with 10MB quota

e2etest/zt_newe2e_file_test.go
```diff
@@ -546,3 +546,53 @@ func (s *FileTestSuite) Scenario_CopyTrailingDotUnsafeDestination(svm *ScenarioV
...
+				CopySyncCommonFlags: CopySyncCommonFlags{
+					Recursive:   pointerTo(true),
+					BlockSizeMB: pointerTo(4.0),
+				},
+			},
+		})
+	}
+
+	// Validate uploaded files are not empty
+	ValidateResource[ContainerResourceManager](svm, shareResource, ResourceDefinitionContainer{
```
gapra-msft, line 595: Shouldnt the share quota error be hit? Can we validate that in the output?

gapra-msft: Also if the share quota is hit, shouldnt some of the files not exist?

e2etest/zt_newe2e_file_test.go
```diff
@@ -546,3 +546,53 @@ func (s *FileTestSuite) Scenario_CopyTrailingDotUnsafeDestination(svm *ScenarioV
 		Body: body,
 	}, false)
 }
+
+// Test empty files are not uploaded when File share quota is hit
+func (s *FileTestSuite) Scenario_UploadFilesWithQuota(svm *ScenarioVariationManager) {
+	quotaMB := int32(10) // 10 MB quota
```
gapra-msft, line 552: quota is in GB

adreed-msft: ditto

e2etest/zt_newe2e_file_test.go
```diff
@@ -546,3 +546,53 @@ func (s *FileTestSuite) Scenario_CopyTrailingDotUnsafeDestination(svm *ScenarioV
...
+	shareName := "testsharewithquota"
```
adreed-msft, line 553: consistent share names can cause problems when testing with static accts, using random names is better; using nil in the ContainerName field automatically creates a random name.

e2etest/zt_newe2e_file_test.go
```diff
@@ -546,3 +546,53 @@ func (s *FileTestSuite) Scenario_CopyTrailingDotUnsafeDestination(svm *ScenarioV
...
+				Quota: &quotaMB},
+		},
+	})
+
+	fileSizeMB := int64(6)
+	fileNames := []string{"file_1.txt", "file_2.txt", "file_3.txt"}
+
+	// Create src obj mapping
+	srcObjs := make(ObjectResourceMappingFlat)
+	// Creat source files
```
adreed-msft, line 567: nit: "Create"

e2etest/zt_newe2e_file_test.go
```diff
@@ -546,3 +546,53 @@ func (s *FileTestSuite) Scenario_CopyTrailingDotUnsafeDestination(svm *ScenarioV
...
+	shareName := "testsharewithquota"
+	shareResource := CreateResource[ContainerResourceManager](svm, GetRootResource(svm, common.ELocation.File()), ResourceDefinitionContainer{
+		ContainerName: &shareName,
+		Properties: ContainerProperties{
+			FileContainerProperties: FileContainerProperties{
+				Quota: &quotaMB},
+		},
+	})
+
+	fileSizeMB := int64(6)
```
adreed-msft, line 562: Since quotas are in gigs, object sizes in megs won't fill a share. To reduce runtime, using a static container that's been pre-filled, or mocking the test may help.

Mocking is probably not feasible within the E2E tests since we don't have a way to spin up a mock endpoint... *todo*.

So, if we went the mock route, a unit test may make more sense.

e2etest/zt_newe2e_file_test.go
adreed-msft, line 595: This should probably fail if the intended error is hit.

adreed-msft (changes requested):
Test doesn't adequately test the change in question

e2etest/zt_newe2e_file_test.go
wonwuakpa-msft: @adreed-msft  I reduced the quota to 1GB to reduce runtime, and number of files to upload to just two files

e2etest/newe2e_task_runazcopy_login_logout.go
```diff
@@ -69,6 +69,15 @@ func RunAzCopyLoginLogout(a Asserter, verb AzCopyVerb) AzCopyStdout {
 	}
 
 	err = command.Wait()
+
+	err = command.Wait()
```
gapra-msft, line 73: this is not needed since its just repeating above right?

adreed-msft (commented):
Looks good-- One concern about test efficiency that I'd like you to take a proper look at, but worst case it can be a remediation item next release.

adreed-msft (approved):

dphulkar-msft (approved):

PR #2919, opened 2025-01-21
Some recommended edits

normesta:
Norm from the doc team. Now that the AzCopy reference content is being removed from docs in favor of being hosted in GitHub, I just wanted to perform a final editorial pass on the content. After reviewing the ref content, it appears I've done this before so there's not a whole lot of new recommendations.

Because AzCopy reference content will no longer appear as official Microsoft Docs pages, there's a bit more flexibility on phrasing. I'm offering up some suggestions in this PR. If you don't want to accept any given suggestion, please let me know and I'll update the PR to exclude the change. Here's some notes about some of these proposed updates:

ADLS Gen 2 is something we have scrubbed out of all Data Lake Storage conceptual docs as the term is no longer recommended as a means to describe the offering as Data Lake Storage Gen 1 is no longer available. The preferred way to refer to this set of capabilities is Azure Data Lake Storage on first mention and then Data Lake Storage for each mention thereafter.

Recommend replacing most instances of the term authentication with the term authorization. In docs, we use the term authorization to describe the granting of permission by way of a token, key, or identity as the task is authorizing access based on allowed permissions versus the action of verifying that an identity is valid.

The official way in docs of referring to OAuth auth is Microsoft Entra ID authorization. Admittedly OATH tokens are passed as part of this form of authorization, but if the intent is to describe which type of authorization is used (Key, SAS, or by way of a user or managed identity), then the branding term for that method of authorization is Microsoft Entra ID. Also suggest leading with the preferred methods first as a matter of promoting good security practices. For example, listing Entra ID before Shared Key etc.

Also noticed that the vast majority of the examples append a SAS token. Given complete support for Microsoft Entra ID and that Entra ID is the preferred method of authorization, I suggest adding text to the beginning that all examples assume that you've authorized by using Entra ID. Then, append a SAS token only to those examples that require one (For example on the source of a copy operation). This PR does not implement that suggestion.


cmd/benchmark.go
```diff
@@ -331,14 +331,14 @@ func init() {
...
-	benchCmd.PersistentFlags().UintVar(&raw.fileCount, common.FileCountParam, common.FileCountDefault, "number of auto-generated data files to use")
+	benchCmd.PersistentFlags().StringVar(&raw.sizePerFile, common.SizePerFileParam, "250M", "Size of each auto-generated data file. Must be "+sizeStringDescription)
+	benchCmd.PersistentFlags().UintVar(&raw.fileCount, common.FileCountParam, common.FileCountDefault, "Number of auto-generated data files to use")
 	benchCmd.PersistentFlags().UintVar(&raw.numOfFolders, "number-of-folders", 0, "If larger than 0, create folders to divide up the data.")
-	benchCmd.PersistentFlags().BoolVar(&raw.deleteTestData, "delete-test-data", true, "if true, the benchmark data will be deleted at the end of the benchmark run.  Set it to false if you want to keep the data at the destination - e.g. to use it for manual tests outside benchmark mode")
+	benchCmd.PersistentFlags().BoolVar(&raw.deleteTestData, "delete-test-data", true, "if true, then the benchmark data will be deleted at the end of the benchmark run.  Set it to false if you want to keep the data at the destination - e.g. to use it for manual tests outside benchmark mode")
```
gapra-msft, line 337:
```suggestion
	benchCmd.PersistentFlags().BoolVar(&raw.deleteTestData, "delete-test-data", true, "If true, then the benchmark data will be deleted at the end of the benchmark run.  Set it to false if you want to keep the data at the destination - e.g. to use it for manual tests outside benchmark mode")
```

gapra-msft:
@normesta Could you make these edits against main? We no longer develop against the dev branch.

normesta:
@gapra-msft - Thx! I was unaware that we could create PRs directly to main. I think at one point I attempted to do that and was told to use the dev branch. I will submit the edits against Main. Thank you!

wonwuakpa-msft:
@normesta Could you link that PR here for review, thanks!

normesta:
Hi @wonwuakpa-msft - content was merged with Main some time back, so there's nothing more to review. Thank you though!

Issue #2489, opened 2023-12-08
Incorrect handling of EINTR in syscalls


ahoenselaar:
##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.

cp [REMOTE_DIR] /tmp/tmpdir

### What problem was encountered?
```
ERR: [P#0-T#0] DOWNLOADFAILED: [REMOTE_FILE] : 000 : File Creation Error interrupted system call
   Dst: /tmp/tmpdir/tmpfile
```

This is likely caused by incorrect handling of EINTR for syscalls, which should be retried, e.g. [here](https://github.com/Azure/azure-storage-azcopy/blob/f84cc60f2b21ef18721375de5dbecfd76630f95d/common/writeThoughFile_linux.go#L183).

### Have you found a mitigation/solution?
No.

vibhansa-msft:
Can you share more details on how you are able to create this scenario or why you feel fallocate is getting EINTR?

ovidiu10:
Hi @gapra-msft & @vibhansa-msft  
we have another customer which encounter same issue 

2024/10/07 11:30:38 AzcopyVersion  10.26.0
2024/10/07 11:30:38 OS-Environment  linux
2024/10/07 11:30:38 OS-Architecture  amd64
2024/10/07 11:30:38 Log times are in UTC. Local time is 7 Oct 2024 11:30:38
2024/10/07 11:30:38 ISO 8601 START TIME: to copy files that changed before or after this job started, use the parameter --include-before=2024-10-07T11:30:33Z or --include-after=2024-10-07T11:30:33Z
2024/10/07 11:30:38 Any empty folders will not be processed, because --recursive was not specified
2024/10/07 11:30:38 Job-Command cp https://XXXXX.file.core.windows.net/YYYYYY-YYYY-.../build/dce-abyss-ui.ui.config-2024.1007.1102421.0004801006.zip?sig=-REDACTED-&st=2024-10-07t10%3a32%3a52z&se=2025-04-07t11%3a02%3a52z&sv=2017-04-17&spr=https&sp=r&sr=f /tmp/AAAAA/chef/cache/cookbooks/installClientPackages/content-package-25.zip 
2024/10/07 11:30:38 Number of CPUs: 8
2024/10/07 11:30:38 Max file buffer RAM 4.000 GB
[...]

PR #2942, opened 2025-02-10
Fixing interrputed system call errors

wonwuakpa-msft:
## Description

  - Retry system call functions that return POSIX `EINTR` error

- Addresses error handling for interrupted file creation system call function

- **Related Links**:
- [Issues](https://github.com/Azure/azure-storage-azcopy/issues/2489)

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Added a test to ensure that no `EINTR` errors are returned in appropriate system calls on Linux


common/writeThoughFile_linux.go
```diff
@@ -180,7 +180,13 @@ func CreateFileOfSizeWithWriteThroughOption(destinationPath string, fileSize int
 		return f, err
 	}
 
-	err = syscall.Fallocate(int(f.Fd()), 0, 0, fileSize)
+	for { // keep retrying when EINTR error occurs
+		err = syscall.Fallocate(int(f.Fd()), 0, 0, fileSize)
```
vibhansa-msft, line 184: If the fallocate fails and we keep running this in a loop, we might overrun the CPU here.
Also, if the fallocate is failing due to any disk full error or some fatal error we shall not retry.

vibhansa-msft: There shall be some sort of max retry count here, say we retry for 5 times and then eventually fail this.

wonwuakpa-msft: Yes, I can cap the number of retries to a threshold. `EINTR` error is only returned when there is an interruption to a system call by a signal

common/writeThoughFile_linux.go
```diff
@@ -180,7 +180,16 @@ func CreateFileOfSizeWithWriteThroughOption(destinationPath string, fileSize int
 		return f, err
 	}
 
-	err = syscall.Fallocate(int(f.Fd()), 0, 0, fileSize)
+	for i := 0; i < 5; i++ { // // Perform up to 5 EINTR error retries
+		err = syscall.Fallocate(int(f.Fd()), 0, 0, fileSize)
+		if err == nil {
```
dphulkar-msft, line 185: You can use a single if statement for err==nil and err!=syscall.EINTR with an OR condition.

common/writeThoughFile_linux.go
```diff
@@ -180,7 +180,16 @@ func CreateFileOfSizeWithWriteThroughOption(destinationPath string, fileSize int
 		return f, err
 	}
 
-	err = syscall.Fallocate(int(f.Fd()), 0, 0, fileSize)
+	for i := 0; i < 5; i++ { // // Perform up to 5 EINTR error retries
```
dphulkar-msft, line 183: You can declare 5 as a constant named EINTR_RETRY_COUNT. Something like this.

dphulkar-msft:
Is it possible to add a UT or E2E test for this change?

vibhansa-msft (approved):

e2etest/zt_newe2e_file_test.go
```diff
@@ -604,3 +604,26 @@ func (s *FileTestSuite) Scenario_UploadFilesWithQuota(svm *ScenarioVariationMana
...
+		GetRootResource(svm, common.ELocation.Local()),
+		ResourceDefinitionObject{Body: body})
+
+	stdOut, _ := RunAzCopy(svm, AzCopyCommand{
+		Verb:       AzCopyVerbCopy,
+		Targets:    []ResourceManager{srcObj, destObj},
+		ShouldFail: false,
+	})
+
+	//ValidateResource[ObjectResourceManager](svm, destObj, ResourceDefinitionObject{
```
dphulkar-msft, line 624: If not required we can remove this.

e2etest/newe2e_task_validation.go
```diff
@@ -217,6 +217,20 @@ func ValidateContainsError(a Asserter, stdout AzCopyStdout, errorMsg []string) {
...
+func ValidateDoesNotContainError(a Asserter, stdout AzCopyStdout, errorMsg []string) {
+	if dryrunner, ok := a.(DryrunAsserter); ok && dryrunner.Dryrun() {
+		return
+	}
+	for _, line := range stdout.RawStdout() {
+		if !checkMultipleErrors(errorMsg, line) {
+			return
+		}
+	}
+	fmt.Println(stdout.String())
```
dphulkar-msft, line 230: We can remove this

common/writeThroughFile_test.go
```diff
@@ -65,3 +68,29 @@ func TestCreateParentDirectoryIfNotExist(t *testing.T) {
...
+	_, err := common.CreateFileOfSizeWithWriteThroughOption(destinationPath, 1,
+		false,
+		tracker,
+		false)
+
+	if err != nil {
+		a.NotEqual(syscall.EINTR, err)
+	} else {
+		a.NoError(err, fmt.Sprintf("Error creating file: %v", err))
+	}
```
dphulkar-msft, line 94:
```suggestion
	if err != nil {
		a.NotEqual(syscall.EINTR, err)
		return
	} 
	a.NoError(err, fmt.Sprintf("Error creating file: %v", err))
	
```

e2etest/newe2e_task_validation.go
wonwuakpa-msft: I wrote this so we can validate that some tests do not error out. In this case, it is used to check for the absence of EINTR errors

e2etest/zt_newe2e_file_test.go
wonwuakpa-msft: I added a unit and e2e test for robustness, should I have one or the other?

dphulkar-msft (approved):

dphulkar-msft:
@wonwuakpa-msft, please trigger the tests from the pipeline. Once all tests have passed successfully, we can proceed with merging this.

PR #2948, opened 2025-02-20
Azcopy memory leak issue starting from Go version 1.22.8 onwards.

dphulkar-msft:
## Description

AzCopy doesn't set MaxConnsPerHost, so net/http may try to create too many connections at once. Before Go [CL 576555](https://go-review.googlesource.com/c/go/+/576555), canceled requests would also cancel connection attempts, causing issues. The CL change keeps connections attempting even if the request is canceled. Setting MaxConnsPerHost to asame as AzCopy use to set the rate limiter semaphone value helped control the number of connections and removed the need for AzCopy's semaphore.

The solution provided by the Go team can be found in the following GitHub issue: https://github.com/golang/go/issues/71308.

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Tested multiple combinations to ensure that AzCopy does not exceed the memory limit:

- Go version 1.24.0, with MaxConnsPerHost set to concurrentDialsPerCpu (i.e., 10) multiplied by runtime.NumCPU().
- Go version 1.24.0, with MaxConnsPerHost set to concurrentDialsPerCpu (i.e., 10) multiplied by runtime.NumCPU(), and the rate limiter logic removed from AzCopy’s code.

In both cases, memory usage remained stable, similar to the behavior observed with Go version 10.22.7.

We are also incorporating memory testing into our release process to ensure consistent performance and stability across versions.


ste/mgr-JobPartMgr.go
```diff
@@ -80,14 +78,11 @@ type IJobPartMgr interface {
...
 	return &http.Client{
 		Transport: &http.Transport{
-			Proxy: common.GlobalProxyLookup,
-			DialContext: newDialRateLimiter(&net.Dialer{
-				Timeout:   30 * time.Second,
-				KeepAlive: 30 * time.Second,
-				DualStack: true,
-			}).DialContext,
+			Proxy:                  common.GlobalProxyLookup,
+			MaxConnsPerHost:        concurrentDialsPerCpu * runtime.NumCPU(),
```
qmuntal, line 85: Not 100% sure, but `MaxConnsPerHost` seems like the perfect fit for `AZCOPY_CONCURRENCY_VALUE` rather than having a hardcoded value.

vibhansa-msft: We do not want to change our current behavior drastically as it may break some of the running scenarios. Changes proposed in the PR appears to be a safer choice to me.

vibhansa-msft (approved):

wonwuakpa-msft (approved):

PR #2950, opened 2025-02-21
Introduced NFS specific flags

dphulkar-msft:
## Description

Added NFS specific flags for NFS support.

- **--nfs** : User will have to specify this flag if they want to perform azcopy copy/sync for NFS share. 
- **--preserve-nfs-info**: User will have to specify this flag if they want to preserve the nfs properties. (Properties list TBD with files team) 
- **--preserve-nfs-permissions**: User will have to specify this flag if they want to preserve nfs permissions. (Permissions list TBD with files team) 

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## How Has This Been Tested?

E2ETest or UT N/A right now as it is just flags addition. 


cmd/constants.go
```diff
@@ -0,0 +1,9 @@
+package cmd
+
+// Common Error and Info messages
+const (
+	PreservePOSIXPropertiesErrMsg = "To use --preserve-posix-properties, both the source and destination must be POSIX-aware. Valid combinations are: Linux -> Blob, Blob -> Linux, or Blob -> Blob."
+	PreservePermissionsInfoMsg    = "Note: The preserve-permissions flag is set to false. As a result, AzCopy will not copy SMB ACLs between the source and destination. For more information, visit: https://aka.ms/AzCopyandAzureFiles."
+
+	PreserveNFSPermissions = "Note: The preserve-nfs-permissions flag is set to false. As a result, AzCopy will not copy NFS permissions between the source and destination."
```
adreed-msft, line 8: Nit: Name should be `PreserveNFSPermissionsDisabledMessage`

cmd/constants.go
```diff
@@ -0,0 +1,9 @@
+package cmd
+
+// Common Error and Info messages
+const (
+	PreservePOSIXPropertiesErrMsg = "To use --preserve-posix-properties, both the source and destination must be POSIX-aware. Valid combinations are: Linux -> Blob, Blob -> Linux, or Blob -> Blob."
+	PreservePermissionsInfoMsg    = "Note: The preserve-permissions flag is set to false. As a result, AzCopy will not copy SMB ACLs between the source and destination. For more information, visit: https://aka.ms/AzCopyandAzureFiles."
```
adreed-msft, line 6: Nit: should indicate disabled in the naming scheme

cmd/copy.go
```diff
@@ -187,6 +187,13 @@ type rawCopyCmdArgs struct {
 
 	// when specified, AzCopy deletes the destination blob that has uncommitted blocks, not just the uncommitted blocks
 	deleteDestinationFileIfNecessary bool
+
+	// Opt-in flag to persist additional NFS properties to Azure Files
+	preserveNFSInfo bool
+	// Opt-in flag to persist additional NFS permissions to Azure Files
+	preserveNFSPermissions bool
```
adreed-msft, line 194: If we're already notating this an NFS copy, do we need separate flags for permissions? Can we potentially be preserving SMB permissions at the same time?

cmd/copy.go
```diff
@@ -187,6 +187,13 @@ type rawCopyCmdArgs struct {
 
 	// when specified, AzCopy deletes the destination blob that has uncommitted blocks, not just the uncommitted blocks
 	deleteDestinationFileIfNecessary bool
+
+	// Opt-in flag to persist additional NFS properties to Azure Files
+	preserveNFSInfo bool
```
adreed-msft, line 192: I'm debating if we shouldn't just merge `--preserve-smb-info`, `--preserve-posix-properties`, and `--preserve-nfs-info`. But that begets a similar question-- Can both NFS and SMB permissions be preserved at the same time? would a user want to do that?

cmd/copy.go
```diff
@@ -267,6 +274,81 @@ func (raw rawCopyCmdArgs) stripTrailingWildcardOnRemoteSource(location common.Lo
...
+//     if the flag is set to false and NFS info is being preserved.
+//   - Ensures that both source and destination locations are NFS-aware for relevant operations.
+//
+// If any of these validations fail, the function returns an error with the respective validation message.
+//
+// Returns:
+// - An error if any validation fails, otherwise nil indicating successful validation.
+func (raw rawCopyCmdArgs) performNFSSpecificValidation(cooked *CookedCopyCmdArgs) (err error) {
+	// TODO remove it
+	glcm.Info("NFS copy support addition in progress")
```
adreed-msft, line 293: todo: remove this

alternatively convert to a preview marker

cmd/copy.go
```diff
@@ -267,6 +274,81 @@ func (raw rawCopyCmdArgs) stripTrailingWildcardOnRemoteSource(location common.Lo
...
+	cooked.isNFSCopy = raw.isNFSCopy
+	cooked.preserveNFSInfo = raw.preserveNFSInfo && areBothLocationsNFSAware(cooked.FromTo)
```
adreed-msft, line 295: sidenote: I like this pattern for the "cook" step-- we should continue doing this in the future; separating individual steps into their own functions.

cmd/copy_helpers.go
```diff
@@ -0,0 +1,73 @@
...
+		(fromTo == common.EFromTo.LocalFile() || fromTo == common.EFromTo.FileLocal()) {
+		return true
+	} else if fromTo == common.EFromTo.FileFile() {
+		return true
+	} else {
+		return false
+	}
+}
+
+func areBothLocationsNFSAware(fromTo common.FromTo) bool {
```
adreed-msft, line 44: Code smell: we have two different places we validate this two different ways, here and validatePreserveNFSPropertyOption.

cmd/copy.go
```diff
@@ -267,6 +274,81 @@ func (raw rawCopyCmdArgs) stripTrailingWildcardOnRemoteSource(location common.Lo
...
+	if err = validatePreserveNFSPropertyOption(cooked.preserveNFSInfo, cooked.FromTo, "preserve-nfs-info"); err != nil {
```
adreed-msft, line 296: This will probably never error out, since this will disable itself even if explicitly set. This check is unnecessary.

cmd/copyEnumeratorInit.go
```diff
@@ -52,10 +52,15 @@ func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 		TrailingDot: cca.trailingDot,
 	}
 	jobPartOrder.CpkOptions = cca.CpkOptions
-	jobPartOrder.PreserveSMBPermissions = cca.preservePermissions
-	jobPartOrder.PreserveSMBInfo = cca.preserveSMBInfo
-	// We set preservePOSIXProperties if the customer has explicitly asked for this in transfer or if it is just a Posix-property only transfer
-	jobPartOrder.PreservePOSIXProperties = cca.preservePOSIXProperties || (cca.ForceWrite == common.EOverwriteOption.PosixProperties())
+	if cca.isNFSCopy {
```
adreed-msft, line 55: False by default, yet a flag behind it is true by default. I suspect this is intentional; iirc not every container can support NFS properties?

adreed-msft: Maybe isNFSCopy could be inferred if preserve nfs perms is set? (assuming we're separating things and not combining them)

cmd/sync.go
```diff
@@ -100,6 +100,9 @@ type rawSyncCmdArgs struct {
 
 	// when specified, AzCopy deletes the destination blob that has uncommitted blocks, not just the uncommitted blocks
 	deleteDestinationFileIfNecessary bool
+	preserveNFSInfo                  bool
```
adreed-msft, line 103: ditto on args here

cmd/sync.go
```diff
@@ -391,6 +381,79 @@ func (raw *rawSyncCmdArgs) cook() (cookedSyncCmdArgs, error) {
...
+//   - Validates the "preserve-nfs-info" flag to ensure it is set correctly for NFS-aware locations.
+//   - Validates the "preserve-nfs-permissions" flag, ensuring that user input is correct and provides feedback
+//     if the flag is set to false and NFS info is being preserved.
+//   - Ensures that both source and destination locations are NFS-aware for relevant operations.
+//
+// If any of these validations fail, the function returns an error with the respective validation message.
+//
+// Returns:
+// - An error if any validation fails, otherwise nil indicating successful validation.
+func (raw rawSyncCmdArgs) performNFSSpecificValidation(cooked *cookedSyncCmdArgs) (err error) {
```
adreed-msft, line 399: Could this be more DRY if we're doing other code cleanup?

cmd/syncEnumerator.go
```diff
@@ -222,6 +223,8 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 		FileAttributes: common.FileTransferAttributes{
 			TrailingDot: cca.trailingDot,
 		},
+		PreserveNFSPermissions: cca.preserveNFSPermisssions,
```
adreed-msft, line 226: This is notably not behind the isNFS flag like copy, is this intentional?

ste/mgr-JobPartTransferMgr.go
```diff
@@ -142,6 +143,9 @@ type TransferInfo struct {
 
 	VersionID  string
 	SnapshotID string
+
+	PreserveNFSPermissions bool
+	PreserveNFSInfo        bool
```
adreed-msft, line 148: ditto

adreed-msft (changes requested):

cmd/copy.go
dphulkar-msft: As per the requirements provided by the Files team, we are tasked with implementing two separate flags to preserve different sets of information.
Each flag should function independently, with the appropriate data being preserved based on the flag that is set to true.
Details are mentioned in this doc. https://microsoftapc-my.sharepoint.com/:w:/g/personal/kviswanath_microsoft_com/EXdKRKQzDDRBkB5fzwhofPMBy1S4U-q6YeAa9Dp-S5NSPQ?e=LzcEIT

cmd/copy.go
dphulkar-msft: No. There cannot be a case wherein a user can preserve both smb and nfs permissions both at the same time.
The valid combinations list is also provided by the Files team in the same doc shared above.
 To enforce this, I will add a validation check: if a user provides both --preserve-permissions and --preserve-nfs-permissions at the same time, AzCopy will return an error same for smb-info and nfs-info.

cmd/copy.go
dphulkar-msft: removed

cmd/copy_helpers.go
dphulkar-msft: removing areBothLocationsNFSAware() as the code logic is duplicated.

cmd/syncEnumerator.go
dphulkar-msft: isNFS check not required here. As we are only passing the values off.

ste/mgr-JobPartTransferMgr.go
dphulkar-msft: isNFS check not required here. As we are only passing the values off.

cmd/sync.go
dphulkar-msft: Yes functionally we are doing the same. But the objects passed rawSyncCmdArgs and cookedSyncCmdArgs are different for which I had to introduce 2 different with same name.

adreed-msft: Uhhhh... Tell ya what. I have an idea but I don't think I can express it well enough over text/in the code suggestion feature. Lemme pop open another PR based on this branch. This'll be a tomorrow item on my side, I'm outta energy tonight.

cmd/copy.go
dphulkar-msft: removing this check areBothLocationsNFSAware() and keeping the validatePreserveNFSPropertyOption() logic

cmd/copy.go
```diff
@@ -267,6 +274,81 @@ func (raw rawCopyCmdArgs) stripTrailingWildcardOnRemoteSource(location common.Lo
...
+// - Validates the preservation of file owner information based on user flags.
+//
+// If any of these validations fail, the function returns an error with the respective validation message.
+//
+// Returns:
+// - An error if any validation fails, otherwise nil indicating successful validation.
+
+func (raw rawCopyCmdArgs) performSMBSpecificValidation(cooked *CookedCopyCmdArgs) (err error) {
+	cooked.preserveSMBInfo = raw.preserveSMBInfo && areBothLocationsSMBAware(cooked.FromTo)
+	if err = validatePreserveSMBPropertyOption(cooked.preserveSMBInfo, cooked.FromTo, "preserve-smb-info"); err != nil {
```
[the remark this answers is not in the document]
dphulkar-msft: removing this check areBothLocationsSMBAware() and keeping the validatePreserveSMBPropertyOption() logic

dphulkar-msft:
New PR : https://github.com/Azure/azure-storage-azcopy/pull/2968

PR #2961, opened 2025-02-27
Fix incorrect count of pending transfers when a job is resumed 

wonwuakpa-msft:
## Description

Customers identified a bug with resume. When resuming a failed with `CompletedWithErrors` job, the reporting metrics (pending transfers, failed transfers, total percent completed) are incorrect. The total percent completed surpasses 100%, the failed transfers is not reset to 0 and pending transfers value is greater than total.

  -  When a job is resumed, the failed transfers reset to 0 
  -  No underflow in the number of pending transfers (from uint type)
  -  No double counting of bytes transferred in the total percent calculation
  -  No undercounting of bytes expected in the total percent calculation

- **Related Links**:
- [ADO Item](https://msazure.visualstudio.com/One/_workitems/edit/30377572/)
- [GitHub Issue](https://github.com/Azure/azure-storage-azcopy/issues/2975)

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Manually tested using the customer reported scenario.


cmd/jobsResume.go
```diff
@@ -161,6 +161,7 @@ func (cca *resumeJobController) ReportProgressOrExit(lcm common.LifecycleMgr) (t
 				return fmt.Sprintf(
 					`
 
+PENDING COUNT FIX:
```
adreed-msft, line 164: accidental debug line?

jobsAdmin/init.go
```diff
@@ -431,13 +434,31 @@ func GetJobSummary(jobID common.JobID) common.ListJobSummaryResponse {
...
 
 	// Add on byte count from files in flight, to get a more accurate running total
-	js.TotalBytesTransferred += jm.SuccessfulBytesInActiveFiles()
+	// Check is added to prevent double counting
+	if js.TotalBytesTransferred+jm.SuccessfulBytesInActiveFiles() <= js.TotalBytesExpected {
+		js.TotalBytesTransferred += jm.SuccessfulBytesInActiveFiles()
+	}
+
+	var glcm = common.GetLifecycleMgr()
+	glcm.Progress(func(format common.OutputFormat) string {
```
adreed-msft, line 443: Debug logging?

jobsAdmin/init.go
```diff
@@ -431,13 +434,31 @@ func GetJobSummary(jobID common.JobID) common.ListJobSummaryResponse {
 	part0PlanStatus := part0.Plan().JobStatus()
 
 	// Add on byte count from files in flight, to get a more accurate running total
-	js.TotalBytesTransferred += jm.SuccessfulBytesInActiveFiles()
+	// Check is added to prevent double counting
+	if js.TotalBytesTransferred+jm.SuccessfulBytesInActiveFiles() <= js.TotalBytesExpected {
```
adreed-msft, line 438: This could still look wrong but I guess it's "less" wrong? Been scratching my head on a more accurate but equally efficient algorithm, and I don't have a good answer.

adreed-msft (approved):

adreed-msft (approved):

jobsAdmin/init.go
```diff
@@ -431,13 +434,19 @@ func GetJobSummary(jobID common.JobID) common.ListJobSummaryResponse {
...
+		js.TotalBytesTransferred += jm.SuccessfulBytesInActiveFiles()
```
vibhansa-msft, line 439: ideally these changes shall be atomic in nature to avoid accidental corruptions

jobsAdmin/init.go
```diff
@@ -568,13 +577,19 @@ func resurrectJobSummary(jm ste.IJobMgr) common.ListJobSummaryResponse {
 	})
 
 	// Add on byte count from files in flight, to get a more accurate running total
-	js.TotalBytesTransferred += jm.SuccessfulBytesInActiveFiles()
+	// Check is added to prevent double counting
+	if js.TotalBytesTransferred+jm.SuccessfulBytesInActiveFiles() <= js.TotalBytesExpected {
```
vibhansa-msft, line 581: if there are multiple threads updating this value, it can still run into incorrect value. Imagine  a situation where you checked the condition it was true and you went ahead to add the value but by that time, condition values might have changed.

jobsAdmin/init.go
wonwuakpa-msft: Yes, that function is an atomic operation under the hood

jobsAdmin/init.go
wonwuakpa-msft: I have added mutex locking mechanism to prevent the value changing from the time it is checked against the condition to the time it is used.

ste/mgr-JobPartMgr.go
```diff
@@ -314,6 +316,7 @@ func (jpm *jobPartMgr) ScheduleTransfers(jobCtx context.Context) {
 		// If the transfer was failed, then while rescheduling the transfer marking it Started.
 		if ts == common.ETransferStatus.Failed() {
 			jppt.SetTransferStatus(common.ETransferStatus.Restarted(), true)
+			atomic.AddUint32(&jpm.atomicTransfersFailed, ^uint32(0)) // Adding uint32 max is effectively subtracting 1
```
adreed-msft, line 319: Until it's 0, in which, you have Nuclear Ghandi. Earlier in the function sets the condition up perfectly at line 236. This will result in a value close to the max of uint32 if this function is called with any restarted transfers

jobsAdmin/init.go
adreed-msft: FWIW, `js` isn't actually global state here.

adreed-msft (approved):

ste/mgr-JobPartMgr.go
wonwuakpa-msft: Ok I added a check to prevent this overflow issue

seanmcc-msft (approved):

adreed-msft (approved):

jobsAdmin/init.go
```diff
@@ -28,13 +28,15 @@ import (
...
 	"os"
+	"sync"
 	"time"
 
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 	"github.com/Azure/azure-storage-azcopy/v10/ste"
 )
 
 var steCtx = context.Background()
+var mu sync.Mutex // Prevent inconsistent state between check and update of TotalBytesTransferred variable
```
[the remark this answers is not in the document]
wonwuakpa-msft: I thought the mutex is better placed here than as a struct field because it would need to have a json representation like the other fields which I believe they don't support. Also, it's locking the check operation/calculation of the total bytes transferred variable not necessarily a job summary instance itself

PR #2962, opened 2025-03-03
Default to using the customized retry policy

adreed-msft:
## Description

- **Feature / Bug Fix**: The extra platform-specific check for the wsarecv error does not occur if no special retry policy is specified.

- **Related Links**:
- Resolves an issue I noticed with #2916 while writing scenario generators for memory testing

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -89,14 +89,16 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
+		// referencing https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/policy/policy.go#L114
+		rscList := "408;429;500;502;503;504"
 		if retryStatusCodes != "" {
-			retryStatusCodes = retryStatusCodes + ";408;429;500;502;503;504"
-			rsc, err := ste.ParseRetryCodes(retryStatusCodes)
-			if err != nil {
-				return err
-			}
-			ste.RetryStatusCodes = rsc
+			rscList += ";" + retryStatusCodes
 		}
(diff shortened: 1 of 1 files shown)


ste/xferRetryHelper.go
```diff
@@ -37,17 +37,19 @@ func getShouldRetry() func(*http.Response, error) bool {
...
-			// no status codes specified to compare to
-			if len(storageErrorCodes) == 0 {
-				return true
-			}
-			// compare to status codes
-			errorCode := getErrorCode(resp)
-			if errorCode != "" {
-				if _, ok = storageErrorCodes[errorCode]; ok {
+		if resp != nil {
+			if storageErrorCodes, ok := RetryStatusCodes[resp.StatusCode]; ok {
```
vibhansa-msft, line 41: Should we put explicit logs here saying we are retrying due to custom policy etc.

gapra-msft (approved):

ste/xferRetryHelper.go
adreed-msft: Sure, I think that's a good idea. Let me do that.

azure-pipelines.yml
```diff
@@ -51,6 +51,7 @@ jobs:
         condition: eq(variables.type, 'linux')
 
       - script: |
+          sudo apt-get install libsecret-1-dev
```
dphulkar-msft, line 54: Already doing the same on line:44

dphulkar-msft (approved):

vibhansa-msft (approved):

PR #2968, opened 2025-03-05
Introduce flags for NFS over REST support

dphulkar-msft:
## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?


vibhansa-msft (approved):

cmd/constants.go
```diff
@@ -0,0 +1,63 @@
+// Copyright © 2025 Microsoft <dphulkar@microsoft.com>
```
gapra-msft, line 1: Should we point this to azcopydev email?

cmd/constants.go
```diff
@@ -0,0 +1,63 @@
...
+	// networking is traditionally done in base 10 units (not base 2).
+	// E.g. "gigabit ethernet" means 10^9 bits/sec, not 2^30. So by using base 10 units
+	// we give the best correspondence to the sizing of the user's network pipes.
+	// See https://networkengineering.stackexchange.com/questions/3628/iec-or-si-units-binary-prefixes-used-for-network-measurement
+	// NOTE that for everything else in the app (e.g. sizes of files) we use the base 2 units (i.e. 1024 * 1024) because
+	// for RAM and disk file sizes, it is conventional to use the power-of-two-based units.
+	base10Mega = 1000 * 1000
+)
+
+// credentials relates consts
```
gapra-msft, line 56:
```suggestion
// credentials related consts
```

cmd/constants.go
dphulkar-msft: done

e2etest/newe2e_task_runazcopy_parameters.go
```diff
@@ -3,11 +3,12 @@ package e2etest
 import (
 	"errors"
 	"fmt"
-	"github.com/Azure/azure-storage-azcopy/v10/common"
 	"os"
 	"reflect"
 	"strings"
 	"time"
+
+	"github.com/Azure/azure-storage-azcopy/v10/common"
```
gapra-msft, line 11: nit. This change can be reverted

cmd/sync.go
```diff
@@ -837,6 +920,12 @@ func init() {
 		"volume mounted on Linux using SMB protocol, this flag will have to be explicitly set to true. Only the attribute bits supported by Azure Files "+
 		"will be transferred; any others will be ignored. This flag applies to both files and folders, unless a file-only filter is specified "+
 		"(e.g. include-pattern). The info transferred for folders is the same as that for files, except for Last Write Time which is never preserved for folders.")
+
+	syncCmd.PersistentFlags().BoolVar(&raw.isNFSCopy, "nfs", false, "False by default. Specifies whether the copy operation is an NFS copy. TODO: Add flag description")
+	//TBD: should we mark this flag hidden?
```
gapra-msft, line 911: I am in favor of marking the flag hidden since we are replacing it with a new flag.

e2etest/newe2e_task_runazcopy_parameters.go
dphulkar-msft: done

cmd/sync.go
dphulkar-msft: I have created an ADO task for it to track separately as I have to discuss it with Daniel and Team.
https://msazure.visualstudio.com/One/_workitems/edit/31823327/

cmd/sync.go
```diff
@@ -837,6 +920,12 @@ func init() {
...
+	_ = syncCmd.PersistentFlags().MarkHidden("preserve-smb-info")
+	syncCmd.PersistentFlags().BoolVar(&raw.preserveInfo, PreserveInfoFlag, false, "Preserves properties. TODO: Add flag description")
```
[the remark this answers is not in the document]
dphulkar-msft: I have created an ADO task for it to track separately as I have to discuss it with Daniel and Team.
https://msazure.visualstudio.com/One/_workitems/edit/31823327/

cmd/helpers.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/spf13/cobra"
+)
+
+func validatePreserveNFSPropertyOption(toPreserve bool, fromTo common.FromTo, flagName string) error {
+	// preserverInfo will be true by default for NFS-aware locations unless specified false.
+	// 1. Upload (Windows/Linux -> Azure File)
+	// 2. Download (Azure File -> Windows/Linux)
+	// 3. S2S (Azure File -> Azure File)
+	// TODO: More combination checks to be added later
+	if toPreserve && !(fromTo == common.EFromTo.LocalFile() ||
```
adreed-msft, line 38: code dupe: areBothLocationsNFSAware

cmd/helpers.go
```diff
@@ -0,0 +1,160 @@
...
+		fromTo == common.EFromTo.FileLocal() ||
+		fromTo == common.EFromTo.FileFile()) {
+		return fmt.Errorf("%s is set but the job is not between %s-aware resources", flagName, common.Iff(flagName == PreserveInfoFlag, "permission", "NFS"))
+	}
+
+	if toPreserve && (fromTo.IsUpload() || fromTo.IsDownload()) &&
```
adreed-msft, line 44: ditto

cmd/helpers.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return nil
+}
+
+func validatePreserveSMBPropertyOption(toPreserve bool, fromTo common.FromTo, flagName string) error {
+	// preserverInfo will be true by default for SMB-aware locations unless specified false.
+	// 1. Upload (Windows/Linux -> Azure File)
+	// 2. Download (Azure File -> Windows/Linux)
+	// 3. S2S (Azure File -> Azure File)
+	if toPreserve && flagName == PreservePermissionsFlag && (fromTo == common.EFromTo.BlobBlob() || fromTo == common.EFromTo.BlobFSBlob() || fromTo == common.EFromTo.BlobBlobFS() || fromTo == common.EFromTo.BlobFSBlobFS()) {
```
adreed-msft, line 57: code dupe: areBothLocationsSMBAware

cmd/copy.go
```diff
@@ -1606,6 +1635,13 @@ func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 		}
 	}
 
+	if err := validateProtocolCompatibility(ctx, cca.FromTo,
```
adreed-msft, line 1638: nit: this feels like a cook-time validation but I think it relies on enough post-cook that it makes sense to be here; use your judgement

cmd/copy_sync_linux_test.go
```diff
@@ -0,0 +1,373 @@
...
+	"testing"
+
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCopy_NFSSpecificValidationForFlags(t *testing.T) {
+	a := assert.New(t)
+
+	tests := []struct {
```
adreed-msft, line 15: we love testing matrixes

cmd/copy_sync_linux_test.go
```diff
@@ -0,0 +1,373 @@
...
+package cmd
+
+import (
+	"testing"
+
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCopy_NFSSpecificValidationForFlags(t *testing.T) {
```
adreed-msft, line 12: What about error cases?

cmd/copy_sync_linux_test.go
```diff
@@ -0,0 +1,373 @@
...
+	for _, tt := range tests {
+		t.Run(tt.testScenario, func(t *testing.T) {
+			tt.rawCommand.setMandatoryDefaults()
+			cooked, _ := tt.rawCommand.cook()
+			tt.expected(&cooked)
+		})
+	}
+}
+
+func TestCopy_SMBSpecificValidationForFlags(t *testing.T) {
```
adreed-msft, line 102: ditto

cmd/copy_sync_linux_test.go
```diff
@@ -0,0 +1,373 @@
+//go:build linux
+
+package cmd
```
adreed-msft, line 3: this file feels generally very repetitious and like it's testing the same things the same ways over again? Perhaps we could compress to one test.

cmd/copy_sync_windows_test.go
```diff
@@ -0,0 +1,382 @@
+//go:build windows
+
+package cmd
```
adreed-msft, line 3: This feels the same way. The test itself isn't reliant upon anything that doesn't compile into the executable, so maybe items in the test matrix could be flagged with a target GOOS?

cmd/sync.go
```diff
@@ -377,6 +369,94 @@ func (raw *rawSyncCmdArgs) cook() (cookedSyncCmdArgs, error) {
...
+//
+// The function checks the following:
+//   - Validates the "preserve-info" flag to ensure it is set correctly for NFS-aware locations.
+//   - Validates the "preserve-permissions" flag, ensuring that user input is correct and provides feedback
+//     if the flag is set to false and NFS info is being preserved.
+//   - Ensures that both source and destination locations are NFS-aware for relevant operations.
+//
+// Returns:
+// - An error if any validation fails, otherwise nil indicating successful validation.
+func (raw rawSyncCmdArgs) performNFSSpecificValidation(cooked *cookedSyncCmdArgs) (err error) {
```
adreed-msft, line 385: If we don't need validation that specifically differs, perhaps taking in pointers to the flags and doing a deref-set e.g.

```go
*x = "foobar"
```

makes sense to reduce code dupe

e2etest/zt_basic_copy_sync_remove_test.go
```diff
@@ -937,6 +937,7 @@ func TestBasic_SyncLMTSwitch_PreferServiceLMT(t *testing.T) {
 		anonymousAuthOnly,
 		params{
 			preserveSMBInfo: to.Ptr(false),
+			preserveInfo:    to.Ptr(false),
```
adreed-msft, line 940: if we're replacing preserveSMBInfo, maybe removing that flag makes sense in most tests

adreed-msft (changes requested):
lots of little things overall here-- some repeats from the prior iteration.

vibhansa-msft (approved):

cmd/copy.go
```diff
@@ -212,6 +200,93 @@ func blockSizeInBytes(rawBlockSizeInMiB float64) (int64, error) {
...
+//   - Validates the "preserve-info" flag to ensure it is set correctly for NFS-aware locations.
+//   - Validates the "preserve-permissions" flag, ensuring that user input is correct and provides feedback
+//     if the flag is set to false and NFS info is being preserved.
+//   - Ensures that both source and destination locations are NFS-aware for relevant operations.
+//
+// Returns:
+// - An error if any validation fails, otherwise nil indicating successful validation.
+func (raw rawCopyCmdArgs) performNFSSpecificValidation(cooked *CookedCopyCmdArgs) (err error) {
+	if (raw.preserveSMBInfo && runtime.GOOS == "linux") || raw.preserveSMBPermissions {
+		return fmt.Errorf(InvalidFlagsForNFSMsg)
```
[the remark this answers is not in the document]
dphulkar-msft: This is an invalid scenario where user is trying to use --nfs flag along with smb-flag(--preserve-smb-info and --preserve-smb-permissions) 
The first condition applies only in case of linux because for in windows --preserve-smb-info is true by default. But in case of linux user has explicitly set this value to true also with --nfs flag which is not correct.

Regarding the error message, I will provide a more detailed error message after discussing it further with Daniel.

cmd/copy.go
dphulkar-msft: yes correct. Its a cook-time validation but this validation requires cred info and destination service client. As a resultx this validation is done at this stage.

cmd/helpers.go
dphulkar-msft: I agree there are some repeated validation in areBothLocationsNFSAware() and validatePreserveNFSPropertyOption() but these will be repeated in case of validating preserve-ingo flag but for preserve-permissions we are not using the areBothLocationsNFSAware() check which will be checked as part of   validatePreserveNFSPropertyOption()

Talking about below code patch:
```
isUserPersistingPermissions := raw.preservePermissions
	if cooked.preserveInfo && !isUserPersistingPermissions {
		glcm.Info(PreserveNFSPermissionsDisabledMsg)
	}
	if err = validatePreserveNFSPropertyOption(isUserPersistingPermissions,
		cooked.FromTo,
		PreservePermissionsFlag); err != nil {
		return err
	}
```
I agree that we can adjust the logic to avoid duplication. However, my goal is to minimize any additional code changes to prevent potential issues that could disrupt the existing workflow for users.

cmd/helpers.go
dphulkar-msft: same reason as above

cmd/helpers.go
dphulkar-msft: same reason as above

cmd/sync.go
dphulkar-msft: done

e2etest/zt_basic_copy_sync_remove_test.go
dphulkar-msft: done

cmd/copy_sync_linux_test.go
dphulkar-msft: added

cmd/copy_sync_linux_test.go
dphulkar-msft: combined into one file

cmd/sync.go
dphulkar-msft: done

cmd/sync.go
```diff
@@ -792,6 +831,15 @@ func init() {
 				glcm.EnableCancelFromStdIn()
 			}
 
+			preserveInfoDefaultVal := GetPreserveInfoFlagDefault(cmd, raw.isNFSCopy)
+			if cmd.Flags().Changed(PreserveInfoFlag) && cmd.Flags().Changed(PreserveSMBInfoFlag) || cmd.Flags().Changed(PreserveInfoFlag) {
+				// we give preserdence to raw.preserveInfo flag value if both flags are set
```
gapra-msft, line 836:
```suggestion
				// we give precedence to raw.preserveInfo flag value if both flags are set
```

cmd/copy.go
```diff
@@ -212,6 +200,51 @@ func blockSizeInBytes(rawBlockSizeInMiB float64) (int64, error) {
...
+// The function performs the following checks:
+// - Validates the "preserve-info" flag to ensure both source and destination are SMB-aware.
+// - Validates the "preserve-posix-properties" flag, ensuring both locations are POSIX-aware if set.
+// - Ensures that the "preserve-permissions" flag is correctly set if SMB information is preserved.
+// - Validates the preservation of file owner information based on user flags.
+//
+// Returns:
+// - An error if any validation fails, otherwise nil indicating successful validation.
+
+func (raw rawCopyCmdArgs) performSMBSpecificValidation(cooked *CookedCopyCmdArgs) (err error) {
```
gapra-msft, line 216: This is just moving the validation logic into its own method right?

cmd/helpers.go
```diff
@@ -0,0 +1,209 @@
...
+	"context"
+	"fmt"
+	"runtime"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azfile/file"
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"github.com/spf13/cobra"
+)
+
+func validatePreserveNFSPropertyOption(toPreserve bool, fromTo common.FromTo, flagName string) error {
```
gapra-msft, line 32: Should we maybe name this file copyValidation or something more specific?

cmd/sync.go
```diff
@@ -377,6 +371,50 @@ func (raw *rawSyncCmdArgs) cook() (cookedSyncCmdArgs, error) {
...
+// The function performs the following checks:
+// - Validates the "preserve-info" flag to ensure both source and destination are SMB-aware.
+// - Validates the "preserve-posix-properties" flag, ensuring both locations are POSIX-aware if set.
+// - Ensures that the "preserve-permissions" flag is correctly set if SMB information is preserved.
+// - Validates the preservation of file owner information based on user flags.
+//
+// Returns:
+// - An error if any validation fails, otherwise nil indicating successful validation.
+
+func (raw rawSyncCmdArgs) performSMBSpecificValidation(cooked *cookedSyncCmdArgs) (err error) {
```
gapra-msft, line 387: Is this validation code mostly duplicated with copy? If so, can we deduplicate it?

cmd/copy.go
dphulkar-msft: yes correct. For NFS related validation there is another function introduced perforNFSSpecificValidation()

cmd/sync.go
dphulkar-msft: In sync we don't check this part of code while in copy we does. But I can accommodate the common code in one part and remaining can handle separately. Will make this change as part of the next PR.

```
if err = validatePreserveOwner(raw.preserveOwner, cooked.FromTo); err != nil {
		return err
	}
	cooked.preservePermissions = common.NewPreservePermissionsOption(isUserPersistingPermissions,
		raw.preserveOwner,
		cooked.FromTo)
```

cmd/helpers.go
dphulkar-msft: sure will make this change as part of next PR. Keeping a note of it.

gapra-msft (approved):
Approved as long as my comments are addressed before NFS goes into main

PR #2976, opened 2025-03-12
Deleting the folders if not present in source

samskan:
## Description

If folder is deleted, need to recursively delete the folder and its contents

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/cmd/syncProcessor.go
+++ b/cmd/syncProcessor.go
@@ -32,6 +32,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 	"github.com/Azure/azure-storage-azcopy/v10/ste"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
 )
 
 // extract the right info from cooked arguments and instantiate a generic copy transfer processor from it
@@ -379,6 +380,10 @@ func (b *remoteResourceDeleter) delete(object StoredObject) error {
 				_, err = blobClient.Delete(b.ctx, nil)
 				return (err == nil)
 			}
+			if UseSyncOrchestrator{
+				cc := bsc.NewContainerClient(b.containerName)
+				IterateThroughFolder(cc,objectPath)
+			}
 		case common.ELocation.File():
 			fsc, _ := sc.FileServiceClient()
 			dirClient := fsc.NewShareClient(b.containerName).NewDirectoryClient(objectPath)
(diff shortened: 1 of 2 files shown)


cmd/syncEnumerator.go
```diff
@@ -365,6 +365,58 @@ func (cca *cookedSyncCmdArgs) InitEnumerator(ctx context.Context, errorChannel c
 	var finalize func() error
 
 	switch cca.fromTo {
+	//based on UseSynOrchestrator flag add case for Blobfs
+	case common.EFromTo.LocalBlobFS():
```
ramkumar-ganapathy, line 369: Sorry, I might not be having the full background here. Why is this code new, and why is this specific to LocalBlobFs()?

cmd/syncOrchestrator.go
```diff
@@ -102,6 +102,7 @@ func (st *SyncTraverser) processor(so StoredObject) error {
 }
```
ramkumar-ganapathy, line 103: A few optimizations we need to consider doing:
1. In the sync orchestrator, when we find out that a directory is present at source, but not at target, we need to trigger a recursive copy on source (similar to what we do in the delete processor, when a directory that is in target is missing in source). This can optimize a lot of cloud requests. And this can be pretty strong optimization.
2. We should consider removing the mutex Lock, Unlock code. That would mean that, we create multiple instances of comparator, one per directory that is being synced. This is also a pretty important CPU optimization. However, this work can be a bit involved and we should do it only after properly benchmarking the performance.

cmd/syncOrchestrator.go
```diff
@@ -414,7 +418,7 @@ func (cca *cookedSyncCmdArgs) runSyncOrchestrator(ctx context.Context) (err erro
 		fi.ModTime(), fi.Size(), noContentProps, noBlobProps, noMetadata, "")
 
 	parallelism := 4
```
ramkumar-ganapathy, line 420: Are we planning to change this and make the values flow in through configuration, or other parameters that are comparable to existing mover logic?

cmd/syncProcessor.go
```diff
@@ -416,3 +425,62 @@ func (b *remoteResourceDeleter) delete(object StoredObject) error {
 		return nil
 	}
 }
+func IterateThroughFolder(containerClient *container.Client, folderPrefix string) error {
```
ramkumar-ganapathy, line 428: Is this brand new code, or was this code adopted from mover/mover-version of azcopy?

cmd/syncProcessor.go
```diff
@@ -416,3 +425,62 @@ func (b *remoteResourceDeleter) delete(object StoredObject) error {
...
+		for _, blobPrefix := range page.Segment.BlobPrefixes {
+			if err := IterateThroughFolder(containerClient, *blobPrefix.Name); err != nil {
+				return err
+			}
+		}
+
+		// Print blobs
+		for _, blob := range page.Segment.BlobItems {
+			glcm.Info(fmt.Sprintf("Deleting extra blob object : %s\n", *blob.Name))
+			blobClient := containerClient.NewBlobClient(*blob.Name)
```
ramkumar-ganapathy, line 449: We seem to be invoking the client directly here, which is fine. But, are we missing some capability by not doing this through the STE?

cmd/syncEnumerator.go
samskan: In syncOrchestrator we use SyncDestinationCompartor. If we dont have this check it goes for newSyncSourceComparator and we see below trace back
image

cmd/syncProcessor.go
samskan: I dont think so. Because in normal case where recursive is true, same kind of iteration is done. You can check parallelList function in zc_traversser_blob.go

cmd/syncOrchestrator.go
samskan: May be once we do the performance testing we can fine tune this

cmd/syncProcessor.go
samskan: The logic of iterating all blobs was present in azcopy. Use same logic to delete the complete folder

cmd/syncOrchestrator.go
samskan: Hope we can consider this optimization in next changes.

PR #3014, opened 2025-04-10
Upload AzCopy binaries to Github

wonwuakpa-msft:
## Description

The previous step of uploading the binaries to blob storage account was more complex than necessary given that GitHub provides its own release functionalities. This change will streamline the binary distribution process. See doc below for more info 

Switched over from uploading AzCopy binaries to a storage account to uploading them on GitHub as assets. Modified the 1ES release pipeline to include this. 

- **Related Links**:

- [Documents](<https://microsoft.sharepoint.com/:w:/t/StorageDevEx/EQ0AD-dZ0MdGhREwSGBDIzMBVx9bQ8Lf3eEhuIccn6_mBQ?e=I1zZZa)
- [Email Subject] "Can we reduce SFI churn by moving to GitHub releases"

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/build-1es-pipeline.yaml
+++ b/build-1es-pipeline.yaml
@@ -761,134 +761,6 @@ extends:
                 displayName: 'Verify exe in zip'
 
       # VerifyArtifacts end here
-
-      - stage: UploadToStorage
-        dependsOn:
-          - AzCopyVersion
-          - VerifyArtifacts
-        jobs:
-          - job: Job
-            pool:
-              name: azcopy-pool
-              image: ubuntu22-1espt
-              os: linux
-            variables:
-              - name: input
-                value: '$(System.DefaultWorkingDirectory)/input'
-              - name: release
-                value: '$(System.DefaultWorkingDirectory)/release'
-              - name: drop
-                value: '$(System.DefaultWorkingDirectory)/drop'
-              - name: publish_to_container
-                value: ${{ parameters.publish_to_container }}
-              - name: publish_to_m1_container
-                value: ${{ parameters.publish_to_m1_container }}
-              - name: azcopy_version
-                value: $[ stageDependencies.AzCopyVersion.GetAzCopyVersion.outputs['SetAzCopyVersion.azcopy_version'] ]
-            steps:
-              - checkout: self
-                displayName: 'Checkout repository'
-
-              - script: |
-                  sudo apt-get clean
-                  sudo apt-get update --fix-missing
-                  sudo apt-get install -y zip unzip
-                displayName: 'Install Dependencies'
-              - task: GoTool@0
-                env:
-                  GO111MODULE: 'on'
-                inputs:
-                  version: $(AZCOPY_GOLANG_VERSION)
-
-              - script: |
-                  go build -tags "netgo" -o azcopy
-                displayName: 'Build AzCopy'
-
-              # DownloadPipelineArtifacts AFTER checkout since checkout wipes out the DefaultWorkingDirectory
-              - task: DownloadPipelineArtifact@2
-                inputs:
-                  artifactName: azcopy-linux-signed
-                  targetPath: $(System.DefaultWorkingDirectory)/input/azcopy-linux-signed
-                  displayName: "Download Linux Signed"
-
-              - task: DownloadPipelineArtifact@2
-                inputs:
-                  artifactName: azcopy-windows-signed
-                  targetPath: $(System.DefaultWorkingDirectory)/input/azcopy-windows-signed
-                  displayName: "Download Windows Signed"
-
-              - task: DownloadPipelineArtifact@2
-                inputs:
-                  artifactName: azcopy-mac-signed
-                  targetPath: $(System.DefaultWorkingDirectory)/input/azcopy-mac-signed
-                  displayName: "Download Mac Signed"
-
-              - script: |
-                  mkdir -p $(release)
-                  cp $(input)/azcopy-linux-signed/azcopy* $(release)
-                  cp $(input)/azcopy-windows-signed/azcopy* $(release)
-                  cp $(input)/azcopy-mac-signed/azcopy* $(release)
-                displayName: 'Prepare release folder'
-
-              - task: DownloadPipelineArtifact@2
-                inputs:
-                  artifactName: azcopy-binaries-linux-amd64
-                  targetPath: $(System.DefaultWorkingDirectory)/input/azcopy-binaries-linux-amd64
-                  displayName: "Download Linux AMD64 Binaries"
-
-              - task: DownloadPipelineArtifact@2
-                inputs:
-                  artifactName: azcopy-binaries-linux-arm64
-                  targetPath: $(System.DefaultWorkingDirectory)/input/azcopy-binaries-linux-arm64
-                  displayName: "Download Linux ARM64 Binaries"
-
-              - task: DownloadPipelineArtifact@2
-                inputs:
-                  artifactName: azcopy-binaries-windows
-                  targetPath: $(System.DefaultWorkingDirectory)/input/azcopy-binaries-windows
-                  displayName: "Download Windows Binaries"
-
-              - task: DownloadPipelineArtifact@2
-                inputs:
-                  artifactName: azcopy-binaries-mac
-                  targetPath: $(System.DefaultWorkingDirectory)/input/azcopy-binaries-mac
-                  displayName: "Download Mac Binaries"
-
-              - script: |
-                  mkdir -p $(drop)
-                  cp $(input)/azcopy-binaries-linux-amd64/azcopy* $(drop)
-                  cp $(input)/azcopy-binaries-linux-arm64/azcopy* $(drop)
-                  cp $(input)/azcopy-binaries-windows/azcopy* $(drop)
-                  cp $(input)/azcopy-binaries-mac/azcopy* $(drop)
-                  cp $(System.DefaultWorkingDirectory)/NOTICE.txt $(drop)
-                displayName: 'Prepare drop folder'
-
-              - task: ArchiveFiles@2
-                displayName: 'Archive drop'
-                inputs:
-                  rootFolderOrFile: '$(drop)'
-                  archiveFile: '$(release)/drop.zip'
-
-              - task: AzureCLI@2
-                inputs:
-                  azureSubscription: 'ESRP KeyVault identity'
-                  scriptType: bash
-                  scriptLocation: inlineScript
-                  inlineScript: |
-                    today=$(date +"%Y%m%d")
-                    if [ $(publish_to_container) = "True" ]; then
-                      container_url="https://azcopyvnextrelease.blob.core.windows.net/%24web/releases/release-$(azcopy_version)-$today"
-                    else
-                      container_url="https://azcopyprivatedrops.blob.core.windows.net/%24web/releases/release-$(azcopy_version)-$today"
-                    fi
-                    echo "Artifacts will be uploaded to: $container_url"
-                    AZCOPY_AUTO_LOGIN_TYPE=AzCLI ./azcopy cp "$(release)/*" "$container_url" --recursive --put-md5=true
-                    
-                    if [ $(publish_to_m1_container) = "True" ]; then
-                      m1_container_url="https://azcopyvnextrelease.blob.core.windows.net/%24web/azcopy-m1-drops/azcopy-$(azcopy_version)-$today"
-                    AZCOPY_AUTO_LOGIN_TYPE=AzCLI ./azcopy cp "$(input)/azcopy-binaries-mac/azcopy_darwin_m1_arm64" "$m1_container_url" --recursive --put-md5=true
-                    fi
-      # UploadToStorage ends here
       - ${{ if eq(parameters.publish_docker_image, true) }}:
           - stage: BuildAndPublishDockerImage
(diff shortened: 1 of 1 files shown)


dphulkar-msft:
Is this change tested? @wonwuakpa-msft

build-1es-pipeline.yaml
```diff
@@ -761,134 +761,6 @@ extends:
...
-                      container_url="https://azcopyprivatedrops.blob.core.windows.net/%24web/releases/release-$(azcopy_version)-$today"
-                    fi
-                    echo "Artifacts will be uploaded to: $container_url"
-                    AZCOPY_AUTO_LOGIN_TYPE=AzCLI ./azcopy cp "$(release)/*" "$container_url" --recursive --put-md5=true
-                    
-                    if [ $(publish_to_m1_container) = "True" ]; then
-                      m1_container_url="https://azcopyvnextrelease.blob.core.windows.net/%24web/azcopy-m1-drops/azcopy-$(azcopy_version)-$today"
-                    AZCOPY_AUTO_LOGIN_TYPE=AzCLI ./azcopy cp "$(input)/azcopy-binaries-mac/azcopy_darwin_m1_arm64" "$m1_container_url" --recursive --put-md5=true
-                    fi
-      # UploadToStorage ends here
```
dphulkar-msft, line 891: Why are we removing this?

wonwuakpa-msft: @dphulkar-msft We decided to no longer upload binaries to the static blob storage

build-1es-pipeline.yaml
```diff
@@ -1028,13 +921,19 @@ extends:
                       changeLogType: 'commitBased'
                       isDraft: ${{ parameters.draft }}
                       isPreRelease: ${{ parameters.prerelease }}
+                      assets: |
```
vibhansa-msft, line 924: Earlier step had two options either to publish on our public container or copy the files to a privatedrop container. With this change we are loosing the ability to give a private drop. We might need to think on that part. One way could be to do a preview releases or pre-releases on github (there is a way to mark a release as a prerelease) (however when we do that we shall not be posting those images on linux repos as well).

vibhansa-msft: I see prerelease is already a flag in our pipeline so we shall use that to post a prerelease to github for private drops, just ensure we do not publish those binaries to linux repos.

adreed-msft: I dislike the idea of publishing every private drop to GitHub, personally. We should still have the storage account as a fallback.

adreed-msft (changes requested):
+1 to private drops

build-1es-pipeline.yaml
dphulkar-msft: Yes, I agree to keep the storage account as a fallback option instead of relying on GitHub for private drops.

wonwuakpa-msft: Noted, I'll make the change to keep uploading private drops to its specific storage container instead of GitHub

vibhansa-msft (approved):

cmd/root.go
```diff
@@ -219,14 +218,14 @@ var rootCmd = &cobra.Command{
...
 
-		if !azcopySkipVersionCheck && !isPipeDownload {
-			// spawn a routine to fetch and compare the local application's version against the latest version available
-			// if there's a newer version that can be used, then write the suggestion to stderr
-			// however if this takes too long the message won't get printed
-			// Note: this function is necessary for non-help, non-login commands, since they don't reach the corresponding
-			// beginDetectNewVersion call in Execute (below)
-			beginDetectNewVersion()
-		}
+		//if !azcopySkipVersionCheck && !isPipeDownload {
```
adreed-msft: We should probably just remove the code too instead of leaving it commented.

cmd/versionChecker_test.go
```diff
@@ -211,24 +211,37 @@ func TestValidateCachedVersion2(t *testing.T) {
...
-	blobClient, err := blob.NewClientWithNoCredential(versionMetadataUrl, &blob.ClientOptions{ClientOptions: options})
-	a.NoError(err)
-
-	downloadBlobResp, err := blobClient.DownloadStream(context.TODO(), nil)
+	latestVersion, err := getGitHubLatestRemoteVersion()
 	a.NoError(err)
+	a.NotNil(latestVersion)
+	a.NotEmpty(latestVersion.original)
+	versionVar, _ := NewVersion(common.AzcopyVersion)
+	a.Equal(latestVersion, versionVar)
```
adreed-msft: I almost wonder if the right answer is to check if it's the same version or later. We already have checks in place so that your branch should be caught up to main, but what about forks, betas, drops, etc?

build-1es-pipeline.yaml
```diff
@@ -1034,16 +1048,22 @@ extends:
...
                       tag: "v$(azcopy_github_tag)"
                       changeLogCompareToRelease: 'lastFullRelease'
-                      changeLogType: 'commitBased'
+                      changeLogType: 'issueBased'
                       isDraft: ${{ parameters.draft }}
                       isPreRelease: ${{ parameters.prerelease }}
+                      assets: |
+                        $(Build.ArtifactStagingDirectory)/azCopy-linux-signed/*.deb
+                        $(Build.ArtifactStagingDirectory)/azCopy-linux-signed/*.rpm
+                        $(Build.ArtifactStagingDirectory)/azCopy-linux-signed/*.tar.gz
```
dphulkar-msft, line 1046: mariner rpm builds are missing. Its published under $(Build.ArtifactStagingDirectory)/azCopy-linux-signed/mariner/*

common/version.go
```diff
@@ -1,6 +1,6 @@
 package common
 
-const AzcopyVersion = "10.30.0~preview.1"
+const AzcopyVersion = "10.29.1"
```
dphulkar-msft: We should keep this to the latest preview version itself. Any reason why we are changing it to 10.29.1?

latest_version.txt
```diff
@@ -0,0 +1 @@
+10.29.1,2025-07-10T17:15:19Z
```
dphulkar-msft: I believe we dont push this file to the repo.

release-pipeline.yml
```diff
@@ -1972,10 +1993,17 @@ stages:
...
                 
                 changeLogCompareToRelease: 'lastFullRelease'
-                changeLogType: 'commitBased'
+                changeLogType: 'issueBased'
                 isDraft: ${{ parameters.draft }}
                 isPreRelease: ${{ parameters.prerelease }}
+                assets: |
+                  $(Build.ArtifactStagingDirectory)/azCopy-linux-signed/*.deb
+                  $(Build.ArtifactStagingDirectory)/azCopy-linux-signed/*.rpm
+                  $(Build.ArtifactStagingDirectory)/azCopy-linux-signed/*.tar.gz
```
dphulkar-msft: same as above comment. Missing mariner

build-1es-pipeline.yaml
```diff
@@ -779,8 +778,6 @@ extends:
                 value: '$(System.DefaultWorkingDirectory)/release'
               - name: drop
                 value: '$(System.DefaultWorkingDirectory)/drop'
-              - name: publish_to_container
```
dphulkar-msft, line 782: Since uploading to containers other than the private drop is no longer supported, we should remove this option from line 5 in the file. By default, uploads will always be directed to the private drop container.

build-1es-pipeline.yaml
```diff
@@ -876,14 +873,10 @@ extends:
                   scriptLocation: inlineScript
                   inlineScript: |
                     today=$(date +"%Y%m%d")
-                    if [ $(publish_to_container) = "True" ]; then
-                      container_url="https://azcopyvnextrelease.blob.core.windows.net/%24web/releases/release-$(azcopy_version)-$today"
-                    else
-                      container_url="https://azcopyprivatedrops.blob.core.windows.net/%24web/releases/release-$(azcopy_version)-$today"
-                    fi
-                    echo "Artifacts will be uploaded to: $container_url"
+                    container_url="https://azcopyprivatedrops.blob.core.windows.net/%24web/releases/release-$(azcopy_version)-$today"
```
dphulkar-msft, line 865: What about the publish_to_m1_container step? I see that it currently uploads to the azcopyvnextrelease container. Should we also redirect the M1 build upload to GitHub?

Additionally, there's a stage named UpdateLatestVersion that needs to be updated as well. It currently updates the latest version in the azcopyvnextrelease container, which should now point to GitHub instead.

latest_version.txt
dphulkar-msft: Oh, will this now be the same file that we refer to for getting the latest AzCopy version? If so, we'll need to push this file to the repo.

release-pipeline.yml
```diff
@@ -1958,7 +1958,28 @@ stages:
...
+            - script: |
+                sudo apt-get clean 
+                sudo apt-get update 
+                sudo apt-get upgrade -y 
+                sudo apt-get install -f --fix-missing 
+                wget https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb
+                sudo dpkg -i packages-microsoft-prod.deb
+                sudo apt install apt-transport-https -y 
+                sudo apt install dotnet-sdk-6.0 -y
+              displayName: 'Update Dependencies'
```
dphulkar-msft: Are all of these dependency installations necessary here? If not, we should remove the unnecessary ones to reduce overhead.

gapra-msft:
> Thought: If the 1ES build pipeline is now stable for the AzCopy release process, we can consider removing the release-pipeline.yml file. Maintaining both files adds unnecessary overhead, as any changes to the release process need to be updated in both places. @gapra-msft Let me know what you think of this.

Yes let's go ahead and delete the file from the repo. We have already disabled it on DevOps anyway

common/version.go
wonwuakpa-msft: I changed it for testing purposes, will revert it now

release-pipeline.yml
wonwuakpa-msft: I believe these are necessary for the PublishArtifacts

build-1es-pipeline.yaml
wonwuakpa-msft: I have updated that stage to no longer upload to the container but overwrite the latest_version.txt file from azcopy version

build-1es-pipeline.yaml
```diff
@@ -1219,16 +1234,9 @@ extends:
...
-                      inlineScript: |
-                        export AZCOPY_AUTO_LOGIN_TYPE=AzCLI
-                        echo "azcopy version: $(azcopy_version)"
-                        #Upload to release version blob file
-                        upload_url="https://azcopyvnextrelease.blob.core.windows.net/%24web/releasemetadata/latest_version.txt"
-                        echo "Version will be uploaded to: $upload_url"
-                        echo $(azcopy_version) | ./azcopy copy "$upload_url" --from-to=PipeBlob --log-level=DEBUG --output-type=text 2>&1 | tee /dev/null
-                        exit ${PIPESTATUS[1]}
\ No newline at end of file
+                  - script: |
```
gapra-msft: What is the intended effect of this step? Isn't this just changing the file locally in the pipeline?

latest_version.txt
gapra-msft: I think this file is supposed to have just the version, right? Why is there a date time here?

wonwuakpa-msft: There is a date time here for caching and validation purposes. If the version is still valid (based on the time stamp) the API call to fetch the new version is not made and the version in the file is used. If the version has expired, the GitHub version is fetched

gapra-msft: is this here for a local test?

build-1es-pipeline.yaml
wonwuakpa-msft: We decided to remove it, it was meant to update the latest version file in our GitHub repo

build-1es-pipeline.yaml
```diff
@@ -1024,6 +1006,27 @@ extends:
                       echo "##vso[task.setvariable variable=azcopy_github_tag]$azcopy_github_tag"
                     displayName: 'AzCopy Tag Name'
 
+                  - task : DownloadPipelineArtifacts@2
```
gapra-msft: I think this should be not plural, when I try to run this it says step does not exist

cmd/root.go
```diff
@@ -375,17 +362,15 @@ func init() {
...
 
-const versionMetadataUrl = "https://azcopyvnextrelease.z22.web.core.windows.net/releasemetadata/latest_version.txt"
-
 // always spins up a new goroutine, because sometimes the aka.ms URL can't be reached (e.g. a constrained environment where
 // aka.ms is not resolvable to a reachable IP address). In such cases, this routine will run for ever, and the caller should
 // just give up on it.
 // We spin up the GR here, not in the caller, so that the need to use a separate GC can never be forgotten
 // (if do it synchronously, and can't resolve URL, this blocks caller for ever)
-func beginDetectNewVersion() chan struct{} {
+func beginDetectNewVersion() (chan struct{}, string) {
```
gapra-msft: I dont think you need to make this return common.AzCopyVersion. Can't you just return that in the version method above?

gapra-msft (approved):

PR #3035, opened 2025-04-30
Fix the TokenStore getting stuck in a read lock

adreed-msft:
## Description

- **Feature / Bug Fix**: When Storage Explorer would call into AzCopy with an OAuth token, it would typically hang after about an hour. This was unrelated to any network lockup, but instead was actually just AzCopy not checking for a new token. The first time GetToken is called, the original token is unexpired, so we exit early, but we never actually release the read lock. Later down the line, we try to grab the write lock upon the second call, and wind up getting deadlocked on that.

- **Related Links**:
- Team thread - StgExp/AzCopy integration

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Added a test guaranteeing that the lock is released with both code paths the function can take

### DIFF
--- a/common/oauthTokenManager.go
+++ b/common/oauthTokenManager.go
@@ -63,10 +63,17 @@ const DefaultActiveDirectoryEndpoint = "https://login.microsoftonline.com"
 
 const TokenCache = "AzCopyTokenCache"
 
+type CredCacheImplementation interface {
+	HasCachedToken() (bool, error)
+	LoadToken() (*OAuthTokenInfo, error)
+	SaveToken(OAuthTokenInfo) error
+	RemoveCachedToken() error
+}
+
 // UserOAuthTokenManager for token management.
 type UserOAuthTokenManager struct {
 	oauthClient *http.Client
-	credCache   *CredCache
+	credCache   CredCacheImplementation
 
 	// Stash the credential info as we delete the environment variable after reading it, and we need to get it multiple times.
 	stashedInfo *OAuthTokenInfo
@@ -479,7 +486,7 @@ func (credInfo *OAuthTokenInfo) Refresh(ctx context.Context) (*Token, error) {
 }
 
 // Single instance token store credential cache shared by entire azcopy process.
-var tokenStoreCredCache = NewCredCacheInternalIntegration(CredCacheOptions{
+var tokenStoreCredCache CredCacheImplementation = NewCredCacheInternalIntegration(CredCacheOptions{
 	KeyName:     "azcopy/aadtoken/" + strconv.Itoa(os.Getpid()),
 	ServiceName: "azcopy",
 	AccountName: "aadtoken/" + strconv.Itoa(os.Getpid()),
@@ -510,8 +517,9 @@ func getAuthorityURL(activeDirectoryEndpoint string) (*url.URL, error) {
 const minimumTokenValidDuration = time.Minute * 5
 
 type TokenStoreCredential struct {
-	token *azcore.AccessToken
-	lock  sync.RWMutex
+	token     *azcore.AccessToken
+	lock      sync.RWMutex
+	credCache CredCacheImplementation
 }
 
 // globalTokenStoreCredential is created to make sure that all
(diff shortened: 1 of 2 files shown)


common/oauthTokenManager.go
```diff
@@ -63,10 +63,17 @@ const DefaultActiveDirectoryEndpoint = "https://login.microsoftonline.com"
 
 const TokenCache = "AzCopyTokenCache"
 
+type CredCacheImplementation interface {
```
wonwuakpa-msft, line 66: whats the purpose of this inteface wrapper?

wonwuakpa-msft (approved):

common/oauthTokenManager.go
```diff
@@ -531,22 +539,26 @@ var globalTokenStoreCredential *TokenStoreCredential
...
 func (tsc *TokenStoreCredential) GetToken(_ context.Context, _ policy.TokenRequestOptions) (azcore.AccessToken, error) {
-	// if the token we've has not expired, return the same.
+	// if the token we have has not expired, return the same.
 	tsc.lock.RLock()
-	if time.Until(tsc.token.ExpiresOn) > minimumTokenValidDuration {
+	if rem := time.Until(tsc.token.ExpiresOn); rem > minimumTokenValidDuration {
+		tsc.lock.RUnlock() // return path, so we must release the read lock here as well.
 		return *tsc.token, nil
 	}
 	tsc.lock.RUnlock()
```
vibhansa-msft, line 548: so like we take a rlock, we unlock and then we again take a write lock here ?

vibhansa-msft (approved):

dphulkar-msft (approved):

common/oauthTokenManager.go
adreed-msft: Yes; many reads can occur at once but only one write at a time. The write is intended for the thread that is refreshing the token.

common/oauthTokenManager.go
adreed-msft: The intent is to ensure that the OS-specific implementations conform to a standard implementation, such that when writing code to interact with cred caches, we don't accidentally call any OS-specific functionality here.

incinfomain:
`[https://www.mql5.com/](url)`

incinfomain:
[
[licenses.zip](https://githu
[Uploading https:%2Flistedcompanyjourney.setlink.set.or.th%2Fresource%2Fdocument%2Fบริษัทจด.pdf…]()
b.com/user-attachments/files/20432141/licenses.zip)
](url)

incinfomain:
[Unicode-Bylaws.pdf](https://github.com/user-attachments/files/20432147/Unicode-Bylaws.pdf)

incinfomain:
[Uploading TradingView_Alerts_Log_2024-11-26.csv…]()

incinfomain:
[Uploading cloud_digital_leader_exam_guide_english.pdf…]()

incinfomain:
[Co-location_service_factsheet_202210.pdf](https://github.com/user-attachments/files/20432189/Co-location_service_factsheet_202210.pdf)

incinfomain:
[Uploading 9484812_TradeCycle.xls…]()

incinfomain:
[CIO Daily 30 Nov 2023.pdf](https://githu
[The-Future-for-UK-Interbank-Payments-eBook.pdf](https://github.com/user-attachments/files/20432193/The-Future-for-UK-Interbank-Payments-eBook.pdf)
b.com/user-attachments/files/20432191/CIO.Daily.30.Nov.2023.pdf)

incinfomain:
[RFExposure_iPad Pro 9.7-inch - Wi-Fi.pdf](https://githu
[2fa_authenticator_secret_code.txt](https://github.com/user-attachments/files/20432198/2fa_authenticator_secret_code.txt)
b.com/user-attachments/files/20432195/RFExposure_iPad.Pro.9.7-inch.-.Wi-Fi.pdf)

PR #3041, opened 2025-05-07
Refactor traverser initialization

adreed-msft:
## Description

InitResourceTraverser (and the functions it calls) have many input parameters which prove to be extremely unwieldly at review and merge conflict time. This PR turns the majority of the parameters into an options bag.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

No fundamental logic change should have occurred.

### DIFF
--- a/cmd/zc_enumerator.go
+++ b/cmd/zc_enumerator.go
@@ -340,7 +340,6 @@ var enumerationCounterFuncNoop enumerationCounterFunc = func(entityType common.E
 type InitResourceTraverserOptions struct {
 	DestResourceType *common.Location // Used by Azure Files
 
-	Context              context.Context        // Required
 	Credential           *common.CredentialInfo // Required for most remote traversers
 	IncrementEnumeration enumerationCounterFunc
 
@@ -367,10 +366,6 @@ type InitResourceTraverserOptions struct {
 }
 
 func (o *InitResourceTraverserOptions) PerformChecks() error {
-	if o.Context == nil {
-		return errors.New("a valid context must be supplied to create a traverser")
-	}
-
 	if o.IncrementEnumeration == nil {
 		o.IncrementEnumeration = enumerationCounterFuncNoop
 	}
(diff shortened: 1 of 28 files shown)


cmd/zc_enumerator.go
```diff
@@ -474,14 +493,14 @@ func InitResourceTraverser(resource common.ResourceString, location common.Locat
...
 			}
-			output = newBlobAccountTraverser(bsc, containerName, *ctx, includeDirectoryStubs, incrementEnumerationCounter, s2sPreserveBlobTags, cpkOptions, preservePermissions, false, excludeContainerNames)
-		} else if listOfVersionIds != nil {
-			output = newBlobVersionsTraverser(r, bsc, *ctx, includeDirectoryStubs, incrementEnumerationCounter, listOfVersionIds, cpkOptions)
+			output = newBlobAccountTraverser(bsc, containerName, opts)
+		} else if opts.ListOfVersionIDs != nil {
+			output = newBlobVersionsTraverser(r, bsc, opts)
 		} else {
-			output = newBlobTraverser(r, bsc, *ctx, recursive, includeDirectoryStubs, incrementEnumerationCounter, s2sPreserveBlobTags, cpkOptions, includeDeleted, includeSnapshot, includeVersion, preservePermissions, false)
+			output = newBlobTraverser(r, bsc, opts)
```
wonwuakpa-msft, line 503: Is it fine to now pass the whole options bag instead of the needed fields?

common/fe-ste-models.go
```diff
@@ -191,16 +191,81 @@ func ValidTrailingDotOptions() []string {
 }
 
 // //////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-var EPermanentDeleteOption = PermanentDeleteOption(3) // Default to "None"
+var EBlobTraverserIncludeOption eBlobTraverserIncludeOption
+
+type eBlobTraverserIncludeOption bool
```
wonwuakpa-msft, line 196: What's the difference between these two eBlobTraverserIncludeOption  and BlobTraverserIncludeOption ? It's a little hard to tell

wonwuakpa-msft (approved):

common/fe-ste-models.go
adreed-msft: Lowercase `eBlobTraverserIncludeOption` is a non-exported type that is attached to uppercase, exported variable `EBlobTraverserIncludeOption`; which is separate from `BlobTraverserIncludeOption`. Because upper-case `EBTIO` is exported, the functions available on it also are exported. This means from a non-common package, `EBTIO` is separate from `BTIO`.

Long story short-- The only thing a dev will see working with this is the enum type itself (`BTIO`, and the variable that exposes the enum `EBTIO`).

cmd/zc_enumerator.go
adreed-msft: The options bag replaces the extra fields.

dphulkar-msft (approved):

PR #3063, opened 2025-05-28
Login, Login Status, Logout code Refactor

gapra-msft:
## Description

This is a refactor of the AzCopy login, login status and logout code in preparation for azcopy as a library.

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

This is a refactor, should be covered by existing tests.

### DIFF
--- a/cmd/credentialUtil.go
+++ b/cmd/credentialUtil.go
@@ -90,7 +90,7 @@ func GetUserOAuthTokenManagerInstance() *common.UserOAuthTokenManager {
 func GetOAuthTokenManagerInstance() (*common.UserOAuthTokenManager, error) {
 	var err error
 	autoOAuth.Do(func() {
-		var lca loginCmdArgs
+		var options LoginOptions
 		autoLoginType := strings.ToLower(common.GetEnvironmentVariable(common.EEnvironmentVariable.AutoLoginType()))
 		if autoLoginType == "" {
 			glcm.Info("Autologin not specified.")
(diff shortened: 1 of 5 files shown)


cmd/credentialUtil.go
```diff
@@ -90,44 +90,50 @@ func GetUserOAuthTokenManagerInstance() *common.UserOAuthTokenManager {
...
-			lca.identityObjectID = common.GetEnvironmentVariable(common.EEnvironmentVariable.ManagedIdentityObjectID())
-			lca.identityResourceID = common.GetEnvironmentVariable(common.EEnvironmentVariable.ManagedIdentityResourceString())
-		case common.EAutoLoginType.Device().String():
-		case common.EAutoLoginType.AzCLI().String():
-		case common.EAutoLoginType.PsCred().String():
-		case common.EAutoLoginType.Workload().String():
+		var loginType common.AutoLoginType
+		err = loginType.Parse(autoLoginType)
+		if err != nil {
+			glcm.Error("Invalid Auto-login type specified: " + autoLoginType)
```
dphulkar-msft, line 111: Should we not return directly from here? As anyways we will be returning from the default in the switch if the value is invalid.

cmd/credentialUtil.go
```diff
@@ -90,44 +90,50 @@ func GetUserOAuthTokenManagerInstance() *common.UserOAuthTokenManager {
...
+		options.LoginType = loginType
+		switch options.LoginType {
+		case common.EAutoLoginType.SPN():
+			options.ApplicationID = common.GetEnvironmentVariable(common.EEnvironmentVariable.ApplicationID())
+			options.CertificatePath = common.GetEnvironmentVariable(common.EEnvironmentVariable.CertificatePath())
+			options.certificatePassword = common.GetEnvironmentVariable(common.EEnvironmentVariable.CertificatePassword())
+			options.clientSecret = common.GetEnvironmentVariable(common.EEnvironmentVariable.ClientSecret())
+		case common.EAutoLoginType.MSI():
+			options.IdentityClientID = common.GetEnvironmentVariable(common.EEnvironmentVariable.ManagedIdentityClientID())
+			options.identityObjectID = common.GetEnvironmentVariable(common.EEnvironmentVariable.ManagedIdentityObjectID())
```
dphulkar-msft, line 124: I noticed that identityObjectID is defined as a local variable, while identityClientID is global. Would it make sense to keep these consistent for clarity? I do understand why values like certificatePassword and clientSecret are kept local, but for identifiers like these, aligning their scope might improve readability and maintainability.

cmd/login.go
```diff
@@ -40,40 +58,57 @@ var lgCmd = &cobra.Command{
...
+	if args.certificatePassword != "" || args.clientSecret != "" {
+		glcm.Info(environmentVariableNotice)
+	}
+
+	err := args.process()
+	if err != nil {
+		// the errors from adal contains \r\n in the body, get rid of them to make the error easier to look at
+		prettyErr := strings.Replace(err.Error(), `\r\n`, "\n", -1)
+		prettyErr += "\n\nNOTE: If your credential was created in the last 5 minutes, please wait a few minutes and try again."
+		glcm.Error("Failed to perform login command: \n" + prettyErr + getErrorCodeUrl(err))
```
dphulkar-msft, line 102: Should we not return from here?

cmd/login.go
```diff
@@ -114,66 +155,68 @@ type loginCmdArgs struct {
...
 	// Persist the token to cache, if login fulfilled successfully.
 
-	switch lca.loginType {
-	case common.EAutoLoginType.SPN().String():
-		if lca.certPath != "" {
-			if err := uotm.CertLogin(lca.tenantID, lca.aadEndpoint, lca.certPath, lca.certPass, lca.applicationID, lca.persistToken); err != nil {
+	switch options.LoginType {
+	case common.EAutoLoginType.SPN():
+		if options.CertificatePath != "" {
+			if err := uotm.CertLogin(options.TenantID, options.AADEndpoint, options.CertificatePath, options.certificatePassword, options.ApplicationID, options.persistToken); err != nil {
```
dphulkar-msft, line 183: Would it make sense to pass the entire options struct as an argument to CertLogin instead of listing out each individual field? Since all the values being passed are already part of the options structure, this could help simplify the function call and make the code a bit cleaner and easier to maintain.

cmd/login.go
```diff
@@ -114,66 +155,68 @@ type loginCmdArgs struct {
...
+			MSIResID: options.IdentityResourceID,
+		}, options.persistToken); err != nil {
 			return err
 		}
 		// For MSI login, info success message to user.
 		glcm.Info("Login with identity succeeded.")
-	case common.EAutoLoginType.AzCLI().String():
-		if err := uotm.AzCliLogin(lca.tenantID, lca.persistToken); err != nil {
+	case common.EAutoLoginType.AzCLI():
+		if err := uotm.AzCliLogin(options.TenantID, options.persistToken); err != nil {
```
dphulkar-msft, line 204: Just a thought—would it be worth considering passing the entire options struct to AzCliLogin, similar to the earlier suggestion for CertLogin? I realize the current argument list is short, but if we anticipate adding more values from the options struct in the future, having it passed directly could help keep the function signature cleaner and more flexible.

cmd/login.go
```diff
@@ -114,66 +155,68 @@ type loginCmdArgs struct {
...
 			return err
 		}
 		glcm.Info("Login with AzCliCreds succeeded")
-	case common.EAutoLoginType.PsCred().String():
-		if err := uotm.PSContextToken(lca.tenantID, lca.persistToken); err != nil {
+	case common.EAutoLoginType.PsCred():
+		if err := uotm.PSContextToken(options.TenantID, options.persistToken); err != nil {
```
dphulkar-msft, line 209: same as above

cmd/login.go
```diff
@@ -114,66 +155,68 @@ type loginCmdArgs struct {
...
 			return err
 		}
 		glcm.Info("Login with Powershell context succeeded")
-	case common.EAutoLoginType.Workload().String():
-		if err := uotm.WorkloadIdentityLogin(lca.persistToken); err != nil {
+	case common.EAutoLoginType.Workload():
+		if err := uotm.WorkloadIdentityLogin(options.persistToken); err != nil {
```
dphulkar-msft, line 214: same as above

cmd/login.go
```diff
@@ -114,66 +155,68 @@ type loginCmdArgs struct {
...
 			return err
 		}
 		glcm.Info("Login with Workload Identity succeeded")
 	default:
-		if err := uotm.UserLogin(lca.tenantID, lca.aadEndpoint, lca.persistToken); err != nil {
+		if err := uotm.UserLogin(options.TenantID, options.AADEndpoint, options.persistToken); err != nil {
```
dphulkar-msft, line 219: ditto

wonwuakpa-msft (approved):

cmd/credentialUtil.go
gapra-msft: you're right - will add the return here

cmd/credentialUtil.go
gapra-msft: I am making it local since we have since deprecated the parameter - so I don't want it to be publicly visible when we expose it in AzCopy as a library

cmd/login.go
gapra-msft: Since these methods are in the common package, passing in the entire options struct would mean I would have to make some of the non public facing options global so that is not quite the best option here.

cmd/login.go
```diff
@@ -114,66 +155,68 @@ type loginCmdArgs struct {
...
+	switch options.LoginType {
+	case common.EAutoLoginType.SPN():
+		if options.CertificatePath != "" {
+			if err := uotm.CertLogin(options.TenantID, options.AADEndpoint, options.CertificatePath, options.certificatePassword, options.ApplicationID, options.persistToken); err != nil {
 				return err
 			}
 			glcm.Info("SPN Auth via cert succeeded.")
 		} else {
-			if err := uotm.SecretLogin(lca.tenantID, lca.aadEndpoint, lca.clientSecret, lca.applicationID, lca.persistToken); err != nil {
+			if err := uotm.SecretLogin(options.TenantID, options.AADEndpoint, options.clientSecret, options.ApplicationID, options.persistToken); err != nil {
```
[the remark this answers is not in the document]
gapra-msft: see above

cmd/login.go
gapra-msft: see above

cmd/login.go
gapra-msft: see above

cmd/login.go
gapra-msft: see above

cmd/login.go
gapra-msft: see above

cmd/login.go
gapra-msft: It's slightly redundant since it's the last part of code to execute, but I will add it in case we ever end up adding code below.

dphulkar-msft (dismissed):

adreed-msft (dismissed):

PR #3074, opened 2025-06-04
Track request lifetimes, add it as a scaling parameter in the concurrency tuner, default to always tuning.

adreed-msft:
## Description

- **Feature / Bug Fix**: 

This half-fixes a bug that's evoked in situations where the number of goroutines we initialize far out-scales the available network bandwidth. It can be evoked by setting `AZCOPY_CONCURRENCY_VALUE` super high and `--cap-mbps` super low.

The janky fix present in this PR is to default the auto-scaler to on. Because it's seeking the optimal output, and trying to not over-scale, scaling past the network's capabilities will cause it to scale down as throughput didn't increase as expected. This _does not_ fix the problem, but it does work around it for the grand and wide majority of users. More PRs are to come of this.

- **Related Links**: StgExp Sync chat

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Manual testing, set benchmark with a 100gb upload and 5mbps cap-mbps. Request lifetime stabilizes out to around a minute max after the concurrency tuner settles. This still isn't great, it's a really severe issue, but it does mean the job _works_, which is way better than it _not working_.

A user manually specifying both could still trigger the underlying behavior.

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -1432,10 +1432,6 @@ func (cca *CookedCopyCmdArgs) getSrcCredential(ctx context.Context, jpo *common.
 // dispatches the job order (in parts) to the storage engine
 func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 	ctx := context.WithValue(context.TODO(), ste.ServiceAPIVersionOverride, ste.DefaultServiceApiVersion)
-	// Make AUTO default for Azure Files since Azure Files throttles too easily unless user specified concurrency value
-	if jobsAdmin.JobsAdmin != nil && (cca.FromTo.From() == common.ELocation.File() || cca.FromTo.To() == common.ELocation.File()) && common.GetEnvironmentVariable(common.EEnvironmentVariable.ConcurrencyValue()) == "" {
-		jobsAdmin.JobsAdmin.SetConcurrencySettingsToAuto()
-	}
 
 	if err := common.VerifyIsURLResolvable(cca.Source.Value); cca.FromTo.From().IsRemote() && err != nil {
 		return fmt.Errorf("failed to resolve source: %w", err)
--- a/common/linkedList.go
+++ b/common/linkedList.go
@@ -37,6 +37,14 @@ func (l *LinkedList[T]) PopRear() {
 	}
 }
 
+func (l *LinkedList[T]) Front() T {
+	return l.front.data
+}
+
+func (l *LinkedList[T]) Back() T {
+	return l.back.data
+}
+
 func (l *LinkedList[T]) Insert(data T) {
 	if l.front == nil {
 		l.front = &linkedListElement[T]{
--- a/common/linkedList.go
+++ b/common/linkedList.go
@@ -1,0 +1,74 @@
+package common
+
+// Singly linked list intended for buckets for ste.RequestLifetimeTracker
+
+type LinkedList[T any] struct {
+	front *linkedListElement[T]
+	back  *linkedListElement[T]
+	len   int64
+}
+
+type linkedListElement[T any] struct {
+	next *linkedListElement[T]
+	data T
+}
+
+func (l *LinkedList[T]) Enum() *LinkedListEnumerator[T] {
+	return &LinkedListEnumerator[T]{
+		index:  0,
+		target: l.back,
+	}
+}
+
+func (l *LinkedList[T]) Len() int64 {
+	return l.len
+}
+
+func (l *LinkedList[T]) PopRear() {
+	if l.back == nil {
+		return
+	}
+
+	l.back = l.back.next
+	l.len--
+
+	if l.back == nil {
+		l.back = l.front
+	}
+}
+
+func (l *LinkedList[T]) Insert(data T) {
+	if l.front == nil {
+		l.front = &linkedListElement[T]{
+			data: data,
+		}
+		l.back = l.front
+	} else {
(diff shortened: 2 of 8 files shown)


cmd/copy.go
```diff
@@ -1432,10 +1432,6 @@ func (cca *CookedCopyCmdArgs) getSrcCredential(ctx context.Context, jpo *common.
 // dispatches the job order (in parts) to the storage engine
 func (cca *CookedCopyCmdArgs) processCopyJobPartOrders() (err error) {
 	ctx := context.WithValue(context.TODO(), ste.ServiceAPIVersionOverride, ste.DefaultServiceApiVersion)
-	// Make AUTO default for Azure Files since Azure Files throttles too easily unless user specified concurrency value
-	if jobsAdmin.JobsAdmin != nil && (cca.FromTo.From() == common.ELocation.File() || cca.FromTo.To() == common.ELocation.File()) && common.GetEnvironmentVariable(common.EEnvironmentVariable.ConcurrencyValue()) == "" {
-		jobsAdmin.JobsAdmin.SetConcurrencySettingsToAuto()
```
wonwuakpa-msft, line 1018: Would the concurrency env var no longer be available for cx to set?

ste/xferLifetimeTracker.go
```diff
@@ -0,0 +1,300 @@
...
+// workers sit in the background and manage the running averages and buckets, ticking with the bucket ticker.
+
+func (r *RequestLifetimeTracker) queueWorker() {
+	requestBucketCount := int64(requestTrackingWindow / requestBucketLifetime)
+
+	for {
+		select { // This thread should die with the program to prevent any hangups
+		case <-r.bucketSwapTicker.C:
+			var rCount, lCount int64
+			var rSum, rAvg, rMax time.Duration
```
gapra-msft, line 136: It might be good to comment what these are tracking - Are the r prefixed ones, historical request metrica and the l prefixed ones live request ones?

gapra-msft: Oh I see how its used below. Yeah would be good to document it here just for code readability.

ste/xferLifetimeTracker.go
```diff
@@ -0,0 +1,300 @@
...
+				}
+
+				maxDiff := float64(rMax) / float64(rAvg) // If the max request lifetime significantly exceeds the average,
+				if maxDiff > 2 {                         // We should probably start to slow down a little.
+					atomic.StoreInt64(r.atomicUnexpectedRequestLifetime, 1)
+				} else {
+					// Before we say no, what about our current live requests?
+					out := int64(0)
+					for _, v := range r.liveRequestInitiationTimes {
+						if float64(time.Now().Sub(v))/float64(rAvg) > 2 { // At bare minimum, a 40 second request would be pretty concerningly unusual.
```
gapra-msft: where do we get this 40 sec number from?

ste/concurrencyTuner.go
```diff
@@ -70,6 +71,9 @@ type autoConcurrencyTuner struct {
 		value  int
 		reason string
 	}
+	// atomicCurrentConcurrency should never be written to, it is a read only reference to the job manager
+	atomicCurrentConcurrency *int32
```
gapra-msft, line 75: where do we read this?

common/linkedList.go
```diff
@@ -0,0 +1,82 @@
+package common
```
dphulkar-msft, line 21: copyright header required

ste/concurrencyTuner.go
```diff
@@ -172,10 +181,46 @@ func (t *autoConcurrencyTuner) worker() {
...
+			concurrency = concurrency * requestLifetimeBackoffMultiplier
+			if concurrency <= 4 { // 4 should be bare minimum
+				concurrency = 4
+			}
+			rateChangeReason = concurrencyReasonHighRequestLifetime
+
+			lastReason = t.setConcurrency(concurrency, rateChangeReason)
+			exitChannel := make(chan bool, 1)
+			go func() {
+				for {
```
dphulkar-msft, line 196: Should we handle it this way just to ensure the goroutine doesn't get stuck if the channel is already full?
`for {
		if requestLifetimeTracker.SimultaneousLiveRequests() < int64(concurrency) {
			select {
			case exitChannel <- true:
			default: // prevent blocking if channel already has a value
			}
			return
		}
		time.Sleep(requestBucketLifetime)
	}
`

ste/xferLifetimeTracker.go
```diff
@@ -0,0 +1,300 @@
+package ste
```
dphulkar-msft, line 21: copyright header

cmd/copy.go
gapra-msft: This particular part of the code is setting the concurrency to Auto for Azure Files by default. Since we are changing the default to Auto for all this code is no longer necessary.

common/linkedList.go
gapra-msft: I can add this since Adele is OOF

ste/concurrencyTuner.go
gapra-msft: Let's address this as a cleanup item but I dont see this being written to or read anywhere so I dont think we need this.

gapra-msft (approved):

gapra-msft (approved):

PR #3078, opened 2025-06-10
Update hardlink flag name and check the sudo permission for the process

dphulkar-msft:
## Description

1. Change flag name from --preserve-hardlinks to --hardlinks flag as it will accept multile values like follow, preserve and skip.
2. Check the sudo permission for user running the azcopy in case of NFS to local linux with --preserve-permissions=false. If sudo permissions are not provided azcopy will fail to set the UID,GID of a file/folder to 0.

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -492,7 +492,7 @@ type cookedSyncCmdArgs struct {
 
 	deleteDestinationFileIfNecessary bool
 	isNFSCopy                        bool
-	hardlinks                        common.HardlinksOption
+	hardlinks                        common.HardlinkHandlingType
 	atomicSkippedSymlinkCount        uint32
 	atomicSkippedSpecialFileCount    uint32
 }
--- a/common/fe-ste-models.go
+++ b/common/fe-ste-models.go
@@ -1559,32 +1559,34 @@ func (pc *PerfConstraint) Parse(s string) error {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-var EHardlinksOption = HardlinksOption(0)
+var EHardlinkHandlingType = HardlinkHandlingType(0)
 
-var DefaultHardlinksOption = EHardlinksOption.Follow()
+var DefaultHardlinkHandlingType = EHardlinkHandlingType.Follow()
 
-type HardlinksOption uint8
+type HardlinkHandlingType uint8
 
 // Copy means copy the files to the destination as regular files
-func (HardlinksOption) Follow() HardlinksOption { return HardlinksOption(0) }
+func (HardlinkHandlingType) Follow() HardlinkHandlingType {
+	return HardlinkHandlingType(0)
+}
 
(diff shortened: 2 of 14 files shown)


common/fe-ste-models.go
```diff
@@ -1559,32 +1559,32 @@ func (pc *PerfConstraint) Parse(s string) error {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-var EPreserveHardlinksOption = PreserveHardlinksOption(0)
+var EHardlinksOption = HardlinksOption(0)
```
gapra-msft: Should this be HardlinkHandlingType to match SymlinkHandlingType?

I am fine if this comment gets resolved after the preview/merge to main

cmd/sync.go
```diff
@@ -925,5 +939,5 @@ func init() {
 	syncCmd.PersistentFlags().BoolVar(&raw.deleteDestinationFileIfNecessary, "delete-destination-file", false, "False by default. Deletes destination blobs, specifically blobs with uncommitted blocks when staging block.")
 	_ = syncCmd.PersistentFlags().MarkHidden("delete-destination-file")
 
-	syncCmd.PersistentFlags().StringVar(&raw.preserveHardlinks, PreserveHardlinksFlag, "follow", "Follow by default. Preserve hardlinks for NFS resources. This flag is only applicable when the source is NFS file share or the destination is NFS file share. Available options: skip, preserve, follow (default 'follow').")
+	syncCmd.PersistentFlags().StringVar(&raw.hardlinks, HardlinksFlag, "follow", "Follow by default. Preserve hardlinks for NFS resources. This flag is only applicable when the source is NFS file share or the destination is NFS file share. Available options: skip, preserve, follow (default 'follow').")
```
gapra-msft, line 942: Is there a good reason to diverge from the existing --preserve-XX pattern?

I'd prefer to stick with the existing pattern to not confuse customers. 

This comment can be addressed after preview if necessary

gapra-msft (approved):
I have a few comments that are non blocking for the preview release.

cmd/sync.go
dphulkar-msft: Since we are introducing this flag with values like follow, preserve, and skip, it makes more sense to name the flag --hardlinks, as --preserve-hardlinks=follow,preserve,skip can be somewhat misleading. For symlinks, we use separate flag names such as --preserve-symlinks and --follow-symlinks, which accept boolean values.

common/fe-ste-models.go
dphulkar-msft: incorporated in this PR

vibhansa-msft (approved):

PR #3083, opened 2025-06-12
Warns the user if multiple AzCopy processes are running 

wonwuakpa-msft:
## Description

This warns customers when they have more than one AzCopy instance running using .pid files

- **Related Links**:
- [ADO Item](https://msazure.visualstudio.com/One/_workitems/edit/32317249/)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/cmd/processesChecker.go
+++ b/cmd/processesChecker.go
@@ -8,34 +8,34 @@ import (
 )
 
 // WarnMultipleProcesses warns if there are multiple AzCopy processes running
-func WarnMultipleProcesses(directory string, currentPid int) error {
+func WarnMultipleProcesses(directory string, currentPid int) {
 	currPidFileName := fmt.Sprintf("%d.pid", currentPid)
 	// Made subdir to not clog main dir
 	pidsSubDir := path.Join(directory, "pids")
-	if err := os.MkdirAll(pidsSubDir, 0755); err != nil {
-		glcm.Error(fmt.Sprintf("error creating pids dir: %v", err))
+	err := os.MkdirAll(pidsSubDir, 0755)
+	if err != nil {
+		return
 	}
 	filePath := path.Join(pidsSubDir, currPidFileName) // E.g "\.azcopy\pids\\XXX.pid"
 
 	dir, err := os.ReadDir(pidsSubDir)
 	if err != nil {
-		glcm.Error(fmt.Sprintf("error reading dir: %v", err))
-		return err
+		return
 	}
 	for _, fileName := range dir {
 		if fileName.Name() != currPidFileName {
 			glcm.Warn(common.ERR_MULTIPLE_PROCESSES)
-			return fmt.Errorf("%w", ErrMultipleProcesses)
+
 		}
 	}
 	// Creates .pid file with specific pid
-	file, err := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
-	if err != nil {
-		glcm.Error(fmt.Sprintf("error creating the .pid file: %v", err))
-		return err
-	}
-	defer file.Close()
-	return nil
+	file, _ := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
+	defer func(file *os.File) {
+		err := file.Close()
+		if err != nil {
+
+		}
+	}(file)
 }
 
 // CleanUpPidFile removes the PID files after the AzCopy run has exited. Ensures we only check against
--- a/cmd/processesChecker_test.go
+++ b/cmd/processesChecker_test.go
@@ -21,17 +21,14 @@ func Test_WarnMultipleProcesses(t *testing.T) {
 	}() // Cleanup
 	a.NoError(err)
 
-	err1 := WarnMultipleProcesses("temp", 1)
-	a.NoError(err1)
+	WarnMultipleProcesses("temp", 1)
 
 	pid1 := path.Join(pidsDir, "1.pid")
 	_, err = os.Stat(pid1)
 	a.NoError(err, "first .pid file should exist")
 
 	// Act
-	err2 := WarnMultipleProcesses("temp", 2) // Additional AzCopy process, err message
-	a.EqualError(err2, common.ERR_MULTIPLE_PROCESSES)
-	a.NotNil(err2)
+	WarnMultipleProcesses("temp", 2) // Additional AzCopy process
 
 	dirEntry, _ := os.ReadDir(pidsDir)
 	// Check only one file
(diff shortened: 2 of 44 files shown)


main.go
```diff
@@ -65,6 +126,11 @@ func main() {
 	// If insufficient arguments, show usage & terminate
 	if len(os.Args) == 1 {
 		cmd.Execute(azcopyLogPathFolder, azcopyJobPlanFolder, 0, jobID)
+		// Output warning if more than one AzCopy process is running
+		pidFilePath := path.Join(azcopyAppPathFolder, "processes.pid")
+		currPid := os.Getpid()
+		warnMultipleProcesses(pidFilePath, currPid)
```
dphulkar-msft: Why are we calling this inside the if block?

main.go
```diff
@@ -21,17 +21,78 @@
...
+	}
+
+	if len(pids) > 1 {
+		glcm.Warn("More than one AzCopy process is running. It is best practice to run a single process per VM.")
+	}
+}
+
+// createAppendPidFile creates and append writes process IDs to a .pid file
+func createAppendPidFile(filePath string, currentPid int) {
+	file, err := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
```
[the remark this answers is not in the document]
dphulkar-msft: @wonwuakpa-msft We should handle race condition

cmd/processesChecker_test.go
```diff
@@ -0,0 +1,41 @@
...
+	err1 := WarnMultipleProcesses("temp", 1)
+	a.NoError(err1)
+
+	pid1 := path.Join(pidsDir, "1.pid")
+	_, err = os.Stat(pid1)
+	a.NoError(err, "first .pid file should exist")
+
+	// Act
+	err2 := WarnMultipleProcesses("temp", 2) // Additional AzCopy process, warning should output
+	a.NoError(err2)
```
gapra-msft: Shouldnt err be NotNil here?
update: oh nvm I see the code just outputs a warning log. Should we consider making this method return a custom error and ignore it in root to make testing easier?

wonwuakpa-msft: @gapra-msft I was thinking it should be a warning and not an error because running multiple AzCopy instances should not show stopping. It warns but still continues the operation. Kind of like what we have with the version checker. So, the testing would check for nil and no error. What do you think?

gapra-msft: Yes, we can keep the same functionality but design this method a little better to make it more testable - the WarnMultipleProcesses function can return an error and then in root where we call it, we can ignore that specific error and log a warning instead of logging an error. Then we can go in and actually test that the WarnMultipleProcesses function actually identifies the other pid file.

wonwuakpa-msft: Ohh I see, I'll make the change

cmd/processesChecker.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"os"
+	"path"
+	"regexp"
+	"strconv"
+)
+
+// WarnMultipleProcesses warns if there are multiple AzCopy processes running
+func WarnMultipleProcesses(directory string, currentPid int) error {
+	currPidFileName := strconv.Itoa(currentPid) + ".pid"
```
adreed-msft: Give sprintf a shot. If you check the [fmt](https://pkg.go.dev/fmt) docs, there's a list of all the magic strings you can use. It's not quite as nice as C#'s string interpolation but it still gives us a nice way to format strings in a maintainable fashion should we ever need to adjust it.

```suggestion
	currPidFileName := fmt.Sprintf("%d.pid", currentPid)
```

cmd/processesChecker.go
```diff
@@ -0,0 +1,56 @@
...
+
+	pidsSubDir := path.Join(directory, "pids") // Made subdir to not clog main dir
+	if err := os.MkdirAll(pidsSubDir, 0755); err != nil {
+		glcm.Error(fmt.Sprintf("error creating pids dir: %v", err))
+	}
+	filePath := path.Join(pidsSubDir, currPidFileName) // E.g "\.azcopy\pids\\XXX.pid"
```
adreed-msft: Creating the PID file feels like it should go after checking, that way we don't have to worry about filtering our own file out.

cmd/processesChecker.go
```diff
@@ -0,0 +1,56 @@
...
+	// Creates .pid file with specific pid
+	file, err := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
+	if err != nil {
+		glcm.Error(fmt.Sprintf("error creating the .pid file: %v", err))
+		return err
+	}
+	defer file.Close()
+
+	// Check if there is a matching .pid file in directory
+	pidFileRegex, _ := regexp.Compile(`\.pid$`)
```
adreed-msft: On its own, the regex feels quite unnecessary. But...

cmd/processesChecker.go
```diff
@@ -0,0 +1,56 @@
...
+	dir, err := os.ReadDir(pidsSubDir)
+	if err != nil {
+		glcm.Error(fmt.Sprintf("error reading dir: %v", err))
+		return err
+	}
+	for _, fileName := range dir {
```
adreed-msft, line 25: If we're gonna have the regex anyway, I think it's worth calling [os.FindProcess](https://pkg.go.dev/os#FindProcess) to check that the PID is still valid and alive (and cleaning the pid file up if not), otherwise, if AzCopy panics and doesn't exit gracefully, we're still guaranteed to clean it up on future runs.

It's not a _perfect_ fix to the problem, because what if another process takes up residence on the same PID?

At the very least, this'd prevent a customer from being left scratching their head after AzCopy crashed and they just. Eternally got this warning for the rest of time until they figured out the pids folder was the problem.

cmd/processesChecker_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	err1 := WarnMultipleProcesses("temp", 1)
+	a.NoError(err1)
+
+	pid1 := path.Join(pidsDir, "1.pid")
+	_, err = os.Stat(pid1)
+	a.NoError(err, "first .pid file should exist")
+
+	// Act
+	err2 := WarnMultipleProcesses("temp", 2) // Additional AzCopy process, warning should output
```
adreed-msft: We could set up a mock LCM (I believe this already happens in the unit tests) to validate this functionality and make sure the expected warning gets output

cmd/processesChecker_test.go
```diff
@@ -0,0 +1,118 @@
...
+//	runCopyAndVerify(a, raw, func(err error) {
+//		runCopyAndVerify(a, raw, func(err error) {
+//			warnMessages := mockedLcm.GatherAllLogs(mockedLcm.warnLog)
+//			a.Equal("more than one AzCopy process is running", warnMessages)
+//			a.True(len(warnMessages) > 0)
+//
+//		})
+//	})
+//
+//}
```
dphulkar-msft: if not required we can remove it

PR #3087, opened 2025-06-16
Add support for transfer data Azure Files NFS <-> Azure Files SMB

dphulkar-msft:
## Description

**SMB → NFS**
1. Symlinks (There will be no symlink for SMB share)
    --preserve-symlinks: N/A fail job
    --follow-symlinks: N/A  fail the job
2. Hardlinks (There will be no symlink for SMB share)
    --hardlinks: N/A fail the job
3. Azure Files SMB will only contain normal files (no symlinks or hardlinks).
4. If --preserve-info is set, preserve creationTime and lastModifiedTime.
5. If --preserve-permissions is set, do not preserve — log error: "Preserving cross-platform permissions is not supported."

**NFS → SMB**
1. Symlinks: Skip and log (no symlink api support for SMB).
2. Hardlinks: Skip and log (no hardlink api support for SMB).
3. For NFS → SMB transfers, --hardlinks must be set to skip. Hardlinked files will always be skipped during this transfer.
4. If --preserve-info is set, preserve creationTime and lastModifiedTime.
5. If --preserve-permissions is set, do not preserve — log error: "Preserving cross-platform permissions is not supported."

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?


common/util.go
```diff
@@ -411,14 +412,19 @@ func IsSystemContainer(containerName string) bool {
...
+	isNFSCopy bool
+	setOnce   sync.Once
+)
 
-func SetNFSFlag(isNFS bool) {
-	// SetNFSFlag sets the global isNFSCopy variable to the given value
-	isNFSCopy = isNFS
+// SetIsNFSCopy sets the flag only once for the process lifetime
+func SetIsNFSCopy(fromTo FromTo) {
+	setOnce.Do(func() {
```
gapra-msft: Can we avoid this global? Having globals like this actually make AzCopy as a library super hard to migrate since we need to reset these on each copy/sync/resume job

jobsAdmin/init.go
```diff
@@ -466,8 +466,8 @@ func resurrectJobSummary(jm ste.IJobMgr) common.ListJobSummaryResponse {
 				js.FolderPropertyTransfers++
 			case common.EEntityType.Symlink():
 				js.SymlinkTransfers++
-			case common.EEntityType.Hardlink():
-				js.HardlinksConvertedCount++
+				// case common.EEntityType.Hardlink():
```
gapra-msft: why is this commented?

e2etest/zt_newe2e_nfs_scenarios_test.go
```diff
@@ -531,7 +530,7 @@ func (s *FilesNFSTestSuite) Scenario_AzureNFSToAzureNFS(svm *ScenarioVariationMa
 				},
 			},
 		})
-
+	fmt.Println("Stdout:--------------", stdOut)
```
gapra-msft: please remove this before merging

cmd/flagsValidation.go
```diff
@@ -30,59 +30,80 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var nfsPermPreserveXfers = map[common.FromTo]bool{
```
gapra-msft, line 33: is this the best way to do this? I feel like it'd be more efficient and less overhead to do a switch case check

cmd/flagsValidation.go
```diff
@@ -346,38 +416,38 @@ func validateProtocolCompatibility(ctx context.Context, fromTo common.FromTo, sr
...
 
-	var protocol string
+	var srcProtocol, dstProtocol string
 
 	// S2S Transfers
 	if fromTo.IsS2S() {
 		switch fromTo {
 		case common.EFromTo.FileFile():
-			protocol = "SMB"
+			srcProtocol, dstProtocol = "SMB", "SMB"
```
gapra-msft: Is this necessary to pass as a string? Could we just pass fromTo?

jobsAdmin/init.go
dphulkar-msft: As with phase-2 coming in we have multiple options for hardlink files like preserve, skip and follow. And we want to club this count based on the type of hardlink preservation mode. Due to this, count logic is now being moved to traversal phase.

cmd/flagsValidation.go
adreed-msft: ditto

common/util.go
dphulkar-msft: done

cmd/flagsValidation.go
dphulkar-msft: This check is for protocol type. I have matched it to location now. So the protocol would be either common.ELocation.File() or common.Elocation.FileNFS()

cmd/flagsValidation.go
dphulkar-msft: just to avoid multiple || statments in if condition I have clubbed them in a map.

```
if toPreserve {
		// The user cannot preserve permissions between SMB->NFS or NFS->SMB transfers.
		if flagName == PreservePermissionsFlag && (fromTo == common.EFromTo.FileNFSFileSMB() || fromTo == common.EFromTo.FileSMBFileNFS()) {
			return fmt.Errorf("--preserve-permissions flag is not supported for cross-protocol transfers (e.g., SMB<->NFS, NFS<->SMB). Please remove this flag and try again.")
		} **else if !nfsPermPreserveXfers[fromTo]** {
			return fmt.Errorf("%s is set but the job is not between %s-aware resources", flagName, common.Iff(flagName == PreserveInfoFlag, "permission", "NFS"))
		} else if (fromTo.IsUpload() || fromTo.IsDownload()) && runtime.GOOS != "linux" {
			return fmt.Errorf("%s is set but persistence for up/downloads is supported only in Linux", flagName)
		}
	}
```

ste/sender-azureFile.go
```diff
@@ -194,7 +194,7 @@ func (u *azureFileSenderBase) Prologue(state common.PrologueState) (destinationM
 		Metadata:    u.metadataToApply,
 	}
 
-	if common.IsNFSCopy() {
+	if common.IsNFSCopy(jptm.FromTo()) {
```
gapra-msft: Can we make this a method on FromTo?

we should be able to call jptm.FromTo().IsNFSAware()

common/util.go
```diff
@@ -411,14 +411,6 @@ func IsSystemContainer(containerName string) bool {
...
-
-func SetNFSFlag(isNFS bool) {
-	// SetNFSFlag sets the global isNFSCopy variable to the given value
-	isNFSCopy = isNFS
-}
-
-func IsNFSCopy() bool {
-	return isNFSCopy
+func IsNFSCopy(fromTo FromTo) bool {
+	return fromTo.From() == ELocation.FileNFS() || fromTo.To() == ELocation.FileNFS()
```
gapra-msft: Will this work when we do something like FileNFS -> FileSMB and vice versa?

adreed-msft (approved):

common/util.go
dphulkar-msft: yes this would work. We can discuss this over a call.

ste/sender-azureFile.go
dphulkar-msft: done

gapra-msft (approved):

PR #3093, opened 2025-06-18
Add standard blob symlink tests

adreed-msft:
## Description

- Testing: Adds simple test cases surrounding persist and follow symlinks

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Code quality improvement (test)
- [ ] Other (describe):

## How Has This Been Tested?

It is tests.

### DIFF
--- a/e2etest/newe2e_resource_managers_file.go
+++ b/e2etest/newe2e_resource_managers_file.go
@@ -860,6 +860,11 @@ func (f *FileObjectResourceManager) Download(a Asserter) io.ReadSeeker {
 	return bytes.NewReader(buf.Bytes())
 }
 
+func (f *FileObjectResourceManager) ReadLink(a Asserter) string {
+	a.Error("Symlinks are unsupported on Files.")
+	return ""
+}
+
 func (f *FileObjectResourceManager) Exists() bool {
 	var err error
 	if f.entityType != common.EEntityType.Folder() {
--- a/e2etest/zt_newe2e_blob_symlink_test.go
+++ b/e2etest/zt_newe2e_blob_symlink_test.go
@@ -1,0 +1,134 @@
+package e2etest
+
+import (
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"path"
+	"runtime"
+)
+
+type BlobSymlinkSuite struct{}
+
+func init() {
+	if runtime.GOOS != "windows" || GlobalConfig.E2EFrameworkSpecialConfig.OverrideWindowsSymlinkSkip {
+		// Windows symlinks are very funny. They need special privileges to create.
+		suiteManager.RegisterSuite(&BlobSymlinkSuite{})
+	}
+}
+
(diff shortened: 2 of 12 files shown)


e2etest/zt_newe2e_blob_symlink_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	source := CreateResource[ContainerResourceManager](svm, GetRootResource(svm, srcLoc), srcDef)
+
+	dest := CreateResource[ContainerResourceManager](svm, GetRootResource(svm, dstLoc), ResourceDefinitionContainer{})
+
+	_, _ = RunAzCopy(
+		svm,
+		AzCopyCommand{
+			Verb: AzCopyVerbCopy, // sync doesn't support symlinks at this time
```
dphulkar-msft, line 56: Should we test the symlink for sync as well? The error scenario. As with NFS coming in we are going to expose the --preserve-symlink and --follow-symlinks flag in case of sync as well.

e2etest/zt_newe2e_blob_symlink_test.go
```diff
@@ -0,0 +1,134 @@
...
+			},
+		},
+	})
+
+	dest := CreateResource[ContainerResourceManager](svm, GetRootResource(svm, common.ELocation.Blob()), ResourceDefinitionContainer{})
+
+	_, _ = RunAzCopy(
+		svm,
+		AzCopyCommand{
+			Verb: AzCopyVerbCopy, // sync doesn't support symlinks at this time
```
dphulkar-msft, line 111: same comment as above

e2etest/zt_newe2e_blob_symlink_test.go
gapra-msft: I think we can skip that for now and add the sync case when support comes for it.

gapra-msft (approved):

dphulkar-msft (approved):

e2etest/newe2e_resource_managers_file.go
```diff
@@ -860,6 +860,11 @@ func (f *FileObjectResourceManager) Download(a Asserter) io.ReadSeeker {
 	return bytes.NewReader(buf.Bytes())
 }
 
+func (f *FileObjectResourceManager) ReadLink(a Asserter) string {
+	a.Error("Symlinks are unsupported on Files.")
```
wonwuakpa-msft, line 873: Based on this, https://learn.microsoft.com/en-us/rest/api/storageservices/create-symbolic-link aren't they supported?

wonwuakpa-msft (approved):

e2etest/newe2e_resource_managers_file.go
gapra-msft: this is for NFS which is coming in this release, we can add tests as part of the NFS effort once that support is fully added

gapra-msft (approved):

PR #3105, opened 2025-06-26
Update retry policy to include copy source error and status codes 

wonwuakpa-msft:
## Description

Modified the retry policy to include new headers `x-ms-copy-source-error-code` and` x-ms-copy-source-status-code
`
- **Related Links**:
- [Issues](https://msazure.visualstudio.com/One/_backlogs/backlog/XStore-DevExp/Epics?showParents=false&workitem=27249523)

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Unit tests in TestGetShouldRetry


ste/mgr-JobPartMgr.go
```diff
@@ -111,6 +111,7 @@ func NewClientOptions(retry policy.RetryOptions, telemetry policy.TelemetryOptio
 		perRetryPolicies = append(perRetryPolicies, NewSourceAuthPolicy(srcCred))
 	}
 	retry.ShouldRetry = GetShouldRetry(&log)
+	retry.StatusCodes = GetRetryStatusCodes(&log)
```
gapra-msft: Why are we adding this variable? We don't want to retry on every retry status code key the user provides, we only want to retry if the status code and error code both match (or just status code if there is no error code shared)

ste/xferRetryHelper.go
```diff
@@ -71,11 +90,27 @@ func GetShouldRetry(log *LogOptions) RetryFunc {
...
+	}
+	if resp.Header["x-ms-copy-source-error-code"] != nil { //nolint:staticcheck
+		errorCodes = append(errorCodes, resp.Header["x-ms-copy-source-error-code"][0]) //nolint:staticcheck
+	} else if resp.Header["X-Ms-Copy-Source-Error-Code"] != nil { //nolint:staticcheck
+		errorCodes = append(errorCodes, resp.Header["X-Ms-Copy-Source-Error-Code"][0]) //nolint:staticcheck
+	}
+	return errorCodes
+}
+
+func getCopySourceStatusCode(resp *http.Response) string {
```
adreed-msft, line 109: [`*http.Header.Values`](https://pkg.go.dev/net/http#Header.Values) is case insensitive, also `len(nil) == 0`.

ste/xferRetryHelper.go
```diff
@@ -71,11 +90,27 @@ func GetShouldRetry(log *LogOptions) RetryFunc {
...
-func getErrorCode(resp *http.Response) string {
+func getErrorCodes(resp *http.Response) []string {
+	// There can be multiple error code headers per response
+	var errorCodes []string
 	if resp.Header["x-ms-error-code"] != nil { //nolint:staticcheck
-		return resp.Header["x-ms-error-code"][0] //nolint:staticcheck
-	} else if resp.Header["X-Ms-Error-Code"] != nil {
-		return resp.Header["X-Ms-Error-Code"][0]
+		errorCodes = append(errorCodes, resp.Header["x-ms-error-code"][0]) //nolint:staticcheck
+	} else if resp.Header["X-Ms-Error-Code"] != nil { //nolint:staticcheck
```
adreed-msft, line 98: ditto

ste/xferRetryHelper.go
```diff
@@ -58,10 +62,25 @@ func GetShouldRetry(log *LogOptions) RetryFunc {
 					} else if !ok && storageErrorCodes["*"] {
 						return true
 					}
+
+				}
+			}
+			// Check if copy source status code is present
+			respStatusCode := getCopySourceStatusCode(resp)
```
adreed-msft, line 69: I'm trying to think of some case where we might not want to retry when the source returns a status code we would typically retry the destination under... But I'm pulling blanks. I *think* this is fine.

ste/xferRetryHelper.go
gapra-msft: I think this is to deal with how the Go SDK parses the response - it usually comes back as camelcase

gapra-msft (approved):

PR #3111, opened 2025-07-02
Warns the cx if job exceeds 10M objects

wonwuakpa-msft:
## Description

Outputs a warning when one job exceeds more than 10 million objects. This does not abort the operation. AzCopy best practice is to run less than that limit per job. 

- **Related Links**:
- [Issues](https://msazure.visualstudio.com/One/_backlogs/backlog/XStore-DevExp/Epics?showParents=true&workitem=32317270)
- 
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Manually tested by setting the max to a smaller value (2 objects). Warning message is outputted in stdout
image


cmd/syncEnumerator.go
```diff
@@ -117,6 +120,10 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 			if entityType == common.EEntityType.File() {
 				atomic.AddUint64(&cca.atomicDestinationFilesScanned, 1)
 			}
+			if cca.atomicScanningStatus == 1 &&
+				atomic.LoadUint64(&cca.atomicDestinationFilesScanned) > common.RECOMMENDED_OBJECTS_COUNT {
+				WarnStdoutAndScanningLog("This job contains more than 10M objects, best practice to run less than this.")
```
gapra-msft: error message should use the common.RecommendedObjectsCount

gapra-msft:
Couple things to watch out for -> what if the customer is running a copy or resume? This logic might be better placed in ste? Something worth considering. Also we need to make sure this is only logged once.

cmd/syncEnumerator.go
```diff
@@ -82,6 +82,9 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 					atomic.AddUint32(&cca.atomicSkippedSymlinkCount, 1)
 				}
 			}
+			if atomic.LoadUint32(&cca.atomicScanningStatus) == 1 && atomic.LoadUint64(&cca.atomicSourceFilesScanned) > common.RECOMMENDED_OBJECTS_COUNT {
+				WarnStdoutAndScanningLog("This job contains more than 10M objects, best practice to run less than this.")
```
dphulkar-msft: Any reason we are not adding this warn message for copy operation? I understand that in case of sync we first enumerate the complete list at the source and destination first and then perform the copy while in copy we parallelly perform the copy with scanning.

cmd/syncEnumerator.go
dphulkar-msft: I'm not entirely sure about this—should we add a test for it? We could consider skipping it, as adding a test would require performing a 10 million record copy job, which would take a significant amount of time to complete.

dphulkar-msft (approved):

cmd/syncEnumerator.go
wonwuakpa-msft: Are you unsure about the warning message or testing it? Copying 10M files in a single test would be very long running and not feasible

dphulkar-msft: About adding a test for it.

common/environment.go
```diff
@@ -52,6 +52,10 @@ func ClearEnvironmentVariable(variable EnvironmentVariable) {
 	_ = os.Setenv(variable.Name, "")
 }
 
+func GetRecommendedMaxObjectsPerJob() uint64 {
```
gapra-msft: I'd move this function to fe-ste-models

ste/jobStatusManager.go
```diff
@@ -128,6 +128,9 @@ func (jm *jobMgr) handleStatusUpdateMessage() {
 			}
 			js.CompleteJobOrdered = js.CompleteJobOrdered || msg.IsFinalPart
 			js.TotalTransfers += msg.TotalTransfers
+			if js.TotalTransfers > uint32(common.GetRecommendedMaxObjectsPerJob()) {
```
gapra-msft: I believe this code will execute after we hit processing of the 10 millionth file. Is there any way we could surface the warning sooner? Like during enumeration?

gapra-msft: Could we somehow calculate this in ExecuteNewCopyJobPartOrder with the order.IsFinalPart and order.partNum? @adreed-msft @dphulkar-msft what are your thoughts here?

gapra-msft: or copyTransferProcessor.dispatchFinalPart?

dphulkar-msft: It would be helpful if AzCopy displayed a warning during the enumeration phase. For copy operations, it's less critical since transfers begin in parallel as files are enumerated in 10K batches. However, for sync operations, the transfer doesn't start until enumeration is fully completed.
In cases where the dataset is large, it would make sense to show a warning as soon as, for example, 10 million files have been scanned at the source.

cmd/syncEnumerator.go
```diff
@@ -117,6 +120,10 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 			if entityType == common.EEntityType.File() {
 				atomic.AddUint64(&cca.atomicDestinationFilesScanned, 1)
 			}
+			if cca.atomicScanningStatus == 1 &&
+				atomic.LoadUint64(&cca.atomicDestinationFilesScanned) > common.GetRecommendedMaxObjectsPerJob() {
```
wonwuakpa-msft: For the sync case, we check during enumeration for the 1 millionth file. @gapra-msft

ste/jobStatusManager.go
wonwuakpa-msft: If we do it in the STE, we would discover the 10 millionth file when it has been sent in batches in a job part not during the initial enumeration. Is this what we prefer? Also, we'd need to track the count of transfers across multiple job parts @gapra-msft @dphulkar-msft

gapra-msft: does the enumerator not track how many requests its sent?

common/fe-ste-models.go
```diff
@@ -1908,3 +1910,17 @@ func (sht *SymlinkHandlingType) Determine(Follow, Preserve bool) error {
 
 	return nil
 }
+
+// /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+var TooManyObjWarningOnce = sync.Once{}
+
+func GetRecommendedMaxObjectsPerJob() uint64 {
```
adreed-msft: This function just returns an exported constant. It's probably not necessary.

jobsAdmin/init.go
```diff
@@ -182,6 +182,12 @@ func ExecuteNewCopyJobPartOrder(order common.CopyJobPartOrderRequest) common.Cop
 	// Supply no plan MMF because we don't have one, and AddJobPart will create one on its own.
 	// Add this part to the Job and schedule its transfers
 
+	// Warn if more objects than recommended threshold
+	jm.AddTotalNumFilesProcessed(int64(len(order.Transfers.List)))
+	if jm.GetTotalNumFilesProcessed() > common.RECOMMENDED_OBJECTS_COUNT {
```
adreed-msft, line 187: why use both the exported constant and the function separately

jobsAdmin/init.go
```diff
@@ -521,6 +527,9 @@ func resurrectJobSummary(jm ste.IJobMgr) common.ListJobSummaryResponse {
 		jpp := jpm.Plan()
 		js.CompleteJobOrdered = js.CompleteJobOrdered || jpp.IsFinalPart
 		js.TotalTransfers += jpp.NumTransfers
+		if js.TotalTransfers > uint32(common.GetRecommendedMaxObjectsPerJob()) {
```
adreed-msft: constant numerics generally will type morph into the type they're used alongside

ste/jobStatusManager.go
adreed-msft: > I believe this code will execute after we hit processing of the 10 millionth file. Is there any way we could surface the warning sooner? Like during enumeration?

Yes-- Enumerators actually have an enumeration counter function we could plug into to serve that purpose. Though, it's worth noting this timing difference will only really be major for something like list of files.

adreed-msft (dismissed):
some question marks but otherwise is in a decent state

common/fe-ste-models.go
```diff
@@ -1908,3 +1910,17 @@ func (sht *SymlinkHandlingType) Determine(Follow, Preserve bool) error {
...
+	return RECOMMENDED_OBJECTS_COUNT
+}
+
+func WarnIfTooManyObjects(oncer *sync.Once) {
```
gapra-msft: Do we need to pass in the oncer here? We can just directly reference the var above right?

wonwuakpa-msft: nope we don't need to

ste/jobStatusManager.go
wonwuakpa-msft: Yes, for sync there's atomic tracking of the source files scanned. It is used in the increment enumeration function to determine if the max is hit and output a warning

cmd/syncEnumerator.go
gapra-msft: if sync also calls into CopyJobPartOrderReq in jobsAdmin, we can skip this check and just have it there.

ste/jobStatusManager.go
```diff
@@ -128,6 +128,9 @@ func (jm *jobMgr) handleStatusUpdateMessage() {
 			}
 			js.CompleteJobOrdered = js.CompleteJobOrdered || msg.IsFinalPart
 			js.TotalTransfers += msg.TotalTransfers
+			if js.TotalTransfers > common.RECOMMENDED_OBJECTS_COUNT {
```
gapra-msft: I dont think we need to add this here - just the one in ExecuteNewCopyJobPartOrder is sufficient. 

What scenario does this execute in that the above doesnt?

PR #3116, opened 2025-07-09
Clean up of JobsAdmin

gapra-msft:
## Description

- Rips out RPC related code that is no longer accurate to the architecture of AzCopy
- Makes logging a little more clear. Since we only have one job running at a time, there is one global instance of the logger. We had a lot of duplication of the logger in the past. We still do a bit, but this is slowly improving it.
- Moving a few more init things into the client which will be necessary for jobs show

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Refactor

### DIFF
--- a/azcopy/jobsShow.go
+++ b/azcopy/jobsShow.go
@@ -1,0 +1,71 @@
+// Copyright © 2025 Microsoft <wastore@microsoft.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package azcopy
+
+import (
+	"errors"
+	"fmt"
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"github.com/Azure/azure-storage-azcopy/v10/jobsAdmin"
+)
+
+// TODO: (gapra) We should refactor some of the common.XXXX types without the ErrorMsg field if possible.
+
+type GetJobSummaryOptions struct {
+	JobID common.JobID
+}
+
+type JobSummaryResponse common.ListJobSummaryResponse
+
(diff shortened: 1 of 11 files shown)


cmd/copyEnumeratorInit.go
```diff
@@ -193,8 +193,8 @@ func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 				logDstContainerCreateFailureOnce.Do(func() {
 					glcm.Warn("Failed to create one or more destination container(s). Your transfers may still succeed if the container already exists.")
 				})
-				jobsAdmin.JobsAdmin.LogToJobLog(fmt.Sprintf("Failed to create destination container %s. The transfer will continue if the container exists", dstContainerName), common.LogWarning)
-				jobsAdmin.JobsAdmin.LogToJobLog(fmt.Sprintf("Error %s", err), common.LogDebug)
+				common.LogToJobLogWithPrefix(fmt.Sprintf("Failed to create destination container %s. The transfer will continue if the container exists", dstContainerName), common.LogWarning)
+				common.LogToJobLogWithPrefix(fmt.Sprintf("Error %s", err), common.LogDebug)
```
dphulkar-msft: I see we are dumping error log so should it not come under LogError?

dphulkar-msft: %v instead of %s

cmd/copyEnumeratorInit.go
```diff
@@ -230,8 +230,8 @@ func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 						logDstContainerCreateFailureOnce.Do(func() {
 							glcm.Warn("Failed to create one or more destination container(s). Your transfers may still succeed if the container already exists.")
 						})
-						jobsAdmin.JobsAdmin.LogToJobLog(fmt.Sprintf("failed to initialize destination container %s; the transfer will continue (but be wary it may fail).", bucketName), common.LogWarning)
-						jobsAdmin.JobsAdmin.LogToJobLog(fmt.Sprintf("Error %s", err), common.LogDebug)
+						common.LogToJobLogWithPrefix(fmt.Sprintf("failed to initialize destination container %s; the transfer will continue (but be wary it may fail).", bucketName), common.LogWarning)
+						common.LogToJobLogWithPrefix(fmt.Sprintf("Error %s", err), common.LogDebug)
```
dphulkar-msft: ditto

cmd/copyEnumeratorInit.go
```diff
@@ -251,8 +251,8 @@ func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 						logDstContainerCreateFailureOnce.Do(func() {
 							glcm.Warn("Failed to create one or more destination container(s). Your transfers may still succeed if the container already exists.")
 						})
-						jobsAdmin.JobsAdmin.LogToJobLog(fmt.Sprintf("failed to initialize destination container %s; the transfer will continue (but be wary it may fail).", resName), common.LogWarning)
-						jobsAdmin.JobsAdmin.LogToJobLog(fmt.Sprintf("Error %s", err), common.LogDebug)
+						common.LogToJobLogWithPrefix(fmt.Sprintf("failed to initialize destination container %s; the transfer will continue (but be wary it may fail).", resName), common.LogWarning)
+						common.LogToJobLogWithPrefix(fmt.Sprintf("Error %s", err), common.LogDebug)
```
dphulkar-msft: ditto

azcopy/client_test.go
```diff
@@ -1,43 +0,0 @@
-package azcopy
```
wonwuakpa-msft, line 1: Why do we no longer want to test the client?

gapra-msft: We're testing it in e2e test so I am removing it here.

cmd/copyEnumeratorInit.go
gapra-msft: Changing to %v

gapra-msft: This is more of an informational error - the transfer will still try to happen after this point, so we arent logging a true error

dphulkar-msft (approved):

PR #3119, opened 2025-07-11
Handling pre unix epoch timestamps 

shnayak-msft:
## Description
This PR addresses an integer overflow issue in timestamp handling, particularly for dates before 1970 (unix epoch time), which are unsupported due to limitations in int64 and UnixNano usage.

**Root Cause**:
- UnixNano uses int64, supporting ±292 years from 1970, but our implementation only handled positive values.
- Pre-1970 timestamps failed due to incorrect use of time.Unix(0, nanoSeconds) with uint64.
- Even with signed values, the lower bound is 1678, leaving 1601–1678 unsupported.
- Additional overflow occurred in StatxTimestampToFiletime() due to unchecked arithmetic on seconds and nanoseconds.

**Fix:**
- Replaced Unix epoch base with Windows epoch time (1601) in nanoseconds.
- Used uint64 to represent time, extending support from 1601 to 2185.
- Preserved existing struct definitions and ensured backward compatibility.
- Implemented workarounds for Go’s time package limitations on nanosecond precision

**Status:**
- Manual testing passed.
- Some pipeline failures observed due to precision mismatches between Go time and filesystem time (under investigation).

## Type of Change

- [x ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/ste/xfer-anyToRemote-file.go
+++ b/ste/xfer-anyToRemote-file.go
@@ -337,7 +337,7 @@ func anyToRemote_file(jptm IJobPartTransferMgr, info *TransferInfo, pacer pacer,
 			jptm.ReportTransferDone()
 			return
 		}
-		if !lmt.Equal(jptm.LastModifiedTime()) {
+		if common.ToWindowsEpoch(lmt) != jptm.LastModifiedEpochTime() {
 			jptm.LogSendError(info.Source, info.Destination, "File modified since transfer scheduled", 0)
 			jptm.SetStatus(common.ETransferStatus.Failed())
 			jptm.ReportTransferDone()


ste/JobPartPlan.go
```diff
@@ -362,7 +362,7 @@ type JobPartPlanTransfer struct {
 	// We use a dedicated field for this because the alternative (of doing something fancy the names) was too complex and error-prone
 	EntityType common.EntityType
 	// ModifiedTime represents the last time at which source was modified before start of transfer stored as nanoseconds.
-	ModifiedTime int64
+	ModifiedTime uint64
```
gapra-msft, line 365: Please be sure to update the schema version at the top of this file.

PR #3136, opened 2025-07-29
Enable external custom credential provider in the S3toBlob case

otendolkar:
The following changes enable a custom credential provider to be utilized in the case that setting static credentials through env variable is not sufficient, specifically for s3 to blob scenario. It can only be done if using azcopy as a library, in all other cases the provider will be a NULL field wherever it has been added.

-cmd/credentialUtil.go: The change comments out the logic that sets credType to S3PublicBucket when AWS access keys are missing, defaulting instead to S3AccessKey.

-cmd/syncEnumerator.go: Adds support for injecting a custom S3 credential provider into the job context, specifically for sync case

- cmd/zc_traverser_s3.go: Adds support for injecting custom S3 credential providers via context in the S3 traverser to enable flexible authentication in the copy case

-common/credentialFactory.go: Enables the use of the provider in the s3 client creation workflow

-common/rpc-models.go: Adds support for custom credential providers in S3CredentialInfo and CopyJobPartOrderRequest  structures so cmd can pass it along its ste workflow

-JobsAdmin/init.go: Adds provider in the InMemoryTransitJobState struct so that it can be passed through ste workflow

-ste/mgr-JobMgr.go: Adds provider to appropriate jobmgr structs to pass down through ste workflow

-ste/mgr-JobPartMgr.go: Adds provider to appropriate jobpartmgr structs to pass down through ste workflow

-ste/mgr-JobPartTransferMgr.go: Adds provider to appropriate jobparttransfermgr structs to pass down through ste workflow

-ste/sourceInfoProvider-S3.go: Adds provider to s3 credential info for ste client

### DIFF
--- a/cmd/credentialUtil.go
+++ b/cmd/credentialUtil.go
@@ -456,14 +456,16 @@ func doGetCredentialTypeForLocation(ctx context.Context, location common.Locatio
 	}
 
 	if location == common.ELocation.S3() {
-		accessKeyID := common.GetEnvironmentVariable(common.EEnvironmentVariable.AWSAccessKeyID())
-		secretAccessKey := common.GetEnvironmentVariable(common.EEnvironmentVariable.AWSSecretAccessKey())
-		if accessKeyID == "" || secretAccessKey == "" {
-			credType = common.ECredentialType.S3PublicBucket()
-			public = true
-			return
-		}
-
+		//Commenting this block out because checkAuthSafeForTarget has no case to handle public. Copy defaults S3 to access key, so similar functionality should be present for sync
+		/*
+			accessKeyID := common.GetEnvironmentVariable(common.EEnvironmentVariable.AWSAccessKeyID())
+			secretAccessKey := common.GetEnvironmentVariable(common.EEnvironmentVariable.AWSSecretAccessKey())
+			if accessKeyID == "" || secretAccessKey == "" {
+				credType = common.ECredentialType.S3PublicBucket()
+				public = true
+				return
+			}
+		*/
 		credType = common.ECredentialType.S3AccessKey()
 		return
 	}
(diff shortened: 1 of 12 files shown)


ste/mgr-JobMgr.go
```diff
@@ -1046,7 +1048,12 @@ func (jm *jobMgr) scheduleJobParts() {
 				go jm.poolSizer()
 				startedPoolSizer = true
 			}
-			jobPart.ScheduleTransfers(jm.Context())
+
+			inMemoryState := jm.getInMemoryTransitJobState()
+			s3provider := inMemoryState.Provider
```
pbanakar-microsoft, line 1053: Nit: inMemoryState.Provider will be NUL when there is no custom credential provider, right? It is good to have a check to update the context only when inMemoryState.Provider is NOT NUL.

cmd/credentialUtil.go
```diff
@@ -456,14 +456,16 @@ func doGetCredentialTypeForLocation(ctx context.Context, location common.Locatio
...
 	if location == common.ELocation.S3() {
-		accessKeyID := common.GetEnvironmentVariable(common.EEnvironmentVariable.AWSAccessKeyID())
-		secretAccessKey := common.GetEnvironmentVariable(common.EEnvironmentVariable.AWSSecretAccessKey())
-		if accessKeyID == "" || secretAccessKey == "" {
-			credType = common.ECredentialType.S3PublicBucket()
-			public = true
-			return
-		}
-
+		//Commenting this block out because checkAuthSafeForTarget has no case to handle public. Copy defaults S3 to access key, so similar functionality should be present for sync
```
shnayak-msft, line 460: Let's use buildmode.IsMover for this change

otendolkar: Added this change, thanks

PR #3143, opened 2025-08-01
Added login, login status, logout to library

gapra-msft:
## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/azcopy/loginStatus.go
+++ b/azcopy/loginStatus.go
@@ -1,4 +1,4 @@
-// Copyright © 2017 Microsoft <wastore@microsoft.com>
+// Copyright © 2025 Microsoft <wastore@microsoft.com>
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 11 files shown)


adreed-msft (approved):

cmd/copy.go
```diff
@@ -994,7 +995,7 @@ func (cca *CookedCopyCmdArgs) getSrcCredential(ctx context.Context, jpo *common.
 		jpo.S2SSourceCredentialType = srcCredInfo.CredentialType
 
 		if jpo.S2SSourceCredentialType.IsAzureOAuth() {
-			uotm := GetUserOAuthTokenManagerInstance()
+			uotm := azcopy.GetUserOAuthTokenManagerInstance()
```
dphulkar-msft, line 998: Have we finalized whether we want to keep the folder name as azcopy?

cmd/credentialUtil.go
```diff
@@ -127,8 +104,8 @@ func GetOAuthTokenManagerInstance() (*common.UserOAuthTokenManager, error) {
 			return
 		}
 
-		options.persistToken = false
-		if err = options.process(); err != nil {
+		options.PersistToken = false
+		if err = RunLogin(options); err != nil {
 			glcm.Error(fmt.Sprintf("Failed to perform Auto-login: %v.", err.Error()))
```
dphulkar-msft:
```suggestion
			glcm.Error(fmt.Sprintf("Failed to perform Auto-login: %v.", err))
```

azcopy/loginStatus.go
```diff
@@ -0,0 +1,54 @@
...
+	TenantID    string
+	AADEndpoint string
+	LoginType   common.AutoLoginType
+}
+
+func (c Client) GetLoginStatus(_ GetLoginStatusOptions) (LoginStatus, error) {
+	uotm := GetUserOAuthTokenManagerInstance()
+
+	// Get current token info and refresh it with GetTokenInfo()
+	ctx := context.Background()
```
wonwuakpa-msft: why aren't we getting the context like how it was previously done? 

`ctx := context.WithValue(context.TODO(), ste.ServiceAPIVersionOverride, ste.DefaultServiceApiVersion)`

cmd/loginStatus.go
```diff
@@ -91,7 +64,7 @@ func init() {
 					glcm.Info(fmt.Sprintf(format, a...))
 				}
 			}
-			status, _ := RunLoginStatus(commandLineInput)
+			status, _ := RunLoginStatus()
```
wonwuakpa-msft, line 67: Why do we not catch the error with RunLoginStatus() ?

azcopy/loginStatus.go
gapra-msft: let me fix that. We should be doing that

cmd/copy.go
gapra-msft: Yes I think that is the best way - then users can call into the library types with azcopy. by default, (ex: azcopy.NewClient())

cmd/loginStatus.go
gapra-msft: backwards compat. I'm trying not to introduce behavior changes if I can help it, also the error doesnt really matter here since we are checking Info.Valid

PR #3144, opened 2025-08-01
Init LCM refactor

gapra-msft:


common/rpc-models.go
```diff
@@ -171,6 +171,7 @@ type S3CredentialInfo struct {
 	Region   string
 }
 
+// TODO: (gapra) This is only used by the tests, so is this really necessary?
```
dphulkar-msft, line 174: todo remainder

gapra-msft: this is more of a longer term TODO: in a later cleanup effort. I dont want to mess with too much along with AzCopy as a library since its already a big change

Issue #3148, opened 2025-08-04
AzCopy copy Blob to local with --preserve-symlinks fails because of not created directories when using wildcard in URL


wolfgang-desalvador:
### Which version of the AzCopy was used? 

AzCopy Version 10.29.1

### Which platform are you using? (ex: Windows, Mac, Linux)

Linux

`AlmaLinux 8.10 (Cerulean Leopard) - 4.18.0-553.16.1.el8_10.x86_64`

### What command did you run?

More documentation is below, in the section describing reproduce.

The command failing is the following, when copying from Blob to local using a wildcard in URL, as described in the [azcopy copy command documentation](https://github.com/Azure/azure-storage-azcopy/wiki/azcopy_copy):

```bash
azcopy copy --include-directory-stub --preserve-posix-properties --preserve-symlinks  --recursive

A note about using a wildcard character (*) in URLs:

There's only two supported ways to use a wildcard character in a URL. 
- You can use one just after the final forward slash (/) of a URL. This copies all of the files in a directory directly to the destination without placing them into a subdirectory. 
- You can also use one in the name of a container as long as the URL refers only to a container and not to a blob. You can use this approach to obtain files from a subset of containers. 

```

### What problem was encountered?

Symlink creation fails with the following output:

```bash
2025/08/04 09:11:34 ERR: [P#0-T#44] DOWNLOADFAILED: https://XXXXXX.blob.core.windows.net/XXXXX/linux-6.16/scripts/dummy-tools/nm : 000 : symlink ld /mnt/recopy/scripts/dummy-tools/nm: no such file or directory. When creating destination symlink (check operation type, is it really download?). X-Ms-Request-Id:
```

### How can we reproduce the problem in the simplest way?

To reproduce run the following:

```bash
wget "https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-6.16.tar.xz"
tar -xf linux-6.16.tar.xz
azcopy login --identity
azcopy copy linux-6.16 "https://<STORAGE_ACCOUNT_NAME>.blob.core.windows.net/<CONTAINER_NAME>" --include-directory-stub --preserve-posix-properties --preserve-symlinks  --recursive ## This will succeed
azcopy copy "https://<STORAGE_ACCOUNT_NAME>.blob.core.windows.net/<CONTAINER_NAME>/linux-6.16/*" recopy --include-directory-stub --preserve-posix-properties --preserve-symlinks  --recursive ## This will have failed entries
```

[...]

PR #3149, opened 2025-08-04
Add creation of parent directory if not existing

wolfgang-desalvador:
## Description
This Pull Request adds the creation/existence check flow when creating a symlink in a Blob to local scenario. This 

**Related Links**:
- [AzCopy copy Blob to local with --preserve-symlinks fails because of not created directories when using wildcard in URL](https://github.com/Azure/azure-storage-azcopy/issues/3148)

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

This has been tested reproducing the same flow described in the issue and confirming job didn't present failed entries


gapra-msft:
@wolfgang-desalvador could you please add a test in e2etest with your use case?

PR #3153, opened 2025-08-07
S3 Traverser: Reusing S3Client in traverser as fix for sync orchestrator scale issue

shnayak-msft:
## Description
S3 Traverser: Reusing S3Client in traverser as fix for sync orchestrator scale issue

1. In sync orchestrator, we create traversers for each path prefix which were creating their own S3Client. This led to failures with scale runs. Using sync.once to make sure that S3Client is created once and reused thereafter.
2. Bug fix: S3 traverser not enumerating directory prefixes for sync orchestrator and double counting files

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/ste/mgr-JobPartMgr.go
+++ b/ste/mgr-JobPartMgr.go
@@ -252,12 +252,12 @@ func (jpm *jobPartMgr) ScheduleTransfers(jobCtx context.Context) {
 	plan := jpm.planMMF.Plan()
 
 	if buildmode.IsMover {
-		// Diagnostic: capture context for panic logging to help identify MMF mapping/unmapping races
-		planFilePath := jpm.filename.GetJobPartPlanPath()
-		partNum := plan.PartNum
-		totalTransfers := plan.NumTransfers
+		// Diagnostic: capture context for panic logging to help identify MMF mapping/unmapping race
 		defer func() {
 			if r := recover(); r != nil {
+				planFilePath := jpm.filename.GetJobPartPlanPath()
+				partNum := plan.PartNum
+				totalTransfers := plan.NumTransfers
 				// Count transfers completed at the time of panic (successfully completed only)
 				completed := atomic.LoadUint32(&jpm.atomicTransfersCompleted)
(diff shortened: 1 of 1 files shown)


common/credentialFactory.go
```diff
@@ -90,7 +90,7 @@ func CreateS3Client(ctx context.Context, credInfo CredentialInfo, option Credent
 	}
 	//support custom credential provider
 	if credInfo.S3CredentialInfo.Provider != nil {
-		fmt.Println("Using custom credentials")
```
shnayak-msft, line 93: Is this needed? I remember we saw this in logs to understand the issue and if i make it info, we may not see it anymore as we set error. We also should not see this a lot with the fix.  I can leave this as is for now.

otendolkar: I think it's fine to remove.  @pbanakar-microsoft any thoughts?

cmd/zc_traverser_s3.go
```diff
@@ -404,3 +382,55 @@ func showS3UrlTypeWarning(s3URLParts common.S3URLParts) {
 }
 
 var s3UrlWarningOncer = &sync.Once{}
+
+// Global S3 client manager for reusing clients across operations
+// This is particularly useful for sync orchestrator which creates many traversers for different path prefixes
+// This allows us to avoid creating a new S3 client for each traverser, improving performance and reducing resource usage.
+// This is a singleton instance, so it can be shared across multiple traversers.
+// It uses sync.Once to ensure that the client is created only once, even if multiple traversers are created concurrently.
```
otendolkar, line 390: Do we plan to use this client manager in STE as well, or will it default to using clientFactory solution. Are there any issues with that?

common/credentialFactory.go
pbanakar-microsoft: I think you can leave it as is for now, to validate the fix is working fine. This should get logged only twice.

PR #3158, opened 2025-08-11
Exit LCM refactor

gapra-msft:


common/lifecyleMgr.go
```diff
@@ -49,10 +49,9 @@ var lcm = func() (lcmgr *lifecycleMgr) {
 // create a public interface so that consumers outside of this package can refer to the lifecycle manager
 // but they would not be able to instantiate one
 type LifecycleMgr interface {
-	OnStart(JobContext) // let the user know the job has started and initial information like log location
-	// print on the same line over and over again, not allowed to float up
-	OnScanProgress(ScanProgress) // only called during sync jobs, print the progress of scanning source and destination
-	OnTransferProgress(progress TransferProgress)
+	JobLifecycleHandler
+	// TODO (gapra) : For copy, sync, resume - we use the OnComplete method to support AzCopy as a library.
```
gapra-msft, line 53: this comment is referring to Exit() below

cmd/copy.go
```diff
@@ -1366,130 +1364,33 @@ func (cca *CookedCopyCmdArgs) ReportProgressOrExit(lcm common.LifecycleMgr) (tot
...
-					jobMan.Log(common.LogError, logStats+"\n"+output)
-				}
-				return output
-			}
+		_, logStats := common.FormatExtraStats(jobSummary.JobType, summary.AverageIOPS, summary.AverageE2EMilliseconds, summary.NetworkErrorPercentage, summary.ServerBusyPercentage)
+		// log to job log
+		jobMan, exists := jobsAdmin.JobsAdmin.JobMgr(summary.JobID)
+		if exists {
+			// Passing this as LogError ensures the stats are always logged.
+			jobMan.Log(common.LogError, logStats+"\n"+common.GetJobSummaryOutputBuilder(jobSummary)(common.EOutputFormat.Text()))
```
wonwuakpa-msft, line 1378: Why do we pass this as an error?

cmd/jobsList.go
```diff
@@ -66,9 +66,7 @@ func init() {
 			}
 
 			err = HandleListJobsCommand(withStatus)
-			if err == nil {
-				glcm.Exit(nil, common.EExitCode.Success())
```
wonwuakpa-msft, line 70: Shouldn't we handle the case when there is no err? And complete with a successful exit code?

cmd/pause.go
```diff
@@ -47,12 +47,11 @@ func init() {
 			if len(args) != 1 {
 				return errors.New("this command only requires jobId")
 			}
-			commandLineInput = (args[0])
+			commandLineInput = args[0]
 			return nil
 		},
 		Run: func(cmd *cobra.Command, args []string) {
 			HandlePauseCommand(commandLineInput)
-			glcm.Exit(nil, common.EExitCode.Success())
```
wonwuakpa-msft, line 55: Why do we remove this but leave the exit on L76 in the pause handler

cmd/sync.go
```diff
@@ -639,65 +639,30 @@ func (cca *cookedSyncCmdArgs) ReportProgressOrExit(lcm common.LifecycleMgr) (tot
...
+			DestinationFilesScanned:  atomic.LoadUint64(&cca.atomicDestinationFilesScanned),
+		}
+		lcm.OnComplete(jobSummary)
 
-			return output
-		}, exitCode)
+		jobMan, exists := jobsAdmin.JobsAdmin.JobMgr(summary.JobID)
+		if exists {
+			_, logStats := common.FormatExtraStats(common.EJobType.Sync(), summary.AverageIOPS, summary.AverageE2EMilliseconds, summary.NetworkErrorPercentage, summary.ServerBusyPercentage)
+			jobMan.Log(common.LogInfo, logStats+"\n"+common.GetJobSummaryOutputBuilder(jobSummary)(common.EOutputFormat.Text()))
```
wonwuakpa-msft, line 664: Shouldn't we use LogInfo for copy like we do for sync?

cmd/copy.go
gapra-msft: // Passing this as LogError ensures the stats are always logged.

cmd/jobsList.go
gapra-msft: It should be covered by the catch all glcm.Exit in main.go if I understand correctly

cmd/pause.go
gapra-msft: That one adds a message, so I didnt remove it.

wonwuakpa-msft (approved):

cmd/sync.go
gapra-msft: Yeah we probably should, but I dont want to introduce breaking changes in this refactor.

cmd/jobsList.go
wonwuakpa-msft: I see. My understanding of this here was to provide immediate feedback on the jobs list success. I also saw this same pattern is done in other commands unless we plan to refactor those away too

gapra-msft: Yeah I'd like to eventually refactor all this away, but it requires a close look at each usage so I'm trying to do it slowly instead of all at once.

gapra-msft:
@dphulkar-msft could you please review this PR?

PR #3160, opened 2025-08-12
Resume cleanup

gapra-msft:
Some cleanup before implementing jobs resume in library


azcopy/jobsResume.go
```diff
@@ -0,0 +1,43 @@
...
+	JobID common.JobID
+
+	SourceSAS      string
+	DestinationSAS string
+}
+
+// ResumeJob resumes a job with the specified JobID.
+func (c Client) ResumeJob(opts ResumeJobOptions) (err error) {
+
+	if opts.JobID.IsEmpty() {
```
adreed-msft: nit: if an option is required it shouldn't be an option; i.e. it should be a required parameter separate from the options bag

cmd/jobsResume.go
```diff
@@ -165,21 +176,26 @@ func init() {
 			return nil
 		},
 		Run: func(cmd *cobra.Command, args []string) {
-			err := resumeCmdArgs.process()
+			includeTransfer := parseTransfers(commandLineArgs.includeTransfer)
+			excludeTransfer := parseTransfers(commandLineArgs.excludeTransfer)
+			if len(includeTransfer) > 0 || len(excludeTransfer) > 0 {
```
adreed-msft, line 183: panic leads to a spooky error to customers; should return an error instead. That all said, this seems like a breaking change.

adreed-msft (changes requested):

cmd/jobsResume.go
gapra-msft: agreed with you there.

azcopy/jobsResume.go
gapra-msft: I'll remove it from the options bag but we'll still need this as a defensive check - a user could just pass in common.JobID{}

azcopy/jobsShow.go
```diff
@@ -23,49 +23,48 @@ package azcopy
...
+func (c Client) GetJobSummary(jobID common.JobID, opts GetJobSummaryOptions) (result JobSummaryResponse, err error) {
+	if jobID.IsEmpty() {
 		return JobSummaryResponse{}, errors.New("get job statistics requires the JobID")
 	}
-	resp := jobsAdmin.GetJobSummary(opts.JobID)
+	resp := jobsAdmin.GetJobSummary(jobID)
 
 	if resp.ErrorMsg != "" {
-		return JobSummaryResponse(resp), fmt.Errorf("failed to get job summary for job %s due to error: %s", opts.JobID, resp.ErrorMsg)
+		return JobSummaryResponse(resp), fmt.Errorf("failed to get job summary for job %s due to error: %s", jobID, resp.ErrorMsg)
```
dphulkar-msft: Since this function already returns an error, and the response is never used when an error is returned, should we just return nil for resp

azcopy/jobsShow.go
```diff
@@ -23,49 +23,48 @@ package azcopy
...
 		return result, errors.New("list job transfers requires the JobID")
 	}
 	status := common.IffNil(opts.WithStatus, common.ETransferStatus.All())
 
-	resp := jobsAdmin.ListJobTransfers(common.ListJobTransfersRequest{JobID: opts.JobID, OfStatus: status})
+	resp := jobsAdmin.ListJobTransfers(common.ListJobTransfersRequest{JobID: jobID, OfStatus: status})
 
 	if resp.ErrorMsg != "" {
-		return ListJobTransfersResponse(resp), fmt.Errorf("failed to list transfers for job %s due to error: %s", opts.JobID, resp.ErrorMsg)
+		return ListJobTransfersResponse(resp), fmt.Errorf("failed to list transfers for job %s due to error: %s", jobID, resp.ErrorMsg)
```
dphulkar-msft: same comment as above

PR #3163, opened 2025-08-14
Extend --include-root flag to include root props for Sync

wonwuakpa-msft:
## Description

This PR extends the behavior of `include-root`. When the customer uses this flag (only for sync) the source root will be included in the transfer.  When it is used with other `preserve--` flags, the root's directory's ACLs and properties will be overwritten to the destination root directory.  This applies to Files and HNS sources.

- **Related Links**:
- [ADO Item](https://msazure.visualstudio.com/One/_workitems/edit/27713309/)

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Tests added to the sync test suite

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -159,6 +159,7 @@ jobs:
         displayName: 'Install dependencies'
       - script: |
           pip install azure-storage-blob==12.12.0
+          pip install azure-storage-blob six
           # set the variable to indicate that the mutex is being acquired
           # note: we set it before acquiring the mutex to ensure we release the mutex. 
           # setting this after can result in an un-broken mutex if someone cancels the pipeline after we acquire the 
@@ -279,19 +280,18 @@ jobs:
           az --version
         displayName: 'Install Azure CLI'
 
-      # - template: azurePipelineTemplates/run_scenarios.yml
-      #   parameters:
-      #     container_name: $(container_name)
-      #     storage_account_name: $(AZCOPY_E2E_ACCOUNT_NAME)
-      #     orp_source_account_name: $(ORP_SOURCE_ACCOUNT_NAME)
-      #     orp_destination_account_name: $(ORP_DESTINATION_ACCOUNT_NAME)
-      #     orp_source_container_name: $(ORP_SOURCE_CONTAINER_NAME)
-      #     destination_container_name: $(DESTINATION_CONTAINER_NAME)
-      #     GOOS: $(GOOS)
-      #     GOARCH: $(GOARCH)
-      #     CGO_ENABLED: $(CGO_ENABLED)
-      #     build_name: $(build_name)
-      #     azcopy_msi_app_id: $(AZCOPY_MSI_APP_ID)
+      - template: azurePipelineTemplates/run_scenarios.yml
+        parameters:
+          container_name: $(container_name)
+          storage_account_name: $(AZCOPY_E2E_ACCOUNT_NAME)
+          orp_source_account_name: $(ORP_SOURCE_ACCOUNT_NAME)
+          orp_destination_account_name: $(ORP_DESTINATION_ACCOUNT_NAME)
+          orp_source_container_name: $(ORP_SOURCE_CONTAINER_NAME)
+          destination_container_name: $(DESTINATION_CONTAINER_NAME)
+          GOOS: $(GOOS)
+          GOARCH: $(GOARCH)
+          CGO_ENABLED: $(CGO_ENABLED)
+          build_name: $(build_name)
           
   - job: Additional_E2E_Test_Windows
     timeoutInMinutes: 360
@@ -343,7 +343,6 @@ jobs:
           GOARCH: $(GOARCH)
           CGO_ENABLED: $(CGO_ENABLED)
           build_name: $(build_name)
-          azcopy_msi_app_id: $(AZCOPY_MSI_APP_ID)
 
     #TODO: Add MacOS E2E tests after creating macos pool
           
--- a/azurePipelineTemplates/run_scenarios.yml
+++ b/azurePipelineTemplates/run_scenarios.yml
@@ -19,88 +19,81 @@ parameters:
     type: string
   - name: build_name
     type: string
-  - name: azcopy_msi_app_id
-    type: string
 
 steps:
-  - task: PowerShell@2
-    displayName: "Run Scenarios"
+  - task: AzureCLI@2
     inputs:
-      targetType: 'inline'
-      script: |
-        # Print "Building executable"
-        echo "Building executable"
-        
-        # Set environment variables
-        $build_name = '${{ parameters.build_name }}'
-        $env:GOOS = '${{ parameters.goos }}'
-        $env:GOARCH = '${{ parameters.goarch }}'
-        $env:AZCOPY_AUTO_LOGIN_TYPE = "MSI"
-        $env:AZCOPY_MSI_CLIENT_ID = '${{ parameters.azcopy_msi_app_id }}'
-        $env:CGO_ENABLED = '${{ parameters.cgo_enabled }}'
+      azureSubscription: azcopytestworkloadidentity
+      scriptType: bash
+      scriptLocation: inlineScript
+      inlineScript: |
+        set -e
+        echo "========== Building executable =========="
 
-        go build -cover -o $build_name
+        go build -cover -o "${{ parameters.build_name }}"
+        if [ ! -f "./${{ parameters.build_name }}" ]; then
+          echo "Error: Executable not found: ${{ parameters.build_name }}"
+          exit 1
+        fi
+        echo "Executable built successfully: ${{ parameters.build_name }}"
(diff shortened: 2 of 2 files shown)


cmd/syncEnumerator.go
```diff
@@ -201,8 +201,18 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 	}
 
 	// decide our folder transfer strategy
-	// sync always acts like stripTopDir=true, but if we intend to persist the root, we must tell NewFolderPropertyOption stripTopDir=false.
-	fpo, folderMessage := NewFolderPropertyOption(cca.fromTo, cca.recursive, !cca.includeRoot, filters, cca.preserveInfo, cca.preservePermissions.IsTruthy(), false, strings.EqualFold(cca.destination.Value, common.Dev_Null), cca.includeDirectoryStubs)
+	var fpo common.FolderPropertyOption
+	var folderMessage string
+	if cca.preserveRootProperties {
+		preservePerms := common.NewPreservePermissionsOption(true, true, cca.fromTo) // must have perservePerms true for persereRootPerms
```
wonwuakpa-msft: Question: is it OK to mandate the preserve-permissions to be true when the user uses this new flag? 
Or should we expect them to use this new flag only when they have preserve-permissions set?

ste/mgr-JobPartTransferMgr.go
```diff
@@ -117,6 +117,7 @@ type TransferInfo struct {
 	PreservePermissions     common.PreservePermissionsOption
 	PreserveInfo            bool
 	PreservePOSIXProperties bool
+	PreserveRootProperties  bool // If true, root dir props will be transferred to destination root dir
```
gapra-msft: why do we need to pass PreserveRootProperties to the ste?

cmd/sync.go
```diff
@@ -1031,4 +1034,9 @@ func init() {
 		"Follow by default. Preserve hardlinks for NFS resources. "+
 			"\n This flag is only applicable when the source is Azure NFS file share or the destination is NFS file share. "+
 			"\n Available options: skip, preserve, follow (default 'follow').")
+
+	syncCmd.PersistentFlags().BoolVar(&raw.preserveRootProperties, "preserve-root-properties", false, "False by default. "+
+		"\n Preserves the root directory and its properties from source to destination. "+
+		"\n When enabled, the destination root properties will be overwritten with the source."+
+		"\n For example, syncing from 'src/file.txt' to 'dest/' will result in 'dest/file.txt' with the props of src on dest.")
```
dphulkar-msft: is this example correct? Should it be for root dir?

cmd/syncEnumerator.go
```diff
@@ -201,8 +201,17 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 	}
 
 	// decide our folder transfer strategy
-	// sync always acts like stripTopDir=true, but if we intend to persist the root, we must tell NewFolderPropertyOption stripTopDir=false.
-	fpo, folderMessage := NewFolderPropertyOption(cca.fromTo, cca.recursive, !cca.includeRoot, filters, cca.preserveInfo, cca.preservePermissions.IsTruthy(), false, strings.EqualFold(cca.destination.Value, common.Dev_Null), cca.includeDirectoryStubs)
+	var fpo common.FolderPropertyOption
+	var folderMessage string
+	if cca.preserveRootProperties {
+		// when preserve-root-properties flag is set, we intend to persist the root, we must tell NewFolderPropertyOption stripTopDir=false.
+		fpo, folderMessage = NewFolderPropertyOption(cca.fromTo, cca.recursive, false, filters, cca.preserveInfo,
```
dphulkar-msft: nitpick: place each argument on newline. Becomes easier to review

cmd/sync.go
dphulkar-msft:
```suggestion
		"\n For example, syncing from 'src/file.txt' to 'dest/' will result in 'dest/file.txt' with the props of source on destination.")
```

wonwuakpa-msft: yes, when the customer specifies this flag, the root directory's ACLs and properties will be copied to the destination root directory

common/rpc-models.go
```diff
@@ -139,6 +139,7 @@ type CopyJobPartOrderRequest struct {
 	CpkOptions                     CpkOptions
 	SetPropertiesFlags             SetPropertiesFlags
 	BlobFSRecursiveDelete          bool
+	PreserveRootProperties         bool // if true, root dir props will be transferred to destination root dir
```
gapra-msft: I dont think this is necessary

cmd/syncEnumerator.go
```diff
@@ -201,8 +203,19 @@ func (cca *cookedSyncCmdArgs) initEnumerator(ctx context.Context) (enumerator *s
 	}
 
 	// decide our folder transfer strategy
-	// sync always acts like stripTopDir=true, but if we intend to persist the root, we must tell NewFolderPropertyOption stripTopDir=false.
-	fpo, folderMessage := NewFolderPropertyOption(cca.fromTo, cca.recursive, !cca.includeRoot, filters, cca.preserveInfo, cca.preservePermissions.IsTruthy(), false, strings.EqualFold(cca.destination.Value, common.Dev_Null), cca.includeDirectoryStubs)
+	var fpo common.FolderPropertyOption
+	var folderMessage string
+	if cca.preserveRootProperties {
```
gapra-msft: instead of doing this, cant we just rely on includeRoot? since we do some inference before hitting this code between includeRoot and preserveRootProperties?

cmd/sync.go
```diff
@@ -324,6 +326,22 @@ func (cooked *cookedSyncCmdArgs) validate() (err error) {
...
+	if cooked.preserveRootProperties {
+		// Both the source and destination must support folders
+		if !cooked.fromTo.AreBothFolderAware() {
+			return fmt.Errorf("preserve-root-properties flag needs both the source and destination resources to be support folders."+
+				"\n Syncing from %s to %s is not supported"+
+				"\n Supported combinations: Local<->File, Local<->BlobFS, BlobFS<->File, File<->File, BlobFS<->BlobFS", cooked.fromTo.From(), cooked.fromTo.To())
+		}
+
+		// For user to use preserveRootProperties flag, preserveInfo and preservePermissions must be set
+		// TODO Reviewers: is it ok to make the preserveRootProperties flag to depend on  preserveInfo and preservePermissions?
```
dphulkar-msft: Are we clear on this? If yes please remove TODO or we can discuss in the meeting.

cmd/sync.go
```diff
@@ -324,6 +326,22 @@ func (cooked *cookedSyncCmdArgs) validate() (err error) {
...
+		if !cooked.preservePermissions.IsTruthy() && !cooked.preserveInfo {
```
dphulkar-msft: Question 1: what all does --preserve-root-properties preserves?
Question 2: Do we need to rely on both the flags?

e2etest/zt_newe2e_sync_test.go
```diff
@@ -561,3 +561,152 @@ func (s *SyncTestSuite) Scenario_TestSyncCreateResources(a *ScenarioVariationMan
 		validateObjectContent: false,
 	})
 }
+
+// Scenario_TestSyncPreserveRootProperties validates that when --preserve-root-properties is set, the destination root folder's properties are overwritten
+// with the source root folder properties.
+func (s *SyncTestSuite) Scenario_TestBlobFSBlobFSPreserveRootProperties(svm *ScenarioVariationManager) {
+	// Use BlobFS (HNS) for ACLs we can validate
+	src := CreateResource[ContainerResourceManager](svm,
+		GetRootResource(svm, common.ELocation.BlobFS()),
```
dphulkar-msft, line 571: Can we add the check for other folder-aware locations?
ELocation.BlobFS(), ELocation.File(), ELocation.Local(), ELocation.FileNFS()

cmd/sync.go
```diff
@@ -340,6 +342,21 @@ func (cooked *cookedSyncCmdArgs) validate() (err error) {
...
+	if cooked.preserveRootProperties {
+		// Both the source and destination must support folders
+		if !cooked.fromTo.AreBothFolderAware() {
+			return fmt.Errorf("preserve-root-properties flag needs both the source and destination resources to be support folders."+
+				"\n Syncing from %s to %s is not supported"+
+				"\n Supported combinations: Local<->File, Local<->BlobFS, BlobFS<->File, File<->File, BlobFS<->BlobFS", cooked.fromTo.From(), cooked.fromTo.To())
+		}
+
+		// For preserve-root-properties to persist the root perms or properties, we need either preserveInfo or preservePermissions flags
+		if !(cooked.preservePermissions.IsTruthy() || cooked.preserveInfo) {
```
gapra-msft: shoudl we also add preserve posix properties here?

Issue #3159, opened 2025-08-12
AzCopy v10.30.0 azcopy remove missing --exclude-path


nreso:
##### Note: The version is visible when running AzCopy without any argument
[AzCopy v10.30.0](https://github.com/Azure/azure-storage-azcopy/releases/tag/v10.30.0)

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows and Linux

##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.
`azcopy remove "$sourceURL" --recursive --exclude-path "myfoldertoexclude" --log-level INFO`

(FYI, I'm using RBAC and not SAS)

### What problem was encountered?
Error retuned: `unknown flag: --exclude-path`

### How can we reproduce the problem in the simplest way?
Create a container with folders and blobs like:
`mystorageaccount/mycontainer/myfolder/subfolder1/blob1.json`
`mystorageaccount/mycontainer/myfolder/subfolder2/blob2.json`
`mystorageaccount/mycontainer/myfolder/myfoldertoexclude/blob3.json`

So `$sourceURL = "https://mystorageaccount.blob.core.windows.net/mycontainer/myfolder`

Using AzCopy v10.30.0 run the command:
`azcopy remove "$sourceURL" --recursive --exclude-path "myfoldertoexclude" --log-level INFO`

### Have you found a mitigation/solution?
`--exclude-path` seems to have been removed in AzCopy v10.30.0. It was in the previous version v10.29.1 and that worked, see [comparison of versions](https://github.com/Azure/azure-storage-azcopy/wiki/azcopy_remove/_compare/e20449e5683b34f33132bf42fec2e6867d545f4c...2c55632e2552ce152606befa3e5e4341a26ca26c).

Using `--exclude-pattern` in v10.30.0 doesn't seem to exclude the folder `myfoldertoexclude`, it deletes it - e.g. running the following still removes the `myfoldertoexclude` folder:
- `azcopy remove "$sourceURL" --recursive --exclude-pattern "myfoldertoexclude" --log-level INFO`
- `azcopy remove "$sourceURL" --recursive --exclude-pattern "/myfoldertoexclude/" --log-level INFO`
- `azcopy remove "$sourceURL" --recursive --exclude-pattern "/myfoldertoexclude*" --log-level INFO`
[...]

PR #3165, opened 2025-08-15
Add missing --exclude-path flag to azcopy remove command

wonwuakpa-msft:
## Description

The `--exclude-path` flag was accidentally removed from the `azcopy remove` command in version 10.30.0, while it remained available in the `azcopy copy` command. This PR restores the missing flag to maintain feature parity between the commands.

- **Related Links**:
   - [Issues](https://github.com/Azure/azure-storage-azcopy/issues/3159)

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Two unit and end-to-end tests that cover removing blobs and files with exclude path flag


cmd/zt_remove_blob_test.go
```diff
@@ -287,6 +287,58 @@ func TestRemoveWithIncludeAndExcludeFlag(t *testing.T) {
 	})
 }
 
+// exclude-path flag limits the scope of the delete
+func TestRemoveWithExcludePathFlag(t *testing.T) {
```
gapra-msft: are these UTs necessary if we have the e2e test?

gapra-msft (approved):

PR #3168, opened 2025-08-15
Clean up with LCM

gapra-msft:
Move CLI related code into cmd - Message Handler

Move AzCopy log closing to init

Remove syslog disable methods to init


common/jobLifecycleHandler.go
```diff
@@ -30,6 +30,14 @@ type JobLifecycleHandler interface {
 	OnTransferProgress(progress TransferProgress)
 
 	OnComplete(summary JobSummary)
+
+	// TODO (gapra): Rename these variables at the end. Choosing not to at the moment to limit the scope of changes.
```
gapra-msft, line 34: I meant to type functions - will fix this before merging, but its not a big deal since its just a comment

azcopy/client.go
```diff
@@ -21,23 +21,27 @@
...
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 	"github.com/Azure/azure-storage-azcopy/v10/jobsAdmin"
 	"github.com/Azure/azure-storage-azcopy/v10/ste"
-	"log"
-	"runtime"
 )
 
 type Client struct {
 	CurrentJobID common.JobID // TODO (gapra): In future this should only be set when there is a current job running. On complete, this should be cleared. It can also behave as something we can check to see if a current job is running
+	logLevel     common.LogLevel
```
wonwuakpa-msft, line 34: Should we also include output-level and output-type flags?

dphulkar-msft (approved):

azcopy/client.go
gapra-msft: Those are AzCopy CLI specific things, so let's keep those in CLI only

wonwuakpa-msft (approved):

cmd/root.go
```diff
@@ -200,14 +195,28 @@ func Initialize(resumeJobID common.JobID, isBench bool) (err error) {
 		Client.CurrentJobID = common.NewJobID()
 	}
 
+	// Run MessagHandler to process messages from Input Watcher
+	if jobsAdmin.JobsAdmin != nil {
+		go jobsAdmin.JobsAdmin.MessageHandler(glcm.MsgHandlerChannel())
```
adreed-msft, line 200: I expect that this'll probably change, but, is this something that an end user of the library would have to do? This seems like a bit of mental overhead and something that isn't exactly. explained?

gapra-msft: No - this is a purely CLI thing. its for the input watcher of the CLI LCM.

PR #3173, opened 2025-08-19
Add --check-version flag

wonwuakpa-msft:
## Description

Instead of version checking happening every time during initialization and execution there is an explicit `check-version` flag to handle this. It no longer happens automatically on every command. `--version` still remains functional but only returns the local apps version.

### Feature / Bug Fix:
- Add new flag
- Remove skip-version-check 

- **Related Links**:
- [ADO Item](https://msazure.visualstudio.com/One/_backlogs/backlog/XStore-DevExp/Epics?showParents=true&workitem=33909434)

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Manually tested
image


e2etest/newe2e_task_runazcopy_parameters.go
```diff
@@ -193,9 +193,8 @@ type RawFlags map[string]string
 // in a test, without rewriting all the flags for every use case.
 
 type GlobalFlags struct {
-	CapMbps          *float64 `flag:"cap-mbps"`
-	TrustedSuffixes  []string `flag:"trusted-microsoft-suffixes"`
-	SkipVersionCheck *bool    `flag:"skip-version-check,default:true"`
+	CapMbps         *float64 `flag:"cap-mbps"`
```
gapra-msft: Can we add an e2e test for this change?

gapra-msft (changes requested):
please add a test

cmd/versionChecker.go
```diff
@@ -159,5 +160,7 @@ func PrintOlderVersion(newest Version, local Version) {
 
 		// output in info mode instead of stderr, as it was crashing CI jobs of some people
 		glcm.Info(executableName + " " + local.original + ": A newer version " + newest.original + " is available to download\n")
+	} else {
+		glcm.Info(fmt.Sprintf("Current azcopy version %s up to date\n", local.original))
```
dphulkar-msft:
```suggestion
		glcm.Info(fmt.Sprintf("Current AzCopy version %s is up to date\n", local.original))
```

cmd/root.go
```diff
@@ -325,6 +317,8 @@ func init() {
 	_ = rootCmd.PersistentFlags().MarkHidden("retry-status-codes")
 	rootCmd.PersistentFlags().StringVar(&debugMemoryProfile, "memory-profile", "", "Export pprof memory profile")
 	_ = rootCmd.PersistentFlags().MarkHidden("memory-profile")
+	rootCmd.PersistentFlags().BoolVar(&checkAzCopyUpdates, "check-version", false,
+		"Check current AzCopy version for a newer version available.")
```
dphulkar-msft:
```suggestion
		"Checks if a newer version of AzCopy is available.")
```

e2etest/zt_version_test.go
```diff
@@ -68,3 +69,54 @@ func TestVersionCommand(t *testing.T) {
...
+	if len(lines) > 1 {
+		skipLine := func() (val int) {
+			for i, line := range lines {
+				if strings.Contains(line, "WARN:") {
+					return i
+				}
+			}
+			return
+		}()
+		lines = lines[skipLine+1:]
```
dphulkar-msft: Instead of testing it this way is it possible to add this test case in our E2Etest. So that we dont have to add the handling from line 93-101 manually as its already handled there.

gapra-msft (approved):
please address dhanashree's comments, otherwise looks good to me

cmd/root.go
```diff
@@ -298,9 +293,6 @@ func init() {
 			trustedSuffixesAAD+"'. \n Any listed here are added to the default. For security, you should only put Microsoft Azure domains here. "+
 			"\n Separate multiple entries with semi-colons.")
 
-	rootCmd.PersistentFlags().BoolVar(&SkipVersionCheck, "skip-version-check", false,
```
gapra-msft, line 322: We can't simply delete the flag - we need to mark it hidden and simply ignore the user input

gapra-msft: We still need to bring this line back.

gapra-msft: Add a test that calls azcopy with skip-version-check and make sure that doenst cause the job to fail

e2etest/zt_version_test.go
```diff
@@ -68,3 +69,54 @@ func TestVersionCommand(t *testing.T) {
 		t.FailNow()
 	}
 }
+
+func TestCheckVersionCommand(t *testing.T) {
```
gapra-msft: Since youve added the e2e test I think you can delete this

cmd/root.go
```diff
@@ -186,6 +186,20 @@ var rootCmd = &cobra.Command{
...
+			// Either wait till this routine completes or timeout and do not print if it exceeds 8s
+			// Spawn a routine to fetch & compare the local application's version against the latest version available
+			case <-beginDetectNewVersion():
+				// noop
+			case <-time.After(time.Second * 8):
+				// don't wait too long
+			}
+		}
+		return nil
+	},
```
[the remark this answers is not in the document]
gapra-msft: What is the intended use? Are we intending this to only be called when azcopy --check-version is called? or also azcopy copy src dest --check-version?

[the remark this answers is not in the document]
gapra-msft: If the latter, why not just do it in initialize?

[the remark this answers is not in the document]
wonwuakpa-msft: I was intending the former, only when azcopy --check-version is called without subcommands. The e2e tests also reflect this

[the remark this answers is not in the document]
gapra-msft: If the former, then it seems like we are doing this wrong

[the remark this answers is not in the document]
gapra-msft: Based on what Copilot is saying, then this code is run for all commands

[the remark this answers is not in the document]
wonwuakpa-msft: I think copilot is wrong in its suggestion :/ RunE hook is only run for the root azcopy _without_ subcommands. It is not inherited by the other commands.  If we put it in PersistentPreRunE, then it would run first for all the subcommands of the root command.

gapra-msft (approved):

Issue #3176, opened 2025-08-20
10.30.0: Azcopy commands create file "latest_version.txt" with azcopy version and timestamp


RobinDinkel:
### Which version of the AzCopy was used? 

10.30.0

### Which platform are you using? (ex: Windows, Mac, Linux)

Ubuntu 24.04

### What command did you run?

azcopy --version

### What problem was encountered?

The command creates a file in the current directory called latest_version.txt with the azcopy version and a timestamp. This is unwanted, unexpected behavior. The command line output is enough.

### How can we reproduce the problem in the simplest way?

Run azcopy --version or even something like azcopy --help.

### Have you found a mitigation/solution?

No.

wonwuakpa-msft:
Hi, @RobinDinkel thanks for reporting this. We will include a fix for this in our patch release of 10.30 sometime next month.

PR #3179, opened 2025-08-20
Remove latest_version.txt creation in user curr dir

wonwuakpa-msft:
## Description

As of 10.30.0, Version checking is done every time azcopy is initialized and executed. Calling `azcopy --help` and `azcopy --version` without any subcommands invokes the version checker but without properly setting the log folder. Because of this, the text file is created in the user's current directory.

In v10.31, we plan to replace automatic version checking, with `check-version` flag. And that will also fix the reported issue. This PR is to address regression

Fix `latest_version.txt` file being wrongly created in user's current directory. 

- **Related Links**:
   - [Issues](https://github.com/Azure/azure-storage-azcopy/issues/3176)

## Type of Change

- [X] Bug fix

## How Has This Been Tested?

Wrote a unit test to validate version text is made in proper location. 


cmd/root.go
```diff
@@ -346,6 +347,11 @@ func beginDetectNewVersion() chan struct{} {
 			return
 		}
 
+		// In the unlikely chance log path folder is empty, make sure we don't write to user's current directory
```
adreed-msft: InitializeFolders gets called on the AzCopy Client Initialize function, also called at the end of PersistentPreRunE.

Double calling it will probably make things worse.

cmd/versionChecker.go
```diff
@@ -121,6 +121,7 @@ func (v Version) EqualTo(v2 Version) bool {
 func (v Version) CacheRemoteVersion(remoteVer Version, filePath string) error {
 	if v.OlderThan(remoteVer) || v.EqualTo(remoteVer) {
 		expiry := time.Now().Add(24 * time.Hour).Format(versionFileTimeFormat)
+		// make sure filepath is absolute filepath so WriteFile is not written to customers current directory
```
adreed-msft, line 124: This is just a comment, not an additional condition.

adreed-msft (changes requested):
This is not right at all, and probably won't resolve the customer's case. Some of the changes do nothing effectual-- some of the changes probably worsen the issue. The test is probably fine, but we should follow up with the customer a bit more about their system's setup, maybe the home directory lives in another env var?

We should probably have a sensible default if we can't pick it up, i.e. /.azcopy or C:\\.azcopy

cmd/root.go
gapra-msft: have we understood the root cause of why LogPathFolder is not set? Is it because we're running this async? I know we call it from two locations in the codebase. @wonwuakpa-msft take a look at where this is called and reason about why this may be happening.

wonwuakpa-msft: @adreed-msft, @gapra-msft It's a weird edge case. I found the root cause of the issue is that with certain commands like azcopy --help and azcopy --version, they do not actually reach the Initialize() and PersistentPreRunE hook of the root command.  The PersistentPreRunE is not run on just the root command so it will not run on just `azcopy`. It runs ONLY before all the children of the root commands. These two commands instead go to the InitalizeAndExecute function. However, this function does not actually do any initialization work. So, it begins version checking without the necessary set. The reason this recently became a problem is due to the AzCopy as a library refactoring because we no longer have the LogPathFolder as a global var. So, the LogPathFolder is never set and is left as empty. Thus, during version checking the latest_version.txt file is created in the user's current directory.

gapra-msft: oh I see

gapra-msft: I think then - let's just run the InitializeFolders code before the version check call. It still shouldnt exist in the version check method itself

gapra-msft: The version check's method's job is to just check the version. It should be the callers responsibility to make sure this is set.

cmd/root.go
```diff
@@ -262,6 +263,12 @@ func InitializeAndExecute() {
 		glcm.Error(err.Error())
 	} else {
 		if !SkipVersionCheck && !isPipeDownload {
+			// In the unlikely chance log path folder is empty, make sure we don't write to user's current directory
```
gapra-msft: I think this comment could be reworded. The log path folder being empty is not an "unlikely" condition, it's expected when things like --version or --help are called. I'd specify this here. Mention how initializing folders happens in the Client init which doesnt get called in this path

adreed-msft:
OK, yep, I went and double checked, I was in fact wrong.

gapra-msft (approved):

gapra-msft:
@wonwuakpa-msft re-run the pipeline and once it passes we should be good to merge this PR

wonwuakpa-msft:
> This is not right at all, and probably won't resolve the customer's case. Some of the changes do nothing effectual-- some of the changes probably worsen the issue. The test is probably fine, but we should follow up with the customer a bit more about their system's setup, maybe the home directory lives in another env var?
> 
> We should probably have a sensible default if we can't pick it up, i.e. /.azcopy or C:\.azcopy
Changes are correct and in line with what we want

PR #3185, opened 2025-08-22
Gapra/resume init

gapra-msft:
In this PR, I've moved the glcm instantiation to the cmd folder, since that is where all azcopy CLI related code should live. There is no need for the CLI lcm to be instantiated during usage of azcopy as a library. 

As a result, I've also needed to set up some mocked LCMs in some UTs since they call into methods that assume that the lcm was magically instantiated

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -52,11 +52,6 @@ var OutputLevel common.OutputVerbosity
 var CapMbps float64
 var SkipVersionCheck bool
 
-// It's not pretty that this one is read directly by credential util.
-// But doing otherwise required us passing it around in many places, even though really
-// it can be thought of as an "ambient" property. That's the (weak?) justification for implementing
-// it as a global
-var TrustedSuffixes string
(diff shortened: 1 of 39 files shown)


cmd/root.go
```diff
@@ -178,21 +158,43 @@ var rootCmd = &cobra.Command{
...
 	currPid := os.Getpid()
 	AsyncWarnMultipleProcesses(cmd.GetAzCopyAppPath(), currPid)
 	jobsAdmin.BenchmarkResults = isBench
 	Client, err = azcopy.NewClient(azcopy.ClientOptions{CapMbps: CapMbps})
 	if err != nil {
 		return err
 	}
-	Client.CurrentJobID = resumeJobID
-	if Client.CurrentJobID.IsEmpty() {
+	if !isMigratedToLibrary {
```
gapra-msft, line 178: This will all eventually get migrated to the appropriate commands that need this.

Issue #3182, opened 2025-08-21
azcopy list  with parameter --output-type json is not only returning json parseable output


krzysztofstepien-tomtom:
##### Note: The version is visible when running AzCopy without any argument

10.30.0

### Which platform are you using? (ex: Windows, Mac, Linux)

Linux, Mac

##### Note: Please remove the SAS to avoid exposing your credentials. If you cannot remember the exact command, please retrieve it from the beginning of the log file.

```
azcopy list --machine-readable --output-type json "blob_dir_url"
```

### What problem was encountered?

Latest azcopy command is creating additional directory `pid` inside `.azcopy`. This is probably causing the problem that for the azcopy commands which should return json only, there are some additional json unparseable output, e.g
```
{"TimeStamp":"2025-08-19T04:12:21.116834023Z","MessageType":"ListObject","MessageContent":"{\"Path\":\"idm_0.bckp\",\"ContentLength\":\"2147483647\"}","PromptDetails":{"PromptType":"","ResponseOptions":null,"PromptTarget":""}},
{"TimeStamp":"2025-08-19T04:12:21.116958046Z","MessageType":"EndOfJob","MessageContent":"","PromptDetails":{"PromptType":"","ResponseOptions":null,"PromptTarget":""}},
INFO - Found blob: WARN: more than one AzCopy process is running. It is best practice to run a single process per VM., parsing line to json object...
{"TimeStamp":"2025-08-19T04:12:21.167888706Z","MessageType":"Info","MessageContent":"WARN: more than one AzCopy process is running. It is best practice to run a single process per VM.","PromptDetails":{"PromptType":"","ResponseOptions":null,"PromptTarget":""}}
```

Once i downgrade the version (f.e to the `10.29.1` or `10.28.0`), the output is always json parseable

This is probably caused by the following line (introduced in the latest release):
https://github.com/Azure/azure-storage-azcopy/blob/main/cmd/processChecker.go#L26

### How can we reproduce the problem in the simplest way?
Sample bash script which i launched in two separate terminals:
```
#!/bin/bash
set -euxo pipefail

for i in {1..100}; do
    output=$(azcopy list --machine-readable --output-type json "<<provide_link>>")

    if echo "$output" | jq empty 2>/dev/null; then
        echo "Valid JSON:"
        echo "$output" | jq
    else
        echo "Error: Invalid JSON detected at iteration $i"
        echo "Problematic output:"
        echo "$output"
    fi
done
```

[...]

PR #3188, opened 2025-08-27
Fixes to warning when multiple Azcopy processes running

wonwuakpa-msft:
## Description

In v10.30, We introduced warning the cx when multiple AzCopy instances are being run. This PR includes some minor fixes to it. 

  - Fix bug with JSON output formatting. Warning now happens after the output format and output verbosity are set. 
  - Added additional cleanup to remove non active or stale AzCopy processes
  - Filtered the warning to only happen on commands that are sent to the STE
  
- **Related Links**:
- [Issues](https://github.com/Azure/azure-storage-azcopy/issues/3182)

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Manually tested pids folder is cleaned up. Also tested warning does not happen for commands that are not sent to the STE.


cmd/processChecker.go
```diff
@@ -5,25 +5,84 @@ import (
...
+// isProcessRunning checks if a process with the given PID is running
+func isProcessRunning(pid int) bool {
+	// In Unix, this can falsely return a Process for pid even when it does not exist
+	// We perform a signal check to test
+	process, err := os.FindProcess(pid)
+	if err != nil {
+		return false
+	}
+	if runtime.GOOS == "windows" {
+		return true
```
gapra-msft: why are we always returning true for Windows?

dphulkar-msft: +1

cmd/processChecker.go
```diff
@@ -5,25 +5,84 @@ import (
...
+			// Not a valid PID, remove the file
+			os.Remove(path.Join(pidsSubDir, fileName))
+			continue
+		}
+		if pid == currentPid { // Skip current process
+			continue
+		}
+		if !isProcessRunning(pid) {
+			// Process is not running, remove the stale PID file
+			os.Remove(path.Join(pidsSubDir, fileName))
```
dphulkar-msft: Should we check the os.Remove() err here?

cmd/processChecker.go
```diff
@@ -5,25 +5,84 @@ import (
...
+
 // WarnMultipleProcesses warns if there are multiple AzCopy processes running
 func WarnMultipleProcesses(directory string, currentPid int) {
 	currPidFileName := fmt.Sprintf("%d.pid", currentPid)
-	// Made subdir to not clog main dir
-	pidsSubDir := path.Join(directory, "pids")
+	pidsSubDir := path.Join(directory, "pids") // Made subdir to not clog main dir
 	err := os.MkdirAll(pidsSubDir, 0755)
 	if err != nil {
 		return
```
dphulkar-msft, line 96: Should we log the error and then return? Same comment for below line:73 and line:78

cmd/root.go
```diff
@@ -185,13 +185,20 @@ var rootCmd = &cobra.Command{
 
 		isBench := cmd.Use == "bench [destination]"
 
-		return Initialize(resumeJobID, isBench)
+		// We only care to warn about multiple AzCopy processes for commands sent to STE
+		sentToSte := []string{"copy", "sync", "bench [destination]", "resume [jobID]", "remove [resourceURL]", "set-properties [source]"}
```
dphulkar-msft: Should we include list command ?

common/fe-ste-models.go
```diff
@@ -59,7 +59,7 @@ const (
 	FILE_NOT_FOUND            = "The specified file was not found."
 	EINTR_RETRY_COUNT         = 5
 	RECOMMENDED_OBJECTS_COUNT = 10000000
-	ERR_MULTIPLE_PROCESSES    = "more than one AzCopy process is running. It is best practice to run a single process per VM."
+	ERR_MULTIPLE_PROCESSES    = "More than one AzCopy process is running. It is best practice to run a single process per VM."
```
dphulkar-msft:
```suggestion
	WARN_MULTIPLE_PROCESSES    = "More than one AzCopy process is running. It is best practice to run a single process per VM."
```

cmd/root.go
wonwuakpa-msft: I don't think we should because they are not sent to the STE. The `jobs list` appends files it gets from the job plan folder after it does some transformations and sorts them. The other `list container URL` command traverses the resource URL and its children to create list objects that are used to build the job summary.

cmd/processChecker.go
wonwuakpa-msft: Yes I will add err handling

cmd/processChecker.go
wonwuakpa-msft: for Windows, the func os.FindProcess returns error when the process with given PID is not running. I handle that err and return above so if it gets to this can safely return true.

There is a slight difference in Unix. The func will always succeed and return the process object, even when the process does not exist and is not running. So, I had to verify with a signal. 
 
 https://go.dev/pkg/os/?m=all,old#FindProcess

cmd/processChecker.go
```diff
@@ -5,25 +5,84 @@ import (
...
 	"os"
 	"path"
+	"runtime"
+	"strconv"
+	"strings"
+	"syscall"
 )
 
+// isProcessRunning checks if a process with the given PID is running
+func isProcessRunning(pid int) bool {
```
dphulkar-msft, line 15: If this handling is required only for Linux, update the function header to clearly indicate that it applies specifically to Linux and the reasoning. Verify the behavior for macOS as well, and if it is the same for both(mac and windows), then implement the handling only for Linux within this function. Ensure that this function is called conditionally, only when runtime.GOOS == "linux".

cmd/processChecker.go
dphulkar-msft: add the details in the code comment

cmd/processChecker.go
```diff
@@ -5,34 +5,106 @@ import (
...
+			continue
+		}
+		if pid == currentPid { // Skip current process
+			continue
+		}
+		if !isProcessRunning(pid) {
+			// Process is not running, remove the stale PID file
+			err := os.Remove(path.Join(pidsSubDir, fileName))
+			if err != nil {
+				glcm.Info("failed to remove stale pid file" + err.Error())
```
gapra-msft: Should we add a message that this is just informational - and to clean up the pids file manually?

gapra-msft (approved):

gapra-msft (approved):

cmd/processChecker.go
```diff
@@ -5,42 +5,120 @@ import (
...
+	// https://go.dev/pkg/os/?m=all,old#FindProcess
+	err = process.Signal(syscall.Signal(0))
+	return err == nil
+}
+
+// cleanupStalePidFiles removes PID files for processes that are no longer running
+func cleanupStalePidFiles(pidsSubDir string, currentPid int) error {
+	f, err := os.Open(pidsSubDir)
+	if err != nil {
+		glcm.Info("could not open pids dir. you can delete files in pids dir manually. \n AzCopy should continue as normal." + err.Error())
```
dphulkar-msft: I think we should write the message more clear as this is going to be displayed on the console and the user should be clear what is happening and why to delete.
Something like below
```
glcm.Info("AzCopy could not open the temporary pids directory used for tracking running jobs. " +
	"You can safely delete the files in the pids directory manually; AzCopy will recreate them as needed. " +
	"Job execution will continue normally without impact. Details: " + err.Error())
```
	
	Just a suggestion. re-write as you want.

dphulkar-msft: Also should we display the exact location where this pid dir exists as user might not be aware of where it has been created.

cmd/processChecker.go
```diff
@@ -5,42 +5,120 @@ import (
...
+			continue
+		}
+		if pid == currentPid { // Skip current process
+			continue
+		}
+		if !isProcessRunning(pid) {
+			// Process is not running, remove the stale PID file
+			err := os.Remove(path.Join(pidsSubDir, fileName))
+			if err != nil {
+				glcm.Info("could not clean up stale pid file. you can delete files in pids dir manually \n AzCopy should continue as normal." + err.Error())
```
dphulkar-msft: Do we need to display all these error message on console? Or can we log it into the log file?

cmd/processChecker.go
wonwuakpa-msft: Sure, I can add more detail to the info message

gapra-msft (approved):

PR #3199, opened 2025-08-29
Inject UOTM into methods

gapra-msft:
This PR injects the UOTM into the credential methods instead of just assuming its available globally.

### DIFF
--- a/cmd/make.go
+++ b/cmd/make.go
@@ -86,7 +86,7 @@ func (cookedArgs cookedMakeCmdArgs) process() (err error) {
 		return fmt.Errorf("failed to resolve target: %w", err)
 	}
 
-	credentialInfo, _, err := GetCredentialInfoForLocation(ctx, cookedArgs.resourceLocation, resourceStringParts, false, common.CpkOptions{})
+	credentialInfo, _, err := GetCredentialInfoForLocation(ctx, cookedArgs.resourceLocation, resourceStringParts, false, Client.GetUserOAuthTokenManagerInstance(), common.CpkOptions{})
 	if err != nil {
 		return err
 	}
--- a/cmd/setPropertiesEnumerator.go
+++ b/cmd/setPropertiesEnumerator.go
@@ -38,7 +38,7 @@ func setPropertiesEnumerator(cca *CookedCopyCmdArgs) (enumerator *CopyEnumerator
 
 	var srcCredInfo common.CredentialInfo
 
-	if srcCredInfo, _, err = GetCredentialInfoForLocation(ctx, cca.FromTo.From(), cca.Source, true, cca.CpkOptions); err != nil {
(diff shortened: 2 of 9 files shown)


cmd/make.go
```diff
@@ -86,7 +86,7 @@ func (cookedArgs cookedMakeCmdArgs) process() (err error) {
 		return fmt.Errorf("failed to resolve target: %w", err)
 	}
 
-	credentialInfo, _, err := GetCredentialInfoForLocation(ctx, cookedArgs.resourceLocation, resourceStringParts, false, common.CpkOptions{})
+	credentialInfo, _, err := GetCredentialInfoForLocation(ctx, cookedArgs.resourceLocation, resourceStringParts, false, Client.GetUserOAuthTokenManagerInstance(), common.CpkOptions{})
```
wonwuakpa-msft, line 89: Shouldn't we follow the same pattern with list and jobsResume - i.e passing the uotm variable instead of passing in the getter?

cmd/setPropertiesEnumerator.go
```diff
@@ -38,7 +38,7 @@ func setPropertiesEnumerator(cca *CookedCopyCmdArgs) (enumerator *CopyEnumerator
 
 	var srcCredInfo common.CredentialInfo
 
-	if srcCredInfo, _, err = GetCredentialInfoForLocation(ctx, cca.FromTo.From(), cca.Source, true, cca.CpkOptions); err != nil {
+	if srcCredInfo, _, err = GetCredentialInfoForLocation(ctx, cca.FromTo.From(), cca.Source, true, Client.GetUserOAuthTokenManagerInstance(), cca.CpkOptions); err != nil {
```
wonwuakpa-msft, line 41: Same comment from L89 of make.go

cmd/make.go
gapra-msft: those methods happened to have an instance of the uotm gotten earlier in the method, so I reused it. In this one it didnt so I just called the method directly.

cmd/setPropertiesEnumerator.go
gapra-msft: see above

cmd/zt_credentialUtil_test.go
```diff
@@ -138,30 +136,30 @@ func TestIsPublic(t *testing.T) {
...
-	_, err = ctr.SetAccessPolicy(ctx, &container.SetAccessPolicyOptions{Access: &publicAccess})
-	a.Nil(err)
-
-	// Verify that blob is public.
-	bb, _ := getBlockBlobClient(a, ctr, "")
-	_, err = bb.UploadBuffer(ctx, []byte("I'm a block blob."), nil)
-	a.Nil(err)
-
-	a.True(isPublic(ctx, bb.URL(), common.CpkOptions{}))
+	//a := assert.New(t)
```
gapra-msft, line 139: this is a valid test and its currently being skipped. I did not want to go through the hassle of migrating the methods associated with this test right now. Whenever we do the TODO, we can deal with it.

PR #3202, opened 2025-09-02
LCM refactor

gapra-msft:
Sorry in advance for how large this PR is. 

This PR does two things
1. Consolidate error handling. It no longer assumes that we have a global LCM that can handle error handling for us. Each job will pass its own Error Handler (which is the glcm for now)
2. Move LCM to cmd folder. This is really an AzCopy CLI implementation, so it belongs in cmd.


azcopy/credentialUtil.go
```diff
@@ -221,15 +221,15 @@ func logAuthType(ct common.CredentialType, location common.Location, isSource bo
 }
 
 // isPublic reports true if the Blob URL passed can be read without auth.
-func isPublic(ctx context.Context, blobResourceURL string, cpkOptions common.CpkOptions) (isPublicResource bool) {
+func isPublic(ctx context.Context, blobResourceURL string, cpkOptions common.CpkOptions) (isPublicResource bool, err error) {
 	bURLParts, err := blob.ParseURL(blobResourceURL)
 	if err != nil {
-		return false
+		return false, nil
```
wonwuakpa-msft, line 226: Do we not care about returning the err from parsing the blob URL?

azcopy/credentialUtil.go
```diff
@@ -353,25 +363,37 @@ func doGetCredentialTypeForLocation(ctx context.Context, location common.Locatio
...
+			public, err = isPublic(ctx, uri.String(), cpkOptions)
 			if err != nil {
 				return common.ECredentialType.Unknown(), false, err
 			}
-			if !oAuthTokenExists {
-				return common.ECredentialType.Unknown(), false,
-					common.NewAzError(common.EAzError.LoginCredMissing(), "No SAS token or OAuth token is present and the resource is not public")
+			if public {
+				credType = common.ECredentialType.Anonymous()
+				public = true
```
dphulkar-msft: Since we are already inside 'if public', the variable is guaranteed to be true.
Reassigning public = true here isnt redundant?

cmd/lifecycleMgr.go
```diff
@@ -0,0 +1,732 @@
+package cmd
```
dphulkar-msft, line 21: missing license header

cmd/lifecycleMgr.go
```diff
@@ -0,0 +1,732 @@
...
+	logSanitizer          common.LogSanitizer
+	inputQueue            chan userInput // msgs from the user
+	allowWatchInput       bool           // accept user inputs and place then in the inputQueue
+	allowCancelFromStdIn  bool           // allow user to send in 'cancel' from the stdin to stop the current job
+	e2eAllowAwaitContinue bool           // allow the user to send 'continue' from stdin to start the current job
+	e2eAllowAwaitOpen     bool           // allow the user to send 'open' from stdin to allow the opening of the first file
+	closeFunc             func()         // used to close logs before exiting
+	waitForUserResponse   chan bool
+	msgHandlerChannel     chan *common.LCMMsg
+	OutputVerbosityType   common.OutputVerbosity
```
dphulkar-msft: Should we keep this local variable? As we have SetOutputVerbosity() API exposed

gapra-msft: yeah this can be made local

common/output.go
```diff
@@ -99,9 +78,11 @@ type JsonOutputTemplate struct {
 	PromptDetails  PromptDetails
 }
 
-func newJsonOutputTemplate(messageType OutputMessageType, messageContent string, promptDetails PromptDetails) *JsonOutputTemplate {
```
wonwuakpa-msft, line 102: Do we no longer need this function?

wonwuakpa-msft (approved):
small comments

common/output.go
gapra-msft: its still used by lifecycle manager - we'll need to clean this up a bit at the end

azcopy/credentialUtil.go
gapra-msft: I didnt want to change any of the logic in case we relied on swallowing the error. The point of this error refactor was to pass the CPK error backwards instead of relying on the magical glcm.Error call

dphulkar-msft (approved):

PR #3203, opened 2025-09-03
Added support for jobs resume progress tracking in library

gapra-msft:
Use new job progress tracking logic. Copy and sync will also move to using this progress tracker eventually, so we can remove it from the glcm code. 

See this for design of job lifecycle manager
https://msazure.visualstudio.com/One/_wiki/wikis/One.wiki/846803/JobLifecycleManager


azcopy/jobsResume.go
```diff
@@ -233,3 +269,86 @@ func getSourceAndDestinationServiceClients(
...
+
+func (r *resumeProgressTracker) CompletedEnumeration() bool {
+	return true // resume does not enumerate, so this is always true
+}
+
+func (r *resumeProgressTracker) GetJobID() common.JobID {
+	return r.jobID
+}
+
+func (r *resumeProgressTracker) GetElapsedTime() time.Duration {
```
gapra-msft, line 350: TODO: I should probably keep a variable for elapsed time that gets set once jobDone is true in CheckProgress. That way Elapsed time is a true measure of when the job ended.

azcopy/jobLifecycleManager.go
```diff
@@ -0,0 +1,233 @@
+package azcopy
```
dphulkar-msft, line 21: msft licensing missing

azcopy/jobLifecycleManager.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	// Signal completion (non-blocking)
+	select {
+	case j.completionChan <- struct{}{}:
+	default:
+	}
+}
+
+// TODO : rename interface method to OnError to match other methods
```
dphulkar-msft, line 106: todo remainder

azcopy/jobLifecycleManager.go
```diff
@@ -0,0 +1,233 @@
...
+						// OnComplete will mark the job as done to bring down the progress reporter and then call the user provided Handler
+						j.OnComplete()
+					}
+				}
+			case <-ctx.Done():
+				cancelCalled = true
+				j.handler.Info("Cancellation requested. Beginning clean shutdown...")
+				if !reporter.CompletedEnumeration() {
+					answer := j.handler.Prompt("The enumeration (source only for copy, source/destination comparison for sync) is not complete, "+
+						"cancelling the job at this point means it cannot be resumed.",
```
dphulkar-msft, line 180: Should we mention jobID here?

common/constants.go
```diff
@@ -0,0 +1,12 @@
+package common
```
dphulkar-msft, line 21: msft licensing missing

common/output.go
```diff
@@ -439,6 +439,39 @@ Final Job Status: %v%s%s
...
+Total Number of Transfers: %v
+Number of File Transfers Completed: %v
+Number of Folder Transfers Completed: %v
+Number of File Transfers Failed: %v
+Number of Folder Transfers Failed: %v
+Number of File Transfers Skipped: %v
+Number of Folder Transfers Skipped: %v
+Total Number of Bytes Transferred: %v
+Percent Complete (approx): %.1f
+Final Job Status: %v
```
dphulkar-msft, line 458: Thought: We are currently repeating the string line:445-458 across multiple jobs. Instead of duplicating it, can we define it once as a variable and reuse it wherever needed?

gapra-msft: I'll add an item to clean up this code for after the library refactor

azcopy/jobLifecycleManager.go
gapra-msft: This todo is not critical right now, and I want to limit changes to files - so we can do this after sync and copy are done

azcopy/jobLifecycleManager.go
gapra-msft: this will diverge from what customers are experiencing today, so choosing not to do this right now

azcopy/jobLifecycleManager.go
```diff
@@ -0,0 +1,253 @@
...
+
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"github.com/Azure/azure-storage-azcopy/v10/jobsAdmin"
+)
+
+type JobProgressTracker interface {
+	// Start - calls OnStart
+	Start()
+	// CheckProgress checks the progress of the job and returns the number of transfers completed so far and whether the job is done
+	CheckProgress() (uint32, bool)
```
adreed-msft, line 38: I feel like outputting a struct may be better? Some basic detalis like

- How many transfers are scheduled?
- How many transfers are completed?
- Are we finished enumerating?
- How many bytes are scheduled/transferred? i.e. what's the "density" of the currently completed work?

might make more sense than just, how many transfers and are we done?

azcopy/jobLifecycleManager.go
```diff
@@ -0,0 +1,253 @@
...
+	// CompletedEnumeration checks whether the enumeration is complete
+	CompletedEnumeration() bool // Whether we should prompt before cancelling
```
adreed-msft, line 40: If the prior comment, then this should be merged into that ig.

azcopy/jobLifecycleManager.go
```diff
@@ -0,0 +1,253 @@
...
+	default:
+	}
+}
+
+// TODO : rename interface method to OnError to match other methods
+func (j *JobLifecycleManager) Error(err string) {
+	j.OnError(err)
+}
+
+func (j *JobLifecycleManager) OnError(err string) {
```
adreed-msft, line 111: Code dupey with OnComplete

azcopy/jobLifecycleManager.go
```diff
@@ -0,0 +1,253 @@
...
+	// Execute completion functions
+	for _, fn := range j.completionFuncs {
+		if fn != nil {
+			fn()
+		}
+	}
+
+	// Signal completion (non-blocking)
+	select {
+	case j.completionChan <- struct{}{}:
```
adreed-msft: closing the channel is the idiomatic way to handle "completion" signals, as otherwise a thread needs to continually write values to a channel if there are multiple potential listeners

azcopy/jobLifecycleManager.go
```diff
@@ -0,0 +1,253 @@
...
+	// Execute completion functions
+	for _, fn := range j.completionFuncs {
+		if fn != nil {
+			fn()
+		}
+	}
+
+	// Signal error (non-blocking)
+	select {
+	case j.errorChan <- err:
```
adreed-msft: ditto

azcopy/jobLifecycleManager.go
```diff
@@ -0,0 +1,253 @@
...
+			return errors.New(lastError)
+		}
+		return nil
+	}
+
+	// wait until OnComplete or OnError is called
+	select {
+	case <-j.completionChan:
+		return nil
+	case errMsg := <-j.errorChan:
```
adreed-msft: in this case, writing the error, closing the channel, and then re-checking would be idiomatic.

adreed-msft (approved):

common/output.go
```diff
@@ -439,6 +439,39 @@ Final Job Status: %v%s%s
...
+Number of Symlink Transfers: %v
+Total Number of Transfers: %v
+Number of File Transfers Completed: %v
+Number of Folder Transfers Completed: %v
+Number of File Transfers Failed: %v
+Number of Folder Transfers Failed: %v
+Number of File Transfers Skipped: %v
+Number of Folder Transfers Skipped: %v
+Total Number of Bytes Transferred: %v
+Percent Complete (approx): %.1f
```
wonwuakpa-msft, line 457: Should we also add the Elapsed time metric for jobs of the `Show` type in the job summary?

cmd/jobsShow.go
```diff
@@ -140,44 +140,10 @@ func PrintJobProgressSummary(summary common.ListJobSummaryResponse) {
 				return string(buf)
 			}, common.EOutputMessageType.GetJobSummary())
 	}
-	glcm.Exit(func(format common.OutputFormat) string {
```
wonwuakpa-msft, line 143: Should we refactor `PrintJobTransfers` in the same way to use OnComplete?

azcopy/jobLifecycleManager.go
gapra-msft: The LifecycleManager doesnt need to know most of this stuff though, so I'm not so sure this is useful to return. It only cares about the number of total transfers scheduled and if the job is done

azcopy/jobLifecycleManager.go
gapra-msft: The part of the LifecycleManager that cares about Completed Enumeration (on cancel) is separate from the part that cares about the progress updates, so I think it makes sense to keep them separate

azcopy/jobLifecycleManager.go
```diff
@@ -0,0 +1,230 @@
...
+	// Signal completion by closing the channel
+	close(j.doneChan)
+}
+
+func (j *JobLifecycleManager) OnComplete() {
+	j.markComplete(nil)
+}
+
+// TODO : rename interface method to OnError to match other methods
+func (j *JobLifecycleManager) Error(err string) {
```
wonwuakpa-msft, line 107: Do we need this method if it just calls OnError?

azcopy/jobLifecycleManager.go
wonwuakpa-msft: Adding to this -- In our codebase right now, [this is the information ](https://github.com/Azure/azure-storage-azcopy/blob/c5e0e71eb391dca15ce47aef49987428f2649883/ste/mgr-JobPartMgr.go#L131) we hold relating to the progress of each job part. Do we not need skipped and failed transfers?

wonwuakpa-msft (approved):

common/output.go
gapra-msft: With the refactor, the goal is to only show what we currently show, so I think we can skip this for now. These are longer term improvements we can think about later.

azcopy/jobLifecycleManager.go
gapra-msft: This method is currently there so jobLifecycleManager implements the JobErrorHandler interface that I created in a previous PR.

gapra-msft: We will one day rename that to OnError and this method will go away then

azcopy/jobLifecycleManager.go
gapra-msft: This method is consumed by the jobLifecycleManager, so it exposes only the things the jobLifecycleManager cares about - aka transfers completed and is job done. The stuff the customer cares about are exposed via the handler thats passed to the resume progress tracker.

cmd/jobsShow.go
gapra-msft: I think I need to actually refactor the stuff I moved to using OnComplete back to Exit. I thought this was necessary but its actually not

wonwuakpa-msft: Ok, why isn't it necessary anymore?

PR #3205, opened 2025-09-03
Moved some traverser types into its own package

gapra-msft:
This PR is a simple refactor - moving some key traverser types into its own package. This allows for clearer separation of AzCopy components

### DIFF
--- a/azcopy/clientUtil.go
+++ b/azcopy/clientUtil.go
@@ -1,3 +1,23 @@
+// Copyright © 2025 Microsoft <wastore@microsoft.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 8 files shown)


cmd/list.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/lease"
 	"github.com/Azure/azure-storage-azcopy/v10/azcopy"
+	traverser2 "github.com/Azure/azure-storage-azcopy/v10/traverser"
```
wonwuakpa-msft, line 37: nitpic: Should we name the package import traverser2?

gapra-msft: There is prob a variable named traverser, so the refactor made the import called traverser2. I dont think this is too big a deal so leaving it as is

PR #3211, opened 2025-09-10
Introduce FileSMB for non File s2s or upload/download paths

dphulkar-msft:
## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?
UT added


cmd/flagsValidation.go
```diff
@@ -335,20 +335,22 @@ func validateShareProtocolCompatibility(
 	fromTo common.FromTo,
 ) error {
 
-	direction := "from"
+	location, direction := "source", "from"
```
gapra-msft: nilt but is there a point to having location and direction variables here? Why not just paste this in the error directly instead of formatting a string?

cmd/validators.go
```diff
@@ -57,6 +57,22 @@ func ValidateFromTo(src, dst string, userSpecifiedFromTo string) (common.FromTo,
 		userFromTo = common.EFromTo.FileLocal()
 	} else if userFromTo == common.EFromTo.FileSMBFileSMB() {
 		userFromTo = common.EFromTo.FileFile()
+	} else if userFromTo == common.EFromTo.FileSMBBlob() {
```
gapra-msft, line 63: Should we add a comment here that this is changing the user FileSMB flag to File so we dont have to change too much code elsewhere in Azcopy for the timebeing

dphulkar-msft: sure will add it

cmd/flagsValidation.go
dphulkar-msft: This function is called for both source and destination. So we will need these variables to error out based on the source or destiantion.

gapra-msft: but you know if its source or destination by the code below where you are using it to print right?

gapra-msft: from the FromTo

cmd/flagsValidation.go
```diff
@@ -335,20 +335,22 @@ func validateShareProtocolCompatibility(
...
+		location, direction = "destination", "to"
 	}
 
 	// We can ignore the error if we fail to get the share properties.
 	shareProtocol, _ := getShareProtocolType(ctx, serviceClient, resource, protocol)
 
 	if shareProtocol == common.ELocation.File() {
 		if isSource && fromTo.From() != common.ELocation.File() {
-			return fmt.Errorf("The %s share has SMB protocol enabled. To copy %s a SMB share, use the appropriate --from-to flag value", direction, direction)
+			return fmt.Errorf("The %s share has SMB protocol enabled. "+
```
gapra-msft: can't you just use the appropriate string in the error message here? Checking is Source above and here seems redundant

gapra-msft:
```suggestion
			return errors.New("The source share has SMB protocol enabled. "+
```

cmd/flagsValidation.go
```diff
@@ -335,20 +335,22 @@ func validateShareProtocolCompatibility(
...
+				"To copy %s a SMB share, use the appropriate --from-to flag value", location, direction)
```
gapra-msft:
```suggestion
				"To copy from a SMB share, use the appropriate --from-to flag value")
```

cmd/flagsValidation.go
```diff
@@ -335,20 +335,22 @@ func validateShareProtocolCompatibility(
...
 		}
 		if !isSource && fromTo.To() != common.ELocation.File() {
-			return fmt.Errorf("The %s share has SMB protocol enabled. To copy %s a SMB share, use the appropriate --from-to flag value", direction, direction)
+			return fmt.Errorf("The %s share has NFS protocol enabled. "+
```
gapra-msft:
```suggestion
			return errors.New("The destination share has NFS protocol enabled. "+
```

cmd/flagsValidation.go
```diff
@@ -335,20 +335,22 @@ func validateShareProtocolCompatibility(
...
+				"To copy %s a NFS share, use the appropriate --from-to flag value", location, direction)
```
gapra-msft:
```suggestion
				"To copy to a NFS share, use the appropriate --from-to flag value")
```

cmd/flagsValidation.go
dphulkar-msft: done

cmd/flagsValidation.go
dphulkar-msft: yeah I got ur point now. Updated

dphulkar-msft: done

PR #3222, opened 2025-09-17
Fix symlink direct targets

adreed-msft:
Call out the change on how we deal with symlink direct targets in the Changelog

## Description

- **Feature / Bug Fix**: When using a symlink as the sole target of a transfer, AzCopy reads the symlink as a file and attempts to transfer as though it is following the symlink. This behavior is incorrect.

- **Related Links**:
- Teams thread: StgExp-AzCopy Sync

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Contains automatic tests for the original bug which fail against a old version of AzCopy.

### DIFF
--- a/e2etest/newe2e_resource_managers_local.go
+++ b/e2etest/newe2e_resource_managers_local.go
@@ -358,7 +358,7 @@ func (l *LocalObjectResourceManager) ListChildren(a Asserter, recursive bool) ma
 
 func (l *LocalObjectResourceManager) GetProperties(a Asserter) ObjectProperties {
 	a.HelperMarker().Helper()
-	stats, err := os.Stat(l.getWorkingPath())
+	stats, err := os.Lstat(l.getWorkingPath())
 	if err != nil { // Prevent nil dereferences
 		a.NoError("failed to get stat", err)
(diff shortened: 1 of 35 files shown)


gapra-msft (approved):

e2etest/newe2e_resource_managers_local.go
```diff
@@ -358,7 +358,7 @@ func (l *LocalObjectResourceManager) ListChildren(a Asserter, recursive bool) ma
 
 func (l *LocalObjectResourceManager) GetProperties(a Asserter) ObjectProperties {
 	a.HelperMarker().Helper()
-	stats, err := os.Stat(l.getWorkingPath())
+	stats, err := os.Lstat(l.getWorkingPath())
```
wonwuakpa-msft, line 361: Do we need a conditional here for when we follow vs preserve symlinks?

adreed-msft: I'm gonna lean towards no-- If we're testing an object, we want to know what it is, and what the properties of the symlink itself is, if it is one.

PR #3229, opened 2025-09-22
C2C Specific Changes for Blob to Blob V1 Support

kylemoore-microsoft:
## Description

1) Adding support for Blob/BlobFS sources/targets in syncOrchestrator
2) Making blob acces tier, preserve access tier, preserve permissions public for sync cooked copy args such that they can be set in mover repo

Mover C2C Blob to Blob V1 Feature Support

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

### DIFF
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -133,7 +133,7 @@ func (raw rawSyncCmdArgs) toOptions() (cooked cookedSyncCmdArgs, err error) {
 		backupMode:                       raw.backupMode,
 		isNFSCopy:                        raw.isNFSCopy,
 		putMd5:                           raw.putMd5,
-		s2sPreserveBlobTags:              raw.s2sPreserveBlobTags,
+		S2sPreserveBlobTags:              raw.s2sPreserveBlobTags,
 		cpkByName:                        raw.cpkScopeInfo,
 		cpkByValue:                       raw.cpkInfo,
 		mirrorMode:                       raw.mirrorMode,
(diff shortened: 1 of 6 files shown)


kylemoore-microsoft:
@shnayak-msft want to mention here in mover it does not appear we are using moverDependencies for copy. Check cookCopyCmdArgs in scanner.go. We are directly setting the members of CookedCopyCmdArgs which has all public members. 

I decided to follow the same pattern and change the args I needed on cookedSyncCmdArgs because it was convenient. 

I see now that sync is using moverDependencies to set the cookedSyncCmdArgs. What is the reason for the difference between sync / copy?

cmd/sync.go
```diff
@@ -466,7 +466,7 @@ type cookedSyncCmdArgs struct {
 
 	// options
 	compareHash             common.SyncHashType
-	preservePermissions     common.PreservePermissionsOption
```
shnayak-msft, line 469: I believe these properties are made public so these can be accessed from xdatamove. Instead of exposing them in the struct, can we create public getters like below in moverDependencies.go and access it from xdatamove. I want to condense all dependencies into this one file and avoid exposing it globally. This file will be conditionally built for mover build tag only when we move to main. 

func (cooked *cookedSyncCmdArgs) StripTopDir() bool {
	return cooked.stripTopDir
}

kylemoore-microsoft: Check cookCopyCmdArgs in scanner.go. We are directly setting the members of CookedCopyCmdArgs which has all public members.

I decided to follow the same pattern and change the args I needed on cookedSyncCmdArgs because it was convenient.

I see now that sync is using moverDependencies to set the cookedSyncCmdArgs. What is the reason for the difference between sync / copy?

shnayak-msft: The original mover team duplicated a layer of copy code from AzCopy (scanner.go) to enable error and scanning metrics reporting. It is still not available in azcopy. That time few properties were exposed and also checked into AzCopy main so that is the pattern that stayed. They also exposed other properties in AzCopy later which were not merged then. The problem with this is exposing these properties directly would change many files unnecessarily in AzCopy main and cause conflicts during merge and also gives the impression that we are changed many layers. 

Eventual goal is to use the new AzCopy go package to interact with AzCopy. Until we do that, I wanted to condense dependencies in AzCopy to this one file so it can be removed when we use the go package. Benefit of that is we can build it conditionally as well.

So, whatever is exposed in AzCopy main, we use directly. Whatever is not, we use through moverDependencies.go.

kylemoore-microsoft: Got it. I made the change such that we set all sync flags through moverDependencies.go

PR #3230, opened 2025-09-22
Fix for issue seen with read-only source files on Azure Files targets…

bewurman-msft:
## Description
This is a fix for the issue we saw where some SMB source files set to read-only would fail transfer due to 409 error from the target stating that the file on the target is read-only and cannot be modified. Since the AzFiles REST API has two separate calls for creating a file on the target and transferring data to the created target file, if we create the file with read-only set to true, we can't put any data in it. 
Before, if the target returned a 404 error due to a missing parent folder, we would set the target's read-only bit to false for the Prologue, and set it back to true for the Epilogue, so we saw the issue with all files that don't have targets 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?
Used a source share with some read-only files. Before the change, azcopy copy would fail to transfer any of those files if it tried to transfer them after the folder was created on the target side. After the change, they would all succeed. 

### DIFF
--- a/ste/sender-azureFile.go
+++ b/ste/sender-azureFile.go
@@ -251,6 +251,9 @@ func (u *azureFileSenderBase) Prologue(state common.PrologueState) (destinationM
 		creationProperties.LastWriteTime = &minimalLwt
 	}
 
+	// Set this before file creation
+	createOptions.SMBProperties = &creationProperties
+
 	err := common.DoWithOverrideReadOnlyOnAzureFiles(u.ctx,
 		func() (interface{}, error) {
 			return u.getFileClient().Create(u.ctx, info.SourceSize, createOptions)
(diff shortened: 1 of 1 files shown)


cmd/credentialUtil.go
```diff
@@ -586,14 +586,23 @@ func createClientOptions(logger common.ILoggerResetable, srcCred *common.ScopedT
...
-	return ste.NewClientOptions(policy.RetryOptions{
-		MaxRetries:    ste.UploadMaxTries,
-		TryTimeout:    ste.UploadTryTimeout,
-		RetryDelay:    ste.UploadRetryDelay,
-		MaxRetryDelay: ste.UploadMaxRetryDelay,
-	}, policy.TelemetryOptions{
-		ApplicationID: common.AddUserAgentPrefix(common.UserAgent),
-	}, ste.NewAzcopyHTTPClient(frontEndMaxIdleConnectionsPerHost), logOptions, srcCred, reauthCred)
+	// Job-level/global client if available so we reuse connections and transports.
+	client := common.GetGlobalHTTPClient(logger)
```
shnayak-msft, line 590: This is already present in mover/stage2. Make sure you rebase with mover/stage2

bewurman-msft: Rebased, any discrepancies between this branch and mover/stage2 should be resolved now

PR #3234, opened 2025-09-26
Users/sanvern/c2cprivate networking 09122025

sgv-msft:
## Description
This PR contains following changes.
1. Update of minio v7 package
2. Minio client changes compatible to V7 package
3. Initial Changes related to Private networking for C2C.
4. Use HttpTransport in RoundRobin fashion when multiple private endpoints are use.
5. Minor fixes related to Privatenetwork found during testing.

- **Feature / Bug Fix**:  Feature: Support Private networking for C2C workflow

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ X] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/common/privateNetwork.go
+++ b/common/privateNetwork.go
@@ -110,8 +110,9 @@ func NewRoundRobinTransport(ips []string, host string, cooldownInSecs int, ipRet
 // For each chosen IP, it will retry the same IP rr.perIPRetries times (with a small delay)
 // before marking the IP unhealthy and moving on to the next IP.
 func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	var lastErr error
+	var lastErrMsg string
 	var peIP string
+	var lastErrorCode int
 
 	log.Printf("*****Request Method: %s, Host: %s, Query: %s, Body: %v, URI: %s****", req.Method, req.URL.Host, req.URL.RawQuery, req.Body, req.RequestURI)
 	numPrivateEndpoints := GetGlobalPrivateEndpointIPCount()
@@ -197,8 +198,6 @@ func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, err
 					log.Printf("[Counter=%d Retry=%d] SUCCESS using IP %s", idx, ipAttempt, peIP)
 				}
 
-				globalIPsMutex.Lock()
-				defer globalIPsMutex.Unlock()
 				globalPrivateEndpointIPs[idx].IncrementNumRequests()
 				return resp, nil
 			}
@@ -227,9 +226,15 @@ func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, err
 					errCode = 0
 					errMsg = ""
 				}
+			} else {
+				// No response - parse error from the error object itself
+				errCode = 0
+				errMsg = err.Error()
+				log.Printf("[Counter=%d Retry=%d] Network error with no response, Error Message:%s retryable:%v", idx, ipAttempt, errMsg, isRetryableErr)
 			}
 
-			lastErr = err
+			lastErrMsg = errMsg
+			lastErrorCode = errCode
 
 			// If we still have per-IP attempts left, wait and retry the same IP
 			if (ipAttempt+1 < rr.perIPRetries) && (isRetryableErr || isNetworkRetryableErr) {
@@ -237,24 +242,21 @@ func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, err
 				time.Sleep(rr.perIPRetryDelay)
 				continue
 			}
+		}
 
-			// Exhausted per-IP retries: mark the IP unhealthy, record time, update global map and healthy pool,
-			// then break to pick another IP (outer loop continues).
-			// attempt to mark unhealthy: 0 -> 1
-			if atomic.CompareAndSwapUint32((*uint32)(&entry.ConnectionStatus), uint32(Healthy), uint32(Unhealthy)) {
-				entry.MarkUnhealthy(errCode, errMsg)
-				updateStatus := UpdateGlobalPrivateEndpointIP(idx, entry)
-				log.Printf("Updating Private Endpoint:%s connection state from HEALTHY->UNHEALTHY after error response with Error Code %d ErrorMsg:%s at %v Update Status:%s", peIP, entry.LastErrCode, entry.LastErrMsg, entry.LastChecked, updateStatus)
-			}
-			// break inner loop; outer loop will select another IP (or finish if attempts exhausted)
-			break
+		// Exhausted per-IP retries: mark the IP unhealthy, record time, update global map and healthy pool,
+		// then break to pick another IP (outer loop continues).
+		// attempt to mark unhealthy: 0 -> 1
+		if (lastErrMsg != "") && atomic.CompareAndSwapUint32((*uint32)(&entry.ConnectionStatus), uint32(Healthy), uint32(Unhealthy)) {
+			entry.MarkUnhealthy(lastErrorCode, lastErrMsg)
+			log.Printf("Updating Private Endpoint:%s connection state from HEALTHY->UNHEALTHY after error response with Error Code %d ErrorMsg:%s at %v", peIP, entry.LastErrCode, entry.LastErrMsg, entry.LastChecked)
 		}
 		// continue outer loop to try next IP (if any attempts remain)
 	}
 
 	fmt.Errorf("No healthy Private Endpoint IPs are available")
 	// All attempts exhausted
-	return nil, fmt.Errorf("Request failed after trying all healthy Private Endpoint IPs. Last error from IP %s: %v", peIP, lastErr)
+	return nil, fmt.Errorf("Request failed after trying all healthy Private Endpoint IPs. Last error from IP %s: %v", peIP, lastErrMsg)
 }
 
 // Close cleans up idle connections and stops the periodic refresh goroutine
@@ -307,20 +309,6 @@ func GetGlobalPrivateEndpointIPCount() int {
 		return 0
 	}
 	return len(globalPrivateEndpointIPs)
-}
-
-// UpdateGlobalPrivateEndpointIP updates a specific IP entry in the global list
-func UpdateGlobalPrivateEndpointIP(index uint64, updatedEntry *IPEntry) bool {
-	globalIPsMutex.Lock()
-	defer globalIPsMutex.Unlock()
-
-	if globalPrivateEndpointIPs == nil || index >= uint64(len(globalPrivateEndpointIPs)) || updatedEntry == nil {
-		log.Printf("Updation of PrivateEndpoint List failed for the Index:%d with Length:%d\n ", index, len(globalPrivateEndpointIPs))
-		return false
-	}
-
-	globalPrivateEndpointIPs[index] = updatedEntry
-	return true
 }
 
 // Function to check if private network is enabled or not. By default it should be disabled and return false
--- a/ste/xfer-anyToRemote-file.go
+++ b/ste/xfer-anyToRemote-file.go
@@ -413,8 +413,8 @@ func scheduleSendChunks(jptm IJobPartTransferMgr, srcPath string, srcFile common
 	// If the source is S3, we need to get a pre-signed URL for the source file.
 	// These conditions to decide to go PutBlob path can be improved later for all the sourceTypes.
 	isPrivateNetworkingEnabled := common.IsPrivateNetworkEnabled()
-	isSourceType := jptm.FromTo().From() == common.ELocation.S3()
-	isSourcePrivate := isSourceType && isPrivateNetworkingEnabled
+	isSourceTypeS3 := jptm.FromTo().From() == common.ELocation.S3()
+	isSourcePrivate := isSourceTypeS3 && isPrivateNetworkingEnabled
 
 	if srcInfoProvider.IsLocal() || isSourcePrivate {
 		md5Channel = s.(uploader).Md5Channel()


common/privateNetwork.go
```diff
@@ -0,0 +1,196 @@
...
+		clonedReq.Host = rr.host
+
+		fmt.Println("[RoundTrip] Updated request: %s://%s%s", clonedReq.URL.Scheme, clonedReq.URL.Host, clonedReq.URL.Path)
+		fmt.Println("[RoundTrip] Updated Request Header Host: %s", clonedReq.Host)
+
+		resp, err := rr.transport.RoundTrip(clonedReq)
+		if err == nil {
+			// Success: mark IP healthy
+			atomic.StoreInt32(&entry.unhealthy, 0)
+			rr.refreshHealthyPool()
```
pbanakar-microsoft: This is an overload. For every http request, it will traverse the array linearly introducing latency.
is this needed?
May be you could refresh only when there is change in the healthstatus.

common/credentialFactory.go
```diff
@@ -62,6 +63,31 @@ func (o CredentialOpOptions) panicError(err error) {
...
 }
 
+// Constants for private network transport
+const HttpsRetryAttempts = 5
+const PeCheckCooldownTimeInSecs = 3600
+
+func createS3ClientForPrivateNetwork(credInfo CredentialInfo) (*minio.Client, error) {
+	peIP := privateNetworkArgs.PrivateEndpointIPs
+	baseS3Host := credInfo.S3CredentialInfo.Endpoint
+	// Doublecheck endpoint should contain bucketname : "<bucketname>.s3.<region>.amazonaws.com"
```
pbanakar-microsoft: Is this still a TODO? Are we still checking if the bucket name is mandatory to create s3 client?

common/credentialFactory.go
```diff
@@ -80,27 +106,39 @@ func CreateS3Credential(ctx context.Context, credInfo CredentialInfo, options Cr
...
 		cred := credentials.NewStatic("", "", "", credentials.SignatureAnonymous)
-		return minio.NewWithOptions(credInfo.S3CredentialInfo.Endpoint, &minio.Options{Creds: cred, Secure: true, Region: credInfo.S3CredentialInfo.Region})
+		return minio.New(credInfo.S3CredentialInfo.Endpoint, &minio.Options{Creds: cred, Secure: true, Region: credInfo.S3CredentialInfo.Region})
 	}
 	//support custom credential provider
 	if credInfo.S3CredentialInfo.Provider != nil {
 		fmt.Println("Using custom credentials")
-		creds := credentials.New(credInfo.S3CredentialInfo.Provider)
-		s3Client, err := minio.NewWithCredentials(credInfo.S3CredentialInfo.Endpoint, creds, true, credInfo.S3CredentialInfo.Region)
+		s3Client, err := CreateS3ClientFromProvider(credInfo)
```
pbanakar-microsoft, line 140: This shouldn't be just limited to credentialProvider case. We should make it common to both credProvider and access keys. 
We should be able to create s3 client for private networking even for HMAC/accessKeys.
Whenever we start supporting HMAC keys in storage mover, we wouldn't need to make changes here.

common/credentialFactory.go
```diff
@@ -62,6 +63,31 @@ func (o CredentialOpOptions) panicError(err error) {
 	}
 }
 
+// Constants for private network transport
+const HttpsRetryAttempts = 5
+const PeCheckCooldownTimeInSecs = 3600
```
pbanakar-microsoft: Is this 60mins? Should we keep smaller time here, like 10mins?

common/privateNetwork.go
```diff
@@ -0,0 +1,196 @@
...
+
+// RoundTrip retries request with different IPs up to rr.maxRetries
+func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	var lastErr error
+	var peIP string
+	for attempt := 1; attempt < (rr.maxRetries + 1); attempt++ {
+		healthy := rr.healthyIPs.Load().([]*IPEntry)
+		if len(healthy) == 0 {
+			fmt.Errorf("[Attempt %d] No healthy IPs available", attempt)
+			return nil, fmt.Errorf("no healthy IPs available")
```
pbanakar-microsoft: we need to fail the job in this case, right?

common/privateNetwork.go
```diff
@@ -0,0 +1,196 @@
...
+	PrivateEndpointIPs []string // List of private endpoint IPs
+	BucketName         string   // Bucket Name required to form Endpoint URL
+}
+
+var privateNetworkArgs PrivateNetworkConfig = PrivateNetworkConfig{}
+
+// IPEntry holds one private IP with health info
+type IPEntry struct {
+	IP          string
+	unhealthy   int32 // 0 = healthy, 1 = unhealthy
```
pbanakar-microsoft: Please make this variable name as healthStatus or connectionStatus. And the values can be enum values like healthy, unhealthy.
Having variable name same as one of its value is a bit confusing.

common/privateNetwork.go
```diff
@@ -0,0 +1,196 @@
...
+
+	return nil, fmt.Errorf("all retries for ip %v failed after %d attempts: %w", peIP, rr.maxRetries, lastErr)
+}
+
+// refreshHealthyPool updates the healthy IP list atomically
+func (rr *RoundRobinTransport) refreshHealthyPool() {
+	var healthy []*IPEntry
+	for _, e := range rr.ips {
+		log.Printf("refreshHealthyPool Counter: %d IP Address: %s Unhealthy: %d\n", rr.counter, e.IP, e.unhealthy)
+		if atomic.LoadInt32(&e.unhealthy) == 0 ||
```
pbanakar-microsoft: changing this to enum value of healthy or unhealthy helps to understand code easily.

common/privateNetwork.go
```diff
@@ -0,0 +1,196 @@
...
+	unhealthy   int32 // 0 = healthy, 1 = unhealthy
+	lastChecked time.Time
+}
+
+// RoundRobinTransport implements http.RoundTripper with retries and cooldowns
+type RoundRobinTransport struct {
+	ips        []*IPEntry
+	host       string
+	healthyIPs atomic.Value // []*IPEntry, cached healthy list
+	counter    uint32
```
pbanakar-microsoft: what is this counter for? Should this be initialized to 0 somewhere?

common/privateNetwork.go
```diff
@@ -0,0 +1,196 @@
...
+	var lastErr error
+	var peIP string
+	for attempt := 1; attempt < (rr.maxRetries + 1); attempt++ {
+		healthy := rr.healthyIPs.Load().([]*IPEntry)
+		if len(healthy) == 0 {
+			fmt.Errorf("[Attempt %d] No healthy IPs available", attempt)
+			return nil, fmt.Errorf("no healthy IPs available")
+		}
+
+		idx := atomic.AddUint32(&rr.counter, 1)
```
pbanakar-microsoft: idx and rr.counter values can possible go upto max(uint32).
Can we do mod(%) operator on these values itself, so that idx and rr.counter will never go beyound healthIP max array size?

common/privateNetwork.go
```diff
@@ -0,0 +1,196 @@
...
+		entry := healthy[idx%uint32(len(healthy))]
```
pbanakar-microsoft: its better to do this mod operation in the previous line.

common/privateNetwork.go
```diff
@@ -0,0 +1,196 @@
...
+		peIP := entry.IP
```
pbanakar-microsoft: peIP := entry.IP , := creates a new varialbe peIP, the declaration above is unused.

common/credentialFactory.go
sgv-msft: If we keep it short, retrying faulty connection might cause performance issues. Thats the reason of keeping it 60mins.  We need to fine tune this based on our performance testing.

common/privateNetwork.go
sgv-msft: Yes thats correct!

common/privateNetwork.go
sgv-msft: Counter is used to derive index and element of the PE array for round robin logic.

common/privateNetwork.go
sgv-msft: Goot catch. Will change the logic not to iterate for all successful response.

common/credentialFactory.go
sgv-msft: bucketName is needed to form endpoint for RoundRobinTransport.

common/privateNetwork.go
sgv-msft: Removed linear check and only state is changed then the operation is performed.

common/privateNetwork.go
sgv-msft: Currently mod operator is used	entry := healthy[idx**%**uint64(len(healthy))]

common/privateNetwork.go
```diff
@@ -0,0 +1,270 @@
...
+	PrivateEndpointIPs []string // List of private endpoint IPs
+	BucketName         string   // Bucket Name required to form Endpoint URL
+}
+
+var privateNetworkArgs PrivateNetworkConfig = PrivateNetworkConfig{}
+
+// IPEntry holds one private IP with health info
+type IPEntry struct {
+	IP          string
+	unhealthy   uint32 // 0 = healthy, 1 = unhealthy
```
akchidan: Since we are referring these fields from a different package, these fields must be capitalized

type IPEntry struct {
    IP          string    `json:"ip"`
    Unhealthy   uint32    `json:"unhealthy"`
    LastChecked time.Time `json:"lastChecked"`
    LastErrCode int       `json:"lastErrCode"`
    LastErrMsg  string    `json:"lastErrMsg"`
}

common/privateNetwork.go
sgv-msft: we need the idx for the mod operation, I doubt reordering might work.

common/privateNetwork.go
```diff
@@ -0,0 +1,277 @@
...
+type PrivateNetworkConfig struct {
+	Enabled            bool     // By default private network is disabled unless user explicitly enables it
+	PrivateEndpointIPs []string // List of private endpoint IPs
+	BucketName         string   // Bucket Name required to form Endpoint URL
+}
+
+var privateNetworkArgs PrivateNetworkConfig = PrivateNetworkConfig{}
+
+// PE Health Status Enum
+type HealthStatus uint32
```
pbanakar-microsoft, line 47: This can be uint8. we don't expect too many enum values to grow on this variable.

common/privateNetwork.go
```diff
@@ -0,0 +1,277 @@
...
+	var lastErr error
+	var peIP string
+
+	healthy := rr.healthyIPs.Load().([]*IPEntry)
+	initialHealthyCount := len(healthy)
+	for attempt := 1; attempt <= initialHealthyCount; attempt++ {
+		if len(healthy) == 0 {
+			fmt.Errorf("No healthy Private Endpoint IPs are available", attempt)
+			return nil, fmt.Errorf("no healthy Private Endpoint IPs are available")
+		}
```
pbanakar-microsoft: will this condition be ever met? if len of healthy is 0, it won't enter for loop?

common/privateNetwork.go
```diff
@@ -0,0 +1,378 @@
...
+		entry := globalPrivateEndpointIPs[idx]
+		peIP = entry.IP
+		//log.Printf("Selected Private endpoint IP: %s Unhealth Status: %d LastTime: %v\n",
+		// peIP, entry.ConnectionStatus, entry.LastChecked)
+
+		// Skip if still in cooldown
+		if entry.ConnectionStatus == Unhealthy {
+			if time.Since(entry.LastChecked) >= rr.cooldown {
+				entry.MarkHealthy()
+				updateStatus := UpdateGlobalPrivateEndpointIP(idx, entry)
```
pbanakar-microsoft: This function is not required. Its redundant.
entry variable is a pointer to an array value, entry.Markhealthy() will update the array value itself, there is no need to put the entry back into the array.

common/privateNetwork.go
```diff
@@ -0,0 +1,382 @@
...
+			log.Printf("Global Endpoint List is already initial with Private Endpoints:%d ", len(globalPrivateEndpointIPs))
+		}
+	}
+}
+
+// GetGlobalPrivateEndpointIPCount returns the number of global private endpoint IPs
+func GetGlobalPrivateEndpointIPCount() int {
+	globalIPsMutex.RLock()
+	defer globalIPsMutex.RUnlock()
+
```
pbanakar-microsoft, line 307: lock is not required.
the global PE IP count will not change after initialization of globalPEIParray

common/privateNetwork.go
```diff
@@ -0,0 +1,382 @@
...
+			// Override destination to PE IP:Port and preserve original Host header
+			clonedReq.URL.Scheme = req.URL.Scheme
+			clonedReq.URL.Host = peIP
+			clonedReq.Host = rr.host
+			var errCode int
+			var errMsg string
+			var isRetryableErr bool
+			var isS3AccessDeniedErr bool
+
+			log.Printf("[Counter=%d Retry=%d] Sending request to PrivateEndpoint IP: %s (Host header: %s)", idx, ipAttempt, clonedReq.URL.Host, clonedReq.Host)
```
pbanakar-microsoft, line 156: Please plan to remove the logging in non-err paths after testing.

common/privateNetwork.go
```diff
@@ -0,0 +1,382 @@
...
+						errCode = 0
+						errMsg = ""
+						log.Printf("[Counter=%d Retry=%d] SUCCESS using IP %s without error", idx, ipAttempt, peIP)
+					}
+
+				} else {
+					log.Printf("[Counter=%d Retry=%d] SUCCESS using IP %s", idx, ipAttempt, peIP)
+				}
+
+				globalIPsMutex.Lock()
```
pbanakar-microsoft: globalIPsMutex lock and unlock is not required, since you are already using entry lock and unlock to increment numofrequests.
In general, since RoundTrip always works on the pointer of the array index, you need lock/unlock for that entry, lock/unlock for the global array is not required.

common/privateNetwork.go
```diff
@@ -0,0 +1,382 @@
...
+			}
+
+			// Exhausted per-IP retries: mark the IP unhealthy, record time, update global map and healthy pool,
+			// then break to pick another IP (outer loop continues).
+			// attempt to mark unhealthy: 0 -> 1
+			if atomic.CompareAndSwapUint32((*uint32)(&entry.ConnectionStatus), uint32(Healthy), uint32(Unhealthy)) {
+				entry.MarkUnhealthy(errCode, errMsg)
+				updateStatus := UpdateGlobalPrivateEndpointIP(idx, entry)
+				log.Printf("Updating Private Endpoint:%s connection state from HEALTHY->UNHEALTHY after error response with Error Code %d ErrorMsg:%s at %v Update Status:%s", peIP, entry.LastErrCode, entry.LastErrMsg, entry.LastChecked, updateStatus)
+			}
```
pbanakar-microsoft: Nit: This can be moved outside the inner loop of ip retries, and the above check ipAttempt+1 < rr.perIPRetries will not be requied.
once you come out of inner loop, you have already tried perIPRetries, if there is err or lastErr, you can mark ipEntry unhealthy outside the inner loop.
you woudn't require break; statement as well, the inner loop will exit after perIPRetries.

common/privateNetwork.go
```diff
@@ -0,0 +1,382 @@
...
+				time.Sleep(rr.perIPRetryDelay)
+				continue
+			}
+
+			// Exhausted per-IP retries: mark the IP unhealthy, record time, update global map and healthy pool,
+			// then break to pick another IP (outer loop continues).
+			// attempt to mark unhealthy: 0 -> 1
+			if atomic.CompareAndSwapUint32((*uint32)(&entry.ConnectionStatus), uint32(Healthy), uint32(Unhealthy)) {
+				entry.MarkUnhealthy(errCode, errMsg)
+				updateStatus := UpdateGlobalPrivateEndpointIP(idx, entry)
```
pbanakar-microsoft: same comment as above.
UpdateGlobalPrivateEndpointIP() function is not required, entry is pointer variable, entry.MarkUnHealthy() would have already updated the Global array.

ste/xfer-anyToRemote-file.go
```diff
@@ -410,7 +410,13 @@ func scheduleSendChunks(jptm IJobPartTransferMgr, srcPath string, srcFile common
 	}
 	safeToUseHash := true
 
-	if srcInfoProvider.IsLocal() {
+	// If the source is S3, we need to get a pre-signed URL for the source file.
+	// These conditions to decide to go PutBlob path can be improved later for all the sourceTypes.
+	isPrivateNetworkingEnabled := common.IsPrivateNetworkEnabled()
+	isSourceType := jptm.FromTo().From() == common.ELocation.S3()
```
pbanakar-microsoft: did you mean to name the variable isSourceTypeS3? since this is checking if the source is S3

common/s3ChunkReader.go
```diff
@@ -0,0 +1,385 @@
...
+		GetLifecycleMgr().Info(fmt.Sprintf("s3 BlockingPrefetch: cacheLimiter.WaitUntilAdd failed for chunkId: %v, err: %v", cr.chunkId, err))
+		return err
+	}
+
+	if cr.chunkLogger != nil {
+		cr.chunkLogger.LogChunkStatus(cr.chunkId, EWaitReason.DiskIO())
+	}
+
+	targetBuffer := cr.slicePool.RentSlice(cr.length)
+	GetLifecycleMgr().Info(fmt.Sprintf("S3ChunkReader: Allocated buffer of length: %d for chunkId: %v", cr.length, cr.chunkId))
```
pbanakar-microsoft, line 101: Putting a note to remove all the logging for non-error code paths.

pbanakar-microsoft (approved):

common/privateNetwork.go
sgv-msft: Taken care

common/privateNetwork.go
sgv-msft: Better to have the lock there might be race condition when two clients try to initialize at the same time.

common/privateNetwork.go
sgv-msft: Made it uint32 because of dependency on atomic.CompareAndSwapUint32 library. uin8 doesn't provide this.

common/privateNetwork.go
sgv-msft: Done.

PR #3237, opened 2025-09-26
Catch EADDRNOTAVAILABLE

adreed-msft:
## Description

- **Feature / Bug Fix**: Checks for `EADDRNOTAVAILABLE` in the platform retry policy for `linux` and `darwin`

- **Related Links**:
- PBI # 34972892

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Includes a test for the error catch on `linux` and `darwin`


ste/xferRetryHelper_unix.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"errors"
+	"golang.org/x/sys/unix"
+	"net/http"
+	"strings"
+)
+
+func init() {
+	platformRetryPolicy = func(response *http.Response, err error) bool {
```
gapra-msft, line 33: since linux and darwin are so similar, can we group them together - we can add a comment saying if they do diverge in the future, we can split it into two files.

ste/xferRetryHelper_unix.go
```diff
@@ -0,0 +1,25 @@
+//go:build linux || darwin
```
dphulkar-msft, line 1: missing header

dphulkar-msft (approved):

gapra-msft (approved):

PR #3251, opened 2025-10-09
Refactored and reorganized traverser code

gapra-msft:


azcopy/pathUtils.go
```diff
@@ -0,0 +1,51 @@
+package azcopy
```
dphulkar-msft, line 1: missing header

gapra-msft: will update in the next PR

PR #3257, opened 2025-10-14
Change run_scenarios.yml to use azcli federated credential instead of MSI

dphulkar-msft:
## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/azcopy/pathUtils.go
+++ b/azcopy/pathUtils.go
@@ -1,0 +1,51 @@
+package azcopy
+
+import (
+	"fmt"
+	"net/url"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/sas"
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+)
+
+func GetContainerName(path string, location common.Location) (string, error) {
+	switch location {
+	case common.ELocation.Local():
+		panic("attempted to get container name on local location")
+	case common.ELocation.Blob(),
+		common.ELocation.File(),
+		common.ELocation.FileNFS(),
+		common.ELocation.BlobFS():
+		bURLParts, err := sas.ParseURL(path)
+		if err != nil {
+			return "", err
+		}
+		return bURLParts.ContainerName, nil
+	case common.ELocation.S3():
+		baseURL, err := url.Parse(path)
+
+		if err != nil {
+			return "", err
+		}
+
+		s3URLParts, err := common.NewS3URLParts(*baseURL)
+
+		if err != nil {
+			return "", err
+		}
+
+		return s3URLParts.BucketName, nil
+	case common.ELocation.GCP():
+		baseURL, err := url.Parse(path)
+		if err != nil {
+			return "", err
+		}
+		gcpURLParts, err := common.NewGCPURLParts(*baseURL)
+		if err != nil {
+			return "", err
+		}
+		return gcpURLParts.BucketName, nil
+	default:
+		return "", fmt.Errorf("cannot get container name on location type %s", location.String())
+	}
+}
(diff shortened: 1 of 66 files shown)


azurePipelineTemplates/run_scenarios.yml
```diff
@@ -19,88 +19,81 @@ parameters:
...
+        ########################################################
+        echo "========== Scenario 1: BlobToPipe =========="
+
+        echo "Cleaning up Scenario 1..."
+        AZCOPY_AUTO_LOGIN_TYPE=AzCLI "./${{ parameters.build_name }}" rm "https://${{ parameters.storage_account_name }}.blob.core.windows.net/${{ parameters.container_name }}/myfile.txt"
+
+        echo "Creating a test file and uploading it..."
         echo "Hello, Azure!" > myfile.txt
-        & "./$build_name" "cp" "myfile.txt" "https://${{ parameters.storage_account_name }}.blob.core.windows.net/${{ parameters.container_name }}"
+        AZCOPY_AUTO_LOGIN_TYPE=AzCLI "./${{ parameters.build_name }}" cp myfile.txt "https://${{ parameters.storage_account_name }}.blob.core.windows.net/${{ parameters.container_name }}"
```
wonwuakpa-msft, line 50: Would it be better set the environment auto_login type to AzCLI instead of doing this for each command?

adreed-msft (approved):

azurePipelineTemplates/run_scenarios.yml
```diff
@@ -19,88 +19,81 @@ parameters:
...
+          exit 1
+        fi
+        echo "Executable built successfully: ${{ parameters.build_name }}"
+
+        ########################################################
+        # Scenario 1: Blob → Pipe
+        ########################################################
+        echo "========== Scenario 1: BlobToPipe =========="
+
+        echo "Cleaning up Scenario 1..."
```
wonwuakpa-msft, line 45: I think the clean-up for both the file and the container in the account should happen after the copy operation

azurePipelineTemplates/run_scenarios.yml
```diff
@@ -19,88 +19,81 @@ parameters:
...
+        fi
 
-        ######################################################## Scenario 2 ########################################################
+        ########################################################
+        # Scenario 2: Object Replication Policy Copy
+        ########################################################
+        echo "========== Scenario 2: ORP Source → Destination =========="
 
-        # Test copying from Object Replication Policy source container to destination container
-        # Assign an object replication policy to a storage account.
```
wonwuakpa-msft, line 83: Since it's doing the same thing, why can't we leave the comments in for more context. More detail could be needed

azurePipelineTemplates/run_scenarios.yml
adreed-msft: It seems to me like it's trying to guarantee that the account will be clean and ready for the test... I'm admittedly not a _huge_ fan of using a static account and have debated for a while converting these tests to be ran in some testing framework.

azurePipelineTemplates/run_scenarios.yml
adreed-msft: Probably. An export earlier in the file would make any future changes less of a hassle.

azurePipelineTemplates/run_scenarios.yml
gapra-msft: Agreed on this point - we should eventually move this to the test framework if possible (same with the tests in Test On Ubuntu step). I wouldnt block this PR on these comments at the moment though

azurePipelineTemplates/run_scenarios.yml
gapra-msft: not too worried about this since we should eventually add this to the test framework

PR #3259, opened 2025-10-14
Users/otendolkar/blobmirrorfix

otendolkar:
This commit has the following 2 changes:

1. In GetSyncEnumeratorWithDestComparator, newFpoAwareProcessor sets the destCleanerFunc to nil for a non folder aware source destination pair like S3->Blob or blob->blob, But in case of SyncOrchestrator, we have recursive deletion for a prefix, which  requires a valid deletion processor to be passed to the comparator.

2. In blob traverser, there is a check for if our root is a hns container. However, we handle that case in syncOrch, so we don't want to do it twice. For that reason I have added a check to make sure we are not doing this for mover scenarios.


cmd/syncEnumerator.go
```diff
@@ -526,10 +525,12 @@ func GetSyncEnumeratorWithDestComparator(
 	}
 	destCleanerFunc := newFpoAwareProcessor(fpo, destinationCleaner.removeImmediately)
 
-	if UseSyncOrchestrator && cca.fromTo == common.EFromTo.S3Blob() {
+	if UseSyncOrchestrator && (cca.fromTo == common.EFromTo.S3Blob() || cca.fromTo == common.EFromTo.BlobBlob() || cca.fromTo == common.EFromTo.BlobFSBlob()) {
```
kylemoore-microsoft, line 528: Should this also include BlobFSBlob and BlobBlobFS?

otendolkar: No, it should only include pairs that have flat destinations, BlobFS as a destination will work as is.

kylemoore-microsoft: Oh I see. I misread. Thanks!

pbanakar-microsoft (approved):
Looks good.

PR #3275, opened 2025-11-05
Enable OAuth-only/Acct Key Lookup on the testing framework for static resources

adreed-msft:
## Description

- **Feature / Bug Fix**: Enables OAuth-only setup of the E2E testing framework (no account key copying required)

## Type of Change

- [ ] Bug fix
- [x] New feature (Testing Framework)
- [ ] Other (describe):

## How Has This Been Tested?

Manually. This feature is only enabled for *static* resources on the testing framework, so it shouldn't negatively affect the core product, or automated runs of the test suite.

Some test breaks should be expected for OAuth-only/fallback mode, since some tests *were designed with the use of public access and/or account keys in mind*. Lookup mode is available for that.


e2etest/newe2e_account_registry.go
```diff
@@ -50,10 +50,8 @@ func CreateAccount(a Asserter, accountType AccountType, options *CreateAccountOp
 
 	uuidSegments := strings.Split(uuid.NewString(), "-")
 
-	accountARMClient := &ARMStorageAccount{
-		ARMResourceGroup: CommonARMResourceGroup,
-		AccountName:      DerefOrDefault(opts.CustomName, "azcopynewe2e") + uuidSegments[len(uuidSegments)-1],
-	}
+	accountARMClient := CommonARMResourceGroup.
+		NewStorageAccountARMClient(DerefOrDefault(opts.CustomName, "azcopynewe2e") + uuidSegments[len(uuidSegments)-1])
```
wonwuakpa-msft, line 54: nitpic: Are we reducing flexibility by using a factory method to create the ARM client instead of using the struct?

wonwuakpa-msft (approved):

e2etest/newe2e_account_registry.go
adreed-msft: "Yes". But not in a way I really care enough to do anything about right this second. I'll add this to the backlog for the e2e framework.

gapra-msft (approved):

Issue #3273, opened 2025-10-30
AzCopy v10.31.0 segmentation violation copying from S3 to ADLS


infuerno:
### Which version of the AzCopy was used? 
10.31.0

### Which platform are you using? (ex: Windows, Mac, Linux)
macOS ARM M3 (same error "2-panic" on Linux)

### What command did you run?
azcopy cp https://s3.eu-west-2.amazonaws.com/s3-bucket/s3-directory/* https://adls-storage-account.blob.core.windows.net/adls-container/adls-folder?sas-key

### What problem was encountered?
Command errored with message:
`panic: runtime error: invalid memory address or nil pointer dereference`

Full stack trace in console:
```
INFO: Scanning...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1059afe0c]

goroutine 1 [running]:
github.com/Azure/azure-storage-azcopy/v10/common.(*ServiceClient).FileServiceClient(...)
	github.com/Azure/azure-storage-azcopy/v10/common/util.go:279
github.com/Azure/azure-storage-azcopy/v10/cmd.getShareProtocolType({0x1061ca200, 0x1400031ede0}, 0x0, {{0x140002586e0, 0x3f}, {0x0, 0x0}, {0x0, 0x0}}, 0x0)
	github.com/Azure/azure-storage-azcopy/v10/cmd/flagsValidation.go:364 +0x11c
github.com/Azure/azure-storage-azcopy/v10/cmd.validateShareProtocolCompatibility({0x1061ca200?, 0x1400031ede0?}, {{0x140002586e0, 0x3f}, {0x0, 0x0}, {0x0, 0x0}}, 0x105a3129d?, 0x1, ...)
	github.com/Azure/azure-storage-azcopy/v10/cmd/flagsValidation.go:323 +0x58
github.com/Azure/azure-storage-azcopy/v10/cmd.validateProtocolCompatibility({0x1061ca200, 0x1400031ede0}, 0x603, {{0x140002586e0, 0x3f}, {0x0, 0x0}, {0x0, 0x0}}, {{0x1400014e150, ...}, ...}, ...)
	github.com/Azure/azure-storage-azcopy/v10/cmd/flagsValidation.go:413 +0xd4
github.com/Azure/azure-storage-azcopy/v10/cmd.(*CookedCopyCmdArgs).processCopyJobPartOrders(0x14000207908)
[...]

PR #3276, opened 2025-11-05
Fixed panic for s3 and gcp transfers

gapra-msft:
## Description

Resolves https://github.com/Azure/azure-storage-azcopy/issues/3273

With the NFS over REST changes, we were checking share protocol compatibility when we didnt have to. This is changing the behavior of validateProtocolCompatibility to just validate File/FileNFS locations.

- **Related Links**:
- [Issues](https://github.com/Azure/azure-storage-azcopy/issues/3273)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Added unit tests for the fix

### DIFF
--- a/cmd/flagsValidation.go
+++ b/cmd/flagsValidation.go
@@ -26,6 +26,7 @@ import (
 	"runtime"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azfile/file"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azfile/share"
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


common/util.go
```diff
@@ -223,7 +223,7 @@ func GetServiceClientForLocation(loc Location,
 		return ret, nil
 
 	default:
-		return nil, nil
+		return ret, nil
```
gapra-msft, line 226: this change prevents us from panicking later due to empty clients

cmd/flagsValidation.go
```diff
@@ -383,49 +390,16 @@ func getShareProtocolType(
...
-		case common.EFromTo.FileNFSFileSMB():
-			srcProtocol, dstProtocol = common.ELocation.FileNFS(), common.ELocation.File()
-		case common.EFromTo.FileSMBFileNFS():
-			srcProtocol, dstProtocol = common.ELocation.File(), common.ELocation.FileNFS()
-		}
-
-		// Validate both source and destination
-		if err := validateShareProtocolCompatibility(ctx, src, srcClient, true, srcProtocol, fromTo); err != nil {
+	if fromTo.To().IsFile() {
+		if err := validateShareProtocolCompatibility(ctx, src, dstClient, false, fromTo.To(), fromTo); err != nil {
```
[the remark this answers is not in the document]
wonwuakpa-msft: Small typo, we should use dst here

cmd/flagsValidation.go
```diff
@@ -320,7 +320,10 @@ func validateShareProtocolCompatibility(
 ) error {
 
 	// We can ignore the error if we fail to get the share properties.
-	shareProtocol, _ := getShareProtocolType(ctx, serviceClient, resource, protocol)
+	shareProtocol, err := getShareProtocolType(ctx, serviceClient, resource, protocol)
+	if err != nil {
```
dphulkar-msft: The user might not have granted share-level access, in which case this call could fail. However, by checking for this error, we are forcing the user to provide share-level access, which should not be required.

dphulkar-msft: Instead can we just call validateShareProtocolCompatibility func if from or to is NFS File Share

gapra-msft: addressed this comment - please take a second pass

PR #3278, opened 2025-11-05
Allow disabling cert validation entirely

adreed-msft:
## Description

**Feature / Bug Fix**:

Posed as an alternative to #3277, in this case we _disable certificate validation entirely when requested_. This technically comes with the additional benefit of avoiding any fly-swatting, at the detriment of potentially letting danger in the door.

**Related Links**:

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Other (describe):

## How Has This Been Tested?

manually tested against mitm proxy

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -21,6 +21,7 @@
 package cmd
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -45,7 +46,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var allowInsecureCerts bool
 var outputFormatRaw string
 var outputVerbosityRaw string
 var logVerbosityRaw string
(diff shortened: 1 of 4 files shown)


cmd/constants.go
```diff
@@ -38,6 +38,9 @@ const (
 	PreserveInfoFlag           = "preserve-info"
 	IsNFSProtocolFlag          = "nfs"
 	HardlinksFlag              = "hardlinks"
+
+	// root command flags
+	AllowInsecureCertificatesFlag = "allow-insecure-certs"
```
gapra-msft, line 43: design question - would this be better as an environment variable over a CLI param?

gapra-msft:
If this is trying to fix the same thing that the other PR is, I'd almost prefer just to have the customer use the Go environment variable.

PR #3294, opened 2025-11-17
Allow copies with leading slash or no name virtual dirs

wonwuakpa-msft:
## Description

Handles copies from un-named virtual directories or blob with leading slashes  e.g `//blob`

- **Related Links**:
- [PBI Item](https://msazure.visualstudio.com/One/_workitems/edit/34624337/)
- [Team thread](https://teams.microsoft.com/l/message/19:709bde86ea99454ea08bdf20072fc05f@thread.v2/1757628446852?context=%7B%22contextType%22%3A%22chat%22%7D)

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/azcopy/copy.go
+++ b/azcopy/copy.go
@@ -34,6 +34,8 @@ import (
 
 // CopyOptions contains the optional parameters for the Copy operation.
 type CopyOptions struct {
+	Handler CopyHandler
+
 	IncludeBefore               *time.Time
 	IncludeAfter                *time.Time
 	IncludePatterns             []string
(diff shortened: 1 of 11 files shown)


cmd/copyUtil.go
```diff
@@ -85,6 +85,11 @@ func stripTrailingWildcardOnRemoteSource(source string, location common.Location
 		return
 	}
 
+	if strings.HasSuffix(resourceURL.RawPath, "//*") {
```
adreed-msft: Good catch on the issue itself, but I think we can actually solve this fundamentally rather than just flagging it.

I have a solution in mind (backed by my own testing), but I'd like to see if you can take a crack at it first. Lemme know if you need any help.

adreed-msft (changes requested):

cmd/copyUtil.go
adreed-msft: To give a hint-- Less is more.

PR #3312, opened 2025-12-03
hdi_isfolder shall always be set in lower case on backend

bhansalivikas:
## Description
[Customer ICM](https://portal.microsofticm.com/imp/v5/incidents/details/21000000792858/summary)
Duplicate Hdi_isfolder keys in metadata of directories on HNS account.

- **Feature / Bug Fix**: 
When azcopy retreives properties from backend, Hdi_isfolder is camel cased by Golang. Later code (in case of account to account copy or sync) reuses the same and we end up adding "Hdi" in camel case while setting the properties on destination. When backend returns back the ListBlob API responce it add "hdi_isfolder" for all directories (all in lower case). This creates a duplicate key in metadata ("hdi" and "Hdi", one set by azcopy during sync and one added by backend during list on dfs account).

- **Related Links**:
[ICM](https://portal.microsofticm.com/imp/v5/incidents/details/21000000792858/summary)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?
This needs manual validation


vibhansa-msft:
az storage blob list --account-name vikasfusename --container-name testcntgo1 --auth-mode login --include m --prefix "abcd"
....
When hdi_isfolder is set by azcopy
    "name": "abcd",
    "objectReplicationDestinationPolicy": null,
    "objectReplicationSourceProperties": [],
    "metadata": {
      "hdi_isfolder": "true"
    },
.....
When Hdi_isfolder is set by azcopy
    "name": "abcd",
    "metadata": {
      "Hdi_isfolder": "true",
      "hdi_isfolder": "true"
    },
......

ste/sender-blobFolders.go
```diff
@@ -187,12 +187,12 @@ func (b *blobFolderSender) EnsureFolderExists() error {
 		}
 	}
 
-	// TODO (gapra): figure out better way to deal with hdi_isfolder metadata key capitalization
```
gapra-msft, line 190: we should apply this to is_symlink too

bhansalivikas: While copying symlinks we are creating the metadata with all lower case explicitly:
```
func (s *blobSymlinkSender) SendSymlink(linkData string) error {
	err := s.getExtraProperties()
	if err != nil {
		return fmt.Errorf("when getting additional folder properties: %w", err)
	}
	s.metadataToApply["is_symlink"] = to.Ptr("true")
```

bhansalivikas: In case of blob to blob copy with preserve-symlink flags here is my output:
```
   GET https://vikasfuseblob.blob.core.windows.net/testcntgo1/abcd%2Fa.lnk?...
   Accept: application/xml
   User-Agent: AzCopy/10.32.0~preview.1 azsdk-go-azblob/v1.6.2 (go1.25.1; linux)
   X-Ms-Client-Request-Id: 9602ecc6-11c6-468a-6ec4-373ccafae6e3
   x-ms-version: 2025-07-05
   RESPONSE Status: 200 OK
   ...
   X-Ms-Lease-Status: unlocked
   X-Ms-Meta-Is_symlink: true
   ...
   PUT https://vikasfusename.blob.core.windows.net/testcntgo1/abcd%2Fa.lnk?...
  ...
   x-ms-blob-type: BlockBlob
   x-ms-meta-is_symlink: true
   x-ms-version: 2025-07-05
  ...
```

gapra-msft (approved):

PR #3330, opened 2026-01-05
Users/pbanakar/s3 chunk reader fix

pbanakar-microsoft:
## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/common/privateNetwork.go
+++ b/common/privateNetwork.go
@@ -171,12 +171,12 @@ func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, err
 							isRetryableErr = IsRetryableS3Error(httpS3Err.S3Error.Code)
 							isRetryableErr = httpS3Err.HTTPStatusError.IsRetryable || isRetryableErr
 							isS3AccessDeniedErr = IsAccessDeniedError(httpS3Err.S3Error.Code)
-							log.Printf("[Counter=%d Retry=%d] FAILED with S3 Error, Error Code:%d Error Message:%s retryable:%v", idx, ipAttempt, errCode, errMsg, isRetryableErr)
+							logHTTPError(errCode, "[Counter=%d Retry=%d] FAILED with S3 Error, Error Code:%d Error Message:%s retryable:%v", idx, ipAttempt, errCode, errMsg, isRetryableErr)
 						} else {
 							errCode = httpS3Err.HTTPStatusError.GetErrorCode()
 							errMsg = httpS3Err.GetErrorMessage()
 							isRetryableErr = httpS3Err.HTTPStatusError.IsRetryable
-							log.Printf("[Counter=%d Retry=%d] FAILED HTTP Error, Error Code:%d Error Message:%s retryable:%v", idx, ipAttempt, errCode, errMsg, isRetryableErr)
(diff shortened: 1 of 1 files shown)


common/retryUtils.go
```diff
@@ -154,15 +154,20 @@ func ReadAndValidateResponseBody[T any](body T, responseBuffer *[]byte, logger I
...
 		}
 
-		n, readErr := io.ReadFull(readCloser, *responseBuffer)
-		if readErr != nil && readErr != io.ErrUnexpectedEOF {
-			return n, fmt.Errorf("failed to read response body: %w", readErr)
+		// Read all available data from the body into the caller's buffer
+		// io.ReadFull reads exactly len(buffer) bytes
+		n, err := io.ReadFull(readCloser, *responseBuffer)
+
+		if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
```
sgv-msft, line 165: Did we encounter any issue without io.EOF condition?

common/credentialFactory.go
```diff
@@ -63,7 +63,7 @@ func (o CredentialOpOptions) panicError(err error) {
 }
 
 // Constants for private network transport
-const PeReCheckCooldownTimeInSecs = 3600
+const PeReCheckCooldownTimeInSecs = 600 // 10 minutes - time to wait before rechecking an unhealthy private endpoint
```
sgv-msft, line 66: 10 mins won't be short?  It has overhead of checking unhealthy PE every 10mins during processing

common/privateNetwork.go
```diff
@@ -311,15 +316,15 @@ func GetGlobalPrivateEndpointIPCount() int {
...
-		//fmt.Println("Private Networking is not enabled")
+// IsPrivateNetworkTransfer returns true when private networking is enabled and the specified
+// source requires traffic to traverse the private tunnel (currently S3 sources).
+func IsPrivateNetworkTransfer(source Location) bool {
+	if !privateNetworkArgs.Enabled {
 		return false
 	}
+
+	// Currently, only S3 sources are supported for private network transfers
+	return source == ELocation.S3()
```
sgv-msft, line 327: I think this is to differentiate private networking with FSX and S3 right? how about having Cloud type(AWS S3, GC) configuration in Private networking for the future.

akchidan (approved):

common/retryUtils.go
pbanakar-microsoft: Not really. Added these as addition safety conditions.

common/credentialFactory.go
pbanakar-microsoft: I think, 10mins is good enough for customers to allow fix the issue and for the connection object to be picked up. 
If customer fixes the connection object instantly within 10mins or even early, we will be waiting for 1hr to pick it up.

It is not really an overhead, we promote connection object to healthy after 10mins, if it is not fixes, it will again be demoted to unhealthy. It just one additional https request every 10mins.

common/privateNetwork.go
pbanakar-microsoft: Yes, going forward to introduce any new cloud or any new file types, we can just update this function, the chunkreader ad getblob/putblob should be taken care automatically.

For GCS, we just have to include ELocation.GCP() condition.

PR #3339, opened 2026-01-14
Fix system or event logging

wonwuakpa-msft:
## Description

Due to some refactoring work in 10.31, GetBlob API was being wrongly called on destination blob resources. This will return 403 errors in the event log when using a container level SAS. 

- **Related Links**:
- [Backlog Item](https://msazure.visualstudio.com/One/_workitems/edit/35821853/)
- [Email Subject] [Unauthorized Error Logged Since v10.31.0]

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Manual testing by viewing the Windows event log before and after fix 

### DIFF
--- a/traverser/zc_traverser_file.go
+++ b/traverser/zc_traverser_file.go
@@ -349,11 +349,15 @@ func (t *fileTraverser) Traverse(preprocessor objectMorpher, processor ObjectPro
 		}
 	}
 
+	filterPfx := FilterSet(filters).GetEnumerationPreFilter(t.recursive)
+
 	// Define how to enumerate its contents
 	// This func must be threadsafe/goroutine safe
 	enumerateOneDir := func(dir parallel.Directory, enqueueDir func(parallel.Directory), enqueueOutput func(parallel.DirectoryEntry, error)) error {
 		currentDirectoryClient := dir.(*directory.Client)
-		pager := currentDirectoryClient.NewListFilesAndDirectoriesPager(nil)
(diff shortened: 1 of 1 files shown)


azcopy/copy.go
```diff
@@ -163,6 +164,7 @@ func (c *Client) Copy(ctx context.Context, src, dest string, opts CopyOptions) (
 	// set up the front end scanning logger
 	common.AzcopyScanningLogger = common.NewJobLogger(jobID, c.GetLogLevel(), common.LogPathFolder, "-scanning")
 	common.AzcopyScanningLogger.OpenLog()
+	common.SetForceLogging()
```
gapra-msft: Do we still need to do this here? Shouldnt the fix in credentialUtil be enough?

common/logger.go
```diff
@@ -300,7 +300,7 @@ func IsForceLoggingDisabled() bool {
 	return disableSyslog
 }
 
-func init() {
+func SetForceLogging() {
```
gapra-msft: I feel like this shouldnt be necessary with the isPublic fix right?

azcopy/copy.go
wonwuakpa-msft: This line addresses something else I noticed and decided to add to this PR since its small
We never calling init() i.e the setter for the disable log env var. So, even if a user wanted to disable sys logs it was still printing to it

gapra-msft: init() is automatically called when the application runs I believe.

PR #3340, opened 2026-01-15
Improving error message 

akchidan:
## Description

1. Update IO timeout error message sent as part of warnings to RP
"dial tcp 10.1.0.21:443: i/o timeout" to "The connection to the destination service timed out.443: i/o timeout".

2. Update error message when all PEs are exhausted to 

"The job failed because access to the Amazon S3 bucket could not be established through any configured private connection. Review the private connection settings and permissions, then try again. "

### DIFF
--- a/common/privateNetwork.go
+++ b/common/privateNetwork.go
@@ -251,11 +251,11 @@ func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, err
 			} else {
 				// No response - parse error from the error object itself
 				errCode = 0
-				errMsg = TranslateErrorMessage(err.Error())
+				errMsg = err.Error()
 				log.Printf("[Counter=%d Retry=%d] Network error with no response, Error Message:%s retryable:%v", idx, ipAttempt, errMsg, isRetryableErr)
 			}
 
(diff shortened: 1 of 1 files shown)


common/privateNetwork.go
```diff
@@ -234,11 +251,11 @@ func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, err
 			} else {
 				// No response - parse error from the error object itself
 				errCode = 0
-				errMsg = err.Error()
+				errMsg = TranslateErrorMessage(err.Error())
 				log.Printf("[Counter=%d Retry=%d] Network error with no response, Error Message:%s retryable:%v", idx, ipAttempt, errMsg, isRetryableErr)
 			}
 
-			lastErrMsg = errMsg
+			lastErrMsg = TranslateErrorMessage(errMsg)
```
pbanakar-microsoft: This can be outside the retry logic, orelse we will be translating the same error message retry number of times. 
Any reason why we don't want to do this in agent code, since that is where we send the error message to RP?

pbanakar-microsoft: or you can do it in MarkUnhealthy() function too

PR #3342, opened 2026-01-16
Bugfix for crashes on cancel operation for sync job

sgv-msft:
## Description
Bugfix for crashes seen during cancel operation of sync job with private networking

Root cause: When RoundRobin returns ErrorMessage: "context canceled", roundrobin gets into loop asynchronously to check the PE health on cancel in the meantime syncOrchestrator exit and error channel is closed. After PE retries, we see the crash while logging into SyncOrchestrator error channel which was closed.  
 
Fix for the issue:
In roundrobin function, return response without getting into retry loop when ErrorMessage is  "context canceled" this should fix the issue. 
Avoid crashes in synOrchestrator's error channel in such scenarios with recover function

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?
Tested the fix by cancelling the sync job and verifying no crashes are seen and job reaches terminal stage. Also retriggering the job to ensure job resumes successfully.

### DIFF
--- a/cmd/syncOrchestrator.go
+++ b/cmd/syncOrchestrator.go
@@ -168,6 +168,17 @@ func IsDestinationNotFoundDuringSync(err error) bool {
 
 func writeSyncErrToChannel(errorChannel chan<- TraverserErrorItemInfo, err SyncOrchErrorInfo) {
 	if errorChannel != nil {
+		// Use defer/recover to handle the case where the channel is closed during cancellation.
+		// This prevents "panic: send on closed channel" when the job is being cancelled.
+		defer func() {
+			if r := recover(); r != nil {
+				// Channel was closed, log the error instead of panicking
+				syncOrchestratorLog(
+					common.LogError,
+					fmt.Sprintf("Error channel closed, could not send error: %v", err.ErrorMessage()))
+			}
+		}()
+
 		select {
 		case errorChannel <- err:
 		default:
--- a/common/privateNetwork.go
+++ b/common/privateNetwork.go
@@ -252,6 +252,12 @@ func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, err
 				// No response - parse error from the error object itself
 				errCode = 0
 				errMsg = err.Error()
+
+				// Incase of cancel operation return response without checking for errors.
+				if errMsg == "context canceled" {
+					log.Printf("Returning for cancel operation for Private Endpoint IP %s", peIP)
+					return resp, fmt.Errorf(errMsg)
+				}
 				log.Printf("[Counter=%d Retry=%d] Network error with no response, Error Message:%s retryable:%v", idx, ipAttempt, errMsg, isRetryableErr)
 			}
 


cmd/syncOrchestrator.go
```diff
@@ -168,6 +168,17 @@ func IsDestinationNotFoundDuringSync(err error) bool {
 
 func writeSyncErrToChannel(errorChannel chan<- TraverserErrorItemInfo, err SyncOrchErrorInfo) {
 	if errorChannel != nil {
+		// Use defer/recover to handle the case where the channel is closed during cancellation.
```
pbanakar-microsoft, line 171: Is this change required?
The default: case already logs if the channel is closed.

pbanakar-microsoft: Also based on the stack trace when the crash happened, I thought the issue was in TreeCrawler.go -> workerloop() -> c.output channel.

Can we please verify if we are making the fix for the right channel?

common/privateNetwork.go
```diff
@@ -252,6 +252,12 @@ func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, err
 				// No response - parse error from the error object itself
 				errCode = 0
 				errMsg = err.Error()
+
+				// Incase of cancel operation return response without checking for errors.
+				if errMsg == "context canceled" {
+					log.Printf("Returning for cancel operation for Private Endpoint IP %s", peIP)
+					return resp, fmt.Errorf(errMsg)
```
pbanakar-microsoft, line 259: Nit: should we return nil?

pbanakar-microsoft: Nit: Here we are catching the error from error response. We would still log -  FAILED using IP %s with Error.

Instead we can also check this way even before we go into network error code path.
if req.Context().Err() == context.Canceled {
			return nil, req.Context().Err()
		}

sgv-msft: I think better not to translate/manipulate the response in the roundtrip, instead appropriate error handling should be done by the caller.

cmd/syncOrchestrator.go
sgv-msft: Changes roundtrip function should fix the issue but recover function changes provides a protection against crash where worker takes longer time to return because of unknown async operation(like PE health check in case of Private health check).   Below backtrace points to the **writeSyncErrToChannel** error channel.

1/16/2026, 12:08:43.576 AM
/go/pkg/mod/github.com/!azure/azure-storage-azcopy/v10@v10.28.1-0.20260114182937-fd5358b703d6/common/parallel/TreeCrawler.go:93 +0x59
1/16/2026, 12:08:43.576 AM
created by github.com/Azure/azure-storage-azcopy/v10/common/parallel.(*crawler).runWorkersToCompletion in goroutine 348
1/16/2026, 12:08:43.576 AM
/go/pkg/mod/github.com/!azure/azure-storage-azcopy/v10@v10.28.1-0.20260114182937-fd5358b703d6/common/parallel/TreeCrawler.go:104 +0x9d
1/16/2026, 12:08:43.576 AM
github.com/Azure/azure-storage-azcopy/v10/common/parallel.(*crawler).workerLoop(0xc0002b3200, {0x1e97b30, 0xc000419220}, 0x0?, 0x28)
1/16/2026, 12:08:43.576 AM
/go/pkg/mod/github.com/!azure/azure-storage-azcopy/v10@v10.28.1-0.20260114182937-fd5358b703d6/common/parallel/TreeCrawler.go:177 +0x302
1/16/2026, 12:08:43.576 AM
github.com/Azure/azure-storage-azcopy/v10/common/parallel.(*crawler).processOneDirectory(0xc0002b3200, {0x1e97b30, 0xc000419220}, 0x28)
1/16/2026, 12:08:43.576 AM
/go/pkg/mod/github.com/!azure/azure-storage-azcopy/v10@v10.28.1-0.20260114182937-fd5358b703d6/cmd/syncOrchestrator.go:722 +0x1017
1/16/2026, 12:08:43.576 AM
github.com/Azure/azure-storage-azcopy/v10/cmd.(*cookedSyncCmdArgs).runSyncOrchestrator.func1({0x19a6940, 0xc001f052f0}, 0xc06b540300, 0xc001406840?)
1/16/2026, 12:08:43.576 AM
/go/pkg/mod/github.com/!azure/azure-storage-azcopy/v10@v10.28.1-0.20260114182937-fd5358b703d6/cmd/syncOrchestrator.go:172 +0x9d
1/16/2026, 12:08:43.576 AM
goroutine 319 [running]:

common/privateNetwork.go
sgv-msft: Discussed with Parma offline. For now, agreed upon merging this code change.

cmd/syncOrchestrator.go
sgv-msft: Discussed with Parma offline.  Brainstorming with Parma on this, there is a possibility of hitting this issue in cancel path or failure paths(403) during enumeration.  We decided to push these changes for now and create a task to test to failure scenarios during enumeration for private networking.

PR #3348, opened 2026-01-22
Hardlink preserve support for Local to FileNFS

dphulkar-msft:
## Description

Please find the below doc for the updated hardlink workflow
https://microsoftapc-my.sharepoint.com/:w:/g/personal/dphulkar_microsoft_com/IQCXgFVfSVHqTrJ_MQm5BiCcAf37OzCQOWhvbmerdxWOiLk?e=Hareld

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## How Has This Been Tested?

E2E Tests

### DIFF
--- a/azcopy/zc_processor.go
+++ b/azcopy/zc_processor.go
@@ -57,7 +57,7 @@ type CopyTransferProcessor struct {
 
 	dryrunMode                bool
 	dryrunJobPartOrderHandler func(request common.CopyJobPartOrderRequest) common.CopyJobPartOrderResponse
-	// Separate tracking for files/folder/symlink and hardlink based on processing mode
+	// Separate tracking for files/folder/symlink vs hardlink based on processing mode
 	dispatcher     syncJobPartDispatcher
 	processingMode common.JobProcessingMode
 }
@@ -137,7 +137,6 @@ func (s *CopyTransferProcessor) scheduleTransfer(srcRelativePath, dstRelativePat
 		return nil // skip this one
 	}
 
-	//added
 	s.dispatchPartIfReady()
 
 	// only append the transfer after we've checked and dispatched a part
@@ -195,15 +194,14 @@ func (s *CopyTransferProcessor) dispatchPart() error {
 }
 
 func (s *CopyTransferProcessor) appendTransfer(copyTransfer common.CopyTransfer) error {
-	// This is a hardlink transfer in NFS mode, add it to the separate hardlink transfers batch
+	// This is a hardlink preserve transfer in NFS mode, add it to the separate hardlink transfers batch
 	if s.processingMode == common.EJobProcessingMode.NFS() &&
 		copyTransfer.EntityType == common.EEntityType.Hardlink() &&
 		copyTransfer.TargetHardlinkFile != "" {
 
 		s.dispatcher.PendingHardlinksTransfers.List = append(s.dispatcher.PendingHardlinksTransfers.List, copyTransfer)
-		// Note: Check on this. Do e need to apped the size for hardlink files. As we will only be creating the hardlink
-		//s.dispatcher.PendingHardlinksTransfers.TotalSizeInBytes += uint64(copyTransfer.SourceSize)
 		s.dispatcher.PendingHardlinksTransfers.HardlinksTransferCount++
+
 	} else {
 
 		s.dispatcher.PendingTransfers.List = append(s.dispatcher.PendingTransfers.List, copyTransfer)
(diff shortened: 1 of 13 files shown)


azcopy/zc_processor.go
```diff
@@ -122,33 +137,91 @@ func (s *CopyTransferProcessor) scheduleTransfer(srcRelativePath, dstRelativePat
...
+		copyTransfer.EntityType == common.EEntityType.Hardlink() &&
+		copyTransfer.TargetHardlinkFile != "" {
+
+		s.dispatcher.PendingHardlinksTransfers.List = append(s.dispatcher.PendingHardlinksTransfers.List, copyTransfer)
+		s.dispatcher.PendingHardlinksTransfers.HardlinksTransferCount++
+
+	} else {
+
+		s.dispatcher.PendingTransfers.List = append(s.dispatcher.PendingTransfers.List, copyTransfer)
+		s.dispatcher.PendingTransfers.TotalSizeInBytes += uint64(copyTransfer.SourceSize)
```
wonwuakpa-msft, line 208: Shouldn't we store the TotalSizeInBytes for the hardlink file too?

azcopy/zc_processor.go
```diff
@@ -158,6 +231,33 @@ var NothingScheduledError = errors.New("no transfers were scheduled because no f
...
+
+		// Either file/folder/symlink or hardlink transfers are pending. Whatever is pending will be the final part.
+		if len(s.dispatcher.PendingTransfers.List) > 0 {
+			s.CopyJobTemplate.Transfers = s.dispatcher.PendingTransfers.Clone()
+		} else if len(s.dispatcher.PendingHardlinksTransfers.List) > 0 {
+			s.CopyJobTemplate.JobPartType = common.EJobPartType.Hardlink()
+			s.CopyJobTemplate.Transfers = s.dispatcher.PendingHardlinksTransfers.Clone()
+		}
+	} else {
+		if len(s.dispatcher.PendingTransfers.List) > 0 {
```
wonwuakpa-msft, line 256: Are we only supporting hard link preservation for NFS processing mode specifically FileNFS?

cmd/copyEnumeratorHelper.go
```diff
@@ -9,60 +9,140 @@ import (
...
+	e.Transfers = common.Transfers{}
+	e.PartNum++
 	return nil
 }
 
 // this function shuffles the transfers before they are dispatched
 // this is done to avoid hitting the same partition continuously in an append only pattern
 // TODO this should probably be removed after the high throughput block blob feature is implemented on the service side
-func shuffleTransfers(transfers []common.CopyTransfer) {
+func (d *CopyJobPartDispatcher) shuffleTransfers(transfers []common.CopyTransfer) {
```
wonwuakpa-msft, line 113: I guess it's effectively the same thing, but why are we now making these functions be method receivers of CopyJobPartDispatcher?

common/inodeStore.go
```diff
@@ -0,0 +1,116 @@
...
+	inodeStoreOnce.Do(func() {
+		inodeStoreInstance, err = NewInodeStore()
+	})
+	return inodeStoreInstance, err
+}
+
+func NewInodeStore() (*InodeStore, error) {
+	f, err := os.OpenFile(fmt.Sprintf("%s/inodeStore-%s.txt", filepath.Join(AzcopyJobPlanFolder), NewJobID()), os.O_CREATE|os.O_RDWR|os.O_APPEND, 0644)
+	if err != nil {
+		return nil, err
```
wonwuakpa-msft, line 52: Should we make the error logging more descriptive?

azcopy/zc_processor.go
dphulkar-msft: For hard-linked files, we do not send the file bytes. Instead, we simply create a hard link to the existing file. The TotalSizeInBytes is calculated only for the first hard link.

azcopy/zc_processor.go
dphulkar-msft: yes since hard link preservation requires special handling, we have introduced a processing mode to control the transfer workflow while maintaining backward compatibility.

We now support two processing modes:

MIXED: Uses the existing transfer workflow that is currently in place.

NFS: When the job involves an NFS transfer with --hardlink=preserve, a different transfer workflow is followed, where hard-linked files are processed after all other transfers are completed.

azcopy/zc_processor.go
```diff
@@ -32,6 +32,11 @@ import (
 	"github.com/Azure/azure-storage-azcopy/v10/common"
 )
 
+type syncJobPartDispatcher struct {
```
adreed-msft: The biggest concern that I have about this is memory usage. This _probably_ works fine but with too many hard links, it might severely slow things down. As a consideration, what if we committed hardlink transfers to the plan files immediately, did not process them until all normal transfers were completed, and simply rescanned the plan files for hardlink transfers at the end?

common/inodeStore.go
dphulkar-msft: incorporated in the next PR

cmd/copyEnumeratorHelper.go
dphulkar-msft: Yeah seems like I missed it. For shuffleTransfers specifically, making it a method on CopyJobPartDispatcher isn’t strictly necessary, since it operates only on the provided transfer slice.

For the other functions, however, we are using method receivers because CopyJobPartDispatcher owns the transfer list, and those functions need access to internal state in order to append or modify transfers based on certain conditions.

Given that distinction, I agree that shuffleTransfers does not need to be a receiver method. I’ll adjust this accordingly and incorporate the change in my next PR.

PR #3350, opened 2026-01-23
Users/akchidan/feature/c2c privatenetwork rebase

akchidan:
## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/cmd/syncOrchestrator.go
+++ b/cmd/syncOrchestrator.go
@@ -170,6 +170,7 @@ func writeSyncErrToChannel(errorChannel chan<- TraverserErrorItemInfo, err SyncO
 	if errorChannel != nil {
 		// Use defer/recover to handle the case where the channel is closed during cancellation.
 		// This prevents "panic: send on closed channel" when the job is being cancelled.
+		// [TO DO] - BugFix - Identify the root cause for cases when the error channel is closed prematurely and fix it.
 		defer func() {
 			if r := recover(); r != nil {
 				// Channel was closed, log the error instead of panicking
--- a/ste/mgr-JobPartMgr.go
+++ b/ste/mgr-JobPartMgr.go
@@ -169,6 +169,7 @@ type jobPartMgr struct {
 	planMMF *JobPartPlanMMF // This Job part plan's MMF
 
 	// Cached plan values to prevent accessing unmapped memory after UnmapPlanFile()
+	// this will not impact other non-c2c scenarios, these variables can still be accessed whether unmapping the plan file or not.
 	// These are set once during initialization and never change
 	cachedJobID        common.JobID
 	cachedPartNum      PartNumber


cmd/syncOrchestrator.go
```diff
@@ -168,6 +168,17 @@ func IsDestinationNotFoundDuringSync(err error) bool {
 
 func writeSyncErrToChannel(errorChannel chan<- TraverserErrorItemInfo, err SyncOrchErrorInfo) {
 	if errorChannel != nil {
```
shnayak-msft, line 170: Did we check which writeSyncErrToChannel call is causing the panic? We can keep this defensive check but ideally we should avoid writing to error channel once its closed.

cmd/zt_test.go
```diff
@@ -25,6 +25,14 @@ import (
 	"context"
```
shnayak-msft, line 25: Did we validate the tests that we have updated here?

ste/mgr-JobPartMgr.go
```diff
@@ -168,6 +168,12 @@ type jobPartMgr struct {
 	// When the part is schedule to run (inprogress), the below fields are used
 	planMMF *JobPartPlanMMF // This Job part plan's MMF
 
+	// Cached plan values to prevent accessing unmapped memory after UnmapPlanFile()
+	// These are set once during initialization and never change
+	cachedJobID        common.JobID
```
shnayak-msft, line 174: These are being unconditionally. Will this have any impact other non-c2c scenarios?

testSuite/cmd/clean.go
```diff
@@ -561,7 +564,7 @@ func cleanS3Account(resourceURL string) {
 			defer close(objectsCh)
 
 			// List all objects from a bucket-name with a matching prefix.
-			for object := range s3Client.ListObjectsV2(bucket.Name, "", true, context.Background().Done()) {
+			for object := range s3Client.ListObjects(ctx, bucket.Name, minio.ListObjectsOptions{Prefix: "", Recursive: true}) {
```
shnayak-msft, line 567: It seems like a downgrade from V2. Is this the updated on from v7 minio?

ste/mgr-JobPartMgr.go
pbanakar-microsoft: No, this will not impact other non-c2c scenarios, these variables can still be accessed whether unmapping the plan file or not.

cmd/syncOrchestrator.go
pbanakar-microsoft: This change is only logging for now. Based on our test, we still think there is a bug here which can occur when multiple traversers of different subpath fail with some error and they try to write to this error channel and the error is already closed by parent thread.
Santosh will create bug and investigate further for final solution.

cmd/zt_test.go
pbanakar-microsoft: No. As per the discussion, we will run azcopy test pipelines during the merge and identify if any issues.

testSuite/cmd/clean.go
pbanakar-microsoft: listObjects in v7 internally uses ListObjectsV2.

cmd/syncOrchestrator.go
sgv-msft: As discussed in the meeting will create two bugs one on C2C and other for OnPrem.  As Parma mentioned will investigate on the final solution.

PR #3353, opened 2026-01-26
Adding support for GCP S3 with HMAC

salamianar-microsoft:
## Description
Adding support for GCS S3 with HMAC to Azure Blob migration

**Related Links**
- [Issues](https://msazure.visualstudio.com/One/_workitems/edit/36470920/)
- [Documents](https://microsoft.sharepoint.com/:w:/t/AzureStorage/IQCbl57kVfKpR7FvZ0iZB_8vASD9ihWUdFVW_2E0yqRCZKs?e=oCLKZf&wdOrigin=TEAMS-MAGLEV.null_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1769469300098&web=1)

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## How Has This Been Tested?
Manual testing for copy and sync flow

### DIFF
--- a/common/privateNetwork.go
+++ b/common/privateNetwork.go
@@ -187,7 +187,7 @@ func (rr *RoundRobinTransport) RoundTrip(req *http.Request) (*http.Response, err
 			var isRetryableErr bool
 			var isS3AccessDeniedErr bool
 
-			log.Printf("[Counter=%d Retry=%d] Sending request to PrivateEndpoint IP: %s (Host header: %s)", idx, ipAttempt, clonedReq.URL.Host, clonedReq.Host)
+			// log.Printf("[Counter=%d Retry=%d] Sending request to PrivateEndpoint IP: %s (Host header: %s)", idx, ipAttempt, clonedReq.URL.Host, clonedReq.Host)
 
 			resp, err := rr.transport.RoundTrip(clonedReq)
 			if err == nil {


common/s3URLParts.go
```diff
@@ -64,6 +66,27 @@ const s3EssentialHostPart = "amazonaws.com"
 
 var s3HostRegex = regexp.MustCompile(s3HostPattern)
 
+func GetS3CompatibleSuffix() string {
+	if os.Getenv("S3_COMPATIBLE_ENDPOINT") != "" {
```
pbanakar-microsoft, line 69: This need not be env variable. we could just pass in the endpoint from worker to azcopy code, just by setting some inMemory variable

ste/sourceInfoProvider-S3.go
```diff
@@ -66,6 +67,13 @@ func newS3SourceInfoProvider(jptm IJobPartTransferMgr) (ISourceInfoProvider, err
 		return nil, err
 	}
 
+	// For S3-compatible endpoints (e.g. GCS path-style), the ObjectKey may have a
+	// leading "/" after URL parsing which causes a double-slash in minio's path-style
+	// request URL (bucket//objectKey), resulting in 403 errors.
+	if p.s3URLPart.IsS3CompatibleEndpoint() && strings.HasPrefix(p.s3URLPart.ObjectKey, "/") {
+		p.s3URLPart.ObjectKey = strings.TrimLeft(p.s3URLPart.ObjectKey, "/")
```
pbanakar-microsoft: Please test this with Sync job too. To make sure there is no impact.

pbanakar-microsoft: Also, why was this not an issue so far with AWS S3?

common/s3URLParts.go
```diff
@@ -72,12 +94,141 @@ func IsS3URL(u url.URL) bool {
...
+//	[ fullHost, bucketCapture(with trailing '.' if present OR ""), regionOrDualStack, keywordDomainRoot ]
+//
+// For path-style compatible providers where bucket is not in host, bucketCapture is "" so caller treats it as path-style.
 func findS3URLMatches(host string) (matches []string, isS3Host bool) {
-	matchSlices := s3HostRegex.FindStringSubmatch(host) // If match the first element would be entire host, and then follows the sub match strings.
-	if matchSlices == nil || !strings.Contains(host, s3EssentialHostPart) {
-		return nil, false
+	suffix := GetS3CompatibleSuffix()
+	hostLower := strings.ToLower(host)
+
```
pbanakar-microsoft, line 107: Nit: Looks like this is redundant, since the callers are already doing ToLower() and passing the parameter.

common/privateNetwork.go
```diff
@@ -107,12 +107,24 @@ func SetPrivateNetworkArgs(privateNetworkEnabled bool, privateEndpointIPs []stri
...
 	tr := http.DefaultTransport.(*http.Transport).Clone()
-	tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: false, ServerName: host}
+	tlsCfg := &tls.Config{
+		InsecureSkipVerify: false,
+		ServerName:         tlsHost,
+	}
+	if forceHTTP11 {
+		// GCS regional endpoints send proprietary HTTP/2 frames that Go doesn't recognize.
+		// Force HTTP/1.1 to avoid noisy "unhandled response frame type" warnings.
+		tlsCfg.NextProtos = []string{"http/1.1"}
```
pbanakar-microsoft, line 124: Please check from security point of view also, is this ok? or is it always recommended to use http/2

cmd/syncOrchestrator.go
```diff
@@ -307,6 +311,18 @@ func buildChildPath(baseDir, relativePath string) string {
...
+	syncOrchestratorLog(common.LogDebug, fmt.Sprintf("[PROCESSOR] Before buildChildPath - dir='%s', originalPath='%s'", st.dir, originalPath))
+
+	// Skip directory placeholder objects that represent the current directory itself
+	// These have empty relativePath after prefix trimming and would cause re-enqueueing of the same directory
+	// This issue specifically occurs with GCP S3-compatible storage where directory placeholders are returned
+	if originalPath == "" && so.entityType == common.EEntityType.Folder() && isGCPSource {
+		syncOrchestratorLog(common.LogDebug, fmt.Sprintf("[PROCESSOR] Skipping self-referential directory placeholder for dir='%s' (GCP S3-compatible source)", st.dir))
+		return nil
+	}
+
```
pbanakar-microsoft, line 330: is this fix for <no name> director? For example path - dir1/dir2//dir3/file.txt
Is this the same bugFix which Kyle is working on? if yes, do you need this change or Kyle's changes?

ste/xfer-anyToRemote-file.go
```diff
@@ -577,14 +613,39 @@ func epilogueWithCleanupSendToRemote(jptm IJobPartTransferMgr, s sender, sip ISo
 			lmt, err := sip.GetFreshFileLastModifiedTime()
 			if err != nil {
 				jptm.FailActiveSend("epilogueWithCleanupSendToRemote", err)
+				return
+			}
+
+			scheduledTime := jptm.LastModifiedTime()
+
```
pbanakar-microsoft: same comment as above, move it to a function and have this logic only for mover.build tag, so that it doesn't apply to azcopy users.

pbanakar-microsoft (changes requested):
Please review the comments.

cmd/syncOrchestrator.go
salamianar-microsoft: @pbanakar-microsoft , this fix is for self-referential directory placeholder.

1. When syncing from GCP via the S3 API, GCP returns directory placeholder objects (zero-byte objects with a trailing /, e.g., dir1/).
2. During recursive sync, when the traverser enters dir1/, it lists objects with dir1/ as the prefix. GCP returns the placeholder object dir1/ itself in the listing.
3. After prefix trimming, this placeholder's relativePath becomes "" (empty string).
4. Without this check, the empty-path folder object gets stored in the indexer and re-enqueued as a subdirectory — causing the orchestrator to re-process the same directory, leading to infinite loops or duplicate processing.

I believe the fix Kyle working on it is different from this one. He is fixing if there is object with a//b.txt kind of virtual object, sync is doing infinite loop. I couldn't create that kind of object on GCS but will do additional check once Kyle's fix merged in. 

As a summary: for GCS we would need both: this fix and Kyle's fix

common/privateNetwork.go
salamianar-microsoft: @pbanakar-microsoft  This is safe from a security perspective. HTTP/1.1 and HTTP/2 both run over the same TLS layer — HTTP/2 provides performance benefits (multiplexing, header compression) but no additional security. TLS certificate validation remains enforced "InsecureSkipVerify: false". This fallback only targets GCS regional endpoints that send non-standard HTTP/2 frames incompatible with Go's HTTP/2 implementation. The global GCS endpoint and all AWS S3 endpoints continue using HTTP/2.

common/s3URLParts.go
salamianar-microsoft: @pbanakar-microsoft, Thank you for your feedback. This will be fix as part of the move to use in memory variables for other env variables we are using. (AwsAccessKeyId, AwsSecretAccessKey etc). I will add comment on top of this implementation as TODO

common/s3URLParts.go
salamianar-microsoft: @pbanakar-microsoft, Thanks for the catch. I would prefer to using toLower() inside of this function to not rely on caller's that they will always send the host lower cases. For now, since all callers of this function passing host using toLower(), i cleaned up and added comment on top of it that we are expecting lower case host, in the future if someone create new caller function of this.

ste/sourceInfoProvider-S3.go
salamianar-microsoft: I found this issue as part of sync testing and already verified using both copy and sync. It is not an issue so far for AWS S3 is because of we are not getting sourceURI from user directly. I have seen this for AWS as well when I did testing for AWS S3 With HMAC, where I pass sourceUri. That is the reason I am using IsS3CompatibleEndpoint() check not GCS only.

cmd/syncOrchestrator.go
```diff
@@ -312,6 +316,18 @@ func buildChildPath(baseDir, relativePath string, isDirectory bool) string {
...
 // and stores them in the indexer for later comparison and transfer.
 func (st *SyncTraverser) processor(so StoredObject) error {
+	// Debug: Log original path before transformation
+	originalPath := so.relativePath
+	syncOrchestratorLog(common.LogDebug, fmt.Sprintf("[PROCESSOR] Before buildChildPath - dir='%s', originalPath='%s'", st.dir, originalPath))
+
+	// Skip directory placeholder objects that represent the current directory itself
+	// These have empty relativePath after prefix trimming and would cause re-enqueueing of the same directory
+	// This issue specifically occurs with GCP S3-compatible storage where directory placeholders are returned
+	if originalPath == "" && so.entityType == common.EEntityType.Folder() && isGCPSource {
```
pbanakar-microsoft: Nit: This should have been the first condition in the sequence, so that this entire condition fails fast for all non-gcp sources.

common/sysinfo_linux.go
```diff
@@ -71,8 +72,71 @@ func GetMemAvailable() (int64, error) {
...
+	}
+
+	log.Printf("[getCgroupMemoryLimit] No cgroup memory limit found, returning 0")
+	return 0
+}
+
+// getTotalPhysicalMemoryInternal performs the actual memory reading operation.
+// It checks cgroup memory limits first (for container environments), then falls back
+// to /proc/meminfo. Returns the minimum of the two if both are available.
 func getTotalPhysicalMemoryInternal() (int64, error) {
```
pbanakar-microsoft, line 117: is this giving you required info, it was actually giving the node VM memory capacity?
Can you also confirm, if this is executed only once?

salamianar-microsoft: Hi @pbanakar-microsoft , yes it is giving back the cgroup (container) memory and executed just once. Based on logs it shows just one time execution and also added logs to rolling-stat files as well

salamianar-microsoft: Based on log: 

 [getTotalPhysicalMemoryInternal] Host memory from /proc/meminfo: 67419009024 bytes (62.79 GB)
 [getCgroupMemoryLimit] cgroup v2 /sys/fs/cgroup/memory.max = "32212254720"
[getCgroupMemoryLimit] Using cgroup v2 limit: 32212254720 bytes (30.00 GB)
 [getTotalPhysicalMemoryInternal] Cgroup memory limit: 32212254720 bytes (30.00 GB)
[getTotalPhysicalMemoryInternal] Using CGROUP limit (container memory): 32212254720 bytes (30.00 GB)

PR #3368, opened 2026-02-09
Skip Archive/Glacier objects during enumeration for S3

pbanakar-microsoft:
## Description
Skipping objects with Glacier or Deep archive storage class on AWS S3. 

- **Feature / Bug Fix**: Skipping objects with Glacier or Deep archive storage class on AWS S3. 

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?
Manual test was performed to test different storage class and stats were validated.

### DIFF
--- a/cmd/moverDependencies.go
+++ b/cmd/moverDependencies.go
@@ -922,6 +922,18 @@ func SetJobId(jobId common.JobID) {
 	azcopyCurrentJobID = jobId
 }
 
+func OpenScanningLogger() {
+	// set up the front end scanning logger
+	azcopyScanningLogger = common.NewJobLogger(azcopyCurrentJobID, LogLevel, azcopyLogPathFolder, "-scanning")
+	azcopyScanningLogger.OpenLog()
+}
+
+func CloseScanningLogger() {
+	if azcopyScanningLogger != nil {
+		azcopyScanningLogger.CloseLog()
+	}
+}
+
 // ============================================================================
 // End Utility Functions - Misellaneous
 // ============================================================================


cmd/copyEnumeratorInit.go
```diff
@@ -104,6 +104,11 @@ func (cca *CookedCopyCmdArgs) initEnumerator(jobPartOrder common.CopyJobPartOrde
 				} else if entityType == common.EEntityType.Symlink() {
 					atomic.AddUint32(&cca.atomicSkippedSymlinkCount, 1)
 				}
+			} else if cca.FromTo.From() == common.ELocation.S3() {
+				// Track skipped S3 objects (e.g., Archive/Glacier storage class objects)
+				if entityType == common.EEntityType.Other() {
```
kylemoore-microsoft, line 109: For my understanding, is the pattern to map skipped files to EEntityType.Other()

pbanakar-microsoft: Possibly in future when we want to use EntityType.other() for any other counter increments.
We could add corresponding entitytypes when we see the need for other counter increments in traverser or STE.

cmd/copy.go
```diff
@@ -1466,6 +1468,7 @@ Number of Folder Transfers Skipped: %v
 Number of Symbolic Links Skipped: %v
 Number of Hardlinks Converted: %v
 Number of Special Files Skipped: %v
+Number of Archive/Glacier Objects Skipped: %v
```
shnayak-msft, line 1471: You can add (AWS S3) to provide context here

pbanakar-microsoft: we might use this same counter for Azure archive blobs too in future, hence mentioned Archive/Glacier.

cmd/zc_traverser_s3.go
```diff
@@ -238,6 +238,22 @@ func (t *s3Traverser) Traverse(preprocessor objectMorpher, processor objectProce
...
+		if isArchiveStorageClass(objectInfo.StorageClass) {
+			skipMessage := fmt.Sprintf("Skipping S3 object %s as it is in %s storage class. "+
+				"Objects in archive storage classes must be restored before they can be transferred.",
+				objectInfo.Key, objectInfo.StorageClass)
+			WarnStdoutAndScanningLog(skipMessage)
+			// Increment the file scanned counter since we've scanned this object
+			// (directories cannot have storage classes, so this is definitely a file)
+			t.incrementEnumerationCounter(common.EEntityType.File())
+			// Increment the enumeration counter to track this as a skipped object
+			t.incrementEnumerationCounter(common.EEntityType.Other())
```
shnayak-msft, line 251: Don't use the increment counter to update different counters for same file. If you want to update both file and skip counters for an archive object, do it in the increment counter method. 

if s3 and other:
   file++
   skip++

cmd/sync.go
```diff
@@ -520,6 +520,7 @@ type cookedSyncCmdArgs struct {
 	hardlinks                        common.HardlinkHandlingType
 	atomicSkippedSymlinkCount        uint32
 	atomicSkippedSpecialFileCount    uint32
+	atomicSkippedArchiveFileCount    uint64
```
shnayak-msft, line 523: same comment as before. Add AWS S3 for clarity as we have archive blob type as well. User can infer it from source type but this will keep it clear.

pbanakar-microsoft: The same counter can be used for archive blob type as well right?
do you mean to say, we should have 2 counters - 
atomicSkippedArchiveBlobFileCount for Azure Blob
atomicSkippedArchiveS3FileCount for AWS S3

shnayak-msft: No one is fine

cmd/zc_traverser_s3.go
pbanakar-microsoft: Fixed.

PR #3369, opened 2026-02-09
Hardlink support for NFS->Local and NFS->NFS

dphulkar-msft:
## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## How Has This Been Tested?


ste/downloader-azureFiles.go
```diff
@@ -220,3 +223,33 @@ func (bd *azureFilesDownloader) preserveProperties(info *TransferInfo) (string,
...
+	rt := filepath.ToSlash(root)
+
+	// 2. Use "/" for splitting now that we've normalized
+	relParts := strings.Split(rel, "/")
+	fullPathParts := strings.Split(rt, "/")
+
+	prefixParts := []string{}
+	var j int
+	for j = 0; j < len(fullPathParts)-1; j++ {
+		if relParts[0] != fullPathParts[j] {
```
gapra-msft, line 239: What if there are multiple matching directory names in the path?

ste/downloader-blob.go
```diff
@@ -202,6 +202,10 @@ func (bd *blobDownloader) GenerateDownloadFunc(jptm IJobPartTransferMgr, destWri
 	})
 }
 
+func (a *blobDownloader) CreateHardLink(jptm IJobPartTransferMgr) error {
```
gapra-msft, line 205: Why do we need to define CreateHardlink? Shouldn;t blob and blobFS downloaders just be downloaders and not hardlinkDownloaders?

dphulkar-msft: yeah we can remove this.
Updated the changes in the next upcoming PR

ste/downloader-azureFiles.go
dphulkar-msft: I have updated the target hardlink file path derivation logic in my next PR.

Issue #3371, opened 2026-02-10
Unable to cancel long-running jobs from Storage Explorer with 10.32.0


craxal:
### Which version of the AzCopy was used? 
10.32.0

### Which platform are you using? (ex: Windows, Mac, Linux)
All

### What command did you run?
export AZCOPY_CRED_TYPE=Anonymous;
export AZCOPY_CONCURRENCY_VALUE=AUTO;
./azcopy copy "https://ACCOUNT.blob.core.windows.net/blob-container-01/*?sv=2025-07-05&st=2026-02-10T17%3A19%3A50Z&se=2026-02-17T17%3A34%3A50Z&skoid=830a51ab-cfba-41c3-a07b-f69350bd0240&sktid=72f988bf-86f1-41af-91ab-2d7cd011db47&skt=2026-02-10T17%3A19%3A50Z&ske=2026-02-17T17%3A34%3A50Z&sks=b&skv=2025-07-05&sr=c&sp=rl&sig=REDACTED" "/Users/user/Downloads" --overwrite=prompt --check-md5 FailIfDifferent --from-to=BlobLocal --list-of-files "/var/folders/nn/64vf54q50ld7mxx20s91kv240000gn/T/stg-exp-azcopy-805c18f7-912c-426c-a336-276346d7f483.txt" --recursive --log-level=INFO;
unset AZCOPY_CRED_TYPE;
unset AZCOPY_CONCURRENCY_VALUE;

### What problem was encountered?
Unable to cancel the transfer from Storage Explorer.

### How can we reproduce the problem in the simplest way?
1. Launch Storage Explorer build 20260210.5.
2. Open a blob container with a large or many files.
3. Start a long-running download.

### Have you found a mitigation/solution?
No

wonwuakpa-msft:
https://github.com/Azure/azure-storage-azcopy/pull/3373

wonwuakpa-msft:
We included a fix for this [in our 10.32.2 release](https://github.com/Azure/azure-storage-azcopy/releases/tag/v10.32.2)

craxal:
Integrated and tested.

PR #3373, opened 2026-02-12
Fix regression with cancellations in StgExplorer

wonwuakpa-msft:
## Description

The cancellation workflow was changed as part of our refactoring efforts in v10.32.0 to support AzCopy as a library. However, in the process using cancel from stdin was broken. A new context-based cancellation logic was introduced that disconnected cancels from stdin. Issue is reproducible here by StgExp

#### Root Cause: 

- `watchInputs()` sends `os.Interrupt` to the `lcm.cancelChannel` this functionality has been present even before 10.32
- However, the signal handler in the go routing in `cmd/copy.go` only receives real OS-level signals. So, it will not be able to read from the that channel. 
- In this new cancellation flow, nothing reads from `lcm.cancelChannel` anymore. We previously used to listen on it with `lifecycleMgr.InitiateProgressReporting()` and call `jc.Cancel()`. 
- But, the new code flow uses `joblifecyclemanager.IntiateProgressReporting(ctx...)` which only listens for if the context is done with `ctx.Done()` The job lcm never knows about the cancellation so the job keeps on running. 

#### Fix:
- Introduced a way of listening to the `lcm.cancelChannel` to where we have the signal handling goroutines. With this, cancelling with the stdin cancel will trigger `cancel()` on the the context. And this can by the caught by the new `joblifecyclemanager.IntiateProgressReporting(ctx...)` that triggers and eventual graceful shut down. 

**Related Links**:
- [Issues](https://github.com/Azure/azure-storage-azcopy/issues/3371)
- 
## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Validated from StgExp team cancelations now work. Added a unit test to context cancelation works in both cancel scenarios. 
31c89348-ab21-4655-937f-febdae8fd554

### DIFF
--- a/azcopy/copy.go
+++ b/azcopy/copy.go
@@ -143,12 +143,16 @@ func (c *Client) Copy(ctx context.Context, src, dest string, opts CopyOptions) (
 		copyHandler = common.NewJobUIHooks()
 		common.SetUIHooks(copyHandler)
 	}
-	jobID := common.NewJobID()
-	c.CurrentJobID = jobID
+	jobID := c.CurrentJobID
+	if jobID.IsEmpty() {
+		jobID = common.NewJobID()
+		c.CurrentJobID = jobID
+	}
 	timeAtPrestart := time.Now()
-	common.AzcopyCurrentJobLogger = common.NewJobLogger(jobID, c.GetLogLevel(), common.LogPathFolder, "")
-	common.AzcopyCurrentJobLogger.OpenLog()
-
+	if common.AzcopyCurrentJobLogger == nil { // In the unlikely case, logger is not initialized in root.go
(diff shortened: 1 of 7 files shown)


azcopy/copy.go
```diff
@@ -206,6 +207,13 @@ func (c *Client) Copy(ctx context.Context, src, dest string, opts CopyOptions) (
 		defer jobsAdmin.JobsAdmin.JobMgrCleanUp(jobID)
 
 		if err != nil {
+			if errors.Is(err, context.Canceled) || errors.Is(ctx.Err(), context.Canceled) {
+				// AzCopy run was canceled with (Ctrl + C).
+				waitErr := mgr.Wait()
+				if err != nil {
```
gapra-msft: Should this be checking waitErr?

PR #3381, opened 2026-02-24
Refactor environment variable implementation

adreed-msft:
## Description

In the course of work on the `cap-mbps` work item, I wanted to deprecate the old `AZCOPY_CONCURRENCY_VALUE` environment variable in favor of a more descriptive `AZCOPY_CONCURRENT_ROUTINES` variable. However, we don't have any real support for doing so. The intent of this PR is a few things.

1) Refactor the enumerator portion of environment variables to not be directly attached to an individual `EnvironmentVariable`.
2) Refactor the environment variables' interface to be called from the `EnvironmentVariable` directly, as opposed to calling a function with the environment variable as a parameter. This makes it less frustrating to work with an environment variable.
3) Support deprecating an environment variable, falling back to it's deprecated alternatives, warning about deprecated variables being set or read (in the latter case, panicking to alert a developer and enforce code quality).
4) Introduce ways to distinguish `""` vs unset
5) Introduce ways to check the name of which deprecated environment variable had been used.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Refactor that shouldn't add any functionality-- existing testing should catch issues.


common/environment.go
```diff
@@ -36,20 +37,75 @@ type EnvironmentVariable struct {
...
 
-// GetEnvironmentVariable gets the environment variable or its default value
-func GetEnvironmentVariable(env EnvironmentVariable) string {
-	value := os.Getenv(env.Name)
-	if value == "" {
-		return env.DefaultValue
+func (env EnvironmentVariable) Lookup(intentionDeprecatedLookup ...bool) (val, name string, present bool) {
+	// If we have deprecated this env variable, and we're looking it up without intending to, something's wrong. Let the developer know by panicking.
+	if env.ReplacedBy != "" && !FirstOrZero(intentionDeprecatedLookup) {
+		panic(fmt.Sprintf("Unintentional lookup of deprecated environment variable %s (replaced by %s). If this was intentional, add true to GetEnvironmentVariable parameters.", env.Name, env.ReplacedBy))
```
wonwuakpa-msft: Do we want to panic or just log an error?

wonwuakpa-msft (approved):
It should be good after addressing the copilot comments

common/environment.go
adreed-msft: Panic is my intent here. If we panic when unintentionally calling for a deprecated variable, it will set alarms off. If alarms are set off for a potential mistake, it can be avoided before it reaches the customer.

PR #3383, opened 2026-02-24
Add sync hardlink preserve support for local->NFS

dphulkar-msft:
## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/tool_distributed_mutex.py
+++ b/tool_distributed_mutex.py
@@ -29,7 +29,6 @@ def get_raw_input():
 
 def process():
     action, mutex_url = get_raw_input()
-    print(f"INFO: mutex_url = {mutex_url}")
 
     # check whether the blob exists, if not quit right away to avoid wasting time
     blob_client = BlobClient.from_blob_url(mutex_url)


traverser/zc_enumerator.go
```diff
@@ -295,7 +301,8 @@ func NewStoredObject(morpher objectMorpher, name string,
 		LeaseStatus:        blobProps.LeaseStatus(),
 		LeaseState:         blobProps.LeaseState(),
 		LeaseDuration:      blobProps.LeaseDuration(),
-		TargetHardlinkFile: targetHardlinkFile,
+		TargetHardlinkFile: nfsOptions.TargetHardlinkFile,
+		Inode:              nfsOptions.Inode,
 	}
```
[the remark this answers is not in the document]
dphulkar-msft: fixed

common/inodeStore.go
```diff
@@ -29,88 +29,237 @@ import (
...
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("failed scanning inode store: %w", err)
+	}
+	return index, nil
+}
 
-	// Slow path: exclusive lock
+// writeRecord writes a padded record at the current end of the file.
+// Returns the offset and capacity. Must be called with s.mu write lock held.
+func (s *InodeStore) writeRecord(inode, firstPath, anchor string) (int64, int, error) {
```
adreed-msft: nit: for non obvious tuple returns I tend to do them named.

```suggestion
func (s *InodeStore) writeRecord(inode, firstPath, anchor string) (offset int64, capacity int, error) {
```

common/inodeStore.go
```diff
@@ -29,88 +29,237 @@ import (
...
+}
+
+// overwriteRecord updates the anchor of an existing record in place.
+// If the new content exceeds the record's capacity, relocates to the end of the file.
+// Must be called with s.mu write lock held.
+func (s *InodeStore) overwriteRecord(meta *inodeMeta, inode, firstPath, newAnchor string) error {
+	content := fmt.Sprintf("%s\t%s\t%s", inode, firstPath, newAnchor)
+
+	if len(content)+1 > meta.capacity {
+		// New content doesn't fit — relocate record to end of file (old slot becomes dead space)
```
adreed-msft, line 196: How often do you think we might end up in this situation? Would it be worth having a dedupe operation that'd trigger if it gets bad enough?

common/inodeStore_test.go
```diff
@@ -0,0 +1,896 @@
...
+// newTestInodeStore creates an InodeStore backed by a temp file.
+// It returns the store and a cleanup function. The caller must defer cleanup().
+func newTestInodeStore(t *testing.T) (*InodeStore, func()) {
+	t.Helper()
+	tmpDir := t.TempDir()
+	origFolder := AzcopyJobPlanFolder
+	AzcopyJobPlanFolder = tmpDir
+
+	jobID := NewJobID()
+	store, err := NewInodeStore(jobID)
```
adreed-msft: Another idea. Can we not just pass in something like

```go
type TargetedReadWriteCloser interface {
    io.ReaderAt
    io.WriterAt
    io.Closer
}
```

in as a backing file, and then pass something backed purely in-memory, that way we don't have to worry about cleaning up the filesystem (for the tests that are just testing the INodeStore's functionality?)

azcopy/copy.go
```diff
@@ -259,7 +262,17 @@ func newCopyTransferExecutor(ctx context.Context, jobID common.JobID, src, dst s
 		return nil, err
 	}
 
+	store, err := common.NewInodeStore(jobID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize inode store: %w", err)
+	}
+	// Ensure that resources held by the inode store are eventually released,
+	// even if callers forget to close it explicitly.
+	runtime.SetFinalizer(store, func(s *common.InodeStore) {
```
gapra-msft: Can we just do a defer store.Close? Why are we using runtime.SetFinalizer?

common/inodeStore.go
dphulkar-msft: How often relocation happens depends on how frequently we encounter a new anchor path that exceeds the current record's capacity (i.e., >64 bytes longer than the original). Right now I am not sure how often this will happen.
To keep the inode store simple for now, I'd prefer to defer dead-space management. If we see evidence of bloat in production, we can add a compaction workflow later.

Issue #3398, opened 2026-03-09
[Bug] AMLFS POSIX Properties style is not passed down from CLI flags


wolfgang-desalvador:
### Which version of the AzCopy was used? 
10.32.1

### Which platform are you using? (ex: Windows, Mac, Linux)
Windows and Linux

### What command did you run?
`./azcopy copy LICENSE.txt $BLOB_URL_WITH_SAS  --preserve-posix-properties --posix-properties-style amlfs`

### What problem was encountered?

- POSIX Properties style is not used
- POSIX Properties style accepts any value and is not validated

### How can we reproduce the problem in the simplest way?

- Run a copy with AMLFS POSIX style
- The generated blob has standard Blob POSIX Metadata

### Have you found a mitigation/solution?
Currently in PR

PR #3401, opened 2026-03-10
Proper chaining for AMLFS flag

wonwuakpa-msft:
- **Related Links**:
- [Issues](https://github.com/Azure/azure-storage-azcopy/issues/3398)

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

The initial version of `TestPOSIX_UploadAMLFSStyle` test did not catch this. Why? Because it was only checking the props values after they had been normalized. In our test framework, we always parse the metadata into the standard POSIX style. The test has been extended to include a hook to also validate the props style.

Confirmed fix with AMLFS team.


cmd/copy.go
```diff
@@ -585,6 +589,7 @@ func (raw *rawCopyCmdArgs) toOptions() (cooked CookedCopyCmdArgs, err error) {
 	return cooked, nil
 }
 
+// This method and ones it calls are legacy code not used in copy command flow (as of v10.32)
```
gapra-msft, line 592: Add that some CLI commands still call into this

wonwuakpa-msft: since CI on this PR is done, I'll change it in the 10.32.2 patch release PR

gapra-msft: fine with me

ste/md5Comparer.go
```diff
@@ -42,8 +42,8 @@ var ErrMd5Mismatch = errors.New("the MD5 hash of the data, as we received it, di
 	"This means that either there is a data integrity error OR another tool has failed to keep the stored hash up to date. " +
 	"(NOTE In the specific case of downloading a Page Blob that has been used as a VM disk, the VM has probably changed the content since the hash was set. That's normal, and " +
 	"in that specific case you can simply disable the MD5 check. " +
-	"See the AzCopy options documentation for --check-md5.)" +
-	"Also see Notes section here https://docs.azure.cn/en-us/storage/common/storage-use-azcopy-blobs-download#download-a-blob")
+	"See the AzCopy options documentation for --check-md5.) " +
+	"Also see Notes section here https://learn.microsoft.com/en-us/storage/common/storage-use-azcopy-blobs-download#download-a-blob")
```
dphulkar-msft, line 46: This link gives 404. Can you re-check?

wonwuakpa-msft: Will update link to 'https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-blobs-download#download-a-blob' 

in the next PR (10.32.2 patch PR)

PR #3410, opened 2026-03-16
Fix deadlock in ResumeTransfers, mmap safety in ScheduleTransfers

pbanakar-microsoft:
Bug fixes for resume path in azcopy engine:

- **Deadlock fix**: Changed ResumeTransfers to use READ lock (via buildmode.IsMover) instead of write lock when iterating jobPartMgrs to queue parts. Write lock caused deadlock when partsChannel fills up: ResumeTransfers holds write lock -> reportJobPartDoneHandler needs read lock -> scheduleJobParts blocks on unbuffered jobPartProgress.
- **Mmap safety**: Cached isFinalPart and cachedNumTransfers before the transfer loop in ScheduleTransfers to avoid accessing memory-mapped plan data that may be unmapped during iteration.

### DIFF
--- a/jobsAdmin/JobsAdmin.go
+++ b/jobsAdmin/JobsAdmin.go
@@ -377,7 +377,6 @@ func (ja *jobsAdmin) ResurrectJob(jobId common.JobID,
 	// Search the existing plan files for the PartPlans for the given jobId
 	// only the files which are not empty and have JobId has prefix and DataSchemaVersion as Suffix
 	// are include in the result
-
 	files := func(prefix, ext string) []os.FileInfo {
 		var files []os.FileInfo
 		_ = filepath.Walk(ja.planDir, func(path string, fileInfo os.FileInfo, _ error) error {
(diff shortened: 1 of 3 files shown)


ste/mgr-JobPartMgr.go
```diff
@@ -334,8 +334,14 @@ func (jpm *jobPartMgr) ScheduleTransfers(jobCtx context.Context) {
 
 	jpm.clientInfo()
 
+	// Cache IsFinalPart before the transfer loop, since plan memory may be unmapped
+	// by progressive cleanup after the last ReportTransferDone fires.
+	isFinalPart := plan.IsFinalPart
```
kylemoore-microsoft, line 339: For my understanding, the mmap issue is separate from the deadlock one right? And was this leading to panic?

pbanakar-microsoft: yes, mmap issue race condition is different from deadlock.
I didn't see mmap race condition occurring due to this variable, but it can potentially happen if we don't take this fix.

PR #3412, opened 2026-03-16
Add hardlink preserve support for Download and S2S sync operation

dphulkar-msft:
## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/azcopy/syncComparator.go
+++ b/azcopy/syncComparator.go
@@ -131,15 +131,13 @@ func (f *syncDestinationComparator) ProcessIfNecessary(destinationObject travers
 		return nil
 	}
 
-	sourceObjectInMap, present := f.sourceIndex.IndexMap[destinationObject.RelativePath]
+	// Normalize the key upfront for case-insensitive destinations so the
+	// lookup always matches the lowercase-keyed sourceIndex.
 	srcKey := destinationObject.RelativePath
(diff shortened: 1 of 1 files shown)


azcopy/syncComparator.go
```diff
@@ -504,6 +542,221 @@ func (f *syncSourceComparator) ProcessIfNecessary(sourceObject traverser.StoredO
...
+		} else if first != obj.Inode {
+			destGroupIsMultiSource[destInode] = true
+		}
+	}
+
+	for _, sourceObject := range f.srcPendingHardlinkObjects.IndexMap {
+
+		dstKey := sourceObject.RelativePath
+		destinationObjectInMap, present := f.destinationIndex.IndexMap[dstKey]
+		if !present && f.destinationIndex.IsDestinationCaseInsensitive {
```
adreed-msft: If the destination is case insensitive, we should commit to the destination index in a way that is always case-insensitive, and try to read from it in a way that is always case-insensitve. Thus, we shouldn't need to retry if we just lowercase it before looking up.

PR #3429, opened 2026-04-02
fix: skip inflated folder stats for virtual directory prefixes in sync

kylemoore-microsoft:
## Description

- **Feature / Bug Fix**: Fix inflated `FoldersScanned` and `FoldersNoTransferNeeded` stats in sync mode caused by synthetic virtual directory prefixes being counted as real folders.

  During hierarchical blob listing, when a virtual directory prefix has no real dir stub or HNS directory blob, the blob traverser creates a synthetic `StoredObject` to track directory structure for mirror sync. These synthetic entries were incorrectly incrementing folder enumeration and "no transfer needed" counters, inflating the reported stats.

  This PR:
  - Adds an `isVirtualPrefix` field to `StoredObject` and `minimalStoredObject` to tag synthetic directory prefixes
  - Skips `FoldersScanned` increment in the blob traverser for virtual prefixes when using the sync orchestrator
  - Skips `FoldersNoTransferNeeded` increment in the sync comparator and orchestrator for virtual prefixes
  - Propagates the `isVirtualPrefix` flag through the sync orchestrator's sub-directory tracking

- **Related Links**:
  - [Issues]()
  - [Team thread]()
  - [Documents]()
  - [Email Subject]

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

- Manual testing with non-HNS blob containers that produce virtual directory prefixes during sync
- Verified that `FoldersScanned` and `FoldersNoTransferNeeded` no longer include synthetic virtual directory prefix entries
- Verified that real dir stubs and HNS directories continue to be counted correctly

### DIFF
--- a/cmd/zc_enumerator.go
+++ b/cmd/zc_enumerator.go
@@ -91,6 +91,10 @@ type StoredObject struct {
 	blobSnapshotID string
 	blobDeleted    bool
 
+	// Synthetic virtual directory prefix, not a real dir stub or HNS directory.
+	// Needed by the sync orchestrator for directory tracking but should not increment folder scan counters.
+	isVirtualPrefix bool
+
 	// Lease information
 	leaseState    lease.StateType
 	leaseStatus   lease.StatusType
--- a/cmd/zc_enumerator.go
+++ b/cmd/zc_enumerator.go
@@ -91,8 +91,9 @@ type StoredObject struct {
 	blobSnapshotID string
 	blobDeleted    bool
 
-	// Synthetic virtual directory prefix, not a real dir stub or HNS directory.
-	// Needed by the sync orchestrator for directory tracking but should not increment folder scan counters.
+	// isVirtualPrefix marks synthetic directory prefixes (not real dir stubs or HNS directories)
+	// created during hierarchical listing. These are tracked by the sync orchestrator for
+	// directory comparison but should not inflate folder scan or transfer counters.
 	isVirtualPrefix bool
 
 	// Lease information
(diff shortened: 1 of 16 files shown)


cmd/zc_enumerator.go
```diff
@@ -91,6 +91,10 @@ type StoredObject struct {
 	blobSnapshotID string
 	blobDeleted    bool
 
+	// isVirtualPrefix marks synthetic directory prefixes
+	// it's used to skip incrementing folder stats for virtual prefixes during scanning for Blob to Blob syncs
```
shnayak-msft, line 95: what about s3 to blob?

cmd/zc_traverser_blob.go
```diff
@@ -523,6 +523,7 @@ func (t *blobTraverser) parallelList(containerClient *container.Client, containe
 							noMetadata,
 							containerName,
 						)
+						storedObject.isVirtualPrefix = true
```
shnayak-msft, line 526: Will this be true for HNS as well?

shnayak-msft (approved):

cmd/zc_enumerator.go
kylemoore-microsoft: The same issue may be there for s3 to blob, but I am leaving that out of scope for now. In the mover repo, we just throw out the folder stats for S3 to blob anyways since both are flat

cmd/zc_traverser_blob.go
kylemoore-microsoft: No, the StoredObject for HNS dirs is formed on line 464

pbanakar-microsoft (approved):

PR #3436, opened 2026-04-09
Centralize HTTP client into a shared global instance

shnayak-msft:
Introduce common.GetGlobalHTTPClient() to create a single process-wide HTTP client, replacing per-job client creation in jobMgr and the separate front-end client in traverser/credentialUtil.go.

- Add common/azHttpClient.go with GetGlobalHTTPClient, MaxIdleConnsPerHost tuning logic, and optional connection-reuse tracing via httptrace
- Replace NewAzcopyHTTPClient in ste/mgr-JobMgr.go with the global client
- Update traverser/credentialUtil.go to use the global client, removing the local frontEndMaxIdleConnectionsPerHost constant
- Add NewTracingTransport wrapper for per-label connection stats logging (GotConn reuse/idle metrics, PutIdleConn errors, periodic dump)

Current http client configuration sets the connection reuse to 2 instead of a value based on concurrency. High concurrency leads to a high number of new TCP connections which causes socket errors and high syscall latency impacting performance. This change addresses this issue. 

## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?
Pipeline run - **https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=30434&view=results**

### DIFF
--- a/common/azHttpClient.go
+++ b/common/azHttpClient.go
@@ -1,0 +1,256 @@
+// Copyright © Microsoft <wastore@microsoft.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package common
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptrace"
+	"runtime"
+	"strconv"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+// GlobalHTTPClient is the process-wide HTTP client used by AzCopy when initialized via InitGlobalHTTPClient.
+var (
+	GlobalHTTPClient     *http.Client
+	globalHTTPClientOnce sync.Once
+)
+
+const (
+	maxIdleConnsPerHost_MaxValue = 3000
+	httpTraceTickerInterval      = time.Minute * 1
+)
+
+// GetGlobalHTTPClient initializes and returns the process-global HTTP client exactly once.
+// Subsequent calls return the same client. The logger function, if provided on the first call,
+// will be invoked with status messages.
+func GetGlobalHTTPClient(logger ILoggerResetable) *http.Client {
+	globalHTTPClientOnce.Do(func() {
+		const concurrentDialsPerCpu = 10
+		client := &http.Client{
+			Transport: &http.Transport{
+				Proxy:                  GlobalProxyLookup,
+				MaxConnsPerHost:        concurrentDialsPerCpu * runtime.NumCPU(),
+				MaxIdleConns:           0,
+				MaxIdleConnsPerHost:    GetMaxIdleConnsPerHost(),
+				IdleConnTimeout:        180 * time.Second,
+				TLSHandshakeTimeout:    10 * time.Second,
+				ExpectContinueTimeout:  1 * time.Second,
+				DisableKeepAlives:      false,
+				DisableCompression:     true,
+				MaxResponseHeaderBytes: 0,
+			},
+		}
+		GlobalHTTPClient = client
+		if logger != nil {
+			if tr, ok := client.Transport.(*http.Transport); ok {
+				logger.Log(LogError, // XDM: This is error level on purpose as we want to make sure it is seen in the logs
+					fmt.Sprintf(
+						"GetGlobalHTTPClient: initialized %p MaxIdleConnsPerHost=%d MaxConnsPerHost=%d MaxIdleConns=%d",
+						client, tr.MaxIdleConnsPerHost, tr.MaxConnsPerHost, tr.MaxIdleConns))
+			} else {
+				logger.Log(LogError, fmt.Sprintf("GetGlobalHTTPClient: initialized %p", client))
+			}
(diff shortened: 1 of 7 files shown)


common/azHttpClient.go
```diff
@@ -0,0 +1,256 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+// GlobalHTTPClient is the process-wide HTTP client used by AzCopy when initialized via InitGlobalHTTPClient.
+var (
+	GlobalHTTPClient     *http.Client
```
adreed-msft, line 38: Singleton pattern, but this is exported? I don't see this used anywhere else. I'd recommend un-exporting it to avoid improper use.

common/azHttpClient.go
```diff
@@ -0,0 +1,256 @@
...
+				logger.Log(LogError, fmt.Sprintf("GetGlobalHTTPClient: initialized %p", client))
+			}
+		}
+	})
+	return GlobalHTTPClient
+}
+
+// This is a code duplication of GetMainPoolSize in ste/concurrency.go
+// Ideally we should just move concurrency.go to common package. That is a todo for future.
+func GetMaxIdleConnsPerHost() int {
```
adreed-msft, line 84: This doesn't seem like it ever updates, akin to GetMainPoolSize? Could that prove to eventually be an issue?

adreed-msft (approved):

common/azHttpClient.go
```diff
@@ -0,0 +1,207 @@
...
+	return globalHTTPClient
+}
+
+// GetGlobalHTTPClient returns the process-global HTTP client previously configured
+// via InitGlobalHTTPClient. If InitGlobalHTTPClient has not yet been called, the
+// client is lazily created with a conservative fallback so that test/import paths
+// which bypass azcopy.NewClient still get a working client.
+func GetGlobalHTTPClient() *http.Client {
+	globalHTTPClientOnce.Do(func() {
+		globalHTTPClient = buildGlobalHTTPClient(fallbackMaxIdleConnsPerHost)
```
adreed-msft: should we log anything if this happens?

common/azHttpClient.go
```diff
@@ -0,0 +1,207 @@
...
+	"net/http/httptrace"
+	"runtime"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+// globalHTTPClient is the process-wide HTTP client used by AzCopy. It is configured by
```
al-msft: A repo-wide search for "NewAzcopyHTTPClient" in any case turns up functions in oauthTokenManager.go and mgr-JobPartMgr.go, as well as references in four test files.

I'm still coming up to speed in this repo. Why were those other files left untouched?

al-msft:
Some questions on testing:

1. These changes didn't require any updates to automated tests?
2. Do you see any value in unit-testing the new file, [azHttpClient.go](https://github.com/Azure/azure-storage-azcopy/pull/3436/changes#diff-fbeebe13d2dde6ee638b07003e052db92fa5b93e26b11837d3da24e05ca0ce5b)?

common/azHttpClient.go
```diff
@@ -0,0 +1,207 @@
...
+// InitGlobalHTTPClient initializes the process-global HTTP client exactly once with
+// the supplied limits. Callers should pass the value derived from the STE
+// ConcurrencySettings (e.g. ConcurrencySettings.MaxIdleConnections) so that the HTTP
+// transport honors the same concurrency configuration as the rest of AzCopy.
+// Subsequent calls are no-ops and the originally supplied values continue to apply.
+// The configured client's transport limits are logged later by JobMgr when a job
+// first picks it up, where a job-scoped logger is available.
+func InitGlobalHTTPClient(maxIdleConnsPerHost int) *http.Client {
+	globalHTTPClientOnce.Do(func() {
+		globalHTTPClient = buildGlobalHTTPClient(maxIdleConnsPerHost)
```
adreed-msft, line 56: If get, then init is called, max idle conns never gets updated. What if we just poll getMainPoolSize, and only have a Get function? I don't think that comes with major side effects.

adreed-msft (approved):

common/azHttpClient.go
adreed-msft: *fwiw*, that's effectively where the data is already sourced from anyway so...?

common/azHttpClient.go
shnayak-msft: Done

common/azHttpClient.go
shnayak-msft: Good catch. let me explain the split, and I've added doc comments to both functions so this is clearer in the code:

ste.NewAzcopyHTTPClient is now used only by tests and the standalone cmd binary. Those run outside azcopy.NewClient, so InitGlobalHTTPClient never fires and tests shouldn't depend on global startup wiring from another package. It's exported only because testblob.go is outside the ste package.

common.newAzcopyHTTPClient (lowercase, in oauthTokenManager.go) is a separate client deliberately, for AAD/IMDS endpoints: it uses Dial instead of DialContext, has no MaxConnsPerHost cap, and targets a different host set (login.microsoftonline.com, IMDS, etc.). Token RPS is negligible vs. data-plane RPS, so consolidation wouldn't buy much and would be a real behavior change for the AAD path.

common/azHttpClient.go
shnayak-msft: Yes this is the challenge I was pointing out in an earlier discussion. Calling getMainPoolSize directly collides with package layering: getMainPoolSize lives in ste and returns *ConfiguredInt, an ste type, and ste already imports common. Ideally we should just move concurrency.go to common package, but it will be a bigger change.

Within this PR I've taken the smaller fix that addresses the underlying concern: removed the lazy fallback from GetGlobalHTTPClient so it panics if Init hasn't run. That way the hazard you flagged (Get first → silent misconfig) becomes a loud failure instead.

common/azHttpClient.go
shnayak-msft: We don't have a reliable logger at this point and the runtime stack trace from the panic will tell who called Get too early. Other files like access.go, extensions.go follow this convention as well.

common/azHttpClient.go
adreed-msft: at risk of being a bit less DRY... could just replicate the functionality in common and add a test for if they ever differ. mechanically, this sucks, but it at least ensures The Bad will never happen.

common/azHttpClient.go
al-msft: @shnayak-msft: Thanks for the explanation - especially for documenting it in the code. One thing I'm not clear on is "the standalone cmd binary" you mentioned. If you're referring to the testSuite\cmd package, it seems ste.NewAzcopyHTTPClient is now only used by test code. If so, can we move it out of mgr-JobPartMgr.go, which I believe is a production file, and into a test file? My motivation here - and in general - is partly to minimize the attack surface, complexity, dependencies, etc. of each unit of code - whether that unit is a function, file, package, module, project, or some other granule of production or test code. Simple tests are also important, but when the tests need functionality the production code doesn't, I typically create ulilities specifically for testing and group them with the tests as test code, not production code.

adreed-msft (approved):

adreed-msft (approved):

shnayak-msft:
I have run the e2e pipeline multiple times and identified the issues to be flaky and unrelated. Bypassing and merging.

PR #3440, opened 2026-04-16
Fix HNS container-root ACL sync

kylemoore-microsoft:
## Summary
- Fix container-root HNS ACL transfer in the mover sync orchestrator by identifying the self-reference folder entity (empty relativePath + Folder entityType) and indexing it without re-enqueuing — avoids the infinite loop while still routing the root through SetContainerACL.
- Tighten the container-root emission gate in the blob traverser to only fire when the destination is BlobFS (HNS), since blobFolderSender.SetContainerACL hardcodes the DFS endpoint and only HNS→HNS can actually land the ACL. Emitting for flat-Blob / File / Local destinations was scanned-but-always-filtered downstream, causing Scanned > Processed.
- Plumb destResourceType through the blob traverser so the gate can distinguish HNS destinations from others.

## Test plan
- [x] c2ctest/TestHNSRootACL_CopyAndSync (hns-full-container, hns-subpath) — both pass with smslidingwindow=true
- [x] Re-run with smslidingwindow=false to confirm the standard (non-mover) path still works end-to-end
- [x] Existing dataplane_e2e HNS copy+sync tests


cmd/zc_traverser_blob.go
```diff
@@ -65,6 +64,8 @@ type blobTraverser struct {
 
 	isDFS bool
 
+	destResourceType *common.Location
```
kylemoore-microsoft, line 67: For context, this was already being set by syncEnumerator.go on line 108 so you won't see it in the diff

pbanakar-microsoft (approved):

PR #3450, opened 2026-04-30
Update Mutex env var

wonwuakpa-msft:
## Description

Fix failing Test_on_Ubuntu run in CI by replacing the mutex url env var

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Ran CI

### DIFF
--- a/tool_distributed_mutex.py
+++ b/tool_distributed_mutex.py
@@ -19,19 +19,19 @@ UNLOCK = 'unlock'
 def get_raw_input():
     parser = argparse.ArgumentParser(description='Lock/unlock a distributed mutex (implemented with blob lease)')
     parser.add_argument('action', help='can be "lock" (attempt to acquire lease) or "unlock" (break lease)')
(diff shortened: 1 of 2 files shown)


tool_distributed_mutex.py
```diff
@@ -19,19 +19,19 @@
 def get_raw_input():
     parser = argparse.ArgumentParser(description='Lock/unlock a distributed mutex (implemented with blob lease)')
     parser.add_argument('action', help='can be "lock" (attempt to acquire lease) or "unlock" (break lease)')
-    parser.add_argument('mutex_url', help='points to a blob url (SAS included)')
+    parser.add_argument('mutex1_url', help='points to a blob url (SAS included)')
```
adreed-msft, line 22: If we're just changing the parameter target upstream why do we need to change the name downstream?

adreed-msft (approved):

PR #3453, opened 2026-05-05
Ensure get/set ACLs are on URLs with paths 

wonwuakpa-msft:
This PR addresses a side effect of an azcore bug fix. 

It was found a part of typespec testing (upgraded azblob version from1 v1.6.2 to v1.7.0-beta.1.0 and its dependent packages) 

It modifies how two tests in the old e2e test framework handle ACLs. When the transfer item is the container, AzCopy constructs a pathless DFS URL which is now invalid on the service-side
`HEAD https://<account>.dfs.core.windows.net/<filesystem>?action=getAccessControl.` 
This is a path-level action (Path - Get Properties), so the URL is malformed without a path. 

_What changed?_ 

[As of azcore v1.21.1, ](https://github.com/Azure/azure-sdk-for-go/blame/ee4be650cb54df163b388f7d0c0f79c3201d182e/sdk/azcore/runtime/request.go#L96)supplying a pathless DFS SAS url will be broken by downstream callers of `runtime.JoinPaths`. This code path is hit in AzCopy when making a new FileClient with an empty path. There [was logic added](https://github.com/Azure/azure-sdk-for-go/blame/ee4be650cb54df163b388f7d0c0f79c3201d182e/sdk/azcore/runtime/request.go#L96) in the GO SDK to stop appending a `/` to paths that started with `?`.  So, it no longer returns `https://acct.dfs.core.net/myfs/?sv=2026-04-06` instead now returns `https://acct.dfs.core.windows.net/myfs?sv=2026-04-06`


wonwuakpa-msft:
\azp run

wonwuakpa-msft:
\azp run

go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/Azure/azure-storage-azcopy/v10
 
 require (
 	cloud.google.com/go/storage v1.50.0
-	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.2
+	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.7.0-beta.1.0.20260501171610-11f9a5ca939c
```
wonwuakpa-msft: I will revert back to 1.6.2 before pushing, we do not merge in beta

dphulkar-msft:
Could you add a test for this?

wonwuakpa-msft:
> Could you add a test for this?

This PR is making changes that will allow `TestBasic_OverwriteHNSDirWithChildren`  and `TestProperties_HNSACLs`   to pass after we uprgade azblob package. So, the changes in the PR are covered by these tests

gapra-msft (approved):

wonwuakpa-msft:
\azp run

dphulkar-msft (approved):

PR #3467, opened 2026-06-04
Remove 0-padding in mode with SetUID 

wonwuakpa-msft:
## Description

We were always prepending the x-ms-mode with a '0'. This becomes a problem in the case when the left most bit is non-zero (e.g setuid, getuid or sticky), the mode becomes 15 bits and exceeds the service 12-bit max. The fix I added is to pad the mode with a 0 only if its less than 12 bits.

- **Related Links**:
- [ADO Item](https://msazure.visualstudio.com/One/_backlogs/backlog/XStore-DevExp/Epics?showParents=false&workitem=38129420)

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Ran existing NFS tests


dphulkar-msft:
Can you add a test case for this?

dphulkar-msft (approved):

gapra-msft (approved):

PR #3473, opened 2026-06-10
Add vulnerabilities scan for ACR images

dphulkar-msft:
## Description

https://[mcr.microsoft.com/v2/azcopy/azcopy/tags/list](https://mcr.microsoft.com/v2/azcopy/azcopy/tags/list)

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?


docker/buildandruncontainer.sh
```diff
@@ -6,7 +6,7 @@
 azcopy_version=$(../azcopy --version | awk '{print $3}')
 
 # Construct the Docker image tag using the fetched version
-docker_image_tag="azure-azcopy-ubuntu-x86_64.$azcopy_version"
+docker_image_tag="azure-azcopy-ubuntu-x86_64:$azcopy_version"
```
gapra-msft: Can we double check that all characters that can be in azcopy version (including preview ~ or whatever we do) is a valid tag name?

wonwuakpa-msft (approved):

wonwuakpa-msft (approved):

build-1es-pipeline.yaml
```diff
@@ -1140,25 +1140,25 @@ extends:
 
                   - task: 1ES.BuildContainerImage@1
                     inputs:
-                      image: azcopycontainers.azurecr.io/azure-azcopy-ubuntu-x86_64.$(azcopy_docker_version):latest
+                      image: azcopycontainers.azurecr.io/public/azcopy/linux/amd64:$(azcopy_docker_version)
                       path: '$(System.DefaultWorkingDirectory)'
-                      dockerfile: docker/Dockerfile
+                      dockerfile: docker/DockerfileMariner
```
al-msft, line 1153: I don't see any context provided for this PR, so I have a few obervations and questions.

There are four "Dockerfile*" files in the "docker" directory of this repo:
- Dockerfile and DockerfileArm64, which both use Ubuntu 22.04 as the base image, and
- DockerfileMariner and DockerfileMarinerArm64, which both use Mariner 2.0 ([which reached end of life over a year ago](https://github.com/microsoft/azurelinux/issues/13555)) as the base image.

On main, this task and the one below are intended to build and push container images based on Dockerfile (Ubuntu, amd64), and the next pair of tasks build and push the container images based on DockerfileMariner (Mariner, amd64).

This PR seems to remove the ability to build and push container images based on Ubuntu and now build and push container images based on Mariner twice. This first pair of tasks is now nearly identical to the next, except that the first pair uses `$(azcopy_docker_version)` as the image version, whereas the second pair uses `latest`.
1. Is this PR intended to remove the ability to build and push container images based on Ubuntu? If so, should the Dockerfile and DockerfileArm64 files be deleted?
2. Is this PR intended to preserve the ability to build and push container images based on Mariner, but now push a second image that lists the specific AzCopy version, alongside the existing image that lists its version as "latest"?
3. Is this PR intended to stop pushing container images to paths that indicate the distro they're based on, and instead provide the Mariner-based image for all Linux environments? I'm trying to make sense of the change to the `image` input of these tasks.

This comment can also be applied to the later four tasks this PR updated in this file - the ones that are nearly identical to these four, but use arm64 as the architecture, instead of amd64. Note that on main, the versions of the Ubuntu arm64 tasks seem to incorrectly use Dockerfile, instead of DockerfileArm64.

docker/buildandruncontainer.sh
al-msft, line 1: How is this file actually used? I have the same question for the other two shell scripts this PR updates. I can't find any explicit uses of any of the following shell scripts:
- buildandruncontainer.sh
- buildcontainer.sh (except its use in buildandruncontainer.sh)
- build.sh (except its use in buildcontainer.sh)
- dockerinstall.sh (except its use in buildandruncontainer.sh)
- publishcontainer.sh

build-1es-pipeline.yaml
dphulkar-msft: Q1: Yes. Per PM Daniel and Scott's decision, the first release ships only CBL-Mariner (Azure Linux) images, not Ubuntu. Dockerfile and DockerfileArm64 are now unused and will be deleted.

Q2: Yes. It's the same Mariner image under two tags:$(azcopy_docker_version) (e.g. 10.32.22) and: latest (always the newest release) so customers can pin a version or auto-track latest.

Q3: Yes. The distro is no longer encoded in the path. Images publish to azcopy/linux/<arch> (OS/arch in the path, clean version/latest tag). Including Linux in the naming helps keep us future-proof when we also need to release Windows containers. For Docker image tagging, the general best practice is to use amd64 rather than x86_64.

docker/buildandruncontainer.sh
dphulkar-msft: These are manual utility scripts for building/running/publishing a container by hand, the automated pipeline builds and pushes via the 1ES built-in tasks, so it doesn't call any of them. I updated the scripts in this PR only to keep them consistent with the new Mariner-only, azcopy/linux/<arch> tagging convention. They aren't wired into automation. Let me know if we should remove it altogether as its not part of the pipeline.

wonwuakpa-msft (approved):

gapra-msft (approved):

docker/buildandruncontainer.sh
al-msft: @dphulkar-msft: Thanks for clearing that up for me. I'm not familar enough with the process of updating and testing AzCopy Docker images to make the call about removing these manual scripts.

If we keep them, I'd suggest adding a comment to the top of `buildandruncontainer.sh` (or to `README.md` under `docker`) to explain what you said above (maybe in the same PR that removes `Dockerfile` and `DockerfileArm64`). I'm also wondering if grouping all the manual scripts in a sub-directory of `docker` would be worthwhile, since that would convey the purpose of those scripts at a glance and would hide all the manual scripts from view by default in IDEs like VS Code when someone expands `docker` to view/update the Docker files the pipeline uses.

PR #3493, opened 2026-07-02
Users/pbanakar/streaming mergejoin

pbanakar-microsoft:
## Description

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

### DIFF
--- a/cmd/syncMergeJoin.go
+++ b/cmd/syncMergeJoin.go
@@ -28,6 +28,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -180,15 +181,20 @@ func traverserToChannel(
 			count++
 			// Sort-invariant guard (always on, ~one string compare per object): the
 			// streaming merge-join is only correct if each stream is emitted in strictly
-			// non-decreasing key order. If a traverser ever emits an out-of-order key,
-			// the merge-join can silently mis-classify objects (missed or extra transfers),
-			// so we log it at Error even in production. This is rare (fires only on the
-			// anomaly) and cheap, so it does not spam logs for millions of sorted files.
+			// non-decreasing key order. If a traverser ever emits an out-of-order key, the
+			// two-pointer merge can silently mis-classify objects (missed transfers, or — for
+			// a mirror sync — extra deletes of destination objects). For the allow-listed,
+			// service-sorted endpoints this must never happen, so treat it as a HARD FAILURE
+			// rather than a warning: abort the traversal by returning an error. It propagates
+			// via errCh and is surfaced to the orchestrator as a mergeJoinTraversalError,
+			// failing this directory (and the job) instead of producing incorrect sync results.
 			curKey := mergeJoinChildKey(obj)
 			if count > 1 && curKey < prevKey {
-				mergeJoinSyncOneDirLog(common.LogError,
-					fmt.Sprintf("ORDER VIOLATION in %s: object #%d key=%q < previous key=%q — listing is not lexicographically sorted; merge-join results may be INCORRECT for this directory",
-						label, count, curKey, prevKey), true)
+				orderErr := fmt.Errorf(
+					"ORDER VIOLATION in %s: object #%d key=%q < previous key=%q — listing is not lexicographically sorted; aborting merge-join to avoid incorrect sync results",
+					label, count, curKey, prevKey)
+				mergeJoinSyncOneDirLog(common.LogError, orderErr.Error(), true)
+				return orderErr
 			}
 			prevKey = curKey
 			mergeJoinTraceLog("%s LIST #%d relPath=%q entity=%s size=%d lmt=%s",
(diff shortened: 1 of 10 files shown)


cmd/syncOrchestrator.go
```diff
@@ -612,6 +612,27 @@ func syncOrchestratorHandler(cca *cookedSyncCmdArgs, enumerator *syncEnumerator,
...
 	}
 
+	// Log (once per job) whether the streaming merge-join is used and, if so, what enabled it:
+	// the AZCOPY_USE_STREAMING_MERGE_JOIN env flag (test/blanket) or the per-job subscription
+	// allowlist (featureConfig). This makes it easy to confirm the gating in production logs.
+	if useStreamingMergeJoin(cca) {
+		var reason string
+		switch {
+		case streamingMergeJoinEnabled && cca.useStreamingMergeJoin:
+			reason = "env AZCOPY_USE_STREAMING_MERGE_JOIN + subscription allowlist (featureConfig)"
```
shnayak-msft, line 622: This comment is not meaningful for AzCopy, as it is for mover. We should clarify that in comments. We should use mover logging to check if subscription allowlist is being used.

cmd/syncOrchestrator.go
```diff
@@ -698,6 +719,114 @@ func (cca *cookedSyncCmdArgs) runSyncOrchestrator(enumerator *syncEnumerator, ct
...
+		if useStreamingMergeJoin(cca) {
+			mergeJoinSyncOneDirLog(common.LogDebug,
+				fmt.Sprintf("Processing dir '%s'", dir.(minimalStoredObject).relativePath))
+
+			var subDirs []minimalStoredObject
+			var mergeErr error
+
+			// Channel-based merge-join: bridge source + destination traversers into
+			// back-pressured channels and merge their sorted streams.
+			pt, ptErr := InitResourceTraverser(
```
shnayak-msft, line 746: Why not initialize pt and st once and fork the default/mergejoin implementation based on the flag? There is some opportunity for optimization here.

cmd/syncMergeJoin.go
```diff
@@ -0,0 +1,602 @@
...
+			subDirs, err = mergeJoinHandleBothExist(enumerator, cca, comparator, srcObj, dstObj, srcOrigPath, subDirs)
+			if err != nil {
+				return subDirs, err
+			}
+			srcObj, srcOk = srcNext()
+			dstObj, dstOk = dstNext()
+		}
+	}
+
+	// Drain remaining source-only objects
```
shnayak-msft, line 388: There might be an issue here. please verify.

traverserToChannel closes objCh via defer whether the traversal succeeded or failed, so the merge cannot tell "cleanly exhausted" apart from "aborted mid-listing due to an error" 

If one side's channel closes early on an error, the drain loop might produce incorrect results:

- source failure + mirror → every remaining destination object is deleted synchronously (Problem)
- destination failure → every remaining source object is scheduled for transfer, overwriting objects that may not actually be missing (this might be okay)

The check at the bottom of the function correctly surfaces the error to the orchestrator, but only after the damage. 

Some issues that may cause this in my opinion - SAS/OAuth token expiry mid-listing on long runs, sustained 429/503 throttling past SDK retry budget

You can probe both errCh's non-blockingly on every iteration (main + drain loops) and short-circuit as soon as either has an error. Specifically for the mirror case.

cmd/syncOrchestrator.go
shnayak-msft, line 622: I feel we should not use an AzCopy environment variable here. Since this is explicitly for mover, it should only be invoked by the init parameters in AzCopy initialization path. Just one boolean should control the flow. The naming gives an impression that AzCopy supports this flag.

cmd/syncMergeJoin.go
```diff
@@ -0,0 +1,602 @@
...
+				if elapsed > 30*time.Second {
+					mergeJoinSyncOneDirLog(common.LogWarning,
+						fmt.Sprintf("[SLOW] %s first object took %v (goroutines=%d)", label, elapsed, runtime.NumGoroutine()), true)
+				}
+			}
+			count++
+			// Sort-invariant guard (always on, ~one string compare per object): the
+			// streaming merge-join is only correct if each stream is emitted in strictly
+			// non-decreasing key order. If a traverser ever emits an out-of-order key,
+			// the merge-join can silently mis-classify objects (missed or extra transfers),
```
shnayak-msft: an out-of-order key silently mis-classifies object. logging alone may not be sufficient given the failure mode. We should have a failure at this point, more so for mirror.  If we think this will be rare, all the more reason to do this.

The service contract is sorted for AWS S3, Azure Blob, and ADLS Gen2, so in the intended allow-list this fires only when the invariant is broken by something we can actually have a bug in:

- the newly-added per-page merge in blobTraverser.parallelList or the whole-level buffer-merge in s3Traverser.Traverse
- a non-AWS S3-compatible endpoint whose ListObjectsV2 ordering isn't compliant (the S3ProviderOnPrem branch)
- snapshot/version/deleted-blob inclusion producing equal keys the merge loop can't reason about (not strictly unsorted but breaks SRC==DST handling)

shnayak-msft (approved):

cmd/syncMergeJoin.go
pbanakar-microsoft: Fixed.

cmd/syncMergeJoin.go
pbanakar-microsoft: Fixed.

PR #3482, opened 2026-06-23
TASK 38260338: Updated the pipelines to produce Linux builds capable of complying with the FIPS 140-3 standard.

al-msft:
Microsoft's version of Go provides the FIPS capability. The FIPS-capable builds are produced alongside the existing builds, which don't use Microsoft's version of Go. 

## Description 

**Update: This PR has been split into https://github.com/Azure/azure-storage-azcopy/pull/3496 (test pipeline updates) and https://github.com/Azure/azure-storage-azcopy/pull/3488 (release pipeline updates).**

The CSI driver image is built by another team at Microsoft. That image needs to comply with the FIPS 140-3 standard. A FIPS-compliance checker scans all binaries in the image; AzCopy is one of those binaries. To pass the FIPS-compliance check, each binary must be built with the Microsoft build of Go in the particular way the check expects and pass a run-time check that considers the host OS. The CSI driver image downloads azcopy as an RPM package on Azure Linux 3.0. To reduce the risk of migrating AzCopy from the official build of Go to the Microsoft build of Go, it was decided to initally release FIPS-capable builds of AzCopy only for Azure Linux 3.0, as an RPM package for the AMD64 and ARM64 architectures the CSI driver image supports, alongside the existing builds of AzCopy, and not to release Docker container images containing FIPS-capable builds. 

These changes are intended to match the approach that was taken in previous pipeline updates, especially regarding the assurances and costs of testing. The changes are expected to increase the run-time of the affected pipelines; to mitigate this, parallelism was used (via the "matrix" strategy) where it seemed unlikely to lead to conflicts in the parallel operations, the tests in run_scenarios.yml were not updated to test the FIPS-capable builds (there seemed to be little value in doing so), and only the FIPS-capable build for the AMD64 architecture was tested (following the approach that was taken for the non-FIPS-capable builds). The "unit" tests, end-to-end tests, and Python tests were updated to test the FIPS-capable build, because each of those types of tests includes tests at the integration level or higher. If we later update the set of "unit" tests to ensure they are all true unit tests, we should update the pipeline to stop running those tests on the FIPS-capable build. 

**Cut for Time/Risk:** 
- FIPS-capable builds for any OS other than Linux, any distro other than Azure Linux 3.0, any architectures other than AMD64 and ARM64, and in any format other than an RPM package
- FIPS-capable builds in Docker container images 
- Updating GitHub Actions 
- Logging a message when AzCopy is invoked that states whether FIPS is enabled 
[...]

PR #3496, opened 2026-07-07
TASK 38260338: Updated the test pipeline to produce Linux builds capable of complying with the FIPS 140-3 standard. 

al-msft:
Microsoft's version of Go provides the FIPS capability. The FIPS-capable builds are produced alongside the existing builds, which don't use Microsoft's version of Go.  

## Description  
 
**Update: This PR will be split into multiple smaller PRs, starting with https://github.com/Azure/azure-storage-azcopy/pull/3504.**

The CSI driver image is built by another team at Microsoft. That image needs to comply with the FIPS 140-3 standard. A FIPS-compliance checker scans all binaries in the image; AzCopy is one of those binaries. To pass the FIPS-compliance check, each binary must be built with the Microsoft build of Go in the particular way the check expects and pass a run-time check that considers the host OS. The CSI driver image downloads azcopy as an RPM package on Azure Linux 3.0. To reduce the risk of migrating AzCopy from the official build of Go to the Microsoft build of Go, it was decided to initally release FIPS-capable builds of AzCopy only for Azure Linux 3.0, as an RPM package for the AMD64 and ARM64 architectures the CSI driver image supports, alongside the existing builds of AzCopy, and not to release Docker container images containing FIPS-capable builds.  

This PR updates the test pipeline; the release pipeline is being updated in https://github.com/Azure/azure-storage-azcopy/pull/3488. An earlier draft PR, https://github.com/Azure/azure-storage-azcopy/pull/3482, included the updates to both pipelines.

These changes are intended to match the approach that was taken in previous pipeline updates, especially regarding the assurances and costs of testing. The changes are expected to increase the run-time of the affected pipelines; to mitigate this, parallelism was used (via the "matrix" strategy) where it seemed unlikely to lead to conflicts in the parallel operations, the tests in run_scenarios.yml were not updated to test the FIPS-capable builds (there seemed to be little value in doing so), and only the FIPS-capable build for the AMD64 architecture was tested (following the approach that was taken for the non-FIPS-capable builds). The "unit" tests, end-to-end tests, and Python tests were updated to test the FIPS-capable build, because each of those types of tests includes tests at the integration level or higher. If we later update the set of "unit" tests to ensure they are all true unit tests, we should update the pipeline to stop running those tests on the FIPS-capable build.  

**Cut for Time/Risk:**  
- Updating GitHub Actions  
- Logging a message when AzCopy is invoked that states whether FIPS is enabled  
- Updating the performance-test pipeline to test the FIPS-capable builds  
- Addressing technical debt in the pipelines  
- Updating the pool agent that runs Azure Linux 3.0 to use the same minor version of Python 3 as the other pool agents to run the Python-based AzCopy tests  
- Optimizing the run-time, parallelism, amount of duplication, and the use of parameters and variables in the pipelines  
- Scanning the FIPS-capable binaries for FIPS compliance in the pipeline(s)  

**Related Links**:  
- [TASK 38260338](https://msazure.visualstudio.com/One/_workitems/edit/38260338)  
- [BlobFuse FIPS Teams chat](https://teams.microsoft.com/l/chat/19:4d86c8df4c5d4e2fa324a5064f0852f0@thread.v2/conversations?context=%7B%22contextType%22%3A%22chat%22%7D)
- [AZCopy FIPS Compliance Teams chat](https://teams.microsoft.com/l/chat/19:96584494811d4c0dbeb7f4eb07bc789a@thread.v2/conversations?context=%7B%22contextType%22%3A%22chat%22%7D)
- [FIPS Support in the Microsoft build of Go (Warning: This is misleading in some places)](https://github.com/microsoft/go/blob/microsoft/main/eng/doc/fips/README.md)  

## Type of Change  
  
- [ ] Bug fix  
- [x] New feature  
- [ ] Other (describe): 

## How Has This Been Tested?  

- See the description for general information. 
- The test pipeline (which runs on PRs, including this one) has been updated to additionally test the FIPS-capable AMD64 Linux build in two distinct scenarios: with FIPS enabled and with FIPS disabled. That pipeline has been manually run on this branch several times, and although there are still issues specific to the agent being used to test with FIPS enabled, some tests pass, and the same tests ultimately fail each time. For example, see the test results for https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=30978&view=results. 
  - One unit test, the TestInferContentType test in the ste directory, fails when run in the pipeline on Azure Linux 3.0, regardless of whether it's testing the FIPS build with FIPS enabled or disabled (see https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=30965&view=results - you can view the YAML file that pipeline used to see "FIPS-enabled" jobs in that run were actually running with FIPS disabled); it even fails in the pipeline on Azure Linux 3.0 when testing the non-FIPS build (see https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=30969&view=results - again, see the YAML). 
    - That test passes when run from an Azure Linux 3.0 VM on a FIPS build with FIPS enabled, albeit without the "workload identity" the pipeline uses when running it. 
  - Three "Old E2E" tests, TestManagedDisks_AAA, TestManagedDisks_NoOAuthRequired, and TestPOSIX_UploadAMLFSStyle (Copy-BlobLocal and Copy-BlobBlob - variants "Anonymous-Anonymous" and "OAuthToken-Anonymous"), fail when run in the pipeline on Azure Linux 3.0, regardless of whether they're testing the FIPS build with FIPS enabled or disabled (see https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=30965&view=results - you can view the YAML file that pipeline used to see "FIPS-enabled" jobs in that run were actually running with FIPS disabled); they even fail in the pipeline when testing the non-FIPS build (see https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=30969&view=results - again, see the YAML). 
    - TestPOSIX_UploadAMLFSStyle/Copy-BlobBlob/Anonymous-Anonymous and TestPOSIX_UploadAMLFSStyle/Copy-BlobLocal/Anonymous-Anonymous pass when run from an Azure Linux 3.0 VM on a FIPS build with FIPS enabled, albeit without the "workload identity" the pipeline uses when running them. 
    - I wasn't able to get the other failing tests to pass when run from an Azure Linux 3.0 VM on a FIPS build with FIPS enabled, but I probably have something misconfigured locally; I'm also trying to run them without workload identity (which the pipeline configures). 
  - The "New E2E" tests fail to run in the pipeline on Azure Linux 3.0, regardless of whether they're testing the FIPS build with FIPS enabled or disabled (see https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=30965&view=results - you can view the YAML file that pipeline used to see "FIPS-enabled" jobs in that run were actually running with FIPS disabled); they even fail in the pipeline when testing the non-FIPS build (see https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=30969&view=results - again, see the YAML). The set-up hook "ARM client" fails after trying to using an access token using the wrong tenant. 
    - By temporarily modifiying the pipeline to run these tests without workload identity, I was able to get at least 425 of theses 1094 tests to pass (see https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=31017&view=results). An additional 53 tests passed when run from an Azure Linux 3.0 VM on a FIPS build with FIPS enabled, likewise without the "workload identity" the pipeline uses when running them. Those are jUnit report totals, which count each test that contains sub-tests, scenarios, or variants at lease one extra time.


azurePipelineTemplates/run-ut-and-py.yml
```diff
@@ -0,0 +1,218 @@
...
+          directories:
+          - 'ste'
+          isMutexSet: variables['isMutexSet']
+          fips: $(fips)
+          display_name: $(display_name)
+          with_federated_credential: true
+
+      - script: |
+          baseReportName='smoke_coverage'
+          # It's unclear what the effect of setting CGO_ENABLED to an empty string would be,
```
adreed-msft, line 127: did a little digging, it does *nothing* when set to any value other than `1` or `0`:

https://cs.opensource.google/go/go/+/master:src/cmd/go/internal/cfg/cfg.go;l=181?q=cgo_enabled&ss=go%2Fgo

adreed-msft (approved):

al-msft:
I often need to run AzCopy's test jobs several times to get them to pass. Once we have a working Azure Linux 3.0 image to run the FIPS-enabled tests on, I'm planning to schedule a meeting to discuss the changes in this PR and propose that we consider:

1. omitting the (or reducing the number of) tests of the FIPS build of AzCopy on a FIPS-disabled host, and
2. splitting the "unit" tests and Python tests into separate jobs that can be individually re-run after failing, with one job per directory we unit test and another job for the Python tests. This will reduce the chance of unit or Python tests that passed in one attempt from failing in a subsequent attempt and thus requiring yet another re-run of the job, even though all tests that initially failed may have passed in later attempts.

azurePipelineTemplates/run-e2e.yml
```diff
@@ -1,56 +1,111 @@
...
+        displayName: 'Install Microsoft build of Go'
+        # Don't run if job is canceled
+        condition: and(succeededOrFailed(), eq(variables.type, 'linux-fips'))
+
+      - task: GoTool@0
+        inputs:
+          version: $(AZCOPY_GOLANG_VERSION_COVERAGE)
+        displayName: 'Install Go'
+        # Don't run if job is canceled
+        condition: and(succeededOrFailed(), ne(variables.type, 'linux-fips'))
```
al-msft, line 85: 'OrFailed' should probably be removed from this task and the one above.

al-msft:
This PR will be split into multiple smaller PRs, starting with https://github.com/Azure/azure-storage-azcopy/pull/3504.

azurePipelineTemplates/run-ut-and-py.yml
al-msft, line 1: I'm hoping to avoid introducing this file in my upcoming PRs. This file adds a layer that makes it harder to glance at the repo and get a general sense of the pipeline structure. I've found it best to aim for small-to-medium-sized top-level pipeline files that provide the outline of the pipeline and all the simple logic, and split out granules of more-complex logic or of unwieldy size into small-to-medium-sized bottom-level files, with no intermediate files in between.

azurePipelineTemplates/run-ut-and-py.yml
```diff
@@ -0,0 +1,218 @@
...
+      strategy:
+        matrix:
+          Linux-FIPS-enabled:
+            # Azure Linux 3.0 enables FIPS by default.
+            imageName: 'azurelinux3'
+            build_name: 'azcopy_linux_fips_amd64'
+            type: 'linux-fips'
+            display_name: 'Linux-FIPS-enabled'
+            fips: 'enabled'
+      pool:
```
al-msft, line 40: Conditionally inserting one of the two pool sections based on the matrix entry seems like a better approach than creating this template file just so I can call it twice from azure-pipelines.yml. I think that didn't work the first time I tried; it's worth another try, now that I know the pipeline better.

PR #3502, opened 2026-07-10
Dedupe Prototype to c2c-stage

ProgramPedro:
## Description

**This draft PR is for code review only and is not intended to be merged as-is.**

- Retrieves per-block CRC64 and SHA256 hashes
- Aligns transfer chunks with source block boundaries
- Reuses matching blocks from previously committed destination blobs
- Uses ETag validation and falls back to the original source on failure
- Tracks job-level dedupe hits and estimated source-read savings

This prototype depends on an unreleased `azblob` SDK extension. The machine-local SDK replacement is intentionally not included.

### DIFF
--- a/common/dedupeHashTable.go
+++ b/common/dedupeHashTable.go
@@ -1,0 +1,277 @@
+// Copyright © Microsoft <wastore@microsoft.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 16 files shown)


ProgramPedro:
Superseded by draft PR #3535, which contains the current Mover-compatible implementation. #3535 is for code review only and is not intended to be merged as-is.

PR #3506, opened 2026-07-14
Fix singleChunkReader to handle concurrent Close during read and add related tests

shnayak-msft:
## Description

Fix a bug in singleChunkReader.go where a `Close()` triggered by the HTTP transport (in response to a peer-side connection close e.g. an AFD 240 s idle recycle, or any front-door / proxy that terminates long-lived HTTPS connections mid-request) permanently poisoned the reader. The azcore retry policy would then call `RewindBody` + `Read` on the same reader; the disk re-read succeeded, but the post-read check `if cr.isClosed` fired and returned `errors.New("closed while reading")` on **every** retry. Result: the retry budget was exhausted without the chunk payload ever being retransmitted, and the transfer failed.

Root cause: `isClosed` was never cleared, so a `Close()` from a previous attempt (already handled and unwound by the transport) was indistinguishable from a `Close()` racing with the current `ReadAt`.

Fix, in singleChunkReader.go `blockingPrefetch`:
- Snapshot `wasClosedBefore := cr.isClosed` before releasing `muClose` for `ReadAt`.
- After `ReadAt`, only report `"closed while reading"` when `cr.isClosed && !wasClosedBefore` (a `Close()` that arrived during *this* read).
- On successful re-read of a previously-closed reader, reset `cr.isClosed = false` so any later concurrent `Close()` is still detected correctly.

Semantics preserved: a `Close()` racing with an in-flight `ReadAt` still aborts that read; terminal `Close()` after all retries still frees the pooled buffer and clears `chunkLogger`; lock scoping around `muClose` / `muMaster` is unchanged.

- **Feature / Bug Fix**: Bug fix — `singleChunkReader` now allows the azcore retry policy to successfully re-drive an upload chunk after the HTTP transport closes the request body due to a peer-side connection close. Prior to this fix, every retry failed with `"closed while reading"` and the chunk payload was never actually re-sent (nginx repro: 528 failed `PUT`s carried headers only, 336–494 bytes, zero payload).

- **Related Links**:
- Issues
- Team thread
- Documents
- [Email Subject]

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

**Unit tests** added in zt_singleChunkReader_test.go:

- `TestSingleChunkReader_ReadAfterCloseSucceedsOnRetry` — reads the chunk fully, calls `Close()` (mimicking the HTTP transport reacting to a server-side connection close), then `Seek(0)` + `Read` again (mimicking the azcore retry policy's `RewindBody`). Asserts the second read returns identical bytes. Fails without the fix with `"closed while reading"`; passes with it.
- `TestSingleChunkReader_ConcurrentCloseDuringReadStillFails` — uses a `blockingReaderAt` to hold `ReadAt` open, then calls `Close()` from another goroutine. Asserts the read still fails with `"closed while reading"`, guarding the original safety invariant.

**Wire-level validation** (independent repro): AzCopy 10.32.4 uploading a 400 MiB file with 1 MiB blocks through an nginx proxy configured to close idle HTTPS connections, in front of `azurefd-test.net`. Without the fix, the AzCopy log contained 546 `REQUEST ERROR ... closed while reading` lines and the job stalled at 72.2 % (289 / 400 blocks). The nginx access log confirmed 528 failed `PUT`s carrying only 336–494 bytes (headers, no payload) versus 289 successful `PUT`s carrying the full ~1,048,934-byte body. With the fix applied, retries actually retransmit the chunk and the upload completes end-to-end.


common/zt_singleChunkReader_test.go
al-msft, line 1: It's always great to see unit tests added to this repo.

common/zt_singleChunkReader_test.go
```diff
@@ -0,0 +1,162 @@
...
+	a.Equal(fileData, retryBuf)
+}
+
+// TestSingleChunkReader_ConcurrentCloseDuringReadStillFails verifies that the original
+// safety check for a Close() that races with an in-flight disk read is preserved after
+// the retry-reuse fix.
+func TestSingleChunkReader_ConcurrentCloseDuringReadStillFails(t *testing.T) {
+	a := assert.New(t)
+
+	const chunkLen int64 = 4 * 1024
```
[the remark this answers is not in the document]
al-msft: This test compiles and runs for me, so I don't know what Copilot is talking about. `make` accepts any IntegerType in my environment.

common/zt_singleChunkReader_test.go
```diff
@@ -0,0 +1,162 @@
...
+// the server closed the connection), the azcore retry policy calls RewindBody+Read on
+// the same singleChunkReader. Previously, a stale isClosed=true flag caused every
+// retry to fail with "closed while reading", exhausting the retry budget without
+// actually resending the chunk. After the fix, a Close() followed by Seek(0)+Read()
+// must successfully re-fetch the chunk data from disk.
+func TestSingleChunkReader_ReadAfterCloseSucceedsOnRetry(t *testing.T) {
+	a := assert.New(t)
+
+	// Prepare a synthetic 1 MiB chunk of file data.
+	const chunkLen int64 = 1024 * 1024
```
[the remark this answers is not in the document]
al-msft: This test compiles and runs for me, so I don't know what Copilot is talking about. `make` accepts any IntegerType in my environment.

common/zt_singleChunkReader_test.go
```diff
@@ -0,0 +1,162 @@
...
+	data    []byte
+	release chan struct{}
+}
+
+func (b *blockingReaderAt) ReadAt(p []byte, off int64) (int, error) {
+	b.release <- struct{}{} // signal that we've entered ReadAt
+	<-b.release             // block until told to proceed
+	n := copy(p, b.data[off:])
+	return n, nil
+}
```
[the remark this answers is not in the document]
al-msft: Regarding int64/off: This test compiles and runs for me, but it does look like using an int64 could be problematic if `off` is outside the range of values an int can accommodate.

common/singleChunkReader.go
```diff
@@ -262,6 +272,13 @@ func (cr *singleChunkReader) blockingPrefetch(fileReader io.ReaderAt, isRetry bo
 		return readErr
 	}
 
+	// The reader was previously Close()-d but is now being successfully re-used (retry after
+	// a server-closed connection).  Clear the closed flag so subsequent state (e.g. any later
+	// concurrent Close during a further read) is detected correctly.
+	if wasClosedBefore {
```
al-msft: What if the reader was closed before this read and was also closed during this read (without a readError)? Does this code handle that case correctly? Is that a realistic case?

common/singleChunkReader.go
```diff
@@ -248,7 +257,8 @@ func (cr *singleChunkReader) blockingPrefetch(fileReader io.ReaderAt, isRetry bo
 
 	// now that we have the lock again, see if any error means we can't continue
 	if readErr == nil {
-		if cr.isClosed {
+		if cr.isClosed && !wasClosedBefore {
```
adreed-msft: This feels a bit hacky. If readErr was nil, and we succeeded the read (completely, to cr.length), should we not just write the buffer anyway? There's no need to repeat. The next prefetch attempt comes along, sees the buffer is set, and returns happily.

Secondarily, if cr.isClosed was set, and then the reader got rewinded, the only place that cr.isClosed is checked is *here*-- is it not valid to wipe the flag on rewind?

adreed-msft: Furthermore. We probably shouldn't be getting closed here anyway. Why we're getting closed is *probably* something upstream. To assist in debugging that, could you make a change to `(*singleChunkReader).Close()`'s "early close" log to identify the source of an early close? 

This should be doable with [the runtime package](https://pkg.go.dev/runtime) to pull a stack trace and find out our caller. It would not inflate the logs too much to stick a file name and a line number in there.

adreed-msft: Finally, the behavior around closing is potentially incorrect. This does *not prevent any other action outside of prefetching* at this time. This is not in alignment with the normal semantics of an `io.ReadCloser` in golang. The exception for a rewind is probably fine, though.

adreed-msft (changes requested):

common/singleChunkReader.go
adreed-msft: Actually, big problem, this potentially allows for the leak of a buffer allocation *if* we permit the write, despite being closed here. If this occurs with frequency, AzCopy will deadlock.

shnayak-msft: Good catch. Fixed by dropping the wasClosedBefore snapshot entirely and instead resetting cr.isClosed = false while still holding muClose, immediately before the unlocked ReadAt. Any Close() that races this read now unambiguously sets isClosed, so the plain post-read if cr.isClosed check catches it even when the reader had been closed in a prior attempt. Added TestSingleChunkReader_ClosedBeforeAndConcurrentCloseDuringRetryDoesNotLeak for this case.

Close()'s early-close log now includes the caller via runtime.Caller, formatted as caller=<file:line> <- <file:line>file:line <- file:line. Is this what you intended?

common/singleChunkReader.go
shnayak-msft: Yes, it's realistic. A transport Close() from a prior attempt followed by a second transport Close() racing the retry read (both from peer-side connection resets). The original wasClosedBefore-diff logic handled this incorrectly (the concurrent close was masked). The current code resets isClosed per attempt before the read, so this case now aborts correctly and returns the buffer. New test TestSingleChunkReader_ClosedBeforeAndConcurrentCloseDuringRetryDoesNotLeak pins the behavior, including a cacheLimiter leak assertion.

common/zt_singleChunkReader_test.go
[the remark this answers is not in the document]
shnayak-msft: This is a non-issue. Thanks for checking @al-msft

common/zt_singleChunkReader_test.go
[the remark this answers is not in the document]
shnayak-msft: Leaving it as is for now.

al-msft (approved):

common/singleChunkReader.go
```diff
@@ -240,6 +253,16 @@ func (cr *singleChunkReader) blockingPrefetch(fileReader io.ReaderAt, isRetry bo
...
 
+	// Clear any "closed" state left over from a previous attempt before starting THIS read.
+	// isClosed exists solely to detect a Close() that races with the unlocked ReadAt below
+	// (see the post-read check). A Close() from a previous attempt (e.g. the HTTP transport
+	// reacting to a server-side connection reset such as an AFD 240s idle recycle) must NOT
+	// prevent the azcore retry policy from re-driving this reader via RewindBody+Read; otherwise
+	// every retry fails with "closed while reading" and exhausts the retry budget without ever
+	// retransmitting the chunk. We must clear it while still holding muClose so that any Close()
+	// arriving during the unlocked ReadAt window is unambiguously observed as a concurrent close.
+	cr.isClosed = false
```
adreed-msft, line 264: My thinking was more that we could reset when azcore rewinds this reader-- AzCore's retry policy calls [request.RewindBody](https://github.com/Azure/azure-sdk-for-go/blob/1582e16957de03de00b3c7cc50e522dafa9ff164/sdk/azcore/internal/exported/request.go#L172) on [every retry](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/policy_retry.go#L132).

If we clear isClosed on that condition, it should be ready for the next read attempt, and then should wind up closing naturally.

adreed-msft: Now, I'm not sure if AzCore wraps our reader in some sort of retry handler itself. If it does, relying upon the rewind may be incorrect behavior.

adreed-msft (approved):

al-msft (approved):

e2etest/zt_newe2e_single_chunk_reader_test.go
```diff
@@ -0,0 +1,264 @@
...
+//  4. azcore retries the original StageBlock request. The fixed reader clears closed state
+//     left by the prior attempt before re-reading the source, while still detecting a Close
+//     that races with the current read. Before the fix, this retry failed locally with
+//     "closed while reading" and sent no block data.
+//  5. The proxy captures the retry body and requires all 8 MiB before forwarding it to a
+//     fake Storage origin. The test also requires all 50 blocks and the final block-list
+//     commit, proving that AzCopy recovered from the mid-body reset and completed the upload.
+//
+// The origin and proxy are entirely local; no production fault hook or Storage account is used.
+func TestSingleChunkReader_RetriesAfterProxyClosesReplayedBody(t *testing.T) {
```
al-msft: I'm not sure this file uses the correct naming convention and is located in the correct directory. New E2E tests typically declare the parameter `svm *ScenarioVariationManager`, rather than `t *testing.T`. I've only seen the latter in unit tests and Old E2E tests.

al-msft: I suggest removing "_newe2e" from the file name, drawing on the example set by `zt_version_test.go`, which is also in the `e2etest` directory, declares the parameter `t *testing.T`, and invokes `exec.Command` with `GlobalInputManager{}.GetExecutablePath()`.

This new test also seems to run as part of the "Old_E2E" job: https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=31590&view=logs&j=e3f09d63-5818-5b24-5cd4-d8f699ae5319&t=0cc9ab0e-0405-52fd-c289-75314eef7222&l=2848.

e2etest/zt_newe2e_single_chunk_reader_test.go
```diff
@@ -214,7 +220,7 @@ func TestSingleChunkReader_RetriesAfterProxyClosesReplayedBody(t *testing.T) {
 	defer proxy.Close()
 
 	destination := proxy.URL + "/container/destination.bin?sv=2021-12-02&sr=b&sp=rw&se=2099-01-01&sig=fake"
-	ctx, cancel := context.WithTimeout(context.Background(), 45*time.Second)
+	ctx, cancel := context.WithTimeout(svm.Context(), 45*time.Second)
 	defer cancel()
 	command := exec.CommandContext(
```
al-msft, line 225: I've never seen `exec.CommandContext` used in a New E2E test. @adreed-msft may know whether that has any downsides.

al-msft (approved):

adreed-msft (approved):

e2etest/zt_newe2e_single_chunk_reader_test.go
al-msft: @adreed-msft approved this PR, so I'm resolving this conversation to prevent it from delaying/blocking the merge

e2etest/zt_newe2e_single_chunk_reader_test.go
al-msft: The latest commit addressed this.

PR #3520, opened 2026-07-20
Increasing Code Coverage

al-msft:
## Commits
**Removed unused code.**

This code was orphaned by commit 4aca52c7c2cf0d0fb3a5d05eb9ec4989a04c7f6f.

**My first AzCopy unit test!**

## Description

I'm submitting this first unit test for review now, so I can receive general feedback on my approach to unit testing BEFORE I use it to write a large number of tests.

**Related Links**:
- [This other PR](https://github.com/Azure/azure-storage-azcopy/pull/3511) will allow this test to run in the pipeline.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

1. The new unit test passes locally.
2. Although the new test from the second commit doesn't run in the pipeline (see the "Related Links", above), the following pipeline run supports my belief that the code removed in the first commit was truly unused: https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=31354&view=results

### DIFF
--- a/azcopy/pathUtils_test.go
+++ b/azcopy/pathUtils_test.go
@@ -1,0 +1,58 @@
+package azcopy
+
+import (
+	"testing"
+)
+
+func TestStartsWith(t *testing.T) {
+	scenarios := []struct {
+		name            string
+		s               string
+		potentialPrefix string
+		expected        bool
+	}{
+		// The method we're testing doesn't trim either string, so there's no need to test whitespace.
+
+		// Empty-string Handling:
+		{"Empty String: Empty String and Non-Empty Potential Prefix", "", "non-empty", false},
+		{"Empty String: Empty String and Empty Potential Prefix", "", "", true},
+		{"Empty String: Non-Empty String and Empty Potential Prefix", "non-empty", "", true},
+
+		// Simple Strings to Test the Basic Logic:
+		{"Simple Strings: Longer Potential Prefix", "n", "nn", false},
+		{"Simple Strings: Equal-Length Potential Prefix That Doesn't Match", "n", "o", false},
+		{"Simple Strings: Equal-Length Potential Prefix That Matches (Same Case)", "n", "n", true},
+		{"Simple Strings: Equal-Length Potential Prefix That Matches (Lower vs. Upper Case)", "n", "N", true},
+		{"Simple Strings: Equal-Length Potential Prefix That Matches (Upper vs. Lower Case)", "N", "n", true},
+		{"Simple Strings: Shorter Potential Prefix That Doesn't Match", "nn", "o", false},
+		{"Simple Strings: Shorter Potential Prefix That Matches (Same Case)", "nn", "n", true},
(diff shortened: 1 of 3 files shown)


azcopy/pathUtils_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	for _, scenario := range tests {
+		t.Run(scenario.name, func(t *testing.T) {
+			actual := StartsWith(scenario.s, scenario.potentialPrefix)
+			if actual != scenario.expected {
+				// We're not using `assert` here, because `assert` doesn't handle scenarios properly.
+				// With assert, when a scenario fails, `go test` labels the test `FAIL` and prints the error,
+				// BUT it labels all scenarios `PASS`.
+				t.Error()
+			}
```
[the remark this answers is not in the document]
al-msft: Too slow, Copilot! My latest commit added the error message, and there was/is no misleading comment about `assert`.

azcopy/pathUtils_test.go
```diff
@@ -0,0 +1,58 @@
...
+		// NOTE: English has ligatures like "æ", "œ", and their upper-case versions, which have digraphs like "ae", "oe", etc.
+	}
+
+	for _, scenario := range scenarios {
+		t.Run(scenario.name, func(t *testing.T) {
+			actual := StartsWith(scenario.s, scenario.potentialPrefix)
+			if actual != scenario.expected {
+				// We're not using `assert` here, because `assert` doesn't handle scenarios properly.
+				// With assert, when a scenario fails, `go test` labels the test `FAIL` and prints the error,
+				// BUT it labels all scenarios `PASS`.
```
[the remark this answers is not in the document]
al-msft: WRONG!

azcopy/pathUtils_test.go
```diff
@@ -0,0 +1,58 @@
+package azcopy
+
+import (
+	"testing"
+)
+
+func TestStartsWith(t *testing.T) {
```
adreed-msft, line 7: Extensive!

adreed-msft (approved):

azcopy/pathUtils_test.go
adreed-msft: I'm quite surprised that this isn't just. A feature of Golang's `strings` library. But I feel like we're kinda drifting into testing Go's `strings.EqualFold`, which I'd bet you they already tested pretty extensively...

al-msft: You'll see as I add more unit tests that I have a hard time deciding how much I can get away with not testing when I can't just mock away all the details, so in cases like this, I tend towards testing more than other people would have tested. I have a history of using mocks extensively and validating them strictly, which has pros and cons.

That kind of feedback is what I was after with this PR, so don't hold back, and don't hesitate to reject this PR if you have any concerns at all about any aspect of the approach I took to testing this. I expect to work with Copilot to write many more tests in many more PRs based on the approach and feedback from this PR and the ones that follow.

gapra-msft (approved):

PR #3310, opened 2025-12-01
Migrated copy code

gapra-msft:
## Description

- **Feature / Bug Fix**: (Brief description of the feature or issue being addressed)

- **Related Links**:
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update required
- [ ] Code quality improvement
- [ ] Other (describe):

## How Has This Been Tested?

Thank you for your contribution to AzCopy!

PR #3531, opened 2026-07-29
Fix SIGTERM cancellation during enumeration

ankurmicrosoft309:
## Description

Fixes a cancellation hang when AzCopy receives SIGTERM while copy enumeration is still in progress.

The copy migration in #3310 moved signal handling to context cancellation. The lifecycle manager then invoked the CLI prompt after the context was already canceled. In non-interactive executions such as `timeout`, that prompt could wait indefinitely for stdin, preventing `cancelJob` from running.

#3373 repaired stdin cancellation delivery and repeated context handling, but cancellation during incomplete enumeration could still block before reaching those fixes.

This change emits a non-resumability warning through `JobUIHooks.Warn` and proceeds with cancellation without prompting. Context cancellation is irreversible, so prompting at this point cannot safely resume the operation.

## Tests

- Added a focused lifecycle test with a deliberately blocking prompt hook and explicit channel synchronization for warning capture.
- Added a deterministic process-level E2E test that pauses immediately before enumeration, sends SIGTERM, and fails if AzCopy does not exit within 10 seconds.
- The E2E test verifies the incomplete-enumeration warning so it cannot pass after enumeration completes.
- Added output-marker coverage for normal, split-write, and `io.Copy` streaming paths.
- The signal scenario runs on Linux/macOS and skips on Windows, where `os.Process.Signal(SIGTERM)` is unsupported.

Validation:

- Windows: `go build ./...`
- Windows: focused lifecycle and output-marker tests
- Windows: `go test ./cmd -run '^$'`
- WSL Ubuntu with local Azurite: process-level SIGTERM E2E passed

### DIFF
--- a/azcopy/copy.go
+++ b/azcopy/copy.go
@@ -206,6 +206,7 @@ func (c *Client) Copy(ctx context.Context, src, dest string, opts CopyOptions) (
 		if !t.opts.dryrun {
 			common.GetLifecycleMgr().Info("Scanning...")
 			mgr.InitiateProgressReporting(ctx, t.tpt)
+			common.GetLifecycleMgr().E2EAwaitEnumerationStart(ctx)
 		}
 		err = enumerator.Enumerate()
 		defer jobsAdmin.JobsAdmin.JobMgrCleanUp(jobID)
--- a/azcopy/jobLifecycleManager_test.go
+++ b/azcopy/jobLifecycleManager_test.go
@@ -1,0 +1,67 @@
+package azcopy
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/Azure/azure-storage-azcopy/v10/common"
+	"github.com/stretchr/testify/require"
+)
+
+type incompleteEnumerationProgressTracker struct{}
+
+func (incompleteEnumerationProgressTracker) Start() {}
+
+func (incompleteEnumerationProgressTracker) CheckProgress() (uint32, bool) {
+	return 0, false
+}
+
+func (incompleteEnumerationProgressTracker) CompletedEnumeration() bool {
+	return false
+}
+
+func (incompleteEnumerationProgressTracker) GetJobID() common.JobID {
+	return common.JobID{}
+}
+
+func (incompleteEnumerationProgressTracker) GetElapsedTime() time.Duration {
+	return 0
+}
+
+func TestContextCancellationDoesNotPromptDuringEnumeration(t *testing.T) {
+	promptCalled := make(chan struct{}, 1)
+	releasePrompt := make(chan struct{})
+	defer close(releasePrompt)
+
+	warningMessages := make(chan string, 1)
+	hooks := common.NewJobUIHooks()
+	hooks.Warn = func(message string) {
+		warningMessages <- message
+	}
+	hooks.Prompt = func(string, common.PromptDetails) common.ResponseOption {
+		promptCalled <- struct{}{}
+		<-releasePrompt
+		return common.EResponseOption.Yes()
+	}
+
+	manager := NewJobLifecycleManager(hooks)
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	manager.InitiateProgressReporting(ctx, incompleteEnumerationProgressTracker{})
+
+	waitResult := make(chan error, 1)
+	go func() {
+		waitResult <- manager.Wait()
+	}()
+
+	select {
+	case <-promptCalled:
+		t.Fatal("context cancellation prompted for user input")
+	case err := <-waitResult:
+		require.ErrorContains(t, err, "cancel job requires the JobID")
+		require.Equal(t, "The enumeration is not complete; cancelling the job at this point means it cannot be resumed.", <-warningMessages)
+	case <-time.After(time.Second):
+		t.Fatal("context cancellation did not reach job cancellation")
+	}
+}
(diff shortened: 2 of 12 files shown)


azcopy/jobLifecycleManager_test.go
al-msft, line 1: It's great to see more unit tests. Be aware: This test won't run in the pipeline until https://github.com/Azure/azure-storage-azcopy/pull/3511 has been merged.

ankurmicrosoft309: Thanks for the heads-up. I validated the ran the unit test locally on Windows and WSL. This PR also includes the process-level scenario in the existing Old E2E suite, so we have coverage there as well.

cmd/root.go
```diff
@@ -334,6 +336,8 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&azcopyAwaitAllowOpenFiles, "await-open", false,
 		"Used when debugging, to tell AzCopy to await `open` on stdin, after scanning but before opening the first file. "+
 			"\n Assists with testing cases around file modifications between scanning and usage")
+	rootCmd.PersistentFlags().BoolVar(&azcopyAwaitEnumerationStart, "await-enumeration", false,
+		"Used by E2E tests to pause immediately before enumeration until the command is cancelled.")
```
ankurmicrosoft309, line 340: @al-msft @adreed-msft Would like your opinions on introducing a flag like this in product code, which only serves a purpose in E2E test

azcopy/copy.go
```diff
@@ -206,6 +206,7 @@ func (c *Client) Copy(ctx context.Context, src, dest string, opts CopyOptions) (
 		if !t.opts.dryrun {
 			common.GetLifecycleMgr().Info("Scanning...")
 			mgr.InitiateProgressReporting(ctx, t.tpt)
+			common.GetLifecycleMgr().E2EAwaitEnumerationStart(ctx)
 		}
 		err = enumerator.Enumerate()
 		defer jobsAdmin.JobsAdmin.JobMgrCleanUp(jobID)
```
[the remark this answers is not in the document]
al-msft: Out of scope for this PR?

cmd/root.go
al-msft: I don't like it, but judging by E2EAwaitAllowOpenFiles in cmd/uihooks.go and the urgency of this fix, it may be reasonable to add in this PR.

I've read the description you added here. Is this flag necessary for testing, or does it just prevent the tests from doing things that won't be validated when testing cancellation? If it only prevents unvalidated effort, I'd rather not add this flag, but I'll defer to the other reviewers if they disagree.

al-msft (approved):

cmd/root.go
ankurmicrosoft309: Its only added for the E2ETest. It doesnt impact real life usage of the azcopy. During testing, we want the azcopy to wait in enumeration stage. Without this, the enumeration stage completes before test has ran its assertions and we are not able to test the behaviour of cancellation in enumeration stage. The other option is to not have an E2ETest for this case.

azcopy/copy.go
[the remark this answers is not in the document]
ankurmicrosoft309: Agreed

cmd/root.go
al-msft: Since this is needed to test cancellation in an E2E test, and this fix is urgently needed, and the approach is precedented, adding this flag seems fine for this PR.

adreed-msft: E2E testing is fine; let's make sure it's hidden properly.

e2etest/runner.go
```diff
@@ -32,6 +32,9 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
```
adreed-msft, line 35: We should be doing this on the new e2e framework, not the old declarative one. This is fine for proving for the time being.

adreed-msft (approved):

ankurmicrosoft309:
> ankurmicrosoft309 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree

ankurmicrosoft309:
Already merged manually by @adreed-msft

al-msft:
> Already merged manually by adreed-msft

@ankurmicrosoft309, @adreed-msft: It looks like the main branch hasn't yet been updated to reflect [release 10.32.7](https://github.com/Azure/azure-storage-azcopy/pull/3542). The changes in this PR (#3531) aren't yet in main - nor are the [related changes to hidden flags](https://github.com/Azure/azure-storage-azcopy/pull/3542/changes/96a83216c1e34f291ee93fbaca614ed010c90b23).

PR #3517, opened 2026-07-17
testing PR checks [skip ci]

al-msft:
## Description

Testing [the "[skip ci]" feature](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#skipping-ci-for-individual-commits).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update required
- [ ] Code quality improvement
- [x] Other (describe): Pipeline testing (**this PR is not intended to be merged**)

## How Has This Been Tested?

Included "[skip ci]" in a commit message. Looked at the effect on the checks listed on this PR.

PR #3539, opened 2026-08-04
testing `skip-checks: true`

al-msft:
skip-checks: true

## Description

testing ["skip-checks: true"](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs)

See also: https://docs.github.com/en/pull-requests/reference/status-checks#skipping-and-requesting-checks-for-individual-commits

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Other (describe): testing GitHub (**this PR is not intended to be merged**)

## How Has This Been Tested?

Included "skip-checks: true" in a commit message (after two empty lines). Looked at the effect on the checks listed on this PR.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -296,7 +296,7 @@ jobs:
         # Don't run if job is canceled
         condition: and(succeededOrFailed(), eq(variables.GOOS, 'linux'))
 
-      - task: PowerShell@2
+      - task: PowerShell@2 
         displayName: 'Install Azure CLI'
         inputs:
           targetType: 'inline'


al-msft:
Testing complete

The jobs in the pipeline (other than the Additional_E2E_Tests jobs - why?) are still required to pass, even though the top-level check (for the pipeline run) was skipped. I suggest we update the project settings so only the top-level check is required. It seems to me that the top-level check will only pass if all the jobs in that pipeline run pass. If so, removing the hard-coded job names from the check will provide flexibility for refactoring and may get ⁠[the "[skip ci]"](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#skipping-ci-for-individual-commits) and skip-checks features working.

Related PR: https://github.com/Azure/azure-storage-azcopy/pull/3517

Our "branch protection" settings are likely to blame. See the 7th point under https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule#creating-a-branch-protection-rule.

PR #3540, opened 2026-08-05
Increased Pipeline Responsiveness

al-msft:
An earlier commit stopped several conditional tasks from running after their job was canceled; this commit goes further by stopping them from running after job failure, where appropriate. This commit also corrects a copy+paste mistake that caused conditional tasks in the merged "Additional_E2E_Tests" job to run after job failure.

[Update: The second commit:] **Corrected a pipeline task name.**

The "Golang Vet - Linux" task of the "Build" jobs runs on all OSs. It seems useful for it to do so, so the name was changed to "Golang Vet", rather than limiting the task to run only on Linux.

## Description

The commit title and detail message (which are automatically used as the PR title and inserted above this section) say it all.

**Related Links**
- [Pipeline Conditions](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops#conditions-a-stage-job-or-step-runs-under)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Other (describe): Pipeline optimization

## How Has This Been Tested?

Before the second commit renamed an irrelevant task noticed in the test runs below:
- https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=31491&view=results - A smoke test. The tests that failed in jobs this PR didn't update are irrelevant and can be ignored; the point is that the pipeline still ran the expected jobs and tasks and the jobs and tasks this PR updated ran successfully.
- https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=31493&view=results - A derivative branch that injects a task into the affected jobs that fails at the start; this branch only runs the jobs this PR updated.

After the second commit:
- TODO when running the pipeline to satisfy checks after double-approval

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,7 +38,7 @@ jobs:
           echo 'Running GO Vet'
           go vet 
         workingDirectory: $(System.DefaultWorkingDirectory)
-        displayName: 'Golang Vet - Linux'
+        displayName: 'Golang Vet'
 
       - script: |
           sudo apt-get update
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,14 +44,14 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libsecret-1-dev libglib2.0-dev
         displayName: 'Install dependencies'
-        # Don't run if job is canceled
-        condition: and(succeededOrFailed(), eq(variables.type, 'linux'))
+        # Don't run if job is canceled or fails
+        condition: and(succeeded(), eq(variables.type, 'linux'))
 
       - script: |
           go run "common/manifest/main.go"
         displayName: 'Create resource.syso files'
-        # Don't run if job is canceled
-        condition: and(succeededOrFailed(), eq(variables.type, 'windows'))
+        # Don't run if job is canceled or fails
+        condition: and(succeeded(), eq(variables.type, 'windows'))
         
       - script: |
           go build -tags "netgo" -o "$(Build.ArtifactStagingDirectory)/azcopy_linux_amd64"
@@ -59,8 +59,8 @@ jobs:
           GOARCH: amd64
           GOOS: linux
         displayName: 'Generate Linux AMD64'
-        # Don't run if job is canceled
-        condition: and(succeededOrFailed(), eq(variables.type, 'linux'))
+        # Don't run if job is canceled or fails
+        condition: and(succeeded(), eq(variables.type, 'linux'))
 
       - script: |
           go build -tags "netgo,se_integration" -o "$(Build.ArtifactStagingDirectory)/azcopy_linux_se_amd64"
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
al-msft, line 1: If you're not familiar with pipeline conditions, see the link in the PR description. It's common to add a condition to a task and overlook the fact that you're overriding the default behavior by not "anding" the condition you're interested in with "succeeded()".

From [the link](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops#conditions-a-stage-job-or-step-runs-under) in the PR description:
"*By default, a pipeline job or stage runs if it doesn't depend on any other job or stage, or if all its dependencies completed and [succeeded](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#succeeded). The dependency requirement applies to direct dependencies and to their indirect dependencies, computed recursively...When you specify a condition property for a stage, job, or step, you overwrite the default condition. Your stage, job, or step might run even if the build is canceled. Make sure your conditions take the state of the parent stage or job into account.*"

ankurmicrosoft309: Oh interesting. So if my understanding is correct, the only reason we need to add these conditions is because we want to run the scripts on individual platforms. If these scripts were required to be run on all platforms, not specifiying any condition would be same as specifiying `condition: succeeded()`. Right?

ankurmicrosoft309 (approved):
Happy with this change. We can merge it. I have a suggestion for a future PR.

I am not sure if this is possible or worth the effort, but I would have liked to refactor this file into something like:

```
azure-pipelines.yml
azurePipelineTemplates/
  build-linux-steps.yml
  build-windows-steps.yml
  build-macos-steps.yml
```

And then import the steps yml files into the main file. Maybe we wouldnt need the `condition` in this case? What do you think?

azure-pipelines.yml
al-msft: @ankurmicrosoft309: That's right.

al-msft:
> Happy with this change. We can merge it. I have a suggestion for a future PR.
> 
> I am not sure if this is possible or worth the effort, but I would have liked to refactor this file into something like:
> 
> ```
> azure-pipelines.yml
> azurePipelineTemplates/
>   build-linux-steps.yml
>   build-windows-steps.yml
>   build-macos-steps.yml
> ```
> 
> And then import the steps yml files into the main file. Maybe we wouldnt need the `condition` in this case? What do you think?

We could definitely avoid the platform-specific conditions if we replaced the current job that's used to build on all platforms with one dedicated build job per platform. Looking at the tasks for the current job, I can see some advantages to that refactoring. We'd need to be careful to give the new files names that clearly communicate which pipeline they apply to, because this repo declares multiple pipelines, one of which already took the approach you're suggesting and used names that are similar to those you've listed above (see https://github.com/Azure/azure-storage-azcopy/tree/main/azurePipelineTemplates). Regardless, I've been hoping to move and potentially rename some of the existing pipeline files to clearly convey which pipeline they apply to (but I have higher priorities right now).

PR #3544, opened 2026-08-06
Deduplicated some pipeline configuration

al-msft:
## Commits

**Deduplicated some pipeline configuration**

Grouped the specification of non-federated unit-test directories using pipeline expressions. Also renamed a nearby variable to clarify its specific use.

**Deduplicated a pipeline script**

Extracted a duplicated script into a template parameter (variables aren't supported in this context). Also made some minor formatting updates to nearby strings.

## Description

This is a continuation of https://github.com/Azure/azure-storage-azcopy/pull/3511.

A later PR will dynamically determine the list of non-federated directories to test, so tests in new sub-directories are automatically tested in the pipeline. For example, a script using the following command could identify the non-federated directories to test, and then they could be stored as an array in a runtime parameter. As a variation on the script below, after pruning directories we don't want to test, the script could return all directories containing files that end with "_test.go".
```
find . -type d \( \
-path "./.*" -prune -o \
-path "./azurePipelineTemplates" -prune -o \
-path "./common/manifest" -prune -o \
-path "./docker" -prune -o \
-path "./e2etest" -prune -o \
-path "./setup" -prune -o \
-path "./ste" -prune -o \
-path "./testSuite" -prune -o \
-path "./website" -prune \
\) -o \
-type d -path "*" -print | sort
```

**Related Links**:
- [Azure Pipelines' "replace" function for expressions](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#replace)
- [Azure Pipelines' "split" function for expressions](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#split)
- [All pipeline variables (not parameters) are strings (which is why the variables aren't defined as arrays)](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch)
- [steps.bash definition](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-bash)
- [steps.script definition](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-script)

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=31530&view=results

### DIFF
--- a/azurePipelineTemplates/run-ut.yml
+++ b/azurePipelineTemplates/run-ut.yml
@@ -8,32 +8,34 @@ parameters:
     default: false
   - name: isMutexSet
     type: string
+  - name: test_script
+    type: string
+    default: |
+      set -e
+
+      # Run tests, write coverage data to a file, and pipe output to a file
+      echo "Running tests in ${directory}. Writing coverage data to ${coverageReportName}.txt and test results to ${testReportName}.txt."
+      go test -timeout=1h -v -coverprofile="${coverageReportName}.txt" "./${directory}" | tee "${testReportName}.txt"
+
+      exit_code=${PIPESTATUS[0]} # PIPESTATUS is defined by Bash, so this script must be run with Bash, not sh.
+
+      # Generate the jUnit report
+      echo "Generating jUnit report ${testReportName}.xml"
+      cat "${testReportName}.txt" | $(go env GOPATH)/bin/go-junit-report > "${testReportName}.xml"
+
+      # Convert coverage data to JSON format
+      echo "Generating JSON coverage report ${coverageReportName}.json"
+      $(go env GOPATH)/bin/gocov convert "${coverageReportName}.txt" > "${coverageReportName}.json"
+
+      # Convert coverage data to XML format
+      echo "Generating XML coverage report ${coverageReportName}.xml"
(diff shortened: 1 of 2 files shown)


azurePipelineTemplates/run-ut.yml
al-msft, line 1: GitHub shows a more-accurate diff if you enable the option to hide whitespace changes.

ankurmicrosoft309 (approved):
I cant comment on the specifics of the bash syntax, but everything else looks great. Thank you, this makes the pipeline much cleaner.

azurePipelineTemplates/run-ut.yml
al-msft, line 1: If the `- ${{ else }}:` on line 59 isn't initially visible, this change should be easier to make sense of after you expand that collapsed line. I've moved the test script above the `steps` line as a parameter, and updated the `- ${{ if eq('false', parameters.with_federated_credential) }}:` and `- ${{ else }}:` paths to reference the test script, rather than duplicating it.

PR #3545, opened 2026-08-07
[Mover Branch] Multitenancy support

adreed-msft:
## Description

- **Feature**: Multi-tenancy support via named credentials. Login, logout, etc. all support naming credentials to utilize for any given OAuth action.

Unlike the original spec, we are unable to automagically detect which token to use based upon challenges from a target endpoint-- This feature seems to be disabled on Microsoft's tenant, at least.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Extensive newe2e tests to prove the feature works in almost every permutation @ e2etest/zt_newe2e_multitenancy_test.go


common/cred/keyring_mem.go
```diff
@@ -0,0 +1,40 @@
+package cred
+
+type memKeyring struct {
+	identities map[string]Token
```
kylemoore-microsoft, line 4: As discussed over teams, when only destination identity is present in the map (as it should be for same tenant transfer) x-ms-copy-source-authorization is not set, resulting in trusted access failing for private source accounts

kylemoore-microsoft (changes requested):

common/cred/keyring_mem.go
adreed-msft: I suspect that this is mover-side more than AzCopy side.

kylemoore-microsoft (approved):

PR #3570, opened 2026-09-01
[Telemetry 01/22] Define bounded telemetry event contracts and correlation semantics

ankurmicrosoft309:
## Summary
Resource and job dimensions, command/start/finish event measurements, privacy bounds, resume counter semantics, and their model-level unit tests. No network sender or CLI activation.

This PR now contains layer 1 of the rewritten 22-layer telemetry review stack, replacing its earlier scope. Implementation and its relevant unit tests are included together.

## Planned Event Timing
Later PRs wire up these three events when telemetry is enabled, with best-effort delivery:
- `azcopy.job.started`: at the start of each eligible copy, sync, benchmark, or resume attempt, before enumeration or resumed transfer work.
- `azcopy.job.finished`: once an attempt reaches a terminal outcome, including success, failure, or cancellation, with its status and numeric observations.
- `azcopy.command.invoked`: during pre-run initialization of eligible commands that do not emit job-attempt events (for example, login, list, and jobs list); it records invocation, not successful completion.

Opt-out, dry runs, help/completion/documentation/environment commands, and benchmark cleanup are excluded. This PR defines the contract only; it does not enable sending.

## Review Order
This is the first PR in the stack and targets `main`.
Next existing PR: #3571.

The existing head branch name `telemetry/01-event-contracts` is retained to preserve this PR and its review history. Its current scope matches local branch `telemetry/01-event-contracts`; use this title and diff rather than the legacy branch suffix.

## Included Changes
- Removes HostNICSpeedMbps and HostNICSpeedBucket from the pre-publication resource contract because advertised interface capacity does not reliably represent bandwidth available to AzCopy, especially on cloud VMs.
- Removes the write-only JobFinishedEvent.StartTimestamp field and fixture assignment. Finished events retain EndTimestamp, and duration measurements remain separately calculated; the serialized schema is unchanged.
- Registers the telemetry package in azure-pipelines.yml using the existing run-ut.yml template, including test-result and coverage reporting. This six-line addition lands with the first telemetry tests and is inherited by later layers.
- Renames the internal MetricEvent.attributes() method to properties() and updates its implementations and model tests. Properties are string metadata; measurements remain numeric observations. Struct names and payload keys are unchanged.
- Defines all three events through their event names: job.started, job.finished, and command.invoked.
- Groups all numeric observations, including FailureErrorOtherCount, in JobFinishedEvent.Measurements (JobMeasurements); identity, status, dimensions, and error properties remain on the event.
- Removes the unused Count/StartedCount/FinishedCount/InvokedCount fields and all three redundant occurrence measurements. Start and command events have no numeric measurements; finishes have 49 observations before conditional omissions. Consumers identify events by name.
- Includes option-key isolation, shared event-name constants, one final property-bounding pass, documented application-level size budgets, and UTF-8/zero/negative/short-limit tests.
- Names the model-only test file `events_test.go`; reporter and transport tests arrive in the next layer.
- Keeps the telemetry model separate from ListJobSummaryResponse, with explicit mapping and mapping tests in layer 07.

- `azure-pipelines.yml`
- `telemetry/events.go`
- `telemetry/events_test.go`

## Publication Note
This is a pre-publication schema change. Dependent E2E/CLI assertions and dashboard queries are updated and tested in later local layers, but this push publishes only layers 01-08. Dashboard changes are not deployed by this publication. Existing review-thread resolutions and reviewer approval decisions are not modified.

## Validation
The pipeline YAML was parsed and structurally compared: exactly one telemetry unit-test entry was added, the shared test-result/coverage publishers are retained, and all other pipeline configuration is unchanged. No Azure pipeline run was queued or observed for this validation.

The following focused tests passed on Windows at this exact branch boundary with `AZCOPY_DISABLE_TELEMETRY=true`:

```text
go test ./telemetry -count=1 -timeout=3m -run '.'
```

23 passing test/subtest results; 0 gated helper skips. Whitespace checks passed. Platform-specific Linux/macOS execution and live Azure E2E were not rerun for this publication.

## Scope And Revision
- One incremental commit; no Markdown documentation, dashboards or historical benchmark artifacts.
- Runtime activation and CLI lifecycle wiring are later layers, not part of this PR.
- Head commit: `04d9869a8ba2feb0ff437af4bb8b0d17574c68ca`
- Parent commit: `35dc27add1e640506b1221172f6c0f3d317dff65`
- Incremental diff: 3 files changed, 1078 insertions(+)


al-msft:
This appears to be a subset of https://github.com/Azure/azure-storage-azcopy/pull/3571. @ankurmicrosoft309: Do you recommend people review this first, then review the other files in https://github.com/Azure/azure-storage-azcopy/pull/3571?

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+
+	// FailureErrorCodes is a compact, bounded histogram of the error codes seen
+	// across failed transfers, e.g. "403:5,500:2" (ordered by descending count).
+	// Empty when there were no failures. Contains no PII (only numeric codes).
+	FailureErrorCodes      string
+	FailureErrorOtherCount int64
+	PerformanceConstraint  string
+	PerformanceAdviceCodes []string
+
+	// Measurements (from ListJobSummaryResponse + ElapsedTime)
```
al-msft, line 218: @ankurmicrosoft309: As written, JobFinishedEvent has a large, mostly-flat namespace (number of field names). Do you see any value in grouping all of these remaining fields into a struct named something like "Measurements" and listing a field of that struct type on JobFinishedEvent, instead of listing all these fields directly alongside the other fields in JobFinishedEvent's namespace?

telemetry/events.go
al-msft, line 1: A disclaimer about my review: I'm new to AzCopy and haven't been involved in this telemetry effort, so I won't be able to review the structs in this file and say if there's another data point they can track. I'm also new to Go, so my review of the code will be based on general programming principles and my experience with other (mainly object-oriented) programming languages.

I guess I should also warn that because this is the first of about 11 PRs in this effort and seems to be fundamental to the effort, I'm likely to ask a lot of questions to be sure my evolving, conceptual understanding is correct.

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+
+type JobFinishedEvent struct {
+	Resource       ResourceAttributes
+	Dimensions     JobDimensions
+	JobID          string
+	InvocationID   string
+	StartTimestamp time.Time
+	EndTimestamp   time.Time
+
+	FinishedCount    int64  // monotonic counter increment, always 1
```
al-msft, line 204: Is `FinishedCount` a measurement that can be grouped below with the others? I see it listed in `measurements()`. I'm asking for my conceptual understanding, and also in relation to https://github.com/Azure/azure-storage-azcopy/pull/3570/changes#r3906668130.

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+	JobStatus        string // "Completed" | "CompletedWithErrors" | "Failed" | "Cancelled" | ...
+	TerminalStage    string // initialization | enumeration | transfer | completion | completed
+	JobErrorCategory string // authentication | authorization | throttling | timeout | network | local-io | conflict | not-found | service | azcopy | initialization | enumeration | transfer | completion | unknown
+	JobErrorCode     string // bounded stable code; never raw error text
+
+	// FailureErrorCodes is a compact, bounded histogram of the error codes seen
+	// across failed transfers, e.g. "403:5,500:2" (ordered by descending count).
+	// Empty when there were no failures. Contains no PII (only numeric codes).
+	FailureErrorCodes      string
+	FailureErrorOtherCount int64
```
al-msft, line 214: Is `FailureErrorOtherCount` another measurement? https://github.com/Azure/azure-storage-azcopy/pull/3570/changes#r3907104565

telemetry/events.go
al-msft, line 218: @ankurmicrosoft309: Listing many of the same measurements both here and in ListJobSummaryResponse introduces the problem of keeping the lists in sync. Do you have any thoughts on that? Do you expect it to be rare for items in either list to be added/changed over time?

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+
+type JobStartedEvent struct {
+	Resource   ResourceAttributes
+	Dimensions JobDimensions
+	// JobID correlates the original attempt and all resume attempts. A paired
+	// job.started and job.finished event also share an InvocationID.
+	JobID        string
+	InvocationID string
+	Timestamp    time.Time
+	StartedCount int64 // monotonic counter increment, always 1
```
al-msft, line 189: @ankurmicrosoft309: From this PR and #3571, and branch `ankurmicrosoft309/telemetry`, it looks like `always 1` in this comment and the comment on `FinishedCount` means exactly that: `StartedCount` and `FinishedCount` will always be `1`. In that case, what's the purpose of these two "count" fields; why declare the fields at all, and why send `1` in `telemetry.JobStartedEvent` and `telemetry.JobFinishedEvent` (which I see in your telemetry branch)? Does sending the `1` give the remote code that recevies the events more information than if the counts were not part of the payloads for those events - or is the count expected to be able to vary in the future?

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+	"strconv"
+	"strings"
+	"time"
+	"unicode/utf8"
+)
+
+// ---------------------------------------------------------------------------
+// Resource attributes - machine/system facts, constant for the process.
+// Attached to BOTH metric events.
+// ---------------------------------------------------------------------------
```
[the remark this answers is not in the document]
al-msft: Agreed.

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package telemetry defines the AzCopy telemetry event model and the reporter
+// used to send those events to Azure Monitor (Application Insights).
+//
+// Two events mirror what AzCopy emits per run:
+//
+//	Event 1 (azcopy.job.started)  - emitted at job start.
+//	Event 2 (azcopy.job.finished) - emitted at job completion, with measurements.
```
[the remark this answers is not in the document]
al-msft: I agree this is a discrepancy that needs clarifying.

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+
+// namedMetric is a single numeric measurement to be sent to a backend.
+type namedMetric struct {
+	Name  string
+	Value float64
+	Count int
+}
+
+// MetricEvent is implemented by every telemetry event the telemetry package can
+// process and send. A single Reporter can therefore handle either event type.
```
al-msft, line 277: The word `either` in `either event type` isn't clear to me. Should that be "any"?

This point of confusion seems related to https://github.com/Azure/azure-storage-azcopy/pull/3570/changes#r3902105118 and https://github.com/Azure/azure-storage-azcopy/pull/3570/changes#r3902105161.

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+	}
+	if len(o.EnvVarsSet) > 0 {
+		props["OptEnvVarsSet"] = truncateValue(strings.Join(o.EnvVarsSet, ","))
+	}
+	for key, value := range o.Values {
+		if value != "" {
+			props[key] = truncateValue(value)
+		}
+	}
+}
```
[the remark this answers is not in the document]
al-msft: I have the same concern.

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+	ServerBusyPct                   float64
+	NetworkErrorPct                 float64
+	PercentComplete                 float64 // PercentComplete (0-100); useful especially for cancelled jobs
+}
+
+// namedMetric is a single numeric measurement to be sent to a backend.
+type namedMetric struct {
+	Name  string
+	Value float64
+	Count int
```
al-msft, line 273: From this PR and #3571, and branch `ankurmicrosoft309/telemetry`, it looks like `Count` is always set to `1` and is never read. In that case, what's the purpose of that field; why declare the field at all?

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+
+type CommandInvokedEvent struct {
+	Resource ResourceAttributes
+	Command  string // "login" | "jobs.list" | "remove" | ...
+	Options  OptionAttributes
+	// JobID is present when the invocation has an AzCopy JobID.
+	JobID        string
+	InvocationID string
+	Timestamp    time.Time
+	InvokedCount int64 // monotonic counter increment, always 1
```
al-msft, line 408: https://github.com/Azure/azure-storage-azcopy/pull/3570/changes#r3907504462 also applies here

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+	EventName() string
+	// attributes returns the flattened resource + dimension properties.
+	attributes() map[string]string
+	// measurements returns the numeric data points to send.
+	measurements() []namedMetric
+	// timestamp returns the event time to stamp on the telemetry.
+	timestamp() time.Time
+}
+
+func (JobStartedEvent) EventName() string  { return "azcopy.job.started" }
```
al-msft, line 289: "azcopy.job.started", "azcopy.job.finished", and "azcopy.command.invoked" each appear twice in this file. I suggest extracting those string literals to string constants this file can reference in multiple places.

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+func (JobFinishedEvent) EventName() string { return "azcopy.job.finished" }
+
+func (e JobStartedEvent) timestamp() time.Time  { return e.Timestamp }
+func (e JobFinishedEvent) timestamp() time.Time { return e.EndTimestamp }
+
+func (e JobStartedEvent) attributes() map[string]string {
+	attrs := mergeProps(e.Resource.props(), e.Dimensions.props())
```
al-msft, line 296: I've noticed `truncateValue` isn't used for every value placed into `attrs`, such as the values from `Resource.props()`, `Dimension.props()`, and `JobFinishedEvent.FailureErrorCodes`, but it is used for some values, such as `JobFinishedEvent.PerformanceAdviceCodes` and `CommandInvokedEvent.Options` (in `Options.addTo`).

Should `truncateValue` be used for `JobFinishedEvent.FailureErrorCodes`?

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+		{Name: "azcopy.command.invoked", Value: float64(e.InvokedCount), Count: 1},
+	}
+}
+
+const (
+	maxPropValueLen         = 1024
+	maxIdentifierValueLen   = 128
+	maxHostValueLen         = 256
+	maxOptionValueLen       = 512
+	maxCompactListValueLen  = 512
```
al-msft, line 439: Do these specific lengths match limits defined elsewhere, or are they reasonable maximums?

telemetry/events.go
al-msft: I think I was misled by the use of `truncateValue` on `JobFinishedEvent.PerformanceAdviceCodes`. I now see `boundProperties` calls `truncateValueTo`. It looks like `attrs["PerformanceAdviceCodes"] = truncateValue(strings.Join(e.PerformanceAdviceCodes, ","))` and `Options.addTo`'s calls to `truncateValue` are redundant with `boundProperties`'s calls to `truncateValueTo`.

telemetry/events.go
```diff
@@ -0,0 +1,547 @@
...
+	marker := truncatedPropertyMarker
+	end := maxBytes - len(marker)
+	if end <= 0 {
+		marker = ""
+		end = maxBytes
+	}
+	for end > 0 && !utf8.RuneStart(value[end]) {
+		end--
+	}
+	return value[:end] + marker
```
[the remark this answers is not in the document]
al-msft: Agreed. The PR description promises unit tests in future PRs, but I don't see unit tests for this file in branch `ankurmicrosoft309/telemetry`. It's good practice to add unit tests along with new code, so the tests are less likely to be overlooked, are present and running for each subsequent code change, are more likely to test everything they should, and can help spot oversights in the code (this is a general statement; I'm not implying there are oversights in this PR).

al-msft (changes requested):
I've asked a lot of questions. I've mainly marked this "Request changes" to request updates to the comments that imply there are only two events. It may be that all my other comments can be resolved with answers, rather than code changes.

telemetry/events.go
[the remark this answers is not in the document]
ankurmicrosoft309: Outdated comments at the top, thanks! Resolved now

telemetry/events.go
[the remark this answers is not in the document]
ankurmicrosoft309: Agreed, resolved!

telemetry/events.go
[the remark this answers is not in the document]
ankurmicrosoft309: Fixed by adding "Opt" prefix to option keys.

telemetry/events.go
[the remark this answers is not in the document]
ankurmicrosoft309: Added the relevant unit test as part of the PR now.

ankurmicrosoft309:
> This appears to be a subset of #3571. ankurmicrosoft309: Do you recommend people review this first, then review the other files in #3571?

Yes, please review #3570 first, followed by #3571 and the subsequent PRs in order. I’ve now updated the PRs to form a chain: only #3570 targets main, and each subsequent PR targets the preceding branch.

telemetry/events.go
ankurmicrosoft309: Agreed. I have implemented a dedicated `JobMeasurements` struct. Identity, dimensions, status, and error properties remain on the event.

telemetry/events.go
ankurmicrosoft309: I totally understand and welcome any questions. I am much the same like you, both in terms of azcopy and golang.

telemetry/events.go
ankurmicrosoft309: Since its value is always 1, I have removed the redundant FinishedCount field and also removed it from the payload.

telemetry/events.go
ankurmicrosoft309: Yes, now part of measurement struct

telemetry/events.go
ankurmicrosoft309: ListJobSummaryResponse = operational job summary that exists from before
JobMeasurements = numeric data deliberately included in telemetry

The telemetry struct is intentionally separate from ListJobSummaryResponse as not every summary field should be emitted, and some telemetry measurements are derived from else where. I dont expect a lot of change here tbh but if it becomes a problem over time to keep them in sync, we can always refactor later.

telemetry/events.go
ankurmicrosoft309: Actually, you are write, these 2 are not necessary. Copilot added them for ease of summing them up on KQL aggregation query. But not necessary, and I have removed them. Thanks!

telemetry/events.go
ankurmicrosoft309: Yes, "Any" is correct.

telemetry/events.go
ankurmicrosoft309: Yes, removed. Thanks

telemetry/events.go
ankurmicrosoft309: Makes sense, done!

telemetry/events.go
ankurmicrosoft309: You are right, boundProperties already applies field-specific limits to every property, including FailureErrorCodes. I’ve removed the earlier redundant truncation calls so bounding happens once at the final event boundary. Thank you!

telemetry/events.go
ankurmicrosoft309: They are just reasonable maximums, made up by me completely and we are free to change them

al-msft (approved):

al-msft (changes requested):
This PR will need to add the new "telemetry" directory to the set of unit-test directories in ‎azure-pipelines.yml, so these new unit tests run in the pipeline. https://github.com/Azure/azure-storage-azcopy/pull/3544 reduces that to little more than a one-word change (line 157); since that PR is still awaiting review, in this PR, I recommend copying the section for "mock_server" (which should be lines 223-227 of the azure-pipelines.yml file on this branch).

telemetry/reporter_test.go
al-msft, line 1: Shouldn't this file be named "events_test.go"?

al-msft (approved):
Approved. As a result of some renamings, there are a few variables named `attrs` that we might want to rename to `props` at some point.

telemetry/events_test.go
ankurmicrosoft309: Makes sense. Done!

PR #3571, opened 2026-09-01
[Telemetry 02/22] Send telemetry with bounded envelopes and typed delivery failures

ankurmicrosoft309:
## Summary
Synchronous direct Application Insights sending with connection validation, sanitized errors, payload budgets, and typed delivery failures. The unused OpenTelemetry sender and backend selector have been removed.

## Review Order
Layer 2 of the 22-layer telemetry stack. Depends on #3570; review the incremental diff against `telemetry/01-event-contracts`. Next existing PR: #3572.

## Sending And Timeout Scope
- `ReportEvent` sends one event synchronously. This layer has no background queue, automatic retries, or shutdown/drain API.
- HTTP requests without a caller-provided context deadline receive a five-second fallback. Existing deadlines, including longer ones, and parent cancellation remain intact; the derived fallback context is canceled when the request completes.
- Layer 9 (`telemetry/09-bounded-async-delivery`) owns asynchronous dispatch, start/finish ordering within an attempt, the four-pending-send limit, five-second send contexts, and the independent four-second shutdown drain. Later layers wire these into transfer and CLI lifecycles. These runtime features are not part of this PR or this first-eight publication.

## Included Changes
- Removes the unused batch-send wrapper and its dedicated test, unreachable empty-envelope check, unused partial-ingestion message field, and ignored HTTP-status return value from postEnvelopes. Single-event sending, status validation, timeout/cancellation handling, and typed delivery failures are retained.
- Uses properties() consistently at the model-to-Application-Insights boundary and in wire-format tests. The string map is assigned directly to EventData Properties; numeric values remain in Measurements. This is a naming-only change with no serialized payload changes.
- Inherits telemetry unit-test and coverage registration from PR #3570; no duplicate pipeline entry is introduced in this layer.
- Direct Application Insights EventData envelopes, HTTP/partial-response validation, privacy-safe diagnostics, and typed delivery failures.
- Metadata-only start and command events retain their names, timestamps, and properties without synthetic occurrence measurements. Finished events retain 49 numeric observations before conditional omissions.
- Deadline regressions cover fallback creation and cleanup, preservation of shorter and longer caller deadlines, stalled local HTTP requests, caller cancellation, timeout classification, and no retry.
- Retains payload-budget, omission, property-bounding, and failure/privacy tests. Adds a brief purpose comment to the reporter file.
- Splits tests by production ownership: event contracts stay in `events_test.go`, envelope conversion lives in `send_events_test.go`, and connection/delivery behavior lives in `reporter_test.go`.

- `telemetry/budget_test.go`
- `telemetry/delivery_failure_test.go`
- `telemetry/events_test.go`
- `telemetry/privacy_failure_test.go`
- `telemetry/reporter.go`
- `telemetry/reporter_test.go`
- `telemetry/send_events.go`
- `telemetry/send_events_test.go`

## Validation
At this exact branch boundary on Windows, with `AZCOPY_DISABLE_TELEMETRY=true`:

```text
go test ./telemetry -count=1 -timeout=3m -run '.'
```

99 passing test/subtest results, 0 skips. The rebuilt local top also passed tagged telemetry checks in azcopy/cmd/e2etest and the actual-CLI loopback integration suite. Live Azure E2E, native Linux/macOS execution, and performance acceptance were not run for this change.

## Publication And Revision
- Publishes only layers 01-08 to existing origin branch heads. Later local layers and dashboards are not published or deployed by this operation.
- One incremental commit; no Markdown documentation, dashboards, or historical performance artifacts. Review-thread resolution and approval decisions are unchanged.
- Head commit: `662ede3d69d5747f3d78de5ffe6c8516be50d265`
- Parent commit: `04d9869a8ba2feb0ff437af4bb8b0d17574c68ca`
- Incremental diff: 8 files changed, 1106 insertions(+), 8 deletions(-)


telemetry/reporter_test.go
al-msft, line 1: Please add the new "telemetry" directory to the set of unit-test directories in ‎azure-pipelines.yml, so these new unit tests run in the pipeline. https://github.com/Azure/azure-storage-azcopy/pull/3544 reduces that to little more than a one-word change (line 157); since that PR is still awaiting review, in this PR, I recommend copying the section for "mock_server" (which should be lines 223-227 of the azure-pipelines.yml file on this branch).

telemetry/reporter.go
```diff
@@ -0,0 +1,496 @@
...
+// ---------------------------------------------------------------------------
+// Shared helpers
+// ---------------------------------------------------------------------------
+
+// parseConnectionString extracts key-value pairs from an Application Insights
+// connection string (e.g. "InstrumentationKey=abc;IngestionEndpoint=https://...").
+func parseConnectionString(cs string) map[string]string {
+	m := make(map[string]string)
+	for _, part := range strings.Split(cs, ";") {
+		kv := strings.SplitN(part, "=", 2)
```
al-msft: FYI: It looks like https://pkg.go.dev/strings#Cut could also have been used here; it returns two substrings and a `bool`, which eliminates the need to call `len` on the array `SplitN` returns. I'm only pointing this out because we're both relatively new to golang.

telemetry/reporter.go
```diff
@@ -0,0 +1,496 @@
...
+	if parseErr != nil || parsedEndpoint.Host == "" ||
+		(parsedEndpoint.Scheme != "http" && parsedEndpoint.Scheme != "https") ||
+		parsedEndpoint.User != nil || parsedEndpoint.RawQuery != "" || parsedEndpoint.Fragment != "" {
+		return "", "", errors.New("invalid Application Insights ingestion endpoint")
+	}
+	return endpoint, ikey, nil
+}
+
+// ---------------------------------------------------------------------------
+// OTel backend – real OpenTelemetry SDK with a custom App Insights exporter
```
al-msft: It feels like this file has multiple purposes or a very broad, yet partial, purpose. A clear indication of the file's potential multiple purposes is the division of the file into sections with headers like this. I think separating `sendEventAppInsights` and `eventToEnvelopes` from the other functions in this file was the right idea (this is why I said "partial" purpose). I'd suggest pulling a bit more out of this file.

I'm still learning the conventions for organizing files in golang, and most of my experience is with Java. Given the current code organization and the number of PRs that follow this one, my inclination would be to:
- move all code specific to the App Insights backend - which appears to be `sendEventAppInsights` and `eventToEnvelopes` - to a new file with a name that indicates it is focused on the App Insights backend, and
- move the OTel-specific code from this section of this file to a new file with a name that indicates it is focused on the OTel backend.

telemetry/reporter.go
```diff
@@ -0,0 +1,496 @@
...
+
+type deliveryFailure struct {
+	error
+}
+
+func (failure *deliveryFailure) Unwrap() error {
+	return failure.error
+}
+
+func IsDeliveryFailure(err error) bool {
```
al-msft, line 167: I don't see any uses of `IsDeliveryFailure` or `Unwrap` in non-test code in this PR or the original, large, telemetry PR. Are these methods used indirectly, or are they used in subsequent PRs?

telemetry/reporter.go
```diff
@@ -0,0 +1,496 @@
...
+	props map[string]string
+}
+
+func metricToPoints(m metricdata.Metrics) []metricPoint {
+	var points []metricPoint
+	add := func(value float64, count int, attrs attribute.Set) {
+		points = append(points, metricPoint{
+			name:  m.Name,
+			value: value,
+			props: attrsToProps(attrs),
```
al-msft: I'm confused about the terminology of "**properties**" and "**_attributes_**". It seems that:
1. `MetricEvent` in `events.go` converts the **properties** of the Event structs declared in that file into **_attributes_**.
2. `sendEventOTel` in `reporter.go` converts those **_attributes_** into **properties** of the `appInsightsExporter` struct.
3. `sendEventOTel` in `reporter.go` then converts the **properties** of the `appInsightsExporter` struct into **_attributes_** via `propsToAttributes`, and then calls `reader.Collect` to associate those attributes with metrics.
4. `Export` in `reporter.go` then calls `metricToPoints`, which converts those **_attributes_** into **properties** of the `metricPoint` struct via `attrsToProps` and returns an array of metricPoints.
5. `Export` in `reporter.go` then calls `pointsToEnvelopes`, which uses those **properties** as `Properties` and `Measurements` of the `appInsightsEventData` struct.

Is there a distinction - or more than one distinction - between "**properties**" and "**_attributes_**"? Are the terms varying in the code to keep them in line with the names used in multiple external contexts that aren't all consisent with each other?

al-msft: Would it be possible to use the term "**_attributes_**" in lines 2-4, above, ending with `metricPoint` calling them "**_attributes_**", so line 5 extracts `Properties` and `Measurements` from `metricPoint` "**_attributes_**"?

al-msft (changes requested):
If https://github.com/Azure/azure-storage-azcopy/pull/3570 doesn't address https://github.com/Azure/azure-storage-azcopy/pull/3570#pullrequestreview-5225550348, this PR will need to address https://github.com/Azure/azure-storage-azcopy/pull/3571#discussion_r3906454894. That's my one blocking comment. I also left a few non-blocking comments that may lead to code changes. I'll take another look at this PR once the comments have been resolved.

telemetry/reporter.go
ankurmicrosoft309: Ok thanks. I’ve removed the OTel implementation, which substantially reduces this file’s scope (it was an an alternative design which I didnt end up using). This makes the purpose of the file cleaner. Added a comment at the top as well.

telemetry/reporter.go
```diff
@@ -0,0 +1,479 @@
...
+	// Each numeric measurement becomes its own counter instrument, recorded
+	// with the shared attribute set. The manual reader then collects them for
+	// a single export pass.
+	for _, m := range evt.measurements() {
+		counter, err := meter.Float64Counter(m.Name)
+		if err != nil {
+			return fmt.Errorf("create instrument %s: %w", m.Name, err)
+		}
+		counter.Add(ctx, m.Value, otelmetric.WithAttributes(attrs...))
+	}
```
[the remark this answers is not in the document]
ankurmicrosoft309: Not relevant anymore as OTel sender was removed

telemetry/reporter.go
```diff
@@ -0,0 +1,479 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("marshal envelopes: %w", err)
+	}
+
+	url := endpoint + "/v2.1/track"
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(body))
+	if err != nil {
+		return 0, fmt.Errorf("create request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
```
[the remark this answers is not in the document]
ankurmicrosoft309: Agreed. The runtime agent supplies a bounded send context, but the reporter should also be safe when called independently without a deadline. I have added a five-second fallback for that case.

telemetry/send_events.go
```diff
@@ -0,0 +1,106 @@
...
+// ReportEvents processes and sends one or more generated telemetry events
+// (JobStartedEvent / JobFinishedEvent) using the configured backend. It stops
+// and returns on the first error.
+func (r *Reporter) ReportEvents(ctx context.Context, events ...MetricEvent) error {
+	for _, evt := range events {
+		if err := r.ReportEvent(ctx, evt); err != nil {
+			return fmt.Errorf("report %s: %w", evt.EventName(), err)
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
ankurmicrosoft309: This PR intentionally provides synchronous event sending. Asynchronous dispatch, lifecycle ordering, and bounded timeout belong to the runtime agent introduced in PR # 9. Will make it clearer in the commit description.

telemetry/reporter.go
ankurmicrosoft309: Oh neat, thanks. Done

telemetry/reporter.go
ankurmicrosoft309: `IsDeliveryFailure` is used in subsequent PR (PR #9 in this stack). `Unwrap` is used indirectly by Go’s error inspection APIs, preserving checks such as errors.Is(err, context.DeadlineExceeded)

telemetry/reporter.go
ankurmicrosoft309: Right, that is indeed confusing. I have removed the entire OTel path and renamed attributes to properties since there is no longer any material difference between the two terms now.

Now, we have 1) properties which have string values and 2) measurements which have numeric values.

ankurmicrosoft309:
> If #3570 doesn't address [#3570 (review)](https://github.com/Azure/azure-storage-azcopy/pull/3570#pullrequestreview-5225550348), this PR will need to address [#3571 (comment)](https://github.com/Azure/azure-storage-azcopy/pull/3571#discussion_r3906454894). That's my one blocking comment. I also left a few non-blocking comments that may lead to code changes. I'll take another look at this PR once the comments have been resolved.

Thank you Al for looking into the code. I apologize for presenting code here which was essentially dead code, its been cleaned up now. I have went back to earlier PR and added unit test to the pipeline as well.

telemetry/reporter_test.go
al-msft: This was addressed in https://github.com/Azure/azure-storage-azcopy/pull/3570.

al-msft (approved):

PR #3572, opened 2026-09-01
[Telemetry 03/22] Collect cross-platform host hardware metadata

ankurmicrosoft309:
## Summary
Best-effort local OS, CPU, physical-memory, and mount metadata probes, with deterministic unit tests plus native checks in the existing cross-platform New E2E suite.

NIC link speed is intentionally not collected. Advertised interface capacity does not reliably represent bandwidth available to AzCopy, especially on cloud VMs where VM SKU, storage, routing, and shared-resource limits may be lower.

## Metadata And Collection
- Human-readable OS version, including Windows product/release/build when available.
- CPU model or brand string.
- Total physical memory rounded to whole GiB; `-1` means unavailable.
- Local-path mount category: `nas-nfs`, `nas-smb`, or `local-disk`, through a separate path-specific helper.

Windows reads the registry for OS/CPU, `GlobalMemoryStatusEx` for RAM, and UNC/`GetDriveTypeW` for mount classification. Linux reads `/etc/os-release`, `/proc/cpuinfo`, `/proc/meminfo`, and the longest matching mount in `/proc/self/mountinfo`; mount-point escape sequences are decoded before matching. macOS uses `sysctl` for OS/CPU/RAM and `statfs` for mounts. Other platforms return unknown sentinels.

Physical-memory rounding avoids unsigned addition overflow and returns `-1` when the rounded GiB value cannot fit in the platform `int`, including the published i386 build.

## Tests And CI
- Pure Windows registry/path classification, Linux parsing/mount selection, macOS value normalization, and shared memory rounding live in the untagged `azcopy/internal/hostinfo` package, so the deterministic cases run on every OS.
- The Windows path table injects a fixed-drive resolver and uses an explicit final `else` to assert `localMountType` for every valid drive-root case; UNC and rootless cases remain separately asserted.
- The existing unit-test job runs `azcopy/internal/hostinfo` through the standard `run-ut.yml` template, including JUnit and coverage publication.
- Real OS, CPU, memory, and local-mount probes run in `HostHardwareSuite` through the existing New E2E jobs on Linux, Windows, and macOS. No special Build-matrix test step remains.
- The next layer adds the real Azure firmware detector to the same E2E scenario while keeping its deterministic helpers in the untagged unit package.

Current exact-head validation:
```text
go test ./azcopy/internal/hostinfo -count=1
go test ./e2etest -run '^$' -count=1
golangci-lint run --config=.config/.golangci.yaml --tests=false --max-issues-per-linter=0 --max-same-issues=0 --timeout 5m0s
```

The complete untagged helper suite passed on Windows amd64 and Windows/i386. The `HostHardwareSuite` native scenario body passed directly on Windows, including local mount checks. Linux package/E2E cross-compilation and native/Linux golangci-lint passed. Native Linux/macOS New E2E execution was not rerun for this revision.

Earlier Build run [31756](https://dev.azure.com/azstorage/AzCopy-NextGen/_build/results?buildId=31756) validated the real OS/CPU/RAM/mount probes on Linux, Windows, and macOS at an earlier revision. It is supporting platform evidence, not validation of this exact head or a full-pipeline pass.
## Review Order And Revision
Layer 3 of 22; depends on #3571, next #3573. Local branch `telemetry/03-host-hardware` is published through the existing head `telemetry/03-host-metadata` to preserve review history.

Only layers 01-08 are published. No new PRs were created.
- Head commit: `5f17c49cb64be30e74cd3d849e7771fd1a380a2c`
- Parent commit: `662ede3d69d5747f3d78de5ffe6c8516be50d265`
- Incremental diff: 15 files changed, 1047 insertions(+)

- `azcopy/hostinfo.go`
- `azcopy/hostinfo_darwin.go`
- `azcopy/hostinfo_linux.go`
- `azcopy/hostinfo_other.go`
- `azcopy/hostinfo_windows.go`
- `azcopy/internal/hostinfo/darwin.go`
- `azcopy/internal/hostinfo/darwin_test.go`
- `azcopy/internal/hostinfo/linux.go`
- `azcopy/internal/hostinfo/linux_test.go`
- `azcopy/internal/hostinfo/physical_memory.go`
- `azcopy/internal/hostinfo/physical_memory_test.go`
- `azcopy/internal/hostinfo/windows.go`
- `azcopy/internal/hostinfo/windows_test.go`
- `azure-pipelines.yml`
- `e2etest/zt_newe2e_host_hardware_test.go`


azcopy/hostinfo_other.go
```diff
@@ -0,0 +1,36 @@
...
+
+package azcopy
+
+// On platforms without a dedicated probe implementation, host hardware facts are
+// reported as sentinels (unknown). The telemetry pipeline treats these as
+// "not collected".
+
+func osVersion() string  { return "" }
+func cpuModel() string   { return "" }
+func totalMemoryGB() int { return 0 }
```
al-msft: Why return `0`, instead of the `-1` sentinel value mentioned in the comment in `azcopy/hostinfo.go`?

This question also applies to the three platform-specific hostinfo files that use 0 as the sentinel value.

azcopy/hostinfo.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// probeHostHardware gathers best-effort host hardware facts. It never blocks on
+// the network and never fails: missing values are returned as sentinels.
+func probeHostHardware() hostHardwareInfo {
+	return hostHardwareInfo{
+		osVersion:     osVersion(),
+		cpuModel:      cpuModel(),
+		memoryTotalGB: totalMemoryGB(),
+		nicMbps:       nicSpeedMbps(),
```
al-msft: Is there any value in declaring an interface that exposes the four functions above, along with `localMountType`?

azcopy/hostinfo.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+func physicalMemoryGB(bytes uint64) int {
+	const gib = 1 << 30
+	return int((bytes + gib/2) / gib)
```
al-msft: It took me an embarassing amount of time to look at the unit tests of this function and recognize half a GiB is being added to `bytes` here to ensure the number is rounded (half up) to the nearest GiB. That now makes sense, but what about overflow?

The documentation for `int` says "[int is a signed integer type that is at least 32 bits in size](https://pkg.go.dev/builtin#int)". This repo still [creates an i386 (32-bit) build](https://github.com/Azure/azure-storage-azcopy/blob/main/azurePipelineTemplates/build_windows.yml#L26) and [publishes it to GitHub](https://github.com/Azure/azure-storage-azcopy/releases/tag/v10.32.8). It looks to me like this function and its unit test will need to be updated to prevent overflow. Are we safe in assuming the value of `bytes` won't ever be large enough to cause overflow? I'd expect that to be an erroneous value; per the [Go spec](https://go.dev/ref/spec#Integer_overflow) and a quick hack I made to test this scenario, it would cause this function to return a negative value.

azcopy/hostinfo_hardware_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHostHardwareProbe(t *testing.T) {
+	hardware := probeHostHardware()
+	assert.Equal(t, osVersion(), hardware.osVersion)
+	assert.Equal(t, cpuModel(), hardware.cpuModel)
+	assert.Equal(t, totalMemoryGB(), hardware.memoryTotalGB)
+	assert.GreaterOrEqual(t, hardware.nicMbps, -1)
```
al-msft: It's great to see this test assert the constraint documented in the comment on `azcopy/hostinfo.go`: `Each field carries a sentinel ("" or -1) when it cannot be determined`. That helps enforce that invariant.

I have two questions:
1. Do you see any value in adding a fifth assertion that `hardware.memoryTotalGB` is at least 0 (or -1, depending on the response to https://github.com/Azure/azure-storage-azcopy/pull/3572/changes#r4039406095)?
2. Why leave out `assert.Equal(t, nicSpeedMbps(), hardware.nicMbps)`?

azure-pipelines.yml
```diff
@@ -47,6 +47,18 @@ jobs:
         # Don't run if job is canceled
         condition: and(succeededOrFailed(), eq(variables.type, 'linux'))
 
+      - pwsh: |
+          $output = @(go test -count=1 -timeout=3m -v ./azcopy -run '^TestHostHardware' 2>&1)
+          $testExit = $LASTEXITCODE
+          $output | ForEach-Object { Write-Host $_ }
+          if ($testExit -ne 0) { exit $testExit }
+          $passed = @($output | Where-Object { $_ -match '^--- PASS: TestHostHardware' })
+          if ($passed.Count -eq 0) { throw 'No host hardware tests executed' }
```
al-msft: It look like this line will report a failed test as an unexecuted test. Is that ok?

azure-pipelines.yml
```diff
@@ -47,6 +47,18 @@ jobs:
         # Don't run if job is canceled
         condition: and(succeededOrFailed(), eq(variables.type, 'linux'))
 
+      - pwsh: |
```
al-msft: I know running these tests here is intended to be a workaround for the unit tests not running on Windows or Mac. An argument could be made that they aren't unit tests, but integration tests. If we can re-classify them as integration tests, I'd suggest moving them into the New E2E test suite, like @shnayak-msft did with [e2etest/zt_newe2e_single_chunk_reader_test.go](https://github.com/Azure/azure-storage-azcopy/pull/3506/changes#diff-63db2bfa69f14f5004fe725c9f8e6b714bdcf2aef6b12259feeecb9a29fa983e).

Moving the tests would hook them into the existing reporting of code coverage and test results and remove the need for us to replace a workaround with a long-term solution later (which often means never). I'm not sure if the entry-point into the tests would need to be a higher-level function for the E2E framework to properly manage the tests.

azcopy/hostinfo_windows.go
```diff
@@ -0,0 +1,278 @@
...
+		bitsPerSecond := adapter.transmitBitsPerS
+		if adapter.receiveBitsPerS > bitsPerSecond {
+			bitsPerSecond = adapter.receiveBitsPerS
+		}
+		if bitsPerSecond == 0 {
+			continue
+		}
+
+		mbps := int(bitsPerSecond / 1_000_000)
+		if mbps > best {
```
al-msft: Is it ok that the division by `1_000_000` on the line above can cause this comparison to ignore a speed that's better by less than 1 mbps? If not, the division could be moved later in the function.

azure-pipelines.yml
al-msft: To be clear, only some of the current unit tests seem like integration tests to me. The tests that read files or the Windows registry integrate with the host OS, but the tests of helper methods that stub the registry or otherwise don't interact with the host OS should be runnable on any OS. I'm envisioning the build flags in the current unit test files being removed if we can move the integration tests to the New E2E test suite.

azcopy/internal/hostinfo/windows_test.go
```diff
@@ -0,0 +1,145 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			stripped := stripExtendedLengthPrefix(test.path)
+			assert.Equal(t, test.stripped, stripped)
+			assert.Equal(t, test.root, volumeRoot(stripped))
+			assert.Equal(t, test.unc, isUNCPath(stripped))
+			if test.unc {
+				assert.Equal(t, "nas-smb", localMountType(test.path))
+			} else if test.root == "" {
+				assert.Empty(t, localMountType(test.path))
+			}
```
al-msft, line 108: Could an `else` be added to close the gap here by asserting the localMountType of the remaining test cases? That might require assigning a stub to `var procGetDriveTypeW`, beause this is intended to be a unit test.

_This isn't a blocking comment._

al-msft (changes requested):

azcopy/hostinfo_linux.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	left := strings.Fields(line[:sep])
+	right := strings.Fields(line[sep+len(" - "):])
+	if len(left) < 5 || len(right) < 1 {
+		return "", "", false
+	}
+	// Field 5 (index 4) is the mount point; octal-style escapes (e.g. \040 for
+	// space) are left as-is, which is acceptable for prefix matching of typical
+	// mount roots.
+	return left[4], right[0], true
```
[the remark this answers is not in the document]
ankurmicrosoft309: This is correctly identified and fixed now

azcopy/hostinfo_other.go
ankurmicrosoft309: -1 is a better sentinel value. Changed to that.

azcopy/hostinfo.go
ankurmicrosoft309: These implementations are selected at compile time through build tags, so there is no runtime implementation switching. Interface is not required. It could still be useful for readability, but I will leave it as is for now.

azcopy/hostinfo.go
ankurmicrosoft309: Yes, you are right Al. Definitely overlooked this. Fixed it now!

azcopy/hostinfo_hardware_test.go
ankurmicrosoft309: 1. Added, seems reasonable.
2. I have been on border about capturing nic speed and its usefulness for business metrics. After talking with Raj Singh, I have decided to not capture it altogether.

azure-pipelines.yml
ankurmicrosoft309: Yes, that makes a lot of sense. I will do this

azure-pipelines.yml
ankurmicrosoft309: I ran an experiment locally and failed test is failed properly. Line 54 here does an early exit if the test failed. And assertion failure in go tests does return non-zero.

azcopy/hostinfo_windows.go
ankurmicrosoft309: This code is removed altogether now

azcopy/internal/hostinfo/windows_test.go
ankurmicrosoft309: Makes sense, done!

ankurmicrosoft309: Implemented. WindowsMountType now accepts an injected drive-type resolver. The table test uses a fixed-drive stub and has an explicit final else that asserts local-disk for every valid drive-root case, while also asserting that the resolver receives the expected root.

azure-pipelines.yml
ankurmicrosoft309: Implemented. The pure Windows, Linux, and macOS host transformations and detector parsers now live in untagged azcopy/internal/hostinfo tests, so they run in the standard unit job with JUnit and coverage reporting. Only real registry/Win32, /etc and /proc, sysctl/statfs, and fixed firmware/sysfs probes run in HostHardwareSuite under New E2E on Linux, Windows, and macOS. The bespoke Build-matrix selector has been removed.

al-msft (approved):

PR #3573, opened 2026-09-01
[Telemetry 04/22] Detect Azure VMs locally without metadata HTTP requests

ankurmicrosoft309:
## Summary
Detect Azure public-cloud VMs using a bounded local chassis asset-tag check without metadata-service HTTP calls. Deterministic detector tests run cross-platform, while the real host detector joins the existing New E2E host scenario.

## Implementation
- Windows reads raw SMBIOS through `GetSystemFirmwareTable`; named RSMB provider and raw-table identifiers document the Win32 call arguments. Linux reads `/sys/class/dmi/id/chassis_asset_tag`; other platforms return false.
- Asset-tag matching, bounded file reads, firmware-buffer handling, and SMBIOS parsing live in the untagged internal hostinfo package.
- Detection matches only the Azure public-cloud tag. Missing, unreadable, invalid, or unsupported input returns false, meaning not detected rather than proof that a machine is not Azure.
- No IMDS HTTP request, fallback network probe, raw firmware emission, or new telemetry field is introduced. Resource assembly uses the boolean in a later layer; it is not security attestation.

## Tests And CI
- Untagged unit tests cover tag matching, bounded asset-tag files, SMBIOS records/string indexes, firmware size/read failures, changing buffer sizes, missing terminators, truncation, declared lengths shorter than returned input, and invalid/oversized lengths on every OS.
- The inherited standard `azcopy/internal/hostinfo` unit-test entry publishes JUnit results and coverage. The fuzz target remains available for explicit fuzzing.
- The real fixed sysfs/Win32 detector runs through `HostHardwareSuite` in the existing New E2E jobs on Linux, Windows, and macOS and logs the detected boolean without assuming every runner is an Azure VM.
- No dedicated Build-matrix selector is added.

## Validation
At this exact PR boundary on Windows with `AZCOPY_DISABLE_TELEMETRY=true`:

```text
go test ./azcopy/internal/hostinfo -run '^TestAzurePublicCloud' -count=1
go test ./azcopy -run '^$' -count=1
go test ./e2etest -run '^$' -count=1
golangci-lint run --config=.config/.golangci.yaml --tests=false --max-issues-per-linter=0 --max-same-issues=0 --timeout 5m0s
```

The cross-platform detector unit tests and Windows/GOOS=linux golangci-lint passed. The combined `HostHardwareSuite` scenario body passed directly on Windows and detected the local Azure public-cloud firmware marker. Linux package/E2E cross-compilation passed. Native Linux/macOS New E2E execution was not rerun for this revision.
## Review Order And Revision
Layer 4 of 22; depends on #3572, next #3574. Review the incremental diff against `telemetry/03-host-metadata`.
Local branch `telemetry/04-azure-vm-detection` is published through the existing head `telemetry/04-job-metric-aggregation` to preserve review history. Only layers 01-08 are published; later local layers and dashboards remain unpublished by this operation. This PR is ready for review.
- Head commit: `4470b870e74b4aedec14fc723762cb15dbe075f1`
- Parent commit: `5f17c49cb64be30e74cd3d849e7771fd1a380a2c`
- Incremental diff: 7 files changed, 264 insertions(+)

- `azcopy/azureVMDetection.go`
- `azcopy/azureVMDetection_linux.go`
- `azcopy/azureVMDetection_other.go`
- `azcopy/azureVMDetection_windows.go`
- `azcopy/internal/hostinfo/azure_vm.go`
- `azcopy/internal/hostinfo/azure_vm_test.go`
- `e2etest/zt_newe2e_host_hardware_test.go`


azcopy/azureVMDetection_other.go
```diff
@@ -0,0 +1,7 @@
+//go:build !windows && !linux
```
al-msft, line 1: Should `azcopy/azureVMDetection_linux.go` and `azcopy/azureVMDetection_windows.go` also include build flags?

azcopy/azureVMDetection_windows.go
```diff
@@ -0,0 +1,26 @@
...
+func probeAzureVM() bool {
+	if err := procGetSystemFirmwareTable.Find(); err != nil {
+		return false
+	}
+	return hostinfointernal.AzurePublicCloudFromFirmware(func(buffer []byte) uint32 {
+		var address uintptr
+		if len(buffer) > 0 {
+			address = uintptr(unsafe.Pointer(&buffer[0]))
+		}
+		written, _, _ := procGetSystemFirmwareTable.Call(0x52534d42, 0, address, uintptr(len(buffer)))
```
al-msft, line 22: It took a bit of Internet searching ([1](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemfirmwaretable), [2](https://github.com/microsoft/win32metadata/issues/658), [3](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-enumsystemfirmwaretables)) for me to understand the arguments to `procGetSystemFirmwareTable.Call`.
1. I recommend extracting `0x52534d42` (which, for anyone reading the second link I referenced, is 1381190978 in decimal notation) into a constant named "RSMB".
2. It might also be helpful to extract the second argument, `0`, into a constant. I've found documentation (the third link above) that says: "The raw SMBIOS table provider ('RSMB') currently returns a single table identifier, 0x0000. This corresponds to the raw SMBIOS firmware table."

azcopy/internal/hostinfo/azure_vm.go
```diff
@@ -0,0 +1,80 @@
...
+func isAzurePublicCloudAssetTag(tag string) bool {
+	return len(tag) <= maxChassisAssetTagBytes && strings.TrimSpace(tag) == azurePublicCloudAssetTag
+}
+
+func azurePublicCloudFromSMBIOS(raw []byte) bool {
+	if len(raw) < 8 || len(raw) > maxSMBIOSTableBytes {
+		return false
+	}
+	length := binary.LittleEndian.Uint32(raw[4:8])
+	if uint64(length) > uint64(len(raw)-8) {
```
al-msft, line 47: This seems to guard against (among other things) a negative `length` causing a panic on line 50. That's clever.

Is there a test that covers `length` being positive, but less than `len(raw)-8`, in an otherwise valid `raw`?

al-msft (approved):

azcopy/azureVMDetection_other.go
ankurmicrosoft309: Not technically required. The name of the file ending with _linux and _windows is actually understood by the go toolchain and file is only included for the correct target OS. But for consistency I have included it.

azcopy/azureVMDetection_windows.go
ankurmicrosoft309: Thanks, that makes sense

azcopy/internal/hostinfo/azure_vm.go
ankurmicrosoft309: There wasnt but, I have added now.

al-msft (approved):

PR #3584, opened 2026-09-03
Handle transient AFD gateway responses without Storage error-code headers and add corresponding tests

shnayak-msft:
## Description

- **Feature / Bug Fix**: Addresses a retry-handling gap for transient responses generated by intermediaries such as Azure Front Door. AFD can return `502 Bad Gateway` or `504 Gateway Timeout` HTML responses without an Azure Storage `x-ms-error-code` header. These responses are now evaluated using the wildcard policy configured for their HTTP status code. The change preserves evaluation of `x-ms-copy-source-status-code` when the wildcard policy is disabled.

- **Related Links**:
- Issues
- Team thread
- Documents
- [Email Subject]

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

Added unit coverage for:

- Headerless `502 Bad Gateway` responses being retried.
- Headerless `504 Gateway Timeout` responses being retried.
- An unconfigured `501 Not Implemented` response not being retried.
- Continued evaluation of a retryable copy-source status when the response status wildcard is disabled.

Added an HTTP pipeline integration test that simulates AFD-style responses:

- `502 OriginConnectionAborted`
- `504 OriginTimeout`
- HTML response bodies with an AFD reference header
- No Azure Storage error-code header
- A successful response on the second request

The integration test verifies that the azcore pipeline sends exactly two requests and returns the successful response.

Focused validation:

```text
go test ./ste -run '^TestGetShouldRetry(RetriesAFDGatewayResponse)?$' -count=1
```

Formatting and whitespace validation also passed:

```text
gofmt -d ste/xferRetryHelper.go ste/xferRetryHelper_test.go
git diff --check -- ste/xferRetryHelper.go ste/xferRetryHelper_test.go
```

A Windows AMD64 private-testing binary was also built successfully using the repository’s pipeline-equivalent build settings.

### DIFF
--- a/ste/xferRetryHelper_test.go
+++ b/ste/xferRetryHelper_test.go
@@ -1,14 +1,21 @@
 package ste
 
 import (
+	"context"
 	"errors"
 	"net/http"
+	"net/http/httptest"
 	"runtime"
+	"sync/atomic"
 	"syscall"
 	"testing"
+	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	azruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/bloberror"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestGetShouldRetry(t *testing.T) {
@@ -40,6 +47,12 @@ func TestGetShouldRetry(t *testing.T) {
 					"x-ms-error-code": {code},
 				},
 			},
+			ShouldRetry: retry,
+		}
+	}
+	RespWithoutErrorCodeTest := func(status int, retry bool) *ResponseRetryPair {
+		return &ResponseRetryPair{
+			Resp:        &http.Response{StatusCode: status, Header: make(http.Header)},
 			ShouldRetry: retry,
 		}
 	}
@@ -90,6 +103,28 @@ func TestGetShouldRetry(t *testing.T) {
 			},
 		},
 		{
+			Rules: ParseRules("502; 504"),
+			Tests: []*ResponseRetryPair{
+				RespWithoutErrorCodeTest(http.StatusBadGateway, true),
+				RespWithoutErrorCodeTest(http.StatusGatewayTimeout, true),
+				RespWithoutErrorCodeTest(http.StatusNotImplemented, false),
+			},
+		},
+		{
+			Rules: ParseRules("502: OriginTimeout; 504"),
+			Tests: []*ResponseRetryPair{
+				{
+					Resp: &http.Response{
+						StatusCode: http.StatusBadGateway,
+						Header: http.Header{
+							"x-ms-copy-source-status-code": {"504"},
+						},
+					},
+					ShouldRetry: true,
+				},
+			},
+		},
+		{
 			TestCond: func() bool {
 				return runtime.GOOS == "windows" || runtime.GOOS == "darwin" || runtime.GOOS == "linux"
 			},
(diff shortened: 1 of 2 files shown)


ste/xferRetryHelper_test.go
```diff
@@ -89,6 +102,28 @@ func TestGetShouldRetry(t *testing.T) {
 				RespTest(409, string(bloberror.ContainerBeingDeleted), false),
 			},
 		},
```
al-msft, line 104: I had a hard time seeing the possible combinations of status codes and x-ms-error-codes until I copied the test set above (twice) and ran it with these changes:
```
		{
			Rules: ParseRules("409: ShareAlreadyExists, ShareBeingDeleted, BlobAlreadyExists; 502; 404: BlobNotFound"),
			Tests: []*ResponseRetryPair{
				RespTest(502, string(bloberror.ServerBusy), true), // passes (matches) with or without Shashank's change
				RespWithoutErrorCodeTest(502, true),               // passes (matches) with Shashank's change; fails (does not match) without it
			},
		},
		{
			Rules: ParseRules("409: ShareAlreadyExists, ShareBeingDeleted, BlobAlreadyExists; 502: ServerBusy; 404: BlobNotFound"),
			Tests: []*ResponseRetryPair{
				RespTest(502, string(bloberror.ServerBusy), true), // passes (matches) with or without Shashank's change
				RespWithoutErrorCodeTest(502, false),              // passes (does not match) with or without Shashank's change
			},
		},
```

_This isn't a blocking comment._

ste/xferRetryHelper_test.go
```diff
@@ -89,6 +102,28 @@ func TestGetShouldRetry(t *testing.T) {
...
+				RespWithoutErrorCodeTest(http.StatusBadGateway, true),
+				RespWithoutErrorCodeTest(http.StatusGatewayTimeout, true),
+				RespWithoutErrorCodeTest(http.StatusNotImplemented, false),
+			},
+		},
+		{
+			Rules: ParseRules("502: OriginTimeout; 504"),
+			Tests: []*ResponseRetryPair{
+				{
+					Resp: &http.Response{
```
al-msft, line 117: I extracted a helper method and re-ran this test set with two additional cases (see below). Is it a bug that the second test case passes when "x-ms-copy-source-status-code" matches an HTTP status code that is supposed to be qualified with an "x-ms-error-code" - or is the way the rule is specified just confusing in that case?
```
		{
			Rules: ParseRules("502: OriginTimeout; 504"),
			Tests: []*ResponseRetryPair{
				RespWithCopySourceStatusCodeTest(http.StatusBadGateway, "504", true),
				RespWithCopySourceStatusCodeTest(http.StatusBadGateway, "502", true), // passes (matches) with or without Shashank's changes - Why? The rule is limited to 502s that are OriginTimeouts.
				RespWithCopySourceStatusCodeTest(http.StatusBadGateway, "501", false),
			},
		},
```

ste/xferRetryHelper_test.go
```diff
@@ -43,6 +50,12 @@ func TestGetShouldRetry(t *testing.T) {
 			ShouldRetry: retry,
 		}
 	}
+	RespWithoutErrorCodeTest := func(status int, retry bool) *ResponseRetryPair {
+		return &ResponseRetryPair{
+			Resp:        &http.Response{StatusCode: status, Header: make(http.Header)},
+			ShouldRetry: retry,
+		}
+	}
```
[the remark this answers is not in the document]
al-msft: This looks like the wrong place to put this code. A better place would be just above `for entryNum, v := range matrix {`.

ste/xferRetryHelper_test.go
al-msft: This case also helped, by clearly separating the HTTP status code from the copy source code and the codes in the rules:
`RespWithCopySourceStatusCodeTest(501, "502", true), // passes (matches) with or without Shashank's changes - The HTTP status code is irrelevant; it's the copy-source status code that is being matched against the rules.`

_This isn't a blocking comment._

ste/xferRetryHelper_test.go
```diff
@@ -140,6 +175,59 @@ func TestGetShouldRetry(t *testing.T) {
 	}
 }
 
+func TestGetShouldRetryRetriesAFDGatewayResponse(t *testing.T) {
```
al-msft, line 178: To me, starting up an HTTP server feels too heavyweight for a unit test.

al-msft: @adreed-msft: Does it make sense to move integration tests like this into the "New E2E" test suite?

al-msft (changes requested):
1. Copilot identified some necessary test clean-up: https://github.com/Azure/azure-storage-azcopy/pull/3584/changes#r3929489535
2. Can you confirm it's correct to retry a status code the rules qualify with a storage error code? https://github.com/Azure/azure-storage-azcopy/pull/3584/changes#r3936501610
3. I'm not sure we should continue running integration tests like unit tests: https://github.com/Azure/azure-storage-azcopy/pull/3584/changes#r3936823347

PR #3591, opened 2026-09-15
Path-traversal guard for MSRC 139699 and 143863

adreed-msft:
## Description

- **Feature / Bug Fix**: It is not possible to achieve this with production Azure services, but, if one were to write a blob with a path traversal in the name, they could theoretically perform a path traversal outside of the user's targeted download directory. (i.e. a blob named `foo/../../../bar.txt` could write bar.txt 2 directories earlier into the tree than the user was targeting.)

The bug was reported using Azurite-- so it at least exists in *one* case.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## How Has This Been Tested?

It's not easy to test in an automated fashion since it is dependent upon Azurite, but the regex has fairly thorough testing via a unit test.


common/path_traversal_regex.go
```diff
@@ -0,0 +1,10 @@
+package common
+
+import "regexp"
+
+// PathTraversalNameRegex matches any path segment that is fully comprised of dots
+// ("."), e.g. ".." or "./", on either a forward- or backslash-delimited path.
+// Such a segment is not representable as a real directory name on most filesystems, and, if encountered in
+// a downloaded relative path (e.g. "../foo"), would resolve outside of the destination directory the user
+// targeted. Both the front-end (traverser) and the storage engine (ste) rely on this to detect such names.
+var PathTraversalNameRegex = regexp.MustCompile("(?i)(^|[/\\\\])\\.+($|[/\\\\])")
```
al-msft, line 10: Doesn't ["(?i)" trigger case insensitivity](https://pkg.go.dev/regexp)? Case doesn't seem relevant to dots and slashes.

common/path_traversal_regex_test.go
```diff
@@ -0,0 +1,54 @@
...
+		input       string
+		shouldMatch bool
+	}{
+		{"leading parent traversal", "../foo", true},
+		{"parent traversal mid-path", "a/../foo", true},
+		{"backslash parent traversal", "a\\..\\foo", true},
+		{"current dir traversal", "./foo", true},
+		{"multiple dots at start", ".../foo", true},
+		{"single dot at start", ".", true},
+		{"single dot mid-path", "foo/.", true},
```
al-msft, line 22: It looks like the leaf node was accidentally left off this path. The current path doesn't seem to match the test case name and is identical to the path for test case "trailing current dir".

traverser/path_traversal_regex.go
```diff
@@ -0,0 +1,8 @@
+package traverser
+
+import "github.com/Azure/azure-storage-azcopy/v10/common"
+
+// PathTraversalNameRegex matches any path segment that is fully comprised of dots (".").
+// It is defined in the common package (so it can also be used by the storage engine without
+// an import cycle) and re-exported here for readability from the traverser package.
+var PathTraversalNameRegex = common.PathTraversalNameRegex
```
al-msft, line 8: Where is this used?

common/fe-ste-models.go
```diff
@@ -895,11 +895,13 @@ func (TransferStatus) BlobTierFailure() TransferStatus { return TransferStatus(-
 
 func (TransferStatus) SkippedEntityAlreadyExists() TransferStatus { return TransferStatus(-3) }
 
-func (TransferStatus) SkippedBlobHasSnapshots() TransferStatus { return TransferStatus(-4) }
+func (TransferStatus) SkippedEntityHasInvalidName() TransferStatus { return TransferStatus(-4) }
```
al-msft, line 898: Is there any risk to shifting the TransferStatus integers? In some contexts, a new value would need to be appended to the end to avoid problems caused by re-purposing the underlying value, like `-4`, here.

jobsAdmin/init.go
```diff
@@ -431,6 +431,7 @@ func resurrectJobSummary(jm ste.IJobMgr) common.ListJobSummaryResponse {
 						TransferStatus:     common.ETransferStatus.Failed(),
 						ErrorCode:          jppt.ErrorCode()}) // TODO: Optimize
 			case common.ETransferStatus.SkippedEntityAlreadyExists(),
+				common.ETransferStatus.SkippedEntityHasInvalidName(),
```
al-msft, line 434: Should any of the existing files this PR updates have unit tests that should be updated?

al-msft (changes requested):

